Issue #380, opened 2018-11-07
When Alice sends Swap Request...


D4nte:
When Alice sends a Swap Request, both Alice and Bob transition to `Start` state.

- [ ] Update HTTP API to send a ETH-BTC and BTC-ETH swap request
- [ ] Update HTTP API to return the full State + actions of existing swap requests (list of requests)

Expected actions to be returned in GET swap:
* Alice: none
* Bob: accept, decline

**POST SWAP**
Route should be `swaps/rfc003`

**GET SWAP Response**

```
{
  "state": "Start",
  "swap": {
     "alpha_ledger" : "Bitcoin",
     "beta_ledger" : "Ethereum",
     "alpha_asset" : "Bitcoin",
     "beta_asset" : "Ether",
   },
   "role" : "Alice",
  "_links" : 
  {
    "redeem" : {
      "href" : "swaps/rfc003/:id/redeem"
  },
  ..
}
```

**Actions**
Actions are sublinks (use HAL) pointing to resources allowing the action.
e.g: `POST swaps/rfc003/:id/redeem` 

**GET List swaps?**
Need to list the current ongoing swaps.
Route: `GET /swaps/`

*Response*
See http://stateless.co/hal_specification.html
Using implementation of "collections".
Notes:
- include protocol
- use `_links` to provide link to `GET swap/:id`
- ensure the state name is included

DoD:
- [ ] Implement `GET swaps/:id` with format above
- [ ] Implement `GET swapS` with format above
- [ ] Update e2e test to use (and only use) these APIs. e2e test would now stop at checking that a Started swap exist on both Alice and Bob side.
- Do **not** implement the action sub-endpoints

Test assertions:
- [ ] Both return (http) a swap in the state "Started"
- [ ] Alice does not return any action on the state
- [ ] Bob returns "accept" and "decline" actions on the test (the actions do not need to be actionable)

Notes:
Create `SwapMetadata::for_alice`/`bob` method that takes `Into<Ledger>` parameters and then implement `From<Ledger>` trait on the enum parameters. This will allow the following block to be done generically.
https://github.com/tenx-tech/swap/blob/c3d6b51f0df52ff918f47f6f30fb7e311f954e8a/application/comit_node/src/http_api/swap.rs#L261-L267

Child of #405.
Child of #408.
Blocks #462.
Blocks #382.

PR #422, opened 2018-11-15
Slim down HTTP controller

### LINKED ISSUE #409: Slim down create swap HTTP controller
Currently, a lot of stuff is happening in the controller of `POST /swaps`.

As a preparation work of #380, we should slim down the controller to just send the id + data through a channel and do whatever needs to be done over there. For this, there needs to be a concrete type owned by the `rfc003` module that we can send through the channel. This involves some refactoring regarding the deserialization of the request body.

Child of #405.
Child of #408.

thomaseizinger:
Resolves #409.

This PR basically moves the whole logic of creating a swap into `CreateSwap` which holds the `Receiver` of a channel. The HTTP controller now just sends stuff through this channel is no longer aware of what happens on the other side.

This is the major preparation work for #380.


bonomat (changes requested):
Loads of stuff happening here... and more to come. 
As the code grows we we should be really carefully with naming stuff...

bonomat (dismissed):

bonomat (approved):

LLFourn (commented):
The reservation I have with the deserialize direct to core type approach (as we discussed briefly) is that it is inflexible when you want to set default values for fields based on runtime config file values etc.

I think we can see if this is a concern later down the track when we add such features. Otherwise LGTM.

PR #423, opened 2018-11-15
Remove setup.sh and check-fmt.sh

### LINKED ISSUE #347: Remove check-fmt.sh and setup.sh
With #344, we can now remove two more shell scripts from our directory.

## Why can we remove `setup.sh`?

- Dependencies are now installed as you run commands with cargo-make
- The rust toolchain version is enforced through the `rust-toolchain` file

## Why can remove `check-fmt.sh`?

Creating a `.githooks` folder is more maintainable. (See tasks section below on how to setup)

## Tasks:

- Create `.githooks` folder in the repository root with a `pre-commit` script that does `cargo fmt -- --check`.
- Add a `install-git-hooks` command to cargo-make that will set the [`core.hooksPath`](https://git-scm.com/docs/git-config#git-config-corehooksPath) config of Git to this folder.

thomaseizinger:
These two files are no longer needed as we now have cargo-make. In
order to ensure that all developers have the commit-hooks installed,
we set the config as part of the cargo-make init task which is
executed every time a cargo-make command is run.

Resolves #347.
Supersedes #420.


D4nte (changes requested):

D4nte (approved):
LGTM
Thanks!

PR #430, opened 2018-11-18
Alpha and Beta

bonomat:
So, I thought to give the renaming a try :)

Since we have a bunch of open branches right now, I thought a script would be useful (see below)
In order to solve merge conflicts easily, try to run the script on you branch and then rebase. 
It's probably not the most beautiful script, but it seems to work (on my machine) :)

```
#!/usr/bin/env bash

skip_files=("logging.rs" "routes.rs")
skip_files_list="";
for file in ${skip_files[@]}; do
    skip_files_list=${skip_files_list}$(printf " ! -name %s " ${file});
done

echo $skip_files_list
#replace strings
replace=("Source/Alpha" "source/alpha" "SL/AL" "SA/AA" "SOURCE/ALPHA" "Target/Beta" "target/beta" "TL/BL" "TA/BA" "TARGET/BETA")

# search for every rs file
find . ! -name "*.bak" -iname "*.rs" ! -path "./target/*" ${skip_files_list} -o -iname "*.js" ! -path "./api_tests/node_modules/*" |
while read file_name; do

    if grep -iq "source" ${file_name} || grep -iq "target" ${file_name}; then
        cp "$file_name" "$file_name".bak
        printf "%s contains source or target\n" ${file_name}
        # backup of sed does not as we iterate over the changes
        cp -n "$file_name" "$file_name".bak
        for i in ${replace[@]}; do
            printf "replacing: %20.20s \t\t in \t %s \n" ${i} ${file_name};
            sed -i '' '/source(&self)/!s/'${i}'/g' "$file_name";
        done
    fi
done

cargo fmt


if $(cargo make); then
    echo "Build failed"
else
    echo "Build successful"
fi

# delete bak files at the end
read -p "Are you sure you want to delete all backup files (y/n)? " -n 1 -r
echo
if [[ $REPLY =~ ^[Yy]$ ]]
then
     find . -name '*.bak' -exec rm {} \;
fi

```

### DIFF
--- a/api_tests/e2e/rfc003/btc_eth/test.js
+++ b/api_tests/e2e/rfc003/btc_eth/test.js
@@ -24,7 +24,7 @@ const alice = test_lib.comit_conf("alice", {
 });
 
 const alice_final_address = "0x00a329c0648769a73afac7f9381e08fb43dbea72";
-const target_asset = new BigNumber(web3.utils.toWei("10", 'ether'));
+const beta_asset = new BigNumber(web3.utils.toWei("10", 'ether'));
 const bitcoin_rpc_client = test_lib.bitcoin_rpc_client();
 
 describe('RFC003 Bitcoin for Ether', () => {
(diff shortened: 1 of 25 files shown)


thomaseizinger (dismissed):
Good idea with the script!
Seems to be working well apart from a few occasions :)

D4nte (approved):
Looks good!

thomaseizinger (approved):

PR #443, opened 2018-11-19
Introduce SwapRequestHandler for Bob

### LINKED ISSUE #417: Slim down create swap BAM controller
Currently, a lot of stuff is happening in the controller of BAM/SWAP.

To simplify the logic, when Bob receives a SWAP request over BAM, concrete type should be passed over a channel and the state machine should be started by the receiver.

Child of #405.
Child of #408.

thomaseizinger:
Resolves #417.

This is the 2nd preparation step for doing #380. After this, we only need to "fill-in" the instantiation of the state machine in the `SwapRequestHandler` for Alice and Bob.


LLFourn (approved):
Nice work. Approving but would prefer the enums be renamed to ..`SwapRequestKind` rather than `SwapRequests` (and wherever else you feel it would be appropriate on Asset and Ledger etc).

D4nte (commented):
First pass

D4nte (changes requested):
Couple of comment regarding `unwrap()`.

D4nte (approved):

PR #464, opened 2018-11-21
Allow Bob to accept a swap

thomaseizinger:
This PR adds two routes to the HTTP API:

- accept
- decline

These are the two actions that Bob can perform when presented with a swap request. At the moment, only `accept` is implemented but this is the much more difficult one. If he accepts, his response is used for two things:

- it is sent back to Alice
- his own state machine is advanced into the next state and starts listening for Alice to deploy the first contract

### DIFF
--- a/api_tests/e2e/rfc003/btc_eth/test.js
+++ b/api_tests/e2e/rfc003/btc_eth/test.js
@@ -41,6 +41,15 @@ describe('RFC003 Bitcoin for Ether', () => {
         });
     });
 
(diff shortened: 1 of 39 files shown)


D4nte (changes requested):
Well done @thomaseizinger for finishing off this PR! Few minor comments

bonomat (commented):
Good job. I just added some comments.

LLFourn (approved):

D4nte (approved):

Issue #450, opened 2018-11-20
Create `LedgerEvents` for ERC20 


bonomat:
Implement `LedgerEvents<Ethereum, ERC20>`.

DoD:
- [ ] Test that instantiate the state machine where ERC20 is alpha
- [ ] Test that instantiate the state machine where ERC20 is beta

Child of #399.
Child of #408.

PR #478, opened 2018-11-23
Add Deployed State & Implement LedgerEvents for ERC20

### LINKED ISSUE #448: Introduce HTLC deployed state
Right now our state machine has a single state for deploying and funding an htlc. However, especially for ERC20 tokens we need an separate state, i.e. splitting up `AlphaFunded` into two new states.

This state would be skipped for Bitcoin.

DoD:
- [ ] State machine `Swap` is extended
- [ ] `LedgerEvents` is extended

Note:
e2e will be done with #445 for ERC20

Child of #399.
Child of #408.
Blocks #445.
Blocks #415.

D4nte:
* Create new Deployed states.
* New LqsEvents for Erc20 with a number of placeholders to be sorted with #450 

Resolves #448


thomaseizinger (dismissed):

LLFourn (changes requested):
Nice work on the code reuse in `lqs.rs`. I think there is one change we need to make there to make it cleaner.

I really do hope we are able to delete this stuff though by making the funding and deployment atomic for ERC20 using the random github EVM codes 🤗

LLFourn (approved):

PR #483, opened 2018-11-26
Implement funding query for ERC20

D4nte:
Resolves #450.

Recommendation on how to test this are welcome. I think this needs to be an end-to-end test.

Also, checking with @LLFourn it seems that the tests defined in the original issue are moot. Please comment.

PR #493, opened 2018-11-27
Implement feedback from PR #483

thomaseizinger:
This implements some feedback from PR #483.
This is includes fixing the construction of the transaction for invoking `transfer` on an ERC20 token contract.

To verify that it works, I am now using this function in the tests to transfer the tokens. This imposed a small refactoring of the harness which now passes back the HTLC so that we can call functions on it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,14 +9,12 @@ sudo: required
 rust: nightly-2018-11-14
 addons:
   apt:
+    sources:
+    - sourceline: 'deb http://dl.yarnpkg.com/debian/ stable main'
+      key_url: 'http://dl.yarnpkg.com/debian/pubkey.gpg'
     packages:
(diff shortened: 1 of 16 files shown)


D4nte (changes requested):
I do not understand the reasoning behind the use of hex has a middle man for the payload building.

D4nte (changes requested):
I do not understand the reasoning behind the use of hex has a middle man for the payload building.

bonomat (dismissed):
Happy to discuss the definition of _giant_ ;) 

Looking good. Added some comments regarding your test: I don't think it makes that much sense in that form.

D4nte (dismissed):

LLFourn (approved):
I made one possibly useful comment but approved otherwise.

PR #500, opened 2018-11-28
Remove final address from swap request

### LINKED ISSUE #484: Remove UTXO Final pubkey hash at creation/acceptance time.
"Final" Redeem/Refund address for UTXO assets to be provided at the time of the Redeem/Refund action (instead of swap creation/acceptance).

DoD:
- [x] Alice, swap request creation (HTTP), remove redeem/refund identity if the ledger is Bitcoin.
- [x] Bob, accept action (HTTP), remove redeem/refund identity from body if the ledger is Bitcoin

Pre-Work:
Verify it works with `serde` before being deep in the change.

Note:
For Alice, in `swap.rs`, add new type parameters that would replace `BL::Identity` and `AL::Identity`. Similar change for Bob *OR* Consider adding a new associated type `HttpIdentity` to the `Ledger` trait.

Blocks #387.
Blocks #388.
Child of #405.
Child of #399.
Child of #408.

D4nte:
For Alice, final refund address is removed from the HTTP request
to create a swap.

For Bob, final redeem address is removed from the HTTP request to accept a swap.

Resolves #484.

### DIFF
--- a/api_tests/e2e/rfc003/btc_eth/test.js
+++ b/api_tests/e2e/rfc003/btc_eth/test.js
@@ -59,8 +59,7 @@ describe("RFC003 Bitcoin for Ether", () => {
                     name: "Ether",
                     quantity: beta_asset.toString()
                 },
-                alpha_ledger_refund_identity:
-                    "ac2db2f2615c81b83Fe9366450799b4992931575",
+                alpha_ledger_refund_identity: null,
                 beta_ledger_success_identity: alice_final_address,
                 alpha_ledger_lock_duration: 144
             })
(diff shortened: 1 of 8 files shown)


thomaseizinger (approved):
I would have favored the solution that doesn't involve another associated type but I guess for the demo, it is good enough. Good work Franck :+1:

LLFourn (commented):
Waiting for others to approve. I think this is a fine solution for now but I'd prefer to get rid of the null before demo.

[EDIT] @thomaseizinger already took a look and so we're rolling with it :)

PR #502, opened 2018-11-28
[BTC->ETH] Alice Redeems 🏆

### LINKED ISSUE #387: When Alice redeems beta asset...
...both Alice and Bob transition to `AlphaFundedBetaRedeemed`.

Expected actions to be returned in GET swap:
- Alice: none (for now)
- Bob: `redeem`

DoD:
- [x] New query parameter on the GET request for redeem action to provide Bitcoin final pubkeyhash
- [x] e2e proceeds with redeem transaction
- [x] e2e checks that both node transition
- [x] e2e checks that Bob returns `redeem` action
- [x] e2e checks that `redeem` action returns transaction data

Notes:
`Refund actions` to be handled in different issue.

Child of #408.
Child of #405.
Blocks #388.

D4nte:
- New query parameters on actions (HTTP).
- Fix `htlc_redeemed_or_refunded`
- e2e tests up to `AlphaFundedBetaRedeemed`
- New test in LQS to double check correctness

Resolves #387.

### DIFF
--- a/.idea/encodings.xml
+++ b/.idea/encodings.xml
@@ -1,4 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="Encoding" addBOMForNewFiles="with NO BOM" />
-</project>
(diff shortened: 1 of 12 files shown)


thomaseizinger (approved):
I was part of the pairing so 2nd approve pls.

LLFourn (commented):
Great Work!
I made a comment about the `PartialEq` magic. I think it's a bit too much. I always interpret `==` as meaning actually equal to in all values rather than "is related to".

thomaseizinger (approved):

LLFourn (approved):

PR #507, opened 2018-11-29
Introduce Transaction new types

thomaseizinger:
This allows us to differentiate between Redeem, Refund and Funding
transactions and avoids errors down the line where they could be mixed
up accidentally.


D4nte (approved):
LGTM

PR #509, opened 2018-11-29
[BTC->ERC20] Both parties to BothFunded state 💰️

### LINKED ISSUE #435: BTC->ERC20 swap up to `AlphaFundedBetaDeployed`
DoD:
- [ ] e2e test with BTC->ERC20 where both parties are in AlphaFundedBetaDeployed

Child of #399.
Child of #408.
blocked by #445

LLFourn:
As described

fixes #435 
Note that issue was only up to BetaDeployed but I was on a roll! XD


D4nte (changes requested):
Few changes regarding how we transformed the data returned by actions.
I believe we should not need to transform it.

I would go as far as saying that maybe the `gas_limit` and the `value` should both me decimal (or both hex).

D4nte (approved):
Well done both!

PR #517, opened 2018-12-03
Remove PendingResponses + HttpIdentity

thomaseizinger:


LLFourn (approved):
Note: I think a lot of things here will have to be reworked but it gets rid of the nulls in the HTTP api for now so let's roll with it.

PR #518, opened 2018-12-04
Fix clippy warnings + make build fail if there are warnings

thomaseizinger:
# READ ME FIRST!

I had to use several `#[allow(clippy::...]` in the code to make the warnings go away.
Note that this does **not** mean, that the code is okay! It just means that we make it explicit, that we might have a problem here that needs to be taken care of. Making the build fail IF there is a warning reminds us to either fix them or explicitly allow them, either because they are ok or because it needs a bigger change to get rid of them.


D4nte (approved):
LGTM

Issue #522, opened 2018-12-04
Service health HTTP call 


D4nte:
Implement a "service check" http endpoint to allow the user to know to what services is the comit_node connected.
- btsieve: return IP/port/status
- LND: return Pubkey/status

Best if this is done matching this RFC: https://inadarei.github.io/rfc-healthcheck/

thomaseizinger:
Moved to "New" to be re-discussed as part of https://app.zenhub.com/workspaces/coblox-scrum-5be4c2ea4b5806bc2bf2d351/issues/comit-network/comit-rs/1122.

thomaseizinger:
Closed because there is no need for now.

PR #523, opened 2018-12-04
Connect COMIT node to Lightning node

### LINKED ISSUE #454: [BTC->LN-BTC] LND configuration
DoD:
- [ ] Add lnd configuration to config file
- [ ] When receiving a http swap request, check that lnd is alive (before forwarding)
- [ ] Test both features above with e2e

Note:
Do the lnd alive check for Bob in #456
In the http request, alice user needs to provide pubkey of node as `beta_success_identity`.

Child of #431.
Blocks #456.
Blocks #455.

D4nte:
the `get_info` call in `comit_node.rs` can be removed once #522 is implemented.

Resolves #454.


LLFourn (approved):
LGTM

thomaseizinger (approved):

PR #531, opened 2018-12-05
Make nvm be useable offline and remove duplication

thomaseizinger:
We actually had quite some duplication with the whole nvm stuff.

This is now all done inside harness_lib when we start tests.


D4nte (commented):
Nice! Just need to make it work :p

D4nte (approved):
LGTM

PR #556, opened 2018-12-11
More Rust 2018 Idioms

### LINKED ISSUE #552: Rust 2018 idioms
Continuation of #266:

- remove `mod.rs` where possible
- remove `#[macro_use]` where possible
- remove `use foo;`

D4nte:
* Remove unneeded `use foo;` when foo is an external crate.
* Add `rust_2018_idioms` warnings
* Fix up some lifetimes
* Show case that `main` can return a Result (that is debugged if it's `Err`)

I went through Rust 2018 documentation and I could not see any other change we should do take advantage of Rust 2018, hence, it resolves #552.


bonomat (approved):
Just one question but otherwise looks good to me

application/comit_node/src/bin/comit_node.rs
```diff
@@ -37,7 +37,7 @@ fn main() {
     let state_store = Arc::new(InMemoryStateStore::default());
     let ledger_query_service_api_client = create_ledger_query_service_api_client(&settings);
 
-    let mut runtime = tokio::runtime::Runtime::new().unwrap();
+    let mut runtime = tokio::runtime::Runtime::new()?;
```
LLFourn: Cool. Is this a new feature?

application/comit_node/src/swap_protocols/rfc003/secret.rs
```diff
@@ -96,22 +94,24 @@ impl From<Secret> for SecretHash {
 }
 
 impl Secret {
+    pub const SECRET_LENGTH: usize = 32;
+
     pub fn generate<T: RandomnessSource>(rng: &mut T) -> Secret {
-        let random_bytes = rng.gen_random_bytes(SECRET_LENGTH);
-        let mut secret = [0; SECRET_LENGTH];
+        let random_bytes = rng.gen_random_bytes(Self::SECRET_LENGTH);
+        let mut secret = [0; Self::SECRET_LENGTH];
```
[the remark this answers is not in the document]
LLFourn: Nice. I think `Self` fine.

LLFourn (approved):
LGTM

application/comit_node/src/bin/comit_node.rs
D4nte: Yep https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/question-mark-in-main-and-tests.html

PR #571, opened 2018-12-14
Use config file for test dependencies

### LINKED ISSUE #475: Dependencies for tests should be specified per directory
Right now we have 3 harness.sh scripts :\ (LQS, dry, and e2e). We should instead have a way of specifying what nodes need to be running in each test directory. Possibly a Makefile.toml.

D4nte:
- [x] Configure service dependencies in a local config.toml file
- [x] Create harness.js and remove shell script
- [x] Migrate e2e to harness.js
- [x] Migrate dry to harness.js
- [x] Migrate lqs to harness.js
- [x] Use `log4js` instead of `console.log()`
- [x] Add git hook to run prettier

Resolves #475


api_tests/e2e/rfc003/btc_eth-erc20/test.js
```diff
@@ -12,89 +12,95 @@ const toby_wallet = test_lib.wallet_conf();
...
+        logger.info("Alice created a new swap at %s", swap_location);
+        swap_location.should.be.a("string");
+        alice_swap_href = swap_location;
     });
 
     let bob_swap_href;
 
     it("[Bob] Shows the Swap as Start in /swaps", async () => {
+        // Wait 100ms for Alice to send the swap request over
+        await test_lib.sleep(100);
```
D4nte: poll on `Start`

api_tests/e2e/rfc003/btc_eth/test.js
```diff
@@ -90,6 +151,8 @@ describe("RFC003: Bitcoin for Ether", () => {
     let bob_swap_href;
 
     it("[Bob] Shows the Swap as Start in /swaps", async () => {
+        // Wait 100ms for Alice to send the swap request over
+        await test_lib.sleep(100);
```
D4nte: Poll on `Start`

vendor/bam/tests/place_android_phone_order_request.json
```diff
@@ -3,10 +3,10 @@
   "id": 10,
   "payload": {
     "type": "PLACE-ORDER",
-    "headers" : {
-      "PRODUCT-TYPE" : "PHONE"
+    "headers": {
+      "PRODUCT-TYPE": "PHONE"
     },
-    "body" : {
+    "body": {
```
LLFourn: 🤡 I'm guessing this wasn't intentional. Let's keep prettify to api_tests only 😅.

D4nte: Hum no, it's intentional. Prettify all Json and JS files.

api_tests/e2e/rfc003/btc_eth/config.toml
```diff
@@ -0,0 +1,27 @@
+[ledger.bitcoin]
+docker = "btc"
+rpc_host = "localhost"
+rpc_port = "18443"
+rpc_username = "bitcoin"
+rpc_password = "54pLR_f7-G6is32LP-7nbhzZSbJs_2zSATtZV_r05yg="
+up_time = "10000"
+down_time = "2000"
```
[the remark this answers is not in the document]
D4nte: Hum, yeah I was thinking that. Yes there is repetition which is not great but it does make everything more straightforward/less hardcoded... I'll think about it.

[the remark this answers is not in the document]
D4nte: Done!

vendor/bam/tests/place_android_phone_order_request.json
D4nte: ok, removed...

thomaseizinger:
I thought the point of #475 was to not roll-our-own task runner but use cargo-make to split up the testing into composable tasks and run those?
Don't get me wrong, this is definitely a huge improvement. I just think that no matter how we refactor this stuff, as long as we roll our own, it is gonna be messy at some point because they are custom scripts.

D4nte:
Can you go in more details @thomas_eizinger? I am not sure I understand your exact expectations.

D4nte:
Thanks Thomas. I think we are moving in the right direction, I'll review tomorrow this idea and see how it could be done.
I'll probably do a second PR as I'd like to remove the `npm_wrapper.sh` file.

api_tests/e2e/rfc003/btc_eth-erc20/test.js
```diff
@@ -187,7 +200,8 @@ describe("RFC003: Bitcoin for ERC20", () => {
         await alice.poll_comit_node_until(
             chai,
             alice_swap_href,
-            "AlphaFunded"
+            "AlphaFunded",
+            true
```
bonomat: ❓What's this `true` doing here?

D4nte: @bonomat it's new, it is to say that you want to generate bitcoin blocks when polling. Look at the new definition of `poll_comit_node_until`.

D4nte:
@thomaseizinger we discussed using cargo-make for the dependencies during the grooming.
The issue is that cargo-make does not provide you a full language so you would either have to:
- get a collection of JS files that you could for each cargo-make task
- revert to shell scripting in each cargo make tasks

We would also face challenges around keeping track of PIDs to kill.
At this stage, the team was happy with the current solution (minus few things I am correct now) and prefer not to migrate to cargo-make.

api_tests/e2e/rfc003/btc_eth-erc20/test.js
D4nte: @LLFourn thinks we keep the ticker. I'll see what I can do.

PR #572, opened 2018-12-14
Bob can decline a swap request

### LINKED ISSUE #381: Implement "decline" action on started state for Bob
- [x] Implement http API call to trigger "decline" action
- [x] Bob sends message to Alice, similar to Accept

Test assertions:
- [x] Bob API returns the swap state as "declined"
- [x] Alice API returns the swap state as "declined"

Child of #405.

luckysori:
Resolves #381.

### DIFF
--- a/api_tests/dry/rfc003/test.js
+++ b/api_tests/dry/rfc003/test.js
@@ -98,7 +98,7 @@ describe("RFC003 HTTP API", () => {
             });
     });
 
-    let swap_url_1;
+    let alice_swap_1_href;
     it("[Alice] Should be able to make first swap request via HTTP api", async () => {
         await chai
             .request(alice.comit_node_url())
@@ -128,11 +128,11 @@ describe("RFC003 HTTP API", () => {
                 res.should.have.status(201);
                 swap_location = res.headers.location;
                 swap_location.should.be.a("string");
-                swap_url_1 = swap_location;
+                alice_swap_1_href = swap_location;
             });
     });
 
-    let swap_url_2;
+    let alice_swap_2_href;
     it("[Alice] Should be able to make second swap request via HTTP api", async () => {
         await chai
             .request(alice.comit_node_url())
@@ -162,14 +162,14 @@ describe("RFC003 HTTP API", () => {
                 res.should.have.status(201);
                 swap_location = res.headers.location;
                 swap_location.should.be.a("string");
-                swap_url_2 = swap_location;
+                alice_swap_2_href = swap_location;
             });
     });
 
     it("[Alice] Is able to GET the swap after POSTing it", async () => {
         await chai
             .request(alice.comit_node_url())
-            .get(swap_url_1)
+            .get(alice_swap_1_href)
             .then(res => {
                 res.should.have.status(200);
 
@@ -206,12 +206,14 @@ describe("RFC003 HTTP API", () => {
                     swap.protocol.should.equal("rfc003");
                     swap.state.should.equal("Start");
                     let links = swap._links;
-                    links.self.href.should.be.oneOf([swap_url_1, swap_url_2]);
+                    links.self.href.should.be.oneOf([alice_swap_1_href, alice_swap_2_href]);
                 }
             });
     });
 
-    let swap_link_href;
+    let bob_swap_1_href;
+    let bob_swap_2_href;
+
     it("[Bob] Shows the swaps as Start in /swaps", async () => {
         let res = await chai.request(bob.comit_node_url()).get("/swaps");
         let embedded = res.body._embedded;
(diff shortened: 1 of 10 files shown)


application/comit_node/src/comit_client/bam.rs
```diff
@@ -88,11 +88,27 @@ impl Client for BamClient {
...
+                        match serde_json::from_value(response.get_body().clone()) {
                             Ok(response) => Ok(Ok(response)),
                             Err(_e) => Err(SwapResponseError::InvalidResponse),
                         }
                     }
+                    Status::SE(20) => {
+                        info!("{} declined swap request: {:?}", socket_addr, response);
+                        Ok(Err({
+                            let reason = response
+                                .get_header("REASON")
```
thomaseizinger: I would appreciate it if we can discuss changes to the COMIT protocol first before they are implemented!
ATM, it is completely undocumented. We should flesh out first, how we want things to be before they are implemented.

Regarding this particular change: I don't see a usecase for custom messages provided by the user. Those are only an attack vector for spamming or possible exploits through buffer overflows in other implementations.

If we want to provide reasons for rejections, I would suggest to discuss and specify a limited set of possible reasons (BAD_RATE for example). Everything else is useless in regards to interoperability and automation which is what we are striving for.

PS: I'll try not to work from now on :)

luckysori: This was just my interpretation of this comment: `// TODO: Implement SwapReject::Decline(reason)`.

But I completely agree with you. I have decided instead to keep the field as an enum, which is something that I documented in one of the commit messages as a future improvement (and what you suggest). For now the enum only has one variant `Unspecified`.

Thank you for the comment, but please do stop working :laughing:

D4nte: "I would appreciate it if we can discuss changes to the COMIT protocol first before they are implemented!" -> you do understand that we may do changes to the protocol while you are away? 😇

api_tests/dry/rfc003/test.js
```diff
@@ -62,7 +62,7 @@ describe("RFC003 HTTP API", () => {
             .post('/swaps/rfc003')
             .send({
                 "alpha_ledger": {
-                    "name": "Thomas' wallet",
```
D4nte: rebase and format :)

application/comit_node/src/comit_client/bam.rs
D4nte: Having written the `TODO` comment and had some verbal discussions with the team I realize that I did not express myself clearly enough.

I was proposing to actually have a range of Error code `Status::SE` that could express defined/common reason of rejections such as:
- `RateDeclined`
- `UnsupportedLedger`
- `InternalError`
- `MissingLiquidity` (this is an idea, not feasible at this stage).

Do we agree that `Status::SE` is to be defined by the application level (ie, COMIT being an application level for BAM).

application/comit_node/src/comit_client/mod.rs
```diff
@@ -31,10 +31,18 @@ pub trait ClientFactory<C: Client>: Send + Sync + RefUnwindSafe + Debug {
...
+    Declined {
+        reason: SwapDeclineReason,
+    },
     /// The counterparty rejected the request
     Rejected,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum SwapDeclineReason {
+    Unspecified,
```
D4nte: Shall we implement few reasons the user can pass in the `POST` request?
Also, is there some existing `internal error` cases that we could clean up?
When do we use `Rejected`? is that internal errors only?
I think under `Rejected` we could also have an enum. Something like:
```
pub enum SwapReject {
    Declined {
        pub enum Reason {
            Liquidity,
            Rate,
            Unspecified,
            AlphaTimeout,
            BetaTimeout,
        }
    },
    Rejected {
        pub enum Reason {
            UnsupportedLedger,
            InternalError,
        }
    }
}
```

D4nte: Maybe worth doing a quick meeting/brainstorm about it.

application/comit_node/src/comit_client/bam.rs
LLFourn: @D4nte The rfc says that `SE20`  and above are reserved for the application.

application/comit_node/src/comit_client/mod.rs
LLFourn: @D4nte I think that idea has potential.

application/comit_node/src/comit_client/bam.rs
D4nte: ok cool, so I guess we can use those for COMIT :)

application/comit_node/src/comit_client/mod.rs
D4nte: #583 opened to track

PR #574, opened 2018-12-14
Add mergify configuration

thomaseizinger:
The initial configuration auto-merges all pull requests with:
- a passing build
- at least one approval
- don't have the label "blocked"
- don't have the label "work-in-progress"

After the PR was merged, the HEAD branch will be deleted.


thomaseizinger:
I think mergify will only be active for branches that already contain the config file, i.e. only the ones that are rebased ontop of current master now.

PR #581, opened 2018-12-17
Test should not be controlling bitcoin blocks gen

D4nte:
Implemented some comments coming from #571.

Spent some time trying to make the bitcoin block generation work in the background but it does not seem to be possible with Mocha.

### DIFF
--- a/api_tests/e2e/rfc003/btc_eth-erc20/test.js
+++ b/api_tests/e2e/rfc003/btc_eth-erc20/test.js
@@ -97,11 +97,7 @@ describe("RFC003: Bitcoin for ERC20", () => {
     });
 
     it("[Alice] Should be in Start state after sending the swap request to Bob", async function() {
-        await alice.poll_comit_node_until(
-            chai,
-            alice_swap_href,
-            "Start",
-        );
+        await alice.poll_comit_node_until(chai, alice_swap_href, "Start");
     });
 
     let bob_swap_href;
@@ -197,12 +193,7 @@ describe("RFC003: Bitcoin for ERC20", () => {
 
     it("[Alice] Should be in AlphaFunded state after executing the funding action", async function() {
         this.timeout(10000);
-        await alice.poll_comit_node_until(
-            chai,
-            alice_swap_href,
-            "AlphaFunded",
-            true
-        );
+        await alice.poll_comit_node_until(chai, alice_swap_href, "AlphaFunded");
     });
 
     let bob_deploy_href;
@@ -420,8 +411,7 @@ describe("RFC003: Bitcoin for ERC20", () => {
         await alice.poll_comit_node_until(
             chai,
             alice_swap_href,
-            "BothRedeemed",
-            true
+            "BothRedeemed"
         );
     });
 
--- a/api_tests/harness.js
+++ b/api_tests/harness.js
@@ -1,6 +1,4 @@
 const execSync = require("child_process").execSync;
-const util = require("util");
-const exec = util.promisify(require("child_process").exec);
 const spawn = require("child_process").spawn;
 const Toml = require("toml");
 const fs = require("fs");
@@ -61,19 +59,12 @@ logger.debug(
 );
 global.harness.ledgers_config = ledgers_config;
 
-let docker_container_names = "";
 let ledger_up_time = 0;
 let ledger_down_time = 0;
 {
     if (config.ledgers) {
-        let docker_containers = [];
-
         config.ledgers.forEach(function(ledger) {
             const config_ledger = ledgers_config[ledger];
-
-            if (config_ledger.docker) {
-                docker_containers.push(config_ledger.docker);
-            }
 
             const up_time = config_ledger.up_time;
             if (up_time && up_time > ledger_up_time) {
@@ -85,7 +76,6 @@ let ledger_down_time = 0;
                 ledger_down_time = down_time;
             }
         });
-        docker_container_names = docker_containers.join(" ");
     }
 }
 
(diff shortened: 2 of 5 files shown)


api_tests/harness.js
```diff
@@ -115,49 +106,63 @@ process.once("SIGINT", function(code) {
...
+            fs.openSync(log_dir + "/docker-compose-err.log", "w"),
+        ],
+    });
+}
+
+async function generateBlock(ledgers, ledgers_config) {
+    if (ledgers) {
+        ledgers.forEach(function(ledger) {
+            if (ledger === "bitcoin") {
+                const block_time_ms = ledgers_config[ledger].block_time_ms;
```
LLFourn: ❓  What's the point of `block_time_ms`. This looks like a dead variable.

api_tests/harness.js
```diff
@@ -177,23 +182,24 @@ async function startLedgerQueryService(name, lqs_config) {
...
     );
-    subprocesses.push(subprocess);
 }
 
 // ********************************** //
 // Start services, run test, shutdown //
 // ********************************** //
 
-describe("Starting services", async function() {
+describe("🛠", async function() {
```
LLFourn: 😄

api_tests/harness.js
```diff
@@ -115,49 +106,63 @@ process.once("SIGINT", function(code) {
...
+            "ignore",
+            fs.openSync(log_dir + "/docker-compose.log", "w"),
+            fs.openSync(log_dir + "/docker-compose-err.log", "w"),
+        ],
+    });
+}
+
+async function generateBlock(ledgers, ledgers_config) {
+    if (ledgers) {
+        ledgers.forEach(function(ledger) {
```
bonomat: 🔧 you can go full functional here, ala this: 
```suggestion
        ledgers
            .filter(ledger => ledger === "bitcoin")
            .filter(ledger => ledgers_config[ledger].block_time_ms)
            .forEach(() => {
                return test_lib.btc_generate(1);
            });
```

bonomat:
Can you use a more speaking branch name in the future please?

application/comit_node/src/http_api/rfc003/swap.rs
```diff
@@ -339,7 +339,10 @@ pub fn get_swaps<T: MetadataStore<SwapId>, S: StateStore<SwapId>>(
             response.with_resources("swaps", swaps);
             Ok(warp::reply::json(&response))
         }
-        Err(e) => Err(warp::reject::custom(HttpApiProblemStdError::new(e))),
+        Err(e) => {
+            error!("returning error: {:?}", e);
```
LLFourn: ❌ Log statements like this in controllers are suboptimal IMO:

1. If it's a 400 type error we may not want to log it as `error`. `error` should be for actual problems with the system (not invalid input from users).
2. If we need to log these kind of errors it will only catch this one route (adding a warp filter to log every 500 error on all routes would catch everything and remove this stuff from the controller).
3. If it is a 500 type error we probably want to log it before it's turned into a `HttpApiProblem`, so closer to the source of the error.

My preferred resolution would be to just remove this for now and have a spike to come up with a sustainable approach to logging. @bonomat @luckysori thoughts?

api_tests/harness.js
```diff
@@ -203,29 +209,40 @@ describe("Starting services", async function() {
...
             });
         }
 
         await test_lib.sleep(2000);
     });
 
-    describe("Running Test Suite", async function() {
+    beforeEach(async function() {
+        // FIXME: use setInterval or equivalent.
+        // Note: could not make it work, Mochas was never stopping
```
LLFourn: change this comment to: "Using `setInterval` to create a block every X seconds made the test harness hang (even after stopping the interval) so we just create a block before every test for now.

api_tests/e2e/rfc003/btc_eth-erc20/test.js
```diff
@@ -97,11 +97,7 @@ describe("RFC003: Bitcoin for ERC20", () => {
     });
 
     it("[Alice] Should be in Start state after sending the swap request to Bob", async function() {
-        await alice.poll_comit_node_until(
-            chai,
-            alice_swap_href,
-            "Start",
-        );
+        await alice.poll_comit_node_until(chai, alice_swap_href, "Start");
```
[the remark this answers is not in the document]
D4nte: No, prettier changed it. Maybe it was missed when I introduced prettier.

D4nte:
@bonomat the branch name was due to the fact that I did not know what I was going to do when I created the branch as we decided how to move forward at the stand-up Monday morning.

api_tests/harness.js
D4nte: yes indeed, fixed.

api_tests/harness.js
D4nte: Is that better? I think it renders ok:

image

application/comit_node/src/http_api/rfc003/swap.rs
D4nte: The issue is that without this log it is a pain to investigate when the error appears as there is no error logged.
The only way to investigate is to re-add the error here and re-run the test hoping that the same error appears.
When it's intermittent, then it becomes a debug nightmare.

What about opening a spike and adding a `// FIXME` here?

LLFourn: Isn't the error your logging exactly the same as what is returned over the http api? Would adding the warp filter that logs 500 errors not give you what you need?

D4nte: FIXME + issue created

api_tests/harness.js
```diff
@@ -115,49 +104,60 @@ process.once("SIGINT", function(code) {
...
+    });
+}
+
+async function generateBlock(ledgers) {
+    if (ledgers) {
+        ledgers
+            .filter(ledger => ledger === "bitcoin")
+            .forEach(() => {
+                return test_lib.btc_generate(1);
+            });
```
LLFourn: ❌ this should just be:

```js
if (ledgers.includes("bitcoin")) {
    await test_lib.btc_generate();
}
```

PR #582, opened 2018-12-17
[fix mergify] Rename status check as per our Travis setup

D4nte:
Mergify currently does not merge because of that.


thomaseizinger:
I think the status check has to be without quotes :)

D4nte:
@thomaseizinger you might be right

PR #586, opened 2018-12-18
Check Secret Length in Bitcoin HTLC

### LINKED ISSUE #416: Validate secret length in Bitcoin HTLC
Currently, Alice could choose a secret longer than 32 bytes and lock ETH with it.
Then, when redeeming BTC, she can use this long secret to redeem BTC (no length check).
However, Bob will not be able to redeem ETH due to length check.

DoD:
- [ ] Update BTC HTLC to include a length check of the secret as part of the redeem branch
- [ ] Add integration test with Bitcoin ledger that verify you cannot use a secret longer or shorter than 32 bytes

Child of #405.

D4nte:
Almost done, just need to fix 2 tests once `bx` is installed.

Resolves #416 

- [x] RTM

### DIFF
--- a/application/comit_node/src/swap_protocols/rfc003/secret.rs
+++ b/application/comit_node/src/swap_protocols/rfc003/secret.rs
@@ -18,19 +18,6 @@ pub struct SecretHash([u8; Self::LENGTH]);
 impl SecretHash {
     pub const LENGTH: usize = 32;
 
-    pub fn from_vec(vec: &[u8]) -> Result<SecretHash, FromErr> {
-        if vec.len() != Self::LENGTH {
-            return Err(FromErr::InvalidLength {
-                expected: Self::LENGTH,
-                got: vec.len(),
-            });
-        }
-        let mut data = [0; Self::LENGTH];
-        let vec = &vec[..Self::LENGTH];
-        data.copy_from_slice(vec);
-        Ok(SecretHash(data))
-    }
-
     pub fn raw(&self) -> &[u8; Self::LENGTH] {
         &self.0
     }
@@ -102,7 +89,16 @@ impl FromStr for SecretHash {
 
     fn from_str(s: &str) -> Result<Self, <Self as FromStr>::Err> {
         let vec = hex::decode(s)?;
-        Self::from_vec(&vec)
+        if vec.len() != Self::LENGTH {
+            return Err(FromErr::InvalidLength {
+                expected: Self::LENGTH,
+                got: vec.len(),
+            });
+        }
+        let mut data = [0; Self::LENGTH];
+        let vec = &vec[..Self::LENGTH];
+        data.copy_from_slice(vec);
+        Ok(SecretHash(data))
     }
 }
 
--- a/application/comit_node/src/swap_protocols/rfc003/secret.rs
+++ b/application/comit_node/src/swap_protocols/rfc003/secret.rs
@@ -6,8 +6,35 @@ use std::{
     str::FromStr,
 };
 
+#[derive(PartialEq, Debug)]
+pub enum FromErr {
+    InvalidLength { expected: usize, got: usize },
+    FromHex(hex::FromHexError),
+}
+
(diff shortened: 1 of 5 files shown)


application/comit_node/src/swap_protocols/rfc003/secret.rs
```diff
@@ -6,8 +6,35 @@ use std::{
...
+}
+
 #[derive(Clone, PartialOrd, Ord, PartialEq, Eq, Hash)]
-pub struct SecretHash(pub Vec<u8>);
+pub struct SecretHash([u8; Self::LENGTH]);
+
+impl SecretHash {
+    pub const LENGTH: usize = 32;
+
+    pub fn from_vec(vec: &[u8]) -> Result<SecretHash, FromErr> {
```
bonomat: ❓Can't you make `vec: &[u8]` already a fixe length array and get rid of the check below?

application/comit_node/src/swap_protocols/rfc003/secret.rs
```diff
@@ -94,24 +128,26 @@ impl From<Secret> for SecretHash {
...
+        let mut secret = [0; Self::LENGTH];
         secret.copy_from_slice(&random_bytes[..]);
         Secret::from(secret)
     }
 
-    pub fn from_vec(vec: &[u8]) -> Result<Secret, SecretFromErr> {
-        if vec.len() != Self::SECRET_LENGTH {
-            return Err(SecretFromErr::InvalidLength {
-                expected: Self::SECRET_LENGTH,
+    pub fn from_vec(vec: &[u8]) -> Result<Secret, FromErr> {
```
bonomat: ❓Same as above: Can't you make `vec: &[u8]` already a fixe length array and get rid of the check below?

application/comit_node/tests/rfc003_bitcoin_htlc.rs
```diff
@@ -3,21 +3,75 @@ use bitcoin_rpc_test_helpers::RegtestHelperClient;
...
+
+    fn from_str(s: &str) -> Result<Self, <Self as FromStr>::Err> {
+        let secret = s.as_bytes().to_vec();
+        Ok(CustomSizeSecret(secret))
+    }
+}
+
+fn fund_htlc<S: SecretTrait>(
     client: &BitcoinCoreClient,
+    secret: &S,
```
LLFourn: 🔧  Just pass in the `SecretHash` and you can remove this trait. You could even remove `CustomSizeSecret` and just have a function which returns the unlock parameters for a given vec.

application/comit_node/tests/rfc003_bitcoin_htlc.rs
```diff
@@ -166,3 +219,92 @@ fn redeem_refund_htlc() {
...
+#[test]
+fn redeem_htlc_with_long_secret() -> Result<(), failure::Error> {
+    let _ = pretty_env_logger::try_init();
+    let docker = Cli::default();
+
+    let container = docker.run(BitcoinCore::default());
+    let client = tc_bitcoincore_client::new(&container);
+    client.generate(432).unwrap().unwrap();
+
+    let secret = CustomSizeSecret::from_str("Grandma, Why have you such a big eyes?")?;
```
bonomat: 🤡
```suggestion
     let secret = CustomSizeSecret::from_str("Grandma, why do have you such big eyes?")?;
```

application/comit_node/src/swap_protocols/rfc003/secret.rs
D4nte: No can do for `Secret` as we will need to extract it from a Bitcoin Script that hold `Vec<u8>`

application/comit_node/tests/rfc003_bitcoin_htlc.rs
D4nte: I still need a `hash` function on `CustomSizeSecret` so I removed the trait but kept the struct.

PR #590, opened 2018-12-19
Bob can actually decline a swap request

### LINKED ISSUE #381: Implement "decline" action on started state for Bob

luckysori:
Note: this is a continuation of PR #572 that was merged too early
because I didn't know how mergify works.

This PR gave me the scope creeps :ghost: 

Finally resolves #381.

- Hold `Response` instead of `Frame` inside `ResponseFrameSource`.
- Implement deserialization of `Frame`'s payload into `Response`.
- Remove `From<Frame> for Response`.
- Return `Result<Start<R>, Error>` from `start_state(&self)`. This
  allows to return the `rfc003::Error` held inside the `Error` state
  which is currently the only state which can't produce the `Start`
  struct.
- Remove `get_swap` macro. There's no need for it since the types
  could already be extracted using the macro `with_swap_types`.


api_tests/dry/rfc003/test.js
```diff
@@ -206,12 +206,17 @@ describe("RFC003 HTTP API", () => {
                     swap.protocol.should.equal("rfc003");
                     swap.state.should.equal("Start");
                     let links = swap._links;
-                    links.self.href.should.be.oneOf([swap_url_1, swap_url_2]);
+                    links.self.href.should.be.oneOf([
```
bonomat: ❓Is it really just `onOf` or `allOf`?

api_tests/dry/rfc003/test.js
```diff
@@ -147,29 +147,29 @@ describe("RFC003 HTTP API", () => {
                 },
                 alpha_asset: {
                     name: alpha_asset_name,
-                    quantity: alpha_asset_quantity,
+                    quantity: "100",
```
bonomat: 🔧What do you think of storing this in a variable so that you can reference it lateron instead of needing to hardcode `100` ?

api_tests/dry/rfc003/test.js
luckysori: We are checking that each swap returned by GET /swaps (in this case 2) matches with one of the swaps sent by Alice. Because we store the swaps using a `HashMap` they're not kept in chronological order, so this is a way to check it.

api_tests/dry/rfc003/test.js
luckysori: Yes, good idea :wink:

application/comit_node/src/http_api/rfc003/swap.rs
```diff
@@ -284,19 +283,18 @@ fn handle_get_swap<T: MetadataStore<SwapId>, S: StateStore<SwapId>>(
...
+
+    with_swap_types!(
         &metadata,
-        state_store,
-        id,
-        state,
         (|| {
-            let state =
-                state.ok_or_else(|| HttpApiProblem::with_title_and_type_from_status(500))?;
+            trace!("Fetched metadata of swap with id {}: {:?}", id, metadata);
```
LLFourn: 🔧  If we want to trace log this let's put it in `MetadataStore::get` and remove this one.

application/comit_node/src/http_api/route_factory.rs
```diff
@@ -23,6 +23,7 @@ pub fn create<T: MetadataStore<SwapId>, S: state_store::StateStore<SwapId>>(
     let rfc003_secret_gen = warp::any().map(move || seed.clone() as Arc<dyn SecretSource>);
     let state_store = warp::any().map(move || state_store.clone());
     let sender = warp::any().map(move || sender.clone());
+    let empty_json_if_missing = warp::any().map(|| json!({}));
```
LLFourn: 🔧 Call it `empty_json_body` (it reads better below).

api_tests/dry/rfc003/test.js
```diff
@@ -259,13 +281,81 @@ describe("RFC003 HTTP API", () => {
...
+        let res = await chai
+            .request(bob.comit_node_url())
+            .get(bob_stingy_swap_href);
+        res.should.have.status(200);
+        res.body.state.should.equal("Rejected");
+    });
+
+    it("[Alice] Should be in the Rejected State after Bob declines a swap request providing a reason", async () => {
+        let res = await chai
+            .request(alice.comit_node_url())
```
LLFourn: ❌ Check the string of the reason is the reason bob `POST`'d

vendor/bam/src/json/frame.rs
```diff
@@ -147,7 +140,10 @@ impl FrameHandler<json::Frame, json::Request, json::Response> for JsonFrameHandl
 
                 debug!("Dispatching response frame {:?} to stored handler.", frame);
 
-                sender.send(frame).unwrap();
+                let response =
+                    serde_json::from_value(frame.payload).expect("should always deserialize");
```
thomaseizinger: 🤔 And what if the other party sends garbage json as the frame payload? That can easily happen and I doubt we want to panic here.

vendor/bam/src/json/response.rs
```diff
@@ -101,3 +113,118 @@ impl IntoFrame<Frame> for Response {
...
+                D: Deserializer<'de>,
+            {
+                struct FieldVisitor;
+
+                impl<'de> Visitor<'de> for FieldVisitor {
+                    type Value = Field;
+
+                    fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                        formatter.write_str(
+                            "`status` or
```
LLFourn: what's with the newline here 🤡

vendor/bam/src/json/frame.rs
LLFourn: Ahhh well spotted. (note it panic'd before this PR too)

thomaseizinger: True, my bad then. I think I wrote the old code 😅
I think it actually was the way we have it now at some point but we changed it to accept a `Frame` for symmetric reasons. Hence also the name, Response*Frame*Source.

vendor/bam/src/json/response.rs
```diff
@@ -1,12 +1,15 @@
...
-use std::collections::HashMap;
+use serde::{
+    de::{self, Deserialize, DeserializeOwned, Deserializer, MapAccess, Visitor},
+    Serialize,
+};
+use serde_json::{self, Map, Value as JsonValue};
+use std::{collections::HashMap, fmt};
 
-#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[derive(Serialize, Debug, PartialEq)]
```
thomaseizinger: Huh? What was wrong with the derived Deserialization?
Please note that the way things worked previously is a total mess altogether and still the early PoC that Lloyd and I put together in like 2 weeks.

The whole BAM crate needs a some love through better modelling and use of the type system. Atm we have too many uses of serde_json::Value that we interpret at runtime.

thomaseizinger: This stupid `deserialization_failed` flag for example is never checked :D 

You (@luckysori) seem to be interested in working on this. Have a go and make it more typesafe and less runtime-error prone :)

luckysori: The problem with the old derived deserialization is that it couldn't work for both compact headers such as
```
headers: {
  "HELLO": "WORLD"
}
```
and full headers like
```
headers: {
  "HELLO": {
    "value": "WORLD"
  }
}
```
Maybe there's an alternative way of doing this without having to define the deserialization in full, but I couldn't figure it out.

PR #595, opened 2018-12-19
BADGES!

thomaseizinger:


README.md
```diff
@@ -1,5 +1,8 @@
 # COMIT-rs
 
+[![Build Status](https://travis-ci.com/comit-network/comit-rs.svg?branch=master)](https://travis-ci.com/comit-network/comit-rs)
+[![Mergify Status](https://gh.mergify.io/badges/comit-network/comit-rs.png?style=small)](https://mergify.io)
```
bonomat: ❓Do we really need a badge for mergify? 
Would you mind removing that one for now? We are still just trying it out and I'm not yet 100% convinced :)

thomaseizinger: The badges only exist because of me 😂
We can always remove it if we stop using it :)

thomaseizinger:
Makes me sad panda :(

Issue #606, opened 2018-12-26
Check-in EVM hex code & remove generation


D4nte:
Motivations:
- Contract ethereum hex should be static and part of RFC003
- Generation of the hex from EVM assembly should be easy but **not** part of the comit-rs crate

DoD:
- [ ] ~One private new repository for all HTLCs in coblox (coblox/blockchain_contracts#1 & coblox/blockchain_contracts#3)~ new vendor crate `blockchain_contracts` #864
- [x] Commit EVM hex code in `comit_node` crate #636 
- [x] remove build.rs from comit-rs repo (3 reviews)
- [x] create new tooling repo with templates to facilitate EVM compilation #864
- [x] Some way to get offsets and full hex for particular HTLC (see table in RFCs) for Ethereum #864
- [ ] Offsets for Bitcoin #910 #932 #933 
- [x] Compare with RFCs (and update the latter one if needed) for Ethereum (done, see comit-network/RFCs#46 & comit-network/RFCs#48)
- [ ] Compare with RFCs for Bitcoin #910
- [x] Make offsets in RFCs: start index inclusive and end index exclusive for Ethereum comit-network/RFCs#66
- [ ] Make offsets in RFCs: start index inclusive and end index exclusive for Bitcoin
- [ ] Tests are running
- [ ] Upgrade to latest solc
- [ ] use fix offsets in comit-rs #636 #933 
- [ ] Move crate to own repo

~~Blocked by coblox/COMIT-RFCs#7~~ (closed)
~~Blocked by comit-network/RFCs#21~~ (closed)

D4nte:
Parity uses `solc`/`solcjs`: https://github.com/paritytech/rust_solc

D4nte:
Can't find Rust lib to compile evm assembly/solidity..

bonomat:
> General question: What is is so bad about the docker requirement? Docker is pretty much a thing that all developers have installed, no? :D

exactly, it's only `pretty much a thing that all developers have installed`. Imho it shouldn't be required if you just want to build our code base. As we saw during your hackathon: not everyone has Docker installed.

thomaseizinger:
The hackathon is a completely different environment.

If you want to build things from source, you expect having to install all kinds of stuff, esp. if you are building a project that compiles for your current architecture (like Rust projects do). Codebases that use languages like Java are different and usually much easier to build.

[...]

PR #617, opened 2019-01-04
Update testcontainers to use parity 2.1.3 across the board

D4nte:
testcontainers was updated to default to parity 2.1.3 and bitcoind 0.17.0.
Updating comit-rs to use latest testcontainers version.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -176,6 +176,8 @@ private = true
 script = [
 '''
 docker pull parity/parity:v1.11.11
+docker pull parity/parity:v2.1.3
+docker pull ethereum/solc:0.4.24
 docker pull coblox/bitcoin-core:0.17.0
 '''
 ]
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -20,7 +20,6 @@ dependencies = [
 workspace = false
 dependencies = [
     "start-sccache",
-    "pull-docker-images",
 ]
 
 # Overridden because: added dependencies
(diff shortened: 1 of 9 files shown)


Makefile.toml
```diff
@@ -182,6 +182,8 @@ private = true
 script = [
 '''
 docker pull parity/parity:v1.11.11
+docker pull parity/parity:v2.1.3
```
bonomat: ❌Instead of pulling both, we should rather get rid of one of them... don't you think so?

D4nte: This is supposed to be a quick fix to get the build stable.

The getting rid of is tracked in the testcontainers-rs issue quoted in the description.

LLFourn: Would this be unnecessary once we do #606?

D4nte: It would still be necessary to pull bitcoin and parity images for the e2e tests.

D4nte:
Blocked on publishing new version of `tc_parity_parity` crate to crates.io

thomaseizinger:
I think we could just remove the whole `cargo-make` task and let the images download on-demand? Feels bad to maintain things like that in two different places.

D4nte:
@bonomat the image would be downloaded in any case and we currently do not look at single test performance.
What is good in downloading the images at first is to fail fast in case of outage of docker.io (which do happen).

thomaseizinger:
But is that enough to justify the duplication? I am not sure.

thomaseizinger:
I did but I am starting to regret it as it seems to be a maintenance burden to keep it in sync with what is actually used.

PR #627, opened 2019-01-08
Specify counterparty address in SWAP request

### LINKED ISSUE #555: Allow COMIT node to talk to more than one COMIT node!
At the moment, we have a setting: `remote_comit_node_url = "127.0.0.1:8001"` (which incidentally isn't a url). 

We should allow the user to send swap requests to any other node by specifying their host and (optionally) port when `POST`ing to /swaps/rfc003.

DoD:
- [x] `remote_comit_node_url` does not exist anymore
- [x] Alice accepts socket address in the http swap request
- [x] Alice can do a swap with Bob and then with Charlie (can be dry test)

Note:
Just keep socket addresses as id for now, will do node pubkey id at later stage.

luckysori:
- Remove `remote_comit_node_url`. Not used anymore, since every swap
  request now includes the address of the node the request is being
  sent to. This means that the SWAP partner is no longer set in stone
  at comit node start-up and a comit node can now trade with multiple
  comit nodes.
- Consequently a `peer` field has been added to
  `SwapRequestBody`. The provided address will be deserialized to a
  socket address even if only an IPv4 address is provided, because it
  will be assumed that the user wants to use the default port value,
  i.e. `9939`.

Resolves #555.

---
- [x] RTM (Ready To Merge)


api_tests/e2e/rfc003/btc_eth-erc20/test.js
```diff
@@ -86,6 +89,8 @@ describe("RFC003: Bitcoin for ERC20", () => {
                 alpha_ledger_refund_identity: null,
                 beta_ledger_redeem_identity: alice_final_address,
                 alpha_ledger_lock_duration: 144,
+                counterparty_ip_address: bob.host,
```
D4nte: What about `peer` instead of `counterparty`? Peer terminology is commonly used in **peer**-to-**peer** protocols. Counterparty, not so much.

I would not separate the host and port, I find it odd.
You could try to deserialize to `SocketAddr` and if it fails try to deserialize to `IpAddr` with default port.

application/comit_node/src/http_api/rfc003/swap.rs
```diff
@@ -61,6 +64,14 @@ pub struct SwapRequestBody<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> {
...
     #[serde(flatten)]
     identities: SwapRequestBodyIdentities<AL::Identity, BL::Identity>,
+    #[serde(rename = "counterparty_ip_address")]
+    bob_ip_address: IpAddr,
+    #[serde(default = "default_port", rename = "counterparty_port")]
+    bob_port: u16,
+}
+
+fn default_port() -> u16 {
+    8001
```
D4nte: ❌ this default port is a big decision. We definitely need to come up with something more original and commonly used than 8001.
Idea hats on!

api_tests/e2e/rfc003/btc_eth-erc20/test.js
D4nte: It'd be good to use [ToSocketAddrs](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html) actually because it gives you DNS resolution.

I don't think custom deserialization is too bad. Let me dig a bit more see if I can find something better.

application/comit_node/src/http_api/rfc003/swap.rs
luckysori: Good idea, but the UTF-8 code for :chains: appears to be [`26d3`](http://www.unicode.org/emoji/charts/full-emoji-list.html#26d3).

api_tests/e2e/rfc003/btc_eth-erc20/test.js
D4nte: I think a quick an easy custom derive is fine.

application/comit_node/src/http_api/rfc003/swap.rs
D4nte: Yes, but ports are decimals hence `9939`

api_tests/dry/rfc003/test.js
```diff
@@ -374,4 +375,77 @@ describe("RFC003 HTTP API", () => {
...
+    it("[Charlie] Shows the Swap as Start in /swaps", async () => {
+        let res = await chai.request(charlie.comit_node_url()).get("/swaps");
+
+        let embedded = res.body._embedded;
+        let swap_embedded = embedded.swaps[0];
+        swap_embedded.protocol.should.equal("rfc003");
+        swap_embedded.state.should.equal("Start");
+        let swap_link = swap_embedded._links;
+        swap_link.should.be.a("object");
+        let charlie_swap_href = swap_link.self.href;
```
bonomat: 🙃 
```suggestion
        let swap_href = swap_link.self.href;
```

api_tests/dry/rfc003/test.js
```diff
@@ -374,4 +375,77 @@ describe("RFC003 HTTP API", () => {
...
+        charlie_swap_href.should.be.a("string");
```
bonomat: 🙃 
```suggestion
        swap_href.should.be.a("string");
```

api_tests/dry/rfc003/test.js
```diff
@@ -374,4 +375,77 @@ describe("RFC003 HTTP API", () => {
...
+
+    it("[Alice] Should see both Bob and Charlie in her list of peers after sending a swap request to both of them", async () => {
+        await chai
+            .request(alice.comit_node_url())
+            .get("/peers")
+            .then(res => {
+                res.should.have.status(200);
+
+                let peers = res.body.peers;
+                for (peer of peers) {
```
bonomat: ❌ I see two issues here: 
1) if `peers` is empty, this test passes
2) if `peers` contains `charlie_comit_node_address` or `bob_comit_node_address` twice, this test passes.

This should be fixed please.

api_tests/dry/rfc003/test.js
```diff
@@ -79,6 +84,7 @@ describe("RFC003 HTTP API", () => {
                 alpha_ledger_refund_identity: "",
                 beta_ledger_redeem_identity: "",
                 alpha_ledger_lock_duration: 0,
+                peer_address: "0.0.0.0",
```
LLFourn: Would prefer this be called just `peer` because I think that's what it will be called in the future once we have multiple ways of describing a peer.

api_tests/dry/rfc003/test.js
```diff
@@ -23,8 +23,13 @@ const alpha_ledger_lock_duration = 144;
 
 const alice = test_lib.comit_conf("alice", {});
 const bob = test_lib.comit_conf("bob", {});
+const charlie = test_lib.comit_conf("charlie", {});
 
 const alice_final_address = "0x00a329c0648769a73afac7f9381e08fb43dbea72";
+const bob_comit_node_address =
+    bob.host + ":" + bob.config.comit.comit_listen.split(":")[1];
+const charlie_comit_node_address =
+    charlie.host + ":" + charlie.config.comit.comit_listen.split(":")[1];
```
LLFourn: 🔧 Make a method on charlie to get host and port from the config

api_tests/dry/rfc003/test.js
D4nte: Yes agreed, I already see this field accepting formats like `"pubkey@address"` or even just `"pubkey"`

api_tests/dry/rfc003/test.js
D4nte: 🔧 can't use just do that:
```suggestion
    charlie.config.comit.comit_listen;
```

api_tests/e2e/rfc003/eth-erc20_btc/test.js
```diff
@@ -312,9 +319,9 @@ describe("RFC003: ERC20 for Bitcoin", () => {
             .request(alice.comit_node_url())
             .get(
                 alice_redeem_href +
-                "?address=" +
-                alice_final_address +
-                "&fee_per_byte=20"
+                    "?address=" +
```
bonomat: ❓did you format this manually ?

luckysori: Nope, that's just `prettier` :nail_care: .

PR #632, opened 2019-01-10
Format all JavaScript files

### LINKED ISSUE #628: cargo make `js-format` task does not work in Travis
The task checks `git diff` files (example taken from prettier website).

In Travis, there is no git changes! Hence the `js-format` checks nothing.

Need to implement another way to list the JavaScript files to check (`find` ?).

DoD:
- [ ] All JS files are format checked (or formatted(
- [ ] Works on Mac Os and Linux (if you use `find`)

D4nte:
Use find to get all JavaScript files to format/check-format.

Resolves #628.

---
- [x] RTM (Ready To Merge)

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -111,26 +111,20 @@ command = "cargo"
 args = ["fmt", "--", "--check"]
 
 [tasks.js-format]
-description = "Runs prettier to check appropriate JS code format."
+description = "Runs prettier to format JavaScript code."
 workspace = false
 script = [
 '''
-jsfiles=$(git diff --no-ext-diff --name-only --diff-filter=ACM "*.js" "*.jsx" | tr '\n' ' ')
-[ -z "$jsfiles" ] && exit 0
-
-echo "$jsfiles" | xargs ./api_tests/node_modules/.bin/prettier --write
+find . -not -path '*/node_modules/*' -name '*.js' | xargs ./api_tests/node_modules/.bin/prettier --write
 '''
 ]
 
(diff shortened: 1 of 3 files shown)


Makefile.toml
```diff
@@ -115,10 +115,7 @@ description = "Runs prettier to check appropriate JS code format."
 workspace = false
 script = [
 '''
-jsfiles=$(git diff --no-ext-diff --name-only --diff-filter=ACM "*.js" "*.jsx" | tr '\n' ' ')
-[ -z "$jsfiles" ] && exit 0
-
-echo "$jsfiles" | xargs ./api_tests/node_modules/.bin/prettier --write
+find . -not -path '*/node_modules/*' -name '*.js' | xargs ./api_tests/node_modules/.bin/prettier --write
```
bonomat: ❓Should we run this only on the JS files in `api_tests`?

D4nte: I think it should be done for all JS files, even if we decide to have more JavaScript in the codebase.

I also wanted to do it for the Json files we have but @LLFourn was against it.

Makefile.toml
```diff
@@ -115,10 +115,7 @@ description = "Runs prettier to check appropriate JS code format."
 workspace = false
```
luckysori: :upside_down_face: I can't comment on the line, because it's too far away from the changes, but I think this description should be different to the one below since this command actually modifies the files, i.e. "Runs prettier to format JS code appropriately".

D4nte:
@luckysori approve then!

PR #635, opened 2019-01-10
Make `SecretHash` `Copy`

### LINKED ISSUE #440: Make rfc003::SecretHash implement Copy
Currently, the `SecretHash` implementation for `rfc003` is not Copy because it is based on a `Vec`. However, we know that it is always 32 bytes since it is based on `SHA256`.

We should refactor it to a be represented as `[u8; 32]` and make it implement `Copy`.

Note:
It is an array but just need to derive `Copy`.

DoD:
- [ ] derive `Copy` for `SecretHash`

D4nte:
Resolves #440

---
- [x] RTM

### DIFF
--- a/application/comit_node/src/swap_protocols/rfc003/bitcoin/htlc.rs
+++ b/application/comit_node/src/swap_protocols/rfc003/bitcoin/htlc.rs
@@ -79,7 +79,7 @@ impl Htlc {
         if *expected_secret_hash != got_secret_hash {
             return Err(UnlockingError::WrongSecret {
                 got: got_secret_hash,
-                expected: expected_secret_hash.clone(),
+                expected: *expected_secret_hash,
             });
         }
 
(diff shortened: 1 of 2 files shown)


Makefile.toml
```diff
@@ -204,6 +205,21 @@ private = true
 command = "sccache"
 args = ["--stop-server"]
 
+[tasks.check-evm]
```
LLFourn: So what's all this then?

LLFourn:
@D4nte I think some unrelated changes ended up in this PR XD

Makefile.toml
```diff
@@ -204,6 +205,21 @@ private = true
...
+# Ensure that when Travis compiled the EVM assembly
+# the hex result is in line with what was checked out
+description = "Ensure the EVM contract have not changed"
+workspace = false
+private = true
+script = [
+'''
+changes=$(git status -s |wc -l)
```
bonomat: does this work on Travis?

Makefile.toml
D4nte: oopsie daysie

Makefile.toml
D4nte: Yes, but it's in the wrong PR :)

Issue #646, opened 2019-01-15
Cleanup Bitcoin LQS


bonomat:
With the integration of the new query filter for LQS in #412 we split up the LQS in Ethereum and Bitcoin  related parts. 
Bitcoin hasn't been touched and still contains unnecessary generic types. This ticket is about cleaning up this mess.

Issue #563, opened 2018-12-12
Use new LQS Parameter to retrieve ETH/ERC20 Refund/Redeem tx


D4nte:
Make `comit_node` use the new `log` filter introduced with #412 to retrieve the refund/redeem tx.

Blocked by #412.
Blocked by #476.

Issue #476, opened 2018-11-23
Validation of ERC20 funding transaction


D4nte:
To validate ERC20 funding transaction, we must verify that the contract call emitted a `Transfer` event on the token contract.

DoD:
Update the LQS query to include filtering on `Transfer()` event emitted by the token contract.

Child of #399.
Needs #412 (tracks adding of log parameter to LQS)

bonomat:
Implementation hint: 
* retrieve the transaction receipt for each transaction
* iterate over all logs
* check if `address` field is the erc20 token contract
* verify the `topics` and `data` field

example logs from redeeming HTLC and transferring the tokens. the second log entry is the ERC20 token transfer
```
[ { address: '0xe46FB33e4DB653De84cB0E0E8b810A6c4cD39d59',
    blockHash:
     '0x2822c1a8fcbdd4e1421fa8007216750903dcbd01ab1dde613775ec53805f3b9d',
    blockNumber: 6,
    data: '0x',
    logIndex: 0,
    removed: false,
    topics:
     [ '0xb8cac300e37f03ad332e581dea21b2f0b84eaaadc184a295fef71e81f44a7413' ],
    transactionHash:
     '0xdfd278be493c83828b8845fe292b0736e7e11d8fc2194563472ccd42d56deac4',
    transactionIndex: 0,
    transactionLogIndex: '0x0',
    type: 'mined',
    id: 'log_f28bd087' },
  { address: '0xb4c79daB8f259C7Aee6E5b2Aa729821864227e84',
    blockHash:
     '0x2822c1a8fcbdd4e1421fa8007216750903dcbd01ab1dde613775ec53805f3b9d',
    blockNumber: 6,
    data:
     '0x0000000000000000000000000000000000000000000000000000000000000190',
    logIndex: 1,
    removed: false,
    topics:
     [ '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef',
       '0x000000000000000000000000e46fb33e4db653de84cb0e0e8b810a6c4cd39d59',
       '0x000000000000000000000000d51ecee7414c4445534f74208538683702cbb3e4' ],
    transactionHash:
     '0xdfd278be493c83828b8845fe292b0736e7e11d8fc2194563472ccd42d56deac4',
    transactionIndex: 0,
    transactionLogIndex: '0x1',
    type: 'mined',
    id: 'log_24e7a7b0' } ]
```

PR #647, opened 2019-01-15
Allow filtering for transactions by log in Ethereum

### LINKED ISSUE #412: Allow filtering of "log" in Ethereum blocks
Update ledger query service to add a new query parameter: `log`.

Note:
Log can be retrieved from the `TransactionReceipt`.
Matching is currently only done for transaction, see if transaction receipt can be inlined in transaction to keep code generic.
Decide on an efficient design so that receipt is only retrieved if needed.
1. Avoid multiple http requests for the same receipt
2. Avoid retrieve receipt if not needed. Apply filters on transaction and only if it matches, retrieve receipt and apply filters on receipt

DoD:
- [ ] Add `log` filter to LQS to filter transaction with specific log
- [ ] Only add unit test of `matches` method

Child of #405.

bonomat:
With this PR we provide a new feature for the Ethereum LQS: We can now filter by logs which are emitted by a transaction, e.g. we can filter for the `Redeemed` log of our HTLC

This PR also splits up the LQS into Bitcoin and Ethereum, however, we only focused on implementing the new feature for Ethereum and cleaned up this part at the same time. 

Note: there is no need to review anything inside: `ledger_query_service/src/bitcoin/*`. These files are mostly unchanged (minor adaptions were necessary though) and will be updated with: #646 

Resolves #412


application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -26,6 +27,89 @@ pub struct EthereumTransactionQuery {
     transaction_data_length: Option<usize>,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct EthereumTransactionLogQuery {
```
thomaseizinger: :question: Why is there a need for a different `Query`? Can this not be part of `/transactions` or `/blocks`?

application/ledger_query_service/src/ethereum/block_processor.rs
```diff
@@ -0,0 +1,183 @@
...
+    ))
+}
+
+fn process_block_queries(
+    block_queries: ArcQueryRepository<EthereumBlockQuery>,
+    block: &Block<Transaction>,
+) -> BlockQueryResults {
+    trace!("Processing {:?}", block);
+
+    let block_hash;
```
thomaseizinger: :wrench: `match` is an expression. You can just assign the result to the variable and don't have to declare separately.

application/ledger_query_service/src/ethereum/block_processor.rs
```diff
@@ -0,0 +1,183 @@
...
+    };
+
+    Ok((
+        block_query_results,
+        transaction_query_results,
+        transaction_log_query_results,
+    ))
+}
+
+fn process_block_queries(
```
thomaseizinger: :wrench: Since it is not a trait, it might be worth to return `impl Iterator` and don't `collect` because you are gonna iterate over it again anyway so you can save some performance and intermediary allocations here.

thomaseizinger: Applies to all three functions.

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -26,6 +27,89 @@ pub struct EthereumTransactionQuery {
...
+    logs: Vec<Vec<H256>>,
```
thomaseizinger: :thinking: Should we provide a more user-friendly way here? Like, just `String`s and convert it to the hash ourselves? Might be easier to consume.

Also, these things are called `Topics` in Ethereum if I am not mistaken. I think it would be good to stick to the terminology.

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -26,6 +27,89 @@ pub struct EthereumTransactionQuery {
...
+}
+
+impl EthereumTransactionLogQuery {
+    pub fn matches_block(&self, block: &EthereumBlock<EthereumTransaction>) -> bool {
```
thomaseizinger: :question: Is this used anywhere? Why do we need two ways to check for logs?

api_tests/ledger_query_service/test.js
```diff
@@ -2,23 +2,40 @@ const chai = require("chai");
...
+const actor = require("../lib/actor.js");
+const ethereum = require("../lib/ethereum.js");
 const wallet = require("../lib/wallet.js");
 const lqs_conf = require("../lib/lqs.js");
 
 const bitcoin_rpc_client = bitcoin.create_client();
 const lqs = lqs_conf.create("localhost", 8080);
 const toby_wallet = wallet.create("toby");
 
+const alice = actor.create("alice", {});
```
D4nte:
```suggestion
const alice = actor.create("alice");
```

api_tests/ledger_query_service/config.toml
```diff
@@ -1,5 +1,9 @@
 ledgers = ["bitcoin", "ethereum"]
 
+[comit_node.alice]
+host = "127.0.0.1"
```
D4nte: ❌ this should not be needed if no comit_node is started

api_tests/ledger_query_service/test.js
D4nte: 🔧 why not just using a wallet? const alice_wallet = wallet.create("alice");

application/ledger_query_service/src/ethereum/queries.rs
luckysori: I'd like to preserve the safety of only accepting valid hashes. How can I do that if I allow any string?

application/ledger_query_service/src/ethereum/queries.rs
luckysori: This query is inherently different to the old transaction query, because it only looks at the logs of the transaction receipt.

luckysori: This checks if a block contains a transaction which has emitted the topics included in the query. It allows for skipping blocks that don't need to be looked at.

application/ledger_query_service/src/ethereum/queries.rs
luckysori: This query only looks at transaction receipt logs and allows for skipping blocks which don't contain transactions emitting said logs, so the implementation is very different.

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -18,80 +19,36 @@ use futures::{
...
 
-fn clean_0x(s: &str) -> &str {
-    if s.starts_with("0x") {
-        &s[2..]
-    } else {
-        s
-    }
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct ReceiptQuery {
+    topics: Vec<Topics>,
```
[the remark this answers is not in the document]
bonomat: If you want to filter for the redemption of our ERC20 HTLC and to ensure that the tokens have been transferred this should look like this: 
`topics: [[sha256(Redeemed)], [sha256(Transferred)]]`
You would want these to be in two separate logs but in the same transaction. 

Note: this will evolve in #563

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -123,81 +80,183 @@ impl Query<EthereumTransaction> for EthereumTransactionQuery {
...
+        match self {
+            Self { topics } if topics.is_empty() => false,
+            Self { topics } => topics.iter().all(|topics| {
+                !topics.is_empty()
+                    && transaction_receipt
+                        .logs
+                        .iter()
+                        .any(|tx_log| topics.iter().all(|topic| tx_log.topics.contains(topic)))
+            }),
+        }
```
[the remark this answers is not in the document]
bonomat: If it matches `any` topic.

Regarding contract address: you are right. As mentioned in regards of #476, we can/will have to introduce this check. 
I would suggest to get this PR done and continue with it in the other one

application/ledger_query_service/src/ethereum/queries.rs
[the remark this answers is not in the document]
LLFourn: > You would want these to be in two separate logs but in the same transaction.
I got confused by this because this doesn't seem to be what the code below does.
I read https://codeburst.io/deep-dive-into-ethereum-logs-a8d2047c7371 which has cleared up quite a few things on this topic.

From what I understood, this is actually what we want:

```rust
struct EventMatcher {
    contract_address: Option<Address>,
    topics: Vec<Option<Topic>>
}
struct ReceiptQuery {
      event_matchers : Vec<EventMatcher>
}
```

Then when you match a particular log you do something like:

```rust
if let Some(address) = event_matcher.address {
      if log.address != address {
            return false
      }
}
return event_matcher.topics.iter().enumerate().all(|i, topic| {
      match topic {
            Some(topic)  =>  log.topics[i] == topic,
             None => true
      }
})
```

application/ledger_query_service/src/bitcoin/block_processor.rs
```diff
@@ -301,7 +269,8 @@ mod tests {
                 Box::new(futures::future::ok(QueryMatchResult::no()))
             }
         }
-
+    }
+    impl IsEmpty for GenericBlockQuery {
         fn is_empty(&self) -> bool {
             false
```
thomaseizinger: Lol. :joy:

application/ledger_query_service/src/ethereum/queries.rs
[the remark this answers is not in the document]
LLFourn: Note, this actually says every row of topics must be not empty and be a subset of a log.

PR #675, opened 2019-01-17
Update README + Fix mergify corner case

D4nte:
Update README after compiling comit-rs in docker.
Also fix a mergify corner case where mergify would fail if there is no "review" label to remove.

No codecov as it looks like tarpaulin cannot compile with C-bindings (and codecov is not critical so we do not want to spend more time on this).

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -12,10 +12,18 @@ pull_request_rules:
         strict: smart
         strict_method: rebase
       delete_head_branch:
-  - name: auto add wip
+  - name: auto add wip (and remove review)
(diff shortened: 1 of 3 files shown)


D4nte:
@luckysori @thomaseizinger could you try locally? Looks like it's not possible, getting:
```
/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/zmq-rs-0.1.8/src/lib.rs:527: undefined reference to `zmq_has'
          collect2: error: ld returned 1 exit status
```

PR #678, opened 2019-01-20
Use sscache git until crate is fixed

D4nte:
Use sccache git master until mozilla/sccache#355 gets fixed


thomaseizinger:
Is there a way we can download a specific commit hash? What if the break their master ...

PR #682, opened 2019-01-21
Bitcoin LQS cleanup

### LINKED ISSUE #646: Cleanup Bitcoin LQS

luckysori:
With #647, the ledger query service has been split up into two, since the implementations for bitcoin and ethereum cannot be easily generalised anymore. This PR removes all the unnecessary traits that were still hanging around in `ledger_query_service/src/bitcoin`.

Additionally, `PendingTransaction` has been removed, because the feature was not being used and was poorly implemented to begin with.

Resolves #646.

### DIFF
--- a/application/ledger_query_service/src/bin/ledger_query_service.rs
+++ b/application/ledger_query_service/src/bin/ledger_query_service.rs
@@ -11,14 +11,10 @@ use ethereum_support::web3::{
 };
 use futures::stream::Stream;
 use ledger_query_service::{
-    bitcoin::{
-        block_processor::DefaultBlockProcessor as BitcoinBlockProcessor,
-        queries::{BitcoinBlockQuery, BitcoinTransactionQuery},
-    },
+    bitcoin::{self, bitcoind_zmq_listener::bitcoin_block_listener},
     ethereum::{self, ethereum_web3_block_poller::ethereum_block_listener},
     settings::{self, Settings},
-    BlockProcessor, InMemoryQueryRepository, InMemoryQueryResultRepository, QueryResultRepository,
-    RouteFactory,
+    InMemoryQueryRepository, InMemoryQueryResultRepository, QueryResultRepository, RouteFactory,
 };
 use std::{env::var, sync::Arc};
 use tokio::runtime::Runtime;
@@ -50,11 +46,13 @@ fn create_bitcoin_routes(
     route_factory: &RouteFactory,
     settings: settings::Bitcoin,
 ) -> BoxedFilter<(impl Reply,)> {
+    let block_query_repository =
+        Arc::new(InMemoryQueryRepository::<bitcoin::BlockQuery>::default());
     let transaction_query_repository =
-        Arc::new(InMemoryQueryRepository::<BitcoinTransactionQuery>::default());
-    let block_query_repository = Arc::new(InMemoryQueryRepository::<BitcoinBlockQuery>::default());
-    let transaction_query_result_repository = Arc::new(InMemoryQueryResultRepository::default());
+        Arc::new(InMemoryQueryRepository::<bitcoin::TransactionQuery>::default());
+
     let block_query_result_repository = Arc::new(InMemoryQueryResultRepository::default());
+    let transaction_query_result_repository = Arc::new(InMemoryQueryResultRepository::default());
 
     let bitcoin_rpc_client = bitcoin_rpc_client::BitcoinCoreClient::new(
         settings.node_url.as_str(),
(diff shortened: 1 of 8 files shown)


application/ledger_query_service/src/lib.rs
```diff
@@ -23,47 +23,27 @@ pub use crate::{
...
-    pub fn yes_with_confirmations(confirmations_needed: u32) -> Self {
-        QueryMatchResult::Yes {
-            confirmations_needed,
-        }
-    }
-    pub fn no() -> Self {
-        QueryMatchResult::No
+impl PartialOrd for QueryMatch {
+    fn partial_cmp(&self, other: &QueryMatch) -> Option<Ordering> {
+        Some((self.0).0.cmp(&(other.0).0))
```
thomaseizinger: :wrench: Maybe we can derive `PartialOrd` on `QueryId`? Then we don't have to do this nesting here (so much).

application/ledger_query_service/src/bitcoin/block_processor.rs
```diff
@@ -1,440 +1,70 @@
...
-        pending_transactions: Arc<Mutex<Vec<PendingTransaction>>>,
-        transaction: &T,
-    ) -> impl Future<Item = Vec<QueryMatch>, Error = ()> + Send {
+    stream::futures_ordered(result_futures).filter_map(|x| x)
+}
+
+pub fn check_transaction_queries(
+    transaction_queries: ArcQueryRepository<TransactionQuery>,
+    block: Block,
+) -> impl Stream<Item = QueryMatch, Error = ()> + Send {
```
LLFourn: ❌ I thought the main point of this PR was to remove all this futures stuff from the bitcoin side of things?
`query.matches` should just return the result and these should just go into a vector that's returned.

Apologies if I missed some discussion.

application/ledger_query_service/src/bitcoin/queries.rs
```diff
@@ -1,47 +1,44 @@
...
-    type Item = BitcoinTransaction;
+    type Item = Transaction;
 
     fn expand_result(
         result: &QueryResult,
         client: Arc<BitcoinCoreClient>,
-    ) -> Result<Vec<BitcoinTransaction>, Error> {
-        let mut expanded_result: Vec<BitcoinTransaction> = Vec::new();
+    ) -> Result<Vec<Transaction>, Error> {
+        let mut expanded_result: Vec<Transaction> = Vec::new();
```
LLFourn: 🤡 This type annotation wouldn't be needed if we wrote this as a map.

application/ledger_query_service/src/bitcoin/queries.rs
```diff
@@ -90,61 +82,39 @@ impl Query<BitcoinTransaction> for BitcoinTransactionQuery {
                     };
 
                 if result {
-                    Box::new(futures::future::ok(
-                        QueryMatchResult::yes_with_confirmations(*confirmations_needed),
-                    ))
+                    Box::new(futures::future::ok(true))
                 } else {
-                    Box::new(futures::future::ok(QueryMatchResult::no()))
+                    Box::new(futures::future::ok(false))
```
LLFourn: ❌ this shouldn't return futures anymore.

application/ledger_query_service/src/bitcoin/queries.rs
luckysori: It's actually not needed at all - compiles without it.

application/ledger_query_service/src/bin/ledger_query_service.rs
```diff
@@ -135,13 +132,13 @@ fn create_ethereum_routes(
...
-        let log_query_result_repository = log_query_result_repository.clone();
-
         let block_query_repository = block_query_repository.clone();
-        let log_query_repository = log_query_repository.clone();
         let transaction_query_repository = transaction_query_repository.clone();
+        let log_query_repository = log_query_repository.clone();
+
+        let block_query_result_repository = block_query_result_repository.clone();
+        let transaction_query_result_repository = transaction_query_result_repository.clone();
+        let log_query_result_repository = log_query_result_repository.clone();
```
bonomat: ❓Why randomly resorting them?

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -406,24 +381,6 @@ mod tests {
...
-            topics: vec![vec![]],
-        };
-        let log1 = log(vec![redeem_log_msg]);
-        let receipt1 = transaction_receipt(vec![log1]);
-        assert_that!(query1.matches_transaction_receipt(receipt1)).is_false();
-
-        let query2 = LogQuery { topics: vec![] };
-        let log2 = log(vec![redeem_log_msg]);
-        let receipt2 = transaction_receipt(vec![log2]);
-        assert_that!(query2.matches_transaction_receipt(receipt2)).is_false()
```
bonomat: ❌Why did you remove this test? Shouldn't this still apply, even if you remove the empty check?

application/ledger_query_service/src/bin/ledger_query_service.rs
luckysori: These two blocks were sorted differently before the change, now they're sorted in the same order. Only did it for consistency.

luckysori: Also the query repositories should naturally go before the query result repositories :laughing:

application/ledger_query_service/src/ethereum/queries.rs
luckysori: Maybe I'm understanding it wrong, but if we remove the empty check then empty queries are allowed and they would match everything, so this test would not pass.

bonomat: ah, wasn't aware that this is what we want

bonomat:
I know you did some changes in ethereum LQS as well. Would you mind undoing this? In the phase of #476 I'm changing stuff big times in here and we can avoid annoying merge conflicts:)

bonomat:
> > I know you did some changes in ethereum LQS as well. Would you mind undoing this? In the phase of #476 I'm changing stuff big times in here and we can avoid annoying merge conflicts:)
> 
> I don't think I can. The changes on the ethereum side come from removing the unnecessary `QueryMatchResult`, which is something that should be removed as part of this cleanup.

Ok, then let me merge first :P

PR #684, opened 2019-01-21
Verify erc20 funding using LQS

### LINKED ISSUE #476: Validation of ERC20 funding transaction

bonomat:
Add a new filter to verify erc20 funding.

Fixes #476

### DIFF
--- a/api_tests/ledger_query_service/test.js
+++ b/api_tests/ledger_query_service/test.js
@@ -380,7 +380,10 @@ describe("Test Ledger Query Service API", () => {
                     .request(lqs.url())
                     .post("/queries/ethereum/logs")
                     .send({
-                        topics: [[transfer_topic]],
+                        log_matchers: [{
+                            address: token_contract_address,
+                            topics: [transfer_topic],
+                        }],
                     })
                     .then(res => {
                         res.should.have.status(201);
@@ -396,8 +399,10 @@ describe("Test Ledger Query Service API", () => {
                     .get("")
                     .then(res => {
                         res.should.have.status(200);
-                        res.body.query.topics.should.deep.equal([
-                            [transfer_topic],
+                        res.body.query.log_matchers.should.have.length(1);
+                        res.body.query.log_matchers[0].address.toLowerCase().should.equal(token_contract_address.toLowerCase());
+                        res.body.query.log_matchers[0].topics.should.deep.equal([
+                            transfer_topic,
                         ]);
                         res.body.matches.should.be.empty;
                     });
@@ -419,8 +424,10 @@ describe("Test Ledger Query Service API", () => {
                         return lqs
                             .poll_until_matches(chai, location)
                             .then(body => {
-                                body.query.topics.should.deep.equal([
-                                    [transfer_topic],
+                                body.query.log_matchers.should.have.length(1);
+                                body.query.log_matchers[0].address.toLowerCase().should.equal(token_contract_address.toLowerCase());
+                                body.query.log_matchers[0].topics.should.deep.equal([
+                                    transfer_topic,
                                 ]);
                                 body.matches.should.have.lengthOf(1);
                                 let query_transaction_hash =
--- a/api_tests/ledger_query_service/test.js
+++ b/api_tests/ledger_query_service/test.js
@@ -389,7 +389,8 @@ describe("Test Ledger Query Service API", () => {
                         event_matchers: [
                             {
                                 address: token_contract_address,
-                                data: "0x0000000000000000000000000000000000000000000000000000000000000001",
+                                data:
+                                    "0x0000000000000000000000000000000000000000000000000000000000000001",
                                 topics: [
                                     transfer_topic,
                                     from_address,
--- a/api_tests/ledger_query_service/test.js
+++ b/api_tests/ledger_query_service/test.js
@@ -382,6 +382,7 @@ describe("Test Ledger Query Service API", () => {
                     .send({
                         log_matchers: [{
                             address: token_contract_address,
+                            data: null,
                             topics: [transfer_topic],
                         }],
                     })
--- a/api_tests/ledger_query_service/test.js
+++ b/api_tests/ledger_query_service/test.js
@@ -373,6 +373,12 @@ describe("Test Ledger Query Service API", () => {
             // keccak('Transfer(address,address,uint256)')
             const transfer_topic =
                 "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef";
+            const from_address =
+                "0x000000000000000000000000" +
+                alice_wallet_address.replace("0x", "");
+            const to_address =
+                "0x00000000000000000000000005cbb3fdb5060e04e33ea89c6029d7c79199b4cd";
+
             let location;
             it("LQS should respond with location when creating a valid transaction receipt query", async function() {
                 this.timeout(1000);
@@ -380,11 +386,17 @@ describe("Test Ledger Query Service API", () => {
                     .request(lqs.url())
                     .post("/queries/ethereum/logs")
                     .send({
-                        log_matchers: [{
-                            address: token_contract_address,
-                            data: null,
-                            topics: [transfer_topic],
-                        }],
+                        event_matchers: [
+                            {
+                                address: token_contract_address,
+                                data: "0x0000000000000000000000000000000000000000000000000000000000000001",
+                                topics: [
+                                    transfer_topic,
+                                    from_address,
+                                    to_address,
+                                ],
+                            },
+                        ],
                     })
                     .then(res => {
                         res.should.have.status(201);
@@ -400,20 +412,26 @@ describe("Test Ledger Query Service API", () => {
                     .get("")
                     .then(res => {
                         res.should.have.status(200);
-                        res.body.query.log_matchers.should.have.length(1);
-                        res.body.query.log_matchers[0].address.toLowerCase().should.equal(token_contract_address.toLowerCase());
-                        res.body.query.log_matchers[0].topics.should.deep.equal([
-                            transfer_topic,
-                        ]);
+                        res.body.query.event_matchers.should.have.length(1);
+                        res.body.query.event_matchers[0].address
+                            .toLowerCase()
+                            .should.equal(token_contract_address.toLowerCase());
+                        res.body.query.event_matchers[0].topics.every(topic => {
+                            [transfer_topic, from_address, to_address].includes(
+                                topic.toLowerCase()
+                            );
+                        });
                         res.body.matches.should.be.empty;
                     });
             });
 
             it("LQS should respond with transaction receipt match when requesting on the transfer_topic query after waiting 3 seconds", async function() {
                 this.slow(2000);
-
+                this.timeout(20000);
                 const transfer_token_data =
-                    "0xa9059cbb0000000000000000000000005cbb3fdb5060e04e33ea89c6029d7c79199b4cd90000000000000000000000000000000000000000000000000000000000000001";
+                    "0xa9059cbb" +
+                    to_address.replace("0x", "") +
+                    "0000000000000000000000000000000000000000000000000000000000000001";
                 return alice_wallet
                     .eth()
                     .send_eth_transaction_to(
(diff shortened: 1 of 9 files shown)


application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -148,45 +179,83 @@ impl IsEmpty for TransactionQuery {
...
+                            result = result && (data == &tx_log.data)
+                        }
+
+                        if tx_log.topics.len() == topics.len() {
+                            tx_log
+                                .topics
+                                .iter()
+                                .enumerate()
+                                .for_each(|(index, tx_topic)| {
+                                    let topic = topics.get(index);
```
LLFourn: Since you checked the length is equal, why do `.get(index)` rather than `topics[index]`?

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -148,45 +179,83 @@ impl IsEmpty for TransactionQuery {
...
+                    } if topics.is_empty() => false,
+                    EventMatcher {
+                        address: address,
+                        data: data,
+                        topics: topics,
+                    } => transaction_receipt.logs.iter().any(|tx_log| {
+                        let mut result = true;
+
+                        if let Some(address) = address {
+                            result = result && (address == &tx_log.address);
```
LLFourn: Early return would be cleaner IMO

api_tests/ledger_query_service/test.js
```diff
@@ -380,7 +380,11 @@ describe("Test Ledger Query Service API", () => {
                     .request(lqs.url())
                     .post("/queries/ethereum/logs")
                     .send({
-                        topics: [[transfer_topic]],
+                        log_matchers: [{
```
LLFourn: I guess you are planning on renaming this to "event*" rather than "log*" given your renaming below?

application/ledger_query_service/src/ethereum/queries.rs
bonomat: Good point :D

api_tests/ledger_query_service/test.js
bonomat: that's why this test was failing... :D

application/ledger_query_service/src/ethereum/queries.rs
bonomat: Mhmm... that's not what we want. We want to check all required filters if they are present.

api_tests/ledger_query_service/test.js
```diff
@@ -373,14 +373,31 @@ describe("Test Ledger Query Service API", () => {
...
                 return chai
                     .request(lqs.url())
                     .post("/queries/ethereum/logs")
                     .send({
-                        topics: [[transfer_topic]],
+                        event_matchers: [
+                            {
+                                address: token_contract_address,
+                                data:
+                                    "0x0000000000000000000000000000000000000000000000000000000000000001",
```
thomaseizinger: :question: Is this `data` the transaction payload?

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -27,11 +27,42 @@ pub struct TransactionQuery {
     transaction_data_length: Option<usize>,
 }
 
-type Topics = Vec<H256>;
+type Topic = H256;
```
thomaseizinger: :wrench: I'd prefer a newtype over a type alias.

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -27,11 +27,42 @@ pub struct TransactionQuery {
...
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
-pub struct LogQuery {
-    topics: Vec<Topics>,
+pub struct EventQuery {
+    event_matchers: Vec<EventMatcher>,
+}
+
+// Event Matcher work similar as web3 filters:
```
thomaseizinger: :x: Should be three slashes to be a proper `rust-doc`.

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -142,45 +173,85 @@ fn clean_0x(s: &str) -> &str {
...
+                            data: None,
+                            topics,
+                        } if topics.is_empty() => false,
+                        EventMatcher {
+                            address,
+                            data,
+                            topics,
+                        } => {
+                            transaction_receipt.logs.iter().any(|tx_log| {
+                                let mut result = true;
```
thomaseizinger: This looks messy but there is probably not much benefit in optimizing it further :)

We could make a `good-first-issue` for refactoring tasks like that!

thomaseizinger: @comit-network/rust-devs what do you think about that?

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -347,58 +418,72 @@ mod tests {
             00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000\
             0000000000000000000000000000";
 
+    const CONTRACT_ADDRESS: &str = "0xe46FB33e4DB653De84cB0E0E8b810A6c4cD39d59";
+    const REDEEM_LOG_MSG: &str =
+        "0xB8CAC300E37F03AD332E581DEA21B2F0B84EAAADC184A295FEF71E81F44A7413";
+    const RANDOM_LOG_MSG: &str =
+        "0xB8CAC300E37F03AD332E581DEA21B2F0B84EAAADC184A295FEF71E81F44A7412";
```
thomaseizinger: :x: Can we have an actually different one here? :) Maybe rename it to `UNKNOWN_LOG_MESSAGE`?

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -408,63 +493,200 @@ mod tests {
...
-        let redeem_log_msg =
-            "0xB8CAC300E37F03AD332E581DEA21B2F0B84EAAADC184A295FEF71E81F44A7413".into();
-        let random_log_msg =
-            "0xB8CAC300E37F03AD332E581DEA21B2F0B84EAAADC184A295FEF71E81F44A7412".into();
-
-        let query = LogQuery {
-            topics: vec![vec![redeem_log_msg, random_log_msg]],
+    fn given_a_transaction_receipt_with_address_should_not_match_with_different_address() {
+        let query = EventQuery {
+            event_matchers: vec![EventMatcher {
```
thomaseizinger: :wrench: It might be worth creating a little builder API for the `EventMatcher` to make the tests less verbose. Something like:

```rust
fn for_contract<S: Into<String>>(address: S) -> Self {
...
}

fn with_topic<S: Into<String>>(self, topic: S) -> Self {
...
}
```

thomaseizinger: It doesn't have to handle all the cases (like `None` topics) just the most common ones :)

application/ledger_query_service/src/ethereum/queries.rs
D4nte: leave it like that for now™

api_tests/ledger_query_service/test.js
bonomat: It's part of the log message

application/ledger_query_service/src/ethereum/queries.rs
bonomat: I tried something fancy like this but ended up in compiler errors _moving_ blabla_ into _FnMut_ blabla so I let it go :D   
```
if data.map_or(false, |data| data == txlog.data) {
return false;
}
```

application/ledger_query_service/src/ethereum/queries.rs
bonomat: it is different and completely random :P 
`0xB8CAC300E37F03AD332E581DEA21B2F0B84EAAADC184A295FEF71E81F44A741`**`3`**
`0xB8CAC300E37F03AD332E581DEA21B2F0B84EAAADC184A295FEF71E81F44A741`**`2`**

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -408,63 +493,200 @@ mod tests {
 
     #[test]
     fn given_a_transaction_receipt_should_not_match_empty_query() {
```
luckysori: :x: I think an empty query *should* match any transaction receipt.

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -27,11 +27,42 @@ pub struct TransactionQuery {
...
+// '0x000000000000000000000000d51ecee7414c4445534f74208538683702cbb3e4' ],
+// },
+// .. ] //Other data omitted
+// }
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+struct EventMatcher {
+    address: Option<Address>,
+    data: Option<Bytes>,
+    topics: Vec<Option<Topic>>,
```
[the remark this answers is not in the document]
bonomat: read the comment above ;)
you need to have an option because you might want to ignore some topics in the list. If you want to know all transactions which called `transfer` with the `toAddress = something` but you don't care where it comes from, i.e. `fromAddress = ignored` than you would need this feature.

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -142,45 +173,85 @@ fn clean_0x(s: &str) -> &str {
...
-impl LogQuery {
+impl EventQuery {
     pub fn matches_block(&self, block: &Block<Transaction>) -> bool {
         match self {
-            Self { topics, .. } if topics.is_empty() => false,
-            Self { topics } => topics.iter().all(|topics| {
-                topics
-                    .iter()
-                    .all(|topic| block.logs_bloom.contains_input(Input::Raw(&topic)))
+            Self { event_matchers } if event_matchers.is_empty() => false,
```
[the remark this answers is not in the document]
bonomat: removed

application/ledger_query_service/src/ethereum/queries.rs
```diff
@@ -142,45 +173,85 @@ fn clean_0x(s: &str) -> &str {
...
         match self {
-            Self { topics } if topics.is_empty() => false,
-            Self { topics } => topics.iter().all(|topics| {
-                !topics.is_empty()
-                    && transaction_receipt
-                        .logs
-                        .iter()
-                        .any(|tx_log| topics.iter().all(|topic| tx_log.topics.contains(topic)))
-            }),
+            Self { event_matchers } if event_matchers.is_empty() => false,
```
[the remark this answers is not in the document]
bonomat: removed

application/ledger_query_service/src/ethereum/queries.rs
bonomat: ok, thanks to @thomaseizinger this worked now.

application/ledger_query_service/src/ethereum/queries.rs
bonomat: done

application/ledger_query_service/src/ethereum/queries.rs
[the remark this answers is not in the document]
luckysori: I think I get it. Is it because the topics are indexed? In that case I would have used a `Vec<(u32, Topic)>` but maybe this is fine or even better :stuck_out_tongue:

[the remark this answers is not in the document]
bonomat: true, that would have been another option.

application/ledger_query_service/src/ethereum/queries/event_query.rs
```diff
@@ -0,0 +1,507 @@
...
+                                {
+                                    return false;
+                                }
+
+                                if data.as_ref().map_or(false, |data| data != &tx_log.data) {
+                                    return false;
+                                }
+
+                                if tx_log.topics.len() == topics.len() {
+                                    tx_log.topics.iter().enumerate().for_each(
```
LLFourn: 🔧 Just make this `foreach` an `all` and I think you can completely delete this `result` variable.

bonomat: true, good point 👍

PR #687, opened 2019-01-22
Make actions returned by HTTP API self-descriptive

### LINKED ISSUE #673: Make actions returned from HTTP API self-descriptive
This is implementation ticket for the results of the spike #494.

Currently, the actions returned in the HTTP API are not self-descriptive. This means, a client has to infer from *out-of-band* information, how to interpret the payload. The design proposed in #494 fixes this.

For reference, this is the current list of all actions:

https://github.com/comit-network/comit-rs/blob/d918e5a2cb712eacd20466da7d5e466447884e7c/application/comit_node/src/http_api/rfc003/action.rs#L154-L170

According to that, the following changes have to made to the currently returned actions:

1. Split up the `SendEthereumTransaction` into two distinct actions:
- deploy contract
- invoke contract

2. Make the serialized JSON contain the `type` attribute.

Implementation hint:

The following serde features might be very helpful in doing this:

- `serde(rename_all = "...")`: https://serde.rs/container-attrs.html#rename_all
- `serde(tag = "...", content = "...")`: https://serde.rs/container-attrs.html#tag--content

luckysori:
Other than what the title says, a `network` attribute has been added to the action payload and the `Ethereum` ledger struct now has a `network` field.

Resolves #673.


api_tests/lib/actor.js
```diff
@@ -45,6 +49,55 @@ class Actor {
...
+
+    async do(action) {
+        let network = action.payload.network;
+        if (network != "regtest") {
+            return Error("Expected network regtest, found " + network);
+        }
+
+        switch (action.type) {
+            case "bitcoin-send-amount-to-address":
+                var { to, amount } = action.payload;
```
LLFourn: ❌ use `let`

api_tests/lib/actor.js
```diff
@@ -45,6 +49,55 @@ class Actor {
                 });
         });
     }
+
+    async do(action) {
+        let network = action.payload.network;
+        if (network != "regtest") {
+            return Error("Expected network regtest, found " + network);
```
LLFourn: ❌ I think you're meant to throw Error in js: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

api_tests/lib/actor.js
```diff
@@ -45,6 +49,55 @@ class Actor {
...
+                var { hex } = action.payload;
+
+                return bitcoin_rpc_client.sendRawTransaction(hex);
+                break;
+            case "ethereum-deploy-contract":
+                var { data, amount, gas_limit } = action.payload;
+
+                return this.wallet
+                    .eth()
+                    .deploy_contract(data, new ethutil.BN(amount, 10));
```
LLFourn: ❓ Is the the `ethutil.BN` necessary?

application/comit_node/src/http_api/rfc003/action.rs
```diff
@@ -302,15 +323,17 @@ impl IntoResponseBody for ethereum::SendTransaction {
...
+            network,
         } = self;
         match query_params {
-            GetActionQueryParams::None {} => Ok(ActionResponseBody::SendEthereumTransaction {
-                to: Some(to),
+            GetActionQueryParams::None {} => Ok(ActionResponseBody::EthereumInvokeContract {
+                contract_address: to,
                 data,
-                value,
+                amount,
```
LLFourn: 🔧  Rename this field value.

In both Bitcoin and Ethereum communities it's called`value`.

vendor/ethereum_support/src/network.rs
```diff
@@ -0,0 +1,7 @@
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "lowercase")]
+pub enum Network {
+    Mainnet,
+    Regtest,
```
LLFourn: ❓ Is Regtest have a special designation as a network, like with special network parameters?

thomaseizinger: Ethereum pre-se doesn't have something like a `regtest` mode. I think parity calls it `development`.

application/comit_node/src/http_api/rfc003/action.rs
luckysori: The spike said amount...

api_tests/lib/actor.js
luckysori: Apparently not!

vendor/ethereum_support/src/network.rs
luckysori: I was just going with the names included in the registry.

application/comit_node/src/http_api/rfc003/action.rs
thomaseizinger: I chose `amount` because it seems less ambiguous and more consistent across multiple blockchains.

PR #688, opened 2019-01-22
Introduce ConnectionPool to replace BamClientPool

### LINKED ISSUE #319: Don't block thread when making new connection in BamClientPool
Given we are using a non-blocking web framework now we shouldn't block the controller threads. However, the `comit_client::ClientFactory::client_for`  method currently blocks when it has to create a new connection. It should instead return a Future for that connection.

It should also keep track of when it is already connecting to something and return a future which resolves when the existing connection attempt completes.

Hint:
Use `ItemCache` to return/create a future for a given COMIT node. The future `Item` is a `Arc<ComitClient>`

DoD:
- [ ] `client_for` returns a future

LLFourn:
- Remove abstractions around "ClientFactory" and just have a ConnectionPool
- Client methods require the NodeId of the thing they want to send a message to
- Client is implemented on ProtocolDependencies so it can properly
  establish connections (hack until we can figure out how to do this nicely)
- Handle the states of connections (which could be displayed over the
  HTTP api at some point)

Also, SwapResponseError is now a RequestError because it had nothing to do with swaps in particular and it seems more natural to have errors that you get when doing a Request called RequestError.

This change also makes it possible to receive swap requests on a connection you initiated (testing requires noise/identity changes).

Note that although it would have been nice to unit test this code I don't feel comfortable doing this yet because:
1. I'm not sure that the current behaviour I've implemented is correct
2. We have this BobSpawner dependency that has to be passed through which would have to be mocked out.

I think the correct behaviour might simply be to allow multiple connections from the same node. 

Resolves #319


application/comit_node/src/comit_client/bam.rs
```diff
@@ -150,69 +151,141 @@ fn to_json_value(
     serde_json::to_value(header).expect("converting bam-header to json must not fail")
 }
 
+#[derive(Debug)]
+enum ConnectionState {
```
thomaseizinger: :wrench: I think naming those fields in the enum variants would greatly aid readability.

application/comit_node/src/comit_client/bam.rs
```diff
@@ -150,69 +151,141 @@ fn to_json_value(
...
-        match existing_client {
-            None => {
+        socket_addr: SocketAddr,
+    ) -> Box<dyn Future<Item = Result<Arc<BamClient>, ClientFactoryError>, Error = ()> + Send> {
+        debug!("Trying to get client for {}", socket_addr);
+        let mut clients = self.clients.write().unwrap();
+
+        let existing_connection = clients
+            .entry(socket_addr)
+            .or_insert_with(|| Arc::new(Mutex::new(ConnectionState::Disconnected)));
```
thomaseizinger: :x: If you do this then you no longer need the default on line 161.

application/comit_node/src/comit_client/bam.rs
```diff
@@ -150,69 +151,141 @@ fn to_json_value(
...
-                    comit_node_socket_addr
+
+                let (sender, receiver) = oneshot::channel();
+                *connection_state = ConnectionState::Connecting(vec![sender]);
+
+                let new_connection_state = Arc::clone(&existing_connection);
+
+                let establish_connection = TcpStream::connect(&socket_addr).then(
+                    move |socket| -> Box<dyn Future<Item = (), Error = ()> + Send> {
+                        info!("Connection to {} established", socket_addr);
```
thomaseizinger: :x: You actually don't know if it is successful here. The log should be moved to the `Ok` match arm below.

application/comit_node/src/comit_client/bam.rs
```diff
@@ -150,69 +151,141 @@ fn to_json_value(
...
+                        let final_connection_state = Arc::clone(&new_connection_state);
+                        let mut new_connection_state = new_connection_state.lock().unwrap();
+
+                        let senders: Vec<_> = match *new_connection_state {
+                            ConnectionState::Connecting(ref mut senders) => senders.drain(..).collect(),
```
thomaseizinger: :question: What is this syntax? :O `.drain(..)`

application/comit_node/src/comit_client/bam.rs
```diff
@@ -150,69 +151,141 @@ fn to_json_value(
...
+                                    Ok(())
+                                }))
+                            }
+
+                            Err(e) => {
+                                error!("Error connecting to {}: {:?}", socket_addr, e);
+                                for sender in senders {
+                                    let _ = sender.send(Err(ClientFactoryError::Connection(e.kind())));
+                                }
+                                *new_connection_state = ConnectionState::Disconnected;
```
thomaseizinger: :thinking: Should we have a connection state `Failed`?

application/comit_node/src/comit_client/bam.rs
```diff
@@ -150,69 +151,141 @@ fn to_json_value(
...
+                Box::new(
+                    receiver.map_err(|_e| {
+                        unreachable!("Sender cannot go out of scope before receiver")
+                    }),
+                )
             }
-            Some(client) => {
-                debug!("Retrieved existing client for {}", comit_node_socket_addr);
-                Ok(Arc::clone(&client))
+            ConnectionState::Connecting(ref mut waiting_for_connections) => {
```
thomaseizinger: :wrench: Maybe this one could also need a log message?

application/comit_node/src/swap_protocols/rfc003/alice/spawner.rs
```diff
@@ -75,17 +74,27 @@ impl<T: MetadataStore<SwapId>, S: StateStore<SwapId>, C: comit_client::Client> A
...
-            self.ledger_events.create_ledger_events(),
-            Arc::clone(&self.comit_client_factory),
-            swap_request.bob_socket_address,
-            save_state,
-        );
+        let swap_execution = {
+            let ledger_events = self.ledger_events.clone();
+            self.comit_client_factory
+                .client_for(swap_request.bob_socket_address)
+                .then(move |client| {
```
thomaseizinger: :question: Why is this not just an `and_then`? Then you don't have to match on the result.

application/comit_node/src/swap_protocols/rfc003/error.rs
```diff
@@ -1,11 +1,18 @@
-use crate::{comit_client::SwapResponseError, ledger_query_service};
+use crate::{comit_client, ledger_query_service};
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum Error {
-    SwapResponse(SwapResponseError),
+    CounterpartyClient(comit_client::ClientFactoryError),
```
thomaseizinger: :wrench: Maybe `Connection` would be a better name?

application/comit_node/src/bam_ext.rs
```diff
@@ -0,0 +1,29 @@
+use bam::{
+    self,
+    config::Config,
+    connection::{self, ClosedReason},
+    json,
+};
+use futures::Future;
+use tokio::io::{AsyncRead, AsyncWrite};
+
```
D4nte: ❌ remove new line

application/comit_node/src/comit_client/mod.rs
```diff
@@ -47,19 +49,19 @@ pub enum SwapDeclineReason {
 pub enum SwapResponseError {
     /// The counterparty had an internal error while processing the request
     InternalError,
-    /// The counterparty produced a response that caused at the transport level
+    /// The counterparty produced a response that was caused at the transport
```
D4nte:
```suggestion
    /// The counterparty produced an error that was caused at the transport
```

application/comit_node/src/comit_client/bam.rs
LLFourn: drain takes a Range argument. `..` is the infinite range. (or something 🤗)

application/comit_node/src/comit_client/bam.rs
LLFourn: Probably. I suggest we implement it when we want it for something though.

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+
+type BamClient =
+    Arc<Mutex<bam::client::Client<json::Frame, json::OutgoingRequest, json::Response>>>;
+
+enum ConnectionState {
+    Connected {
+        client: BamClient,
+    },
+    Connecting {
+        waiting_list: Shared<Box<dyn Future<Item = BamClient, Error = io::ErrorKind> + Send>>,
```
thomaseizinger: :upside_down_face: I'd prefer `wait_list` :)

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+    },
+    Disconnected,
+}
+
+impl Debug for ConnectionState {
```
thomaseizinger: :wrench: Instead of doing this, you could also just [`ignore`](https://mcarton.github.io/rust-derivative/Debug.html#ignoring-a-field) the inner fields of `ConnectionState`:

```rust
#[derive(Derivative)]
#[derivative(Debug)]
enum ConnectionState {
    Connected {
        #[derivative(Debug="ignore")]
        client: BamClient,
    },
    Connecting {
        #[derivative(Debug="ignore")]
        waiting_list: Shared<Box<dyn Future<Item = BamClient, Error = io::ErrorKind> + Send>>,
    },
    Disconnected,
}
```

This would allow you to remove the manual `Debug` and `Display` impl.

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/// Keeps track of connections per NodeId.
+/// For now we don't allow multiple connections from the same NodeId.
+#[derive(Default, Debug)]
+pub struct ConnectionPool {
+    connections: RwLock<HashMap<NodeId, Arc<Mutex<ConnectionState>>>>,
+}
+
+impl ConnectionPool {
```
thomaseizinger: :wrench: Can we have 3 functions on `ConnectionPool` for 

1. establishing a new connection
2. putting a caller on the wait_list
3. retrieve the existing client

This would simplify the `client_for` method a bit and only concern it with delegating based on the current state.

For consistency, I would make each method log at the beginning what they are about to do. Currently, adding a caller to the wait_list is missing this log statement.

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+                    error!("Failed to connect to {}: {:?}", node_id, e);
+                    Err(e)
+                }
+            },
+        }
+    }
+
+    /// Takes a server socket and updates the
+    pub fn new_incoming_socket<B: BobSpawner>(&self, socket: TcpStream, bob_spawner: B) {
+        let node_id = match socket.peer_addr() {
```
thomaseizinger: :question: Can this `match` be replaced with `map_err` and `?`?

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+            Ok(node_id) => node_id,
+            Err(e) => {
+                error!("Couldn't get peer address: {:?}", e);
+                return;
+            }
+        };
+        let mut connections = self.connections.write().unwrap();
+
+        let existing_connection = connections
```
thomaseizinger: These lines are duplicated from above but I don't know if it is worth the effort to reduce the duplication.

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+            .entry(node_id)
+            .or_insert_with(|| Arc::new(Mutex::new(ConnectionState::Disconnected)));
+
+        let mut existing_connection_guard = (*existing_connection).lock().unwrap();
+
+        match *existing_connection_guard {
+            // Ignore the new socket -- we already have a connection
```
thomaseizinger: :x: I think the comments are redundant. The log message says exactly the same.

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+
+            // We're trying to connect to the remote node at the same time they
+            // connect to us. We ignore their connection and keep connecting.
+            ConnectionState::Connecting { .. } => {
+                warn!(
+                    "Ignoring incoming connection from {} because we're already connecting",
+                    node_id
+                );
+            }
+            // The most common case. We were disconnected but now we're connected.
```
thomaseizinger: :wrench: Maybe put the common case at the top then? :)

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+                    bob_spawner,
+                    socket,
+                );
+            }
+        }
+    }
+
+    /// Sets the connection_state to "Connected" and spawns the Connection.
+    /// When the connection is finished it sets it to Disconnected.
+    /// `connection_state` and `connection_handle`` must point to the same
```
thomaseizinger: :x: You got a double \` here.

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+    /// ConnectionState.
+    fn transition_to_connected<S: AsyncRead + AsyncWrite + Send + 'static, B: BobSpawner>(
+        node_id: NodeId,
+        connection_state: &mut ConnectionState,
```
thomaseizinger: :question: Is this the `current` ConnectionState? Maybe distinct naming could avoid this confusion.

application/comit_node/src/connection_pool.rs
thomaseizinger: :wrench: This statements sounds like potential `debug_assert!` at the end of the function! :)

(Maybe at the cost of removing the documentation here?)

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+        bob_spawner: B,
+        socket: S,
+    ) -> BamClient {
+        info!(
+            "Established new connection to {} while connection was in the {} state",
+            node_id, connection_state
+        );
+
+        let (client, connection) = {
+            // Set up BAM connection
```
thomaseizinger: :wrench: Maybe this comment should be a function name?

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+
+            *final_connection_state = ConnectionState::Disconnected;
+
+            Ok(())
+        }));
+
+        client
+    }
+
+    pub fn connected_addrs(&self) -> Vec<SocketAddr> {
```
thomaseizinger: :question: Don't you need to filter by `ConnectedState::Connected` here?

application/comit_node/src/connection_pool.rs
```diff
@@ -0,0 +1,286 @@
...
+            Connecting { .. } => write!(f, "connecting"),
+            Disconnected => write!(f, "disconnected"),
+        }
+    }
+}
+
+/// Keeps track of connections per NodeId.
+/// For now we don't allow multiple connections from the same NodeId.
+#[derive(Default, Debug)]
+pub struct ConnectionPool {
```
thomaseizinger: :x: This guy has *A LOT* of logic but not a single test :(

I'd be interesting in doing some pair-programming on that since it seems to be quite a complex and important piece :)

application/comit_node/src/node_id.rs
```diff
@@ -0,0 +1,2 @@
+use std::net::SocketAddr;
+pub type NodeId = SocketAddr;
```
thomaseizinger: :question: Should it really be a type alias and not a new type? Given that we might change this at some point, a new type would be more resilient to refactoring?

application/comit_node/src/swap_protocols/dependencies/client_impl.rs
```diff
@@ -0,0 +1,131 @@
...
+        node_id: NodeId,
+        request: rfc003::Request<AL, BL, AA, BA>,
+    ) -> SwapResponse<AL, BL> {
+        Box::new(
+            self.connection_pool
+                .client_for(node_id, self.clone())
+                .map_err(|e| {
+                    RequestError::Connecting(e)
+                })
+                .and_then(move |client| {
```
thomaseizinger: Looking forward to `async/await`. This is going to look sooooo much better!

application/comit_node/src/connection_pool.rs
LLFourn: This doesn't return a Result. A socket not having a peer_addr is a major yikes but we can just ignore and log sockets that don't.

application/comit_node/src/connection_pool.rs
LLFourn: The logic and the api will change soon so I'm reluctant to test this at all atm.

I'd suggest pair programming the right logic with TDD.

application/comit_node/src/connection_pool.rs
LLFourn: There is no way of asserting this AFAIK. This private function just captures what was duplicated between incoming and outgoing connections. It's rather arbitrary.

application/comit_node/src/connection_pool.rs
LLFourn: Yes. I thought that "current" is implied

application/comit_node/src/connection_pool.rs
LLFourn: Yes. But I'd rather write tests for this new feature before implementing it.

thomaseizinger: Should we remove the function then? :D

application/comit_node/src/connection_pool.rs
LLFourn: This has changed to `spawn_new_connection`.

application/comit_node/src/node_id.rs
LLFourn: I just looked at doing this and actually it wouldn't save any work later down the track because it will completely change with the introduction of noise IDs.

D4nte:
The change is quite big. I see that @luckysori and @thomaseizinger looked into it. If you want my review then let's do a walk through together. If you think you already got enough feedback then happy to leave you to it :)

luckysori:
> The change is quite big. I see that @luckysori and @thomaseizinger looked into it. If you want my review then let's do a walk through together. If you think you already got enough feedback then happy to leave you to it :)

I could to with a walk through too tbh. I just left some comments, but didn't feel confident enough to approve.

D4nte:
> > The change is quite big. I see that @luckysori and @thomaseizinger looked into it. If you want my review then let's do a walk through together. If you think you already got enough feedback then happy to leave you to it :)
> 
> I could to with a walk through too tbh. I just left some comments, but didn't feel confident enough to approve.

I guess it's time to do a walkabout

PR #721, opened 2019-01-24
Remove noisy log messages

### LINKED ISSUE #690: Logs repeat themselves
When running end to end tests, something is causing most of the logs emitted to `comit_node-alice.log` and `comit_node-bob.log` to be repeated a lot of times. This makes these files really hard to read and slows down debugging.

thomaseizinger:
These log statements were introduced at a time where the COMIT node
would still deploy stuff by themselves. However, we no longer that.
Instead, we tell the user, what to do by returning "actions" over the
HTTP API. This means, whatever was logged here is actually returned
to the user and doesn't add any value in the log file.

Fixes #690.


D4nte:
@thomaseizinger  Are we still logging the HTLC at least once?

PR #732, opened 2019-01-31
Noise Handshake Future

### LINKED ISSUE #668: Implement the noise handshake protocol
Internally we have a noise codec implementation in: https://github.com/coblox/tokio-noise-codec
This allows us to use noise as the transport once we have done a handshake.

We still need an asynchronous handshake implementation. It needs to be asynchronous so we can do #319.

hint:
- Use a state machine
- Push code in own crate

DOD: 
- [x] noise-handshake-future implemented for the "XK" handshake pattern. (Initially in comit-rs/vendor)
- [x] tested against the parameters decided in #637 

Consider how hard it would be to generalise across all handshake patterns.

blocked by #637

D4nte:
Resolves #668

### DIFF
--- a/vendor/noise_handshake/src/handshake.rs
+++ b/vendor/noise_handshake/src/handshake.rs
@@ -88,7 +88,7 @@ impl<IO: AsyncRead + AsyncWrite> Future for NoiseHandshake<IO> {
         match self {
             Self {
                 noise: Some(ref mut noise),
-                io: Some(ref _io),
+                io: Some(_),
                 next:
                     Step::Write {
                         ref mut buffer,
@@ -146,10 +146,7 @@ impl<IO: AsyncRead + AsyncWrite> Future for NoiseHandshake<IO> {
                     self.poll()
                 }
             }
-            Self { noise: None, .. } => {
-                panic!("Future is already resolved!");
-            }
-            Self { io: None, .. } => {
+            Self { noise: None, .. } | Self { io: None, .. } => {
                 panic!("Future is already resolved!");
             }
         }
--- a/vendor/noise_handshake/src/handshake.rs
+++ b/vendor/noise_handshake/src/handshake.rs
@@ -186,7 +186,10 @@ mod tests {
     fn handshake() -> Result<(), std::io::Error> {
         let (hs_init, hs_resp) = setup();
 
-        let mut runtime = tokio::runtime::Runtime::new().unwrap();
+        let mut runtime = tokio::runtime::Builder::new()
+            .core_threads(2)
+            .build()
+            .unwrap();
 
         let hs_init = hs_init.map_err(|_| ());
 
--- a/vendor/noise_handshake/src/handshake.rs
+++ b/vendor/noise_handshake/src/handshake.rs
@@ -215,11 +215,13 @@ impl<IO: AsyncRead + AsyncWrite> Future for NoiseHandshake<IO> {
 mod tests {
 
     use super::*;
+    use futures::future::poll_fn;
     use std::thread;
+    use tokio::net::{TcpListener, TcpStream};
 
     fn setup() -> (
-        NoiseHandshake<memsocket::UnboundedSocket>,
-        NoiseHandshake<memsocket::UnboundedSocket>,
+        impl Future<Item = (Session, TcpStream), Error = std::io::Error>,
+        impl Future<Item = (Session, TcpStream), Error = std::io::Error>,
     ) {
         let builder_resp = snow::Builder::new("Noise_XK_25519_ChaChaPoly_BLAKE2s".parse().unwrap());
         let static_keypair_resp = builder_resp.generate_keypair().unwrap();
@@ -236,11 +238,19 @@ mod tests {
             .build_initiator()
             .unwrap();
 
-        let (socket_init, socket_resp) = memsocket::unbounded();
+        let addr = "127.0.0.1:0".parse().unwrap();
 
-        let handshake_init = noise_init.handshake(socket_init);
+        let mut listener = TcpListener::bind(&addr).expect("unable to bind TCP Listener");
+        let addr = listener.local_addr().expect("Did not bind?");
+        dbg!(addr);
 
-        let handshake_resp = noise_resp.handshake(socket_resp);
+        let listener_future = poll_fn(move || listener.poll_accept());
+
+        let handshake_resp =
+            listener_future.and_then(move |(socket, _)| noise_resp.handshake(socket));
+
+        let handshake_init =
+            TcpStream::connect(&addr).and_then(move |socket| noise_init.handshake(socket));
 
         (handshake_init, handshake_resp)
     }
--- a/vendor/noise_handshake/src/handshake.rs
+++ b/vendor/noise_handshake/src/handshake.rs
@@ -11,9 +11,9 @@ pub enum Step {
         len: usize,
     },
     Write {
-        buffer: [u8; NOISE_MAX_SIZE],
+        to_write: [u8; NOISE_MAX_SIZE],
         written_bytes: usize,
-        len: usize,
+        total_bytes: usize,
     },
 }
 
@@ -29,11 +29,11 @@ impl Step {
         let mut buffer = [0u8; NOISE_MAX_SIZE];
         let len = noise
             .write_message(&[], &mut buffer)
-            .expect("Cannot encode the message");
+            .expect("A zero-length message cannot be too long");
         Step::Write {
-            buffer,
+            to_write: buffer,
             written_bytes: 0,
-            len,
+            total_bytes: len,
         }
     }
 }
@@ -78,7 +78,7 @@ impl<IO: AsyncRead + AsyncWrite> NoiseHandshake<IO> {
         let noise = self.noise.take().expect("We know it's a Some");
         let noise = noise
             .into_transport_mode()
-            .expect("Cannot go into transport mode despite handshake being finished");
+            .expect("Should not fail as handshake is finished");
         let io = self.io.take().expect("We know it's a Some");
         (noise, io)
     }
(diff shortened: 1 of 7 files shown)


vendor/noise_handshake/src/handshake.rs
```diff
@@ -0,0 +1,230 @@
...
+                io: Some(io),
+            },
+            _ => panic!(
+                "Noise Session in incorrect state, you should init before starting handshake"
+            ),
+        }
+    }
+}
+
+#[derive(Debug)]
```
thomaseizinger: :wrench: Futures don't usually have to be `Debug` because you are very likely going to spawn them anyway and then you no longer have a reference to them that you could print.

vendor/noise_handshake/src/handshake.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        match self {
+            Self {
+                noise: Some(noise),
+                io: Some(_),
+                ..
+            } if noise.is_handshake_finished() => {
+                let noise = self.noise.take().expect("We know it's a Some");
+                let noise = noise.into_transport_mode().unwrap();
```
thomaseizinger: :x: Change `unwrap` to `expect`.

vendor/noise_handshake/src/handshake.rs
```diff
@@ -0,0 +1,230 @@
...
+                next:
+                    Step::Write {
+                        mut buffer,
+                        mut remaining_bytes,
+                        mut len,
+                    },
+            } if remaining_bytes == 0 => {
+                len = noise.write_message(&[], &mut buffer).unwrap();
+
+                let mut num_bytes_written = 0;
```
thomaseizinger: :wrench: Remove the whole block until `self.poll` and let the block below handle that :)

vendor/noise_handshake/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+#![warn(unused_extern_crates, rust_2018_idioms, missing_debug_implementations)]
+#![deny(unsafe_code)]
+#![feature(stmt_expr_attributes)]
+
+#[macro_use]
+extern crate log;
+#[macro_use]
+extern crate debug_stub_derive;
+
+pub mod handshake;
```
thomaseizinger: :wrench: I'd suggest to just re-export the future struct and not exposing the module to the outside.

vendor/noise_handshake/src/handshake.rs
```diff
@@ -0,0 +1,230 @@
+use futures::*;
+use snow::Session::{self, Handshake};
+use tokio::io::{AsyncRead, AsyncWrite};
+
+const NOISE_MAX_SIZE: usize = 65535;
```
thomaseizinger: Do we actually need that much for the handshake? Instances of the enum will have quite a big memory footprint with this size.

D4nte: Hum, no we apparently do not but we would need to make assumptions.

LLFourn: The way you've implemented this you will be able to handle any noise handshake successfully! So well done on that. It's possible you will need a buffer of 65535 in case the application puts a lot of data on top of the handshake messages.

The thing is, you're completely dropping all the application data below:

```rust
                let mut dec_buffer = [0u8; NOISE_MAX_SIZE];

                *len += try_ready!(io.poll_read(&mut enc_buffer[*len..]));
                if *len > 0 {
                    match noise.read_message(&enc_buffer[..*len], &mut dec_buffer) {
                        Ok(_) => self.next = Step::write(),
                        Err(e) => debug!("Error decoding message: {:?}", e),
                    };
                }
                if noise.is_handshake_finished() {
                    Ok(Async::Ready(self.wrap_up()))
                } else {
                    self.poll()
                }
```
`dec_buffer` is never used for anything.

So the question is does this Future allow for reading and understanding the application data?
In our case we're probably going to want to tell the other node during the handshake we're using `bam-json` or something so we know what codec to use afterwards.

So to properly calculate the size of the noise message you need to know:
1. Where you are in whatever handshake you're using e.g `s, se`, `e`
2. What curve you're using e.g 25519 (32 byte keys)
3. How much application data is meant to be sent at this point in the handshake.

If you get all these then you can figure out exactly how long the buffer should be. I think you could still keep your general structure -- probably just change the arrays in the enums to vecs.

We can leave the whole application data problem until we actually want to implement it. The question is, do you want to get rid of the 65535 and figure out how long each thing is meant to be (assuming 0 length application data) in the PR? We will have to do it eventually. The difficult thing about it won't be the actual calculation but adding the tracking the state of handshake you are in to your future. Can the snow session tell you this?

LLFourn: I agree on point 1. Point 2 doesn't sound very useful.

I don't think it's over optimising to know how much data we need to read from a socket. That seems like something someone reading from a socket should know. In general I don't like throwing things into something and seeing if it's OK as a way of processing stuff.

For now I don't mind merging this and seeing how it works in practice. I'll review the rest of the code ASAP.

D4nte: Ok.. Should we open an issue for point 1 and another one to predict the size of the buffer?

D4nte: Please note that regarding prediction of the size of the buffer I think some of the work should go in the snow library itself.

vendor/noise_handshake/src/handshake.rs
```diff
@@ -0,0 +1,217 @@
...
+
+        (handshake_init, handshake_resp)
+    }
+
+    #[test]
+    fn handshake() -> Result<(), std::io::Error> {
+        let (hs_init, hs_resp) = setup();
+
+        let mut runtime = tokio::runtime::Builder::new()
+            .core_threads(2)
```
LLFourn: ❓ Ahhh is this how you go it to pass travis? When I suggested this I was thinking you would always set it to 1 and then find a solution. But I guess this works 😅

vendor/noise_handshake/src/handshake.rs
```diff
@@ -0,0 +1,217 @@
...
+                        ref mut len,
+                    },
+            } => {
+                let mut dec_buffer = [0u8; NOISE_MAX_SIZE];
+
+                *len += try_ready!(io.poll_read(&mut enc_buffer[*len..]));
+                if *len > 0 {
+                    match noise.read_message(&enc_buffer[..*len], &mut dec_buffer) {
+                        Ok(_) => self.next = Step::write(),
+                        Err(e) => debug!("Error decoding message: {:?}", e),
```
LLFourn: Knowing the size of what we are expecting would remove this weird Err debug thing. If we know what we are looking for we can just return a proper error.

vendor/noise_handshake/src/handshake.rs
thomaseizinger: I think so too. I'd say, open an issue with the problem of "optimize buffer size during snow handshake". There, we can then reference that a contribution to snow is necessary and block the feature on that.

vendor/noise_handshake/src/handshake.rs
```diff
@@ -0,0 +1,217 @@
...
+            Self {
+                noise: Some(ref mut noise),
+                io: Some(_),
+                next:
+                    Step::Write {
+                        ref mut buffer,
+                        ref mut len,
+                        ..
+                    },
+            } if *len == 0 => {
```
LLFourn: ❌  Isn't len only 0 after you create a new `Step::Write`. Why not just write the message and get the len at the point you do `Step::write()` to avoid having three match arms.

D4nte: It's on purpose to make it clearer. Otherwise you end-up doing a lot in the same match arm.

LLFourn: Hmm I think removing a match arm and bloating the other one would actually be clearer. You don't do any IO in that match arm so it doesn't feel like the others.

vendor/noise_handshake/src/handshake.rs
D4nte: #736 open

vendor/noise_handshake/src/handshake.rs
LLFourn: I made a PR demonstrating how you can do this in two match arms #738

D4nte: Ok I did not understand it this way.

D4nte:
Not yet ready, still need to stabilise Travis

vendor/noise_handshake/src/handshake.rs
D4nte: done

D4nte:
Build successful. @LLFourn @thomaseizinger please approve.

vendor/noise_handshake/src/handshake.rs
```diff
@@ -0,0 +1,209 @@
...
+#[allow(missing_debug_implementations)]
+pub enum Step {
+    Read {
+        enc_buffer: [u8; NOISE_MAX_SIZE],
+        len: usize,
+    },
+    Write {
+        buffer: [u8; NOISE_MAX_SIZE],
+        written_bytes: usize,
+        len: usize,
```
LLFourn: 🔧  let's give these more meaningful names:

-` buffer` change to `to_write`
- `len` change to `total_bytes`

In the Read part you could also change `len` to `read_bytes`

vendor/noise_handshake/src/handshake.rs
```diff
@@ -0,0 +1,209 @@
...
+            enc_buffer: [0u8; NOISE_MAX_SIZE],
+            len: 0,
+        }
+    }
+
+    fn write(noise: &mut Session) -> Self {
+        let mut buffer = [0u8; NOISE_MAX_SIZE];
+        let len = noise
+            .write_message(&[], &mut buffer)
+            .expect("Cannot encode the message");
```
LLFourn: ❌ With expect you should say why this can never fail: "A zero-length message is cannot be too long".

vendor/noise_handshake/src/handshake.rs
```diff
@@ -0,0 +1,209 @@
...
+    noise: Option<Session>,
+    io: Option<IO>,
+}
+
+impl<IO: AsyncRead + AsyncWrite> NoiseHandshake<IO> {
+    fn wrap_up(&mut self) -> (Session, IO) {
+        let noise = self.noise.take().expect("We know it's a Some");
+        let noise = noise
+            .into_transport_mode()
+            .expect("Cannot go into transport mode despite handshake being finished");
```
LLFourn: Same thing here.

PR #733, opened 2019-01-31
Add log_level option to settings

### LINKED ISSUE #711: get level from config file once implemented with #136...
https://github.com/comit-network/comit-rs/blob/13b233bbec1d292e15542988b338257442dadf5c/application/comit_node/src/logging.rs#L20-L21

* Found on master

LLFourn:
Fixes #711 

Adds a log_level (default: debug) under each heading in the settings and uses that to determine the relevant LogFilter in `logging.rs`.

### DIFF
--- a/application/comit_node/src/logging.rs
+++ b/application/comit_node/src/logging.rs
@@ -1,3 +1,4 @@
+use crate::settings::ComitNodeSettings;
 use fern::{
     colors::{Color, ColoredLevelConfig},
     Dispatch, FormatCallback,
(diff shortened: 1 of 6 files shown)


application/comit_node/src/logging.rs
```diff
@@ -14,19 +15,21 @@ pub fn set_context<S: ToString>(input: &S) {
     });
 }
 
-pub fn set_up_logging() {
+pub fn set_up_logging(settings: &ComitNodeSettings) {
     Dispatch::new()
         .format(move |out, message, record| formatter(out, message, record))
-        // TODO: get level from config file once implemented with #136
         .level(LevelFilter::Debug)
```
thomaseizinger: :wrench: What about the other way round:

1. Make the standard level be controlled by a single flag in the config file.
2. Remove all those `level_for` specific to the application here.
3. Override certain libraries (like tokio) that we don't want.

LLFourn: I wouldn't mind having a single default one. Would you remove all the section specific ones in favour of it or just make it set the default?

LLFourn:
@D4nte all the log levels are optional so they don't need to change I don't think. What change would you like?

application/comit_node/src/logging.rs
thomaseizinger: Yes I would remove them so they are all automatically set to the default one.

D4nte:
@LLFourn just add some so that we can see it at example + 
Also, shouldn;t

> @D4nte all the log levels are optional so they don't need to change I don't think. What change would you like?

As discussed:
- update config file at `./application/comit_node/config/*` to test the parsing
- make default `info`
- make e2e test `debug` for the relevant modules (up to you which modules, probably same that it is before your PR)

LLFourn:
@D4nte @thomaseizinger Both your comments addressed.

PR #737, opened 2019-02-01
Split up dry tests

### LINKED ISSUE #593: Dry test needs to be split up
the dry tests are now bloated (well done 🎉). We should split this file up into smaller files that each test one area of functionality.

LLFourn:
Fixes #593


api_tests/dry/multiple_peers.js
```diff
@@ -0,0 +1,138 @@
...
+
+it("[Alice] Should be in Start state after sending the swap request to Charlie", async function() {
+    await alice.poll_comit_node_until(
+        chai,
+        alice_swap_with_charlie_href,
+        "Start"
+    );
+});
+
+it("[Charlie] Shows the Swap as Start in /swaps", async () => {
```
luckysori: :question: Should we do the same for Alice's swap request to Bob?

api_tests/dry/multiple_peers.js
```diff
@@ -0,0 +1,138 @@
...
+        .then(res => {
+            res.error.should.equal(false);
+            res.should.have.status(201);
+            swap_location = res.headers.location;
+            swap_location.should.be.a("string");
+            alice_swap_with_charlie_href = swap_location;
+        });
+});
+
+it("[Alice] Should be in Start state after sending the swap request to Charlie", async function() {
```
luckysori: :question: Should we do the same for Alice's swap request to Bob?

api_tests/dry/stateless.js
```diff
@@ -0,0 +1,90 @@
...
+        .send({
+            garbage: true,
+        })
+        .then(res => {
+            res.should.have.status(400);
+            res.body.title.should.equal("Bad Request");
+        });
+});
+
+it("[Bob] Should have no peers before receiving a swap request from Alice", async () => {
```
luckysori: :x: 
```suggestion
it("[Alice] Should have no peers before sending or receiving swap requests", async () => {
```

api_tests/harness.js
```diff
@@ -180,7 +180,7 @@ async function startLedgerQueryService(name, lqs_config) {
 // Start services, run test, shutdown //
 // ********************************** //
 
-describe("🛠", async function() {
+describe("🏃" + test_dir, async function() {
```
luckysori: Hahaha.

api_tests/harness.js
```diff
@@ -228,9 +228,7 @@ describe("🛠", async function() {
         await generateBlock(config.ledgers, ledgers_config);
     });
 
-    describe("👟", async function() {
-        require(test_dir + "/test.js");
-    });
+    require(test_dir + "/test.js");
```
D4nte: ❓ are you sure you are having reliable test run with this change?

api_tests/dry/multiple_peers.js
LLFourn: Nah there's no need to duplicate all these tests. We test this already in the single peer test.

I think we could delete both these tests actually.

api_tests/dry/multiple_peers.js
LLFourn: No. We test this already in the other file. I think we should use different quantities for this test and asset them though.

api_tests/dry/stateless.js
LLFourn: Heh good catch 👍

api_tests/harness.js
LLFourn: Haven't has any problems so far.

LLFourn:
By the way, I changed the harness so it runs all files ending in `.js` in the directory of the test (not just `test.js`).

api_tests/harness.js
```diff
@@ -180,60 +180,68 @@ async function startLedgerQueryService(name, lqs_config) {
...
-        await generateBlock(config.ledgers, ledgers_config);
-    });
+            await util.sleep(2000);
+        });
 
-    describe("👟", async function() {
-        require(test_dir + "/test.js");
-    });
+        beforeEach(async function() {
+            // FIXME: using setInterval to create a block every X seconds made the test harness hang (even after stopping the interval) so we just create a block before every test for now.
```
D4nte:
```suggestion
            // TODO: (fix me) using setInterval to create a block every X seconds made the test harness hang (even after stopping the interval) so we just create a block before every test for now.
```
So that `source of truth` picks it up.

LLFourn: I just fixed the TODO.

PR #738, opened 2019-02-01
Remove a match statement in noise handshake future

LLFourn:
A suggested change to #732

### DIFF
--- a/vendor/noise_handshake/src/handshake.rs
+++ b/vendor/noise_handshake/src/handshake.rs
@@ -25,11 +25,13 @@ impl Step {
         }
     }
 
-    fn write() -> Self {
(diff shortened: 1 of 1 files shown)


vendor/noise_handshake/src/handshake.rs
```diff
@@ -25,11 +25,13 @@ impl Step {
         }
     }
 
-    fn write() -> Self {
+    fn write(noise: &mut Session) -> Self {
+        let mut buffer = [0u8; NOISE_MAX_SIZE];
+        let len = noise.write_message(&[], &mut buffer).unwrap();
```
D4nte: Use `expect`

D4nte:
Integrated in #732

PR #743, opened 2019-02-05
Fix RLP encoding error

### LINKED ISSUE #740: Random RLP encoding failure
Brought in by #725

LLFourn:
Ethereum apparently expects that signature components do not start
with 0x00 (i.e be treated as RLP integers) Why? No body knows.

So we have to wrap the (R,s) values in U256s so they encode properly.

🤦

Fix #740.

### DIFF
--- a/application/comit_node/tests/ethereum_wallet/transaction.rs
+++ b/application/comit_node/tests/ethereum_wallet/transaction.rs
@@ -44,8 +44,12 @@ impl<'a> SignedTransaction<'a> {
 
 impl<'a> Encodable for SignedTransaction<'a> {
     fn rlp_append(&self, stream: &mut RlpStream) {
-        let r = &self.signature.0[0..32];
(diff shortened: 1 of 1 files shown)


application/comit_node/tests/ethereum_wallet/transaction.rs
```diff
@@ -44,8 +44,8 @@ impl<'a> SignedTransaction<'a> {
 
 impl<'a> Encodable for SignedTransaction<'a> {
     fn rlp_append(&self, stream: &mut RlpStream) {
-        let r = &self.signature.0[0..32];
-        let s = &self.signature.0[32..64];
+        let r = U256::from(&self.signature.0[0..32]);
```
D4nte: ❌ as mentioned, please as add a comment 🤯

LLFourn: Done.

PR #745, opened 2019-02-06
Logs for error 500 and refactoring of http_api::rfc003 handlers/routes

### LINKED ISSUE #584: Log 500 errors
When returning a 500 errors, comit_node should log the reason why to assist with local investigation.

The easiest way is likely to be a new filter on warp.

DoD:
- [ ] Go through `HttpProblem` usage and ensure the originating error is logged

D4nte:
- Add more logging when creating an error 500 from a generic error
- split `swap.rs` file
  - `routes.rs` contains the functions used by warp
  - `handlers` contains the handlers used by `routes.rs`
- Update logs to remove deprecated functionalities and log filename+line

I am stopping here. I think `IntoResponseBody` could go in its own file but want to reduce amount of merge work for @LLFourn and @luckysori 

Resolves #584


application/comit_node/src/http_api/problem.rs
```diff
@@ -50,14 +42,16 @@ pub fn unsupported() -> HttpApiProblem {
     HttpApiProblem::new("swap-not-supported").set_status(400)
 }
 
-pub fn serde(_e: &serde_json::Error) -> HttpApiProblem {
-    // FIXME: Use error to give more detail to the user
+pub fn serde(e: &serde_json::Error) -> HttpApiProblem {
+    // TODO (fixme): Use error to give more detail to the user
```
luckysori: :x:  This TODO can be removed.

application/comit_node/src/http_api/rfc003/handlers/get_swaps.rs
```diff
@@ -0,0 +1,49 @@
...
+use http_api_problem::HttpApiProblem;
+use rustic_hal::HalResource;
+
+#[derive(Serialize, Debug)]
+pub struct EmbeddedSwapResource {
+    state: String,
+    protocol: String,
+}
+
+use crate::http_api::rfc003::routes::{swap_path, PROTOCOL_NAME};
```
luckysori: :wrench: This should go at the top with the other `use` statements.

application/comit_node/src/http_api/rfc003/handlers/post_swap.rs
```diff
@@ -0,0 +1,174 @@
...
+                body.identities,
+                id,
+                secret_source,
+            )?,
+            bob_socket_address: body.peer,
+        })
+    }
+}
+
+pub fn handle_post_swap<A: AliceSpawner>(
```
luckysori: :upside_down_face: Since you're reorganising stuff, maybe this could go at the top, since it's the focal point of the file. Same goes for the other handlers.

application/comit_node/src/http_api/rfc003/routes.rs
```diff
@@ -0,0 +1,345 @@
...
+use rustic_hal::HalResource;
+use std::{net::SocketAddr, str::FromStr, sync::Arc};
+use warp::{Rejection, Reply};
+
+pub const PROTOCOL_NAME: &str = "rfc003";
+pub fn swap_path(id: SwapId) -> String {
+    format!("/{}/{}/{}", http_api::PATH, PROTOCOL_NAME, id)
+}
+
+#[derive(Debug, Serialize)]
```
luckysori: :wrench: I think some of these structs and enums should go in a separate file or where they are used, like you did with `EmbeddedSwapResource`. Otherwise, I see this file depending on the handlers and the handlers depending on this file.

application/comit_node/src/http_api/problem.rs
D4nte: Why? What changed?

luckysori: I think you now use the error, which is why you removed the underscore :stuck_out_tongue:

application/comit_node/src/http_api/rfc003/routes.rs
D4nte: Yes that's fine, will do. I just want to avoid `route_factory.rs` depending on the handlers. We'll see if that happens.

D4nte: e.g. What is the best place for `PROTOCOL_NAME` and `swap_path`? @comit-network/rust-devs

application/comit_node/src/http_api/problem.rs
D4nte: ok I guess so, whoever wrote this comment speak now or hold your peace forever.

application/comit_node/src/http_api/rfc003/routes.rs
```diff
@@ -0,0 +1,119 @@
...
+    request_body_kind: SwapRequestBodyKind,
+) -> Result<impl Reply, Rejection> {
+    handle_post_swap(&alice_spawner, secret_source.as_ref(), request_body_kind)
+        .map(|swap_created| {
+            let body = warp::reply::json(&swap_created);
+            let response =
+                warp::reply::with_header(body, header::LOCATION, swap_path(swap_created.id));
+            warp::reply::with_status(response, warp::http::StatusCode::CREATED)
+        })
+        .map_err(|problem| warp::reject::custom(HttpApiProblemStdError::from(problem)))
```
thomaseizinger: :x: Can you please unify all the router functions to use:

```rust
.map_err(into_rejection)
```

where `into_rejction` is:

```rust
fn into_rejection(problem: HttpApiProblem) -> Rejection {
    warp::reject::custom(HttpApiProblemStdError::from(problem))
}
```

application/comit_node/src/logging.rs
```diff
@@ -29,27 +17,22 @@ pub fn set_up_logging(settings: &ComitNodeSettings) {
...
+    };
 
-    LOG_CONTEXT.with(|context| {
-        let context = {
-            match *context.borrow() {
-                Some(ref context) => format!("[{}] ", context),
-                None => "".to_string(),
-            }
-        };
+    let path = match record.file() {
```
thomaseizinger: :wrench: This can be expressed more idiomatically as:

```rust
let path = record.file().unwrap_or_else(|| record.target())
```

application/comit_node/src/logging.rs
```diff
@@ -29,27 +17,22 @@ pub fn set_up_logging(settings: &ComitNodeSettings) {
 }
 
 fn formatter(out: FormatCallback<'_>, message: &Arguments<'_>, record: &Record<'_>) {
-    // configure colors for the whole line
-    let colors_line = ColoredLevelConfig::default()
-        .info(Color::Green)
-        .debug(Color::Blue)
-        .trace(Color::Cyan);
+    let line = match record.line() {
```
thomaseizinger: :wrench: This can be expressed more idiomatically as:

```rust
let line_number = record.line().map(|line| format!(":{}", line)).unwrap_or_else(|| String::new())
```

D4nte: ❓ what makes `map` more idiomatic than `match` ?

thomaseizinger: It probably boils down to personal preference actually :sweat_smile: 
The way I see it is that `match` is like a 'terminal' operation me, like `for_each` on an iterator. Something where I can cause side-effects for example.

But if all I want is to *change* the option and re-assign it, I always try to use the APIs on `Option`, `Result` etc.

It is a bit like doing `for x in list` vs `iter.map`.

The first one allows you to do whatever you want (same as the `match` statement) because you are simply unpacking it whereas the latter describes through the method calls, what is happening.

application/comit_node/src/http_api/problem.rs
thomaseizinger: No it is not resolved.

The comment means: return details to the user like `field blublub was unexpected`.

That requires some more work I guess and should be tracked in its own ticket and not in a TODO :)

I'd suggest: remove the TODO and create a proper ticket.

application/comit_node/src/http_api/rfc003/routes.rs
thomaseizinger: `route_factory.rs`?

application/comit_node/src/http_api/problem.rs
D4nte: #754 opened

Issue #576, opened 2018-12-16
⚗️ Re-architecting state machine


D4nte:
DoD:
* Presentation of the new architecture including upsides and downsides.
* New issues ready to be groomed, under a new Epic.
* Current "Refund" issues updated/deleted.
* How do we get refund with this re-architecture (without new "refundable" states).
* How do we add "secret reveal" before redeem transaction is confirmed (mempool).

LLFourn:
## Forward

While doing this spike I realised this problem can be split into to:

1. What we store in the storage layer is too tightly coupled to our state machine. We shouldn't parameterize our state machine by the `Role` (and even worse have an instance of `role` inside the state machine!).
2. I don't think we shouldn't have a single state machine across both ledgers but rather individual state machines for alpha and beta ledgers.

I decided that fixing (1) first is the most useful because once that's done we can restructure the state machines however we want.

## Problem
Right now we have a very tight coupling between states in the state machine and actions. The HTTP controllers themselves are able to get the current state from the `state_store`. This means that if we want to add new actions or represent new knowledge we have to do so by adding new states. For example:

- To add refund actions we have to create new `Refundable` states for every situation where you could refund one or the other
- To represent the fact that you can learn the secret from a tx in the mempool before it's confirmed you would have to have some new set of states to represent "I've learned the secret but the beta ledger redeem hasn't happened yet"

Doing the above leads to a combinatorial explosion of new states. The purpose of the ticket is to discover a way to decouple *what* we store from *how* we store it. We should store things that make it easy to produce actions and responses for the HTTP api. If the state machine has many complicated states we shouldn't necessarily have to deal with them when producing actions.

## Solution

Instead of storing state machine states we store `Alice` or `Bob` structs which represent the actor's knowledge at the current time. This includes their view of the ledgers, the request/response data and anything else they learn during the execution of the protocol (e.g. learning the secret early from the mempool). From the outside we retrieve `Alice` or `Bob` from the storage layer and ask them for the actions or details about what they know. For example, here's what an `Actor` trait might look like (we may not even need this trait but it's here to help the exposition):

```rust

/// This is what we are able to pull out of the DB. Each variant represents the transactions on the blockchain at the time
enum LedgerState<L: Ledger> {
    None,
    //probably a Deploy in here too but ommitted for clarity
    Funded{
       htlc_location: L::HtlcLocation,
       // we might not even need this if we transition to time based locks rather than block height
       refundable: bool
    },
[...]

PR #752, opened 2019-02-11
Extend GET swaps/rfc003/:id response + state machine re-architecting

### LINKED ISSUE #693: Extend response of HTTP GET for a single swap
This is the implementation ticket for the spike https://github.com/comit-network/comit-rs/issues/497.

The final design is outlined in the following comment: https://github.com/comit-network/comit-rs/issues/497#issuecomment-455029878

luckysori:
The original goal of this PR was to extend the body of the response when sending a get request to `/swaps/rfc003/:id`, as per #693. When attempting to do this, we realised that it would be very difficult to do with the way we store state machine states directly. Therefore the scope of this PR has changed to include part of the famous re-architecting of the state machine, as discussed here #576. 

Fixes #693.


application/comit_node/src/http_api/rfc003/swap.rs
```diff
@@ -273,22 +213,152 @@ fn handle_post_swap<A: AliceSpawner>(
...
+    Swapped,
+    NotSwapped,
+}
+
+#[derive(Debug, Serialize)]
+struct SwapCommunication {
+    current_state: SwapCommunicationState,
     alpha_expiry: Timestamp,
     beta_expiry: Timestamp,
+    alpha_redeem_identity: Option<Value>,
```
thomaseizinger: :question: Should the identities really be a `Value` here?

application/comit_node/src/swap_protocols/rfc003/alice/actions/btc_erc20.rs
```diff
@@ -60,44 +73,25 @@ impl Actions for SwapStates<Alice<Bitcoin, Ethereum, BitcoinQuantity, Erc20Token
     >;
 
     fn actions(&self) -> Vec<Self::ActionKind> {
-        use self::SwapStates as SS;
-        match *self {
-            SS::Accepted(Accepted { ref swap, .. }) => {
-                vec![alice::ActionKind::Fund(swap.fund_action())]
+        let swap_accepted = match self.swap_communication {
+            SwapCommunication::Accepted { ref swap_accepted } => swap_accepted,
+            SwapCommunication::Proposed { .. } | SwapCommunication::Rejected { .. } => {
```
thomaseizinger: :question: Could this line not just be a catch-all?

application/comit_node/src/swap_protocols/rfc003/bitcoin/mod.rs
```diff
@@ -18,9 +18,7 @@ pub use self::{
 };
 
 impl Ledger for Bitcoin {
-    type LockDuration = Blocks;
     type HtlcLocation = OutPoint;
-    type HtlcIdentity = KeyPair;
```
thomaseizinger: :heart_eyes:

application/comit_node/src/swap_protocols/rfc003/bob/actions/mod.rs
```diff
@@ -36,31 +37,33 @@ impl<Accept, Decline, Deploy, Fund, Redeem, Refund>
     }
 }
 
-#[allow(type_alias_bounds)]
-type Response<AL: Ledger, BL: Ledger> =
-    Result<StateMachineResponse<AL::HtlcIdentity, BL::HtlcIdentity>, SwapReject>;
-
-#[derive(Debug, Clone)]
+#[derive(Clone, Derivative)]
+#[derivative(Debug)]
```
thomaseizinger: Why derive `Debug` if all fields are ignored?

application/comit_node/src/swap_protocols/rfc003/bob/mod.rs
```diff
@@ -1,121 +1,162 @@
...
     }
 
-    pub fn decline_action(&self) -> Decline<AL, BL> {
-        Decline::new(self.response_sender.clone())
+    pub fn response_future(&self) -> Option<Box<ResponseFuture<AL, BL>>> {
+        match &self.swap_communication {
+            SwapCommunication::Proposed { response, .. } => {
+                Some(Box::new(response.response_future()))
+            }
+            _ => None,
```
thomaseizinger: :wrench: We could log a `warn` here with the current state because calling this in any state other than `Proposed` will likely yield an error and then this message might be a good hint to discover why.

application/comit_node/src/swap_protocols/rfc003/bob/spawner.rs
```diff
@@ -36,49 +36,38 @@ pub trait BobSpawner: Send + Sync + 'static {
...
-        let save_state = self
-            .state_store
-            .new_save_state(id)
-            .map_err(Error::Storage)?;
+        let swap_seed = Arc::new(self.seed.swap_seed(id));
+        let bob = bob::State::new(swap_request.clone(), swap_seed);
+
+        let response_future = bob
+            .response_future()
+            .expect("bob::State is created in the SwapCommunication::Proposed variant");
```
thomaseizinger: :thinking: The error message knows quite something about the internal state of `Bob` :)

Not sure if we can/should change that.

application/comit_node/src/swap_protocols/rfc003/messages.rs
```diff
@@ -0,0 +1,57 @@
...
+            alpha_ledger: request.alpha_ledger.into(),
+            beta_ledger: request.beta_ledger.into(),
+            alpha_asset: request.alpha_asset.into(),
+            beta_asset: request.beta_asset.into(),
+            role: RoleKind::Alice,
+        }
+    }
+}
+
+pub trait ToFullSwapRequest<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> {
```
thomaseizinger: :thinking: Since the result of the method is named `Request`, maybe this should just be called `ToRequest`?

application/comit_node/src/swap_protocols/rfc003/messages.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn to_full_swap_request(&self, secret_source: &dyn SecretSource) -> Request<AL, BL, AA, BA>;
+}
+
+pub trait ToFullAcceptResponse<AL: Ledger, BL: Ledger> {
```
thomaseizinger: :thinking: Same thing here.

application/comit_node/src/swap_protocols/rfc003/state_machine.rs
```diff
@@ -535,91 +515,3 @@ impl_display!(AlphaRefundedBetaFunded);
 impl_display!(AlphaFundedBetaRedeemed);
```
thomaseizinger: Do we still need that? I think we don't :)

application/comit_node/src/http_api/rfc003/swap.rs
```diff
@@ -256,11 +284,32 @@ struct SwapCommunication {
...
+    htlc_location: Option<Value>,
+    fund_tx: Option<Value>,
+    redeem_tx: Option<Value>,
+    refund_tx: Option<Value>,
+}
+
+impl Default for LedgerState {
+    fn default() -> Self {
+        Self {
+            current_state: HtlcState::NotDeployed,
```
thomaseizinger: If you make `NotDeployed` the `Default` of `HtlcState`, you can `#[derive(Default)]` here.

application/comit_node/src/http_api/rfc003/swap.rs
```diff
@@ -270,94 +319,137 @@ enum SwapCommunicationState {
...
-// refund_tx: refund_transaction.map(|refund_tx| format!("{:?}", refund_tx)),
-//         }
-//     }
-// }
+impl<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> From<rfc003::messages::Request<AL, BL, AA, BA>>
+    for SwapDescription
+{
+    fn from(request: rfc003::messages::Request<AL, BL, AA, BA>) -> Self {
+        Self {
+            alpha_ledger: request.alpha_ledger.to_http_ledger().unwrap(),
```
thomaseizinger: :thinking: Should we really unwrap here?

application/comit_node/src/http_api/rfc003/action.rs
```diff
@@ -451,7 +383,7 @@ pub fn handle_post<T: MetadataStore<SwapId>, S: StateStore<SwapId>>(
     with_swap_types_bob!(
         &metadata,
         (|| match action {
-            PostAction::Accept => serde_json::from_value::<AcceptSwapRequestHttpBody<AL, BL>>(body)
+            PostAction::Accept => serde_json::from_value::<BobAcceptBody>(body)
```
D4nte: ❓  Is that a type aliasing? Is `ROLE` a type aliasing too? Should we have a format convention for type aliasing to make things easier such as all upper-case?

application/comit_node/src/swap_protocols/rfc003/ledger_state.rs
```diff
@@ -0,0 +1,29 @@
+use crate::swap_protocols::rfc003::ledger::Ledger;
+
+// FIXME: Can't get deploy or fund transactions because of the way the state
```
D4nte: Replace with `TODO` if you intend to keep it.

application/comit_node/src/http_api/rfc003/swap.rs
luckysori: I thought about using `Ledger::Identity`s, but that would mean passing type parameters to some of these HTTP structs. Would that be preferable?

application/comit_node/src/swap_protocols/rfc003/bob/actions/mod.rs
luckysori: I will just allow `missing_debug_implementations`.

application/comit_node/src/swap_protocols/rfc003/state_machine.rs
luckysori: We still log the state transition using these. Should that change?

application/comit_node/src/http_api/rfc003/swap.rs
luckysori: Clippy doesn't let you anyway, so I replaced it with a `new()`.

application/comit_node/src/http_api/rfc003/action.rs
luckysori: That is exactly what @LLFourn suggested, which is why it was introduced here.

PR #755, opened 2019-02-11
Run JS check format before commit

D4nte:
Ditto.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -26,7 +26,7 @@ dependencies = [
 [tasks.pre-build]
 workspace = false
 dependencies = [
-    "check-format",
+    "check-rs-format",
     "check-js-format",
     "clippy"
 ]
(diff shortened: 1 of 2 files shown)


Makefile.toml
```diff
@@ -102,6 +102,13 @@ alias = "lqs-tests-flow"
...
 # Custom tasks #
 
+[tasks.check-all-format]
+workspace = false
+dependencies = [
+    "check-format",
+    "check-js-format",
+]
+
 [tasks.check-format]
```
thomaseizinger: :question: Do we maybe want to rename this to `check-rs-format` for consistency reasons?

Makefile.toml
```diff
@@ -102,6 +102,13 @@ alias = "lqs-tests-flow"
 ################
 # Custom tasks #
 
+[tasks.check-all-format]
```
thomaseizinger: :thinking: Not sure if we actually need this task?
The aggregation (i.e. calling both) could also just happen in the `pre-commit` file. I don't see a usecase where you would call both of them manually.

Would make our Makefile simpler.

PR #756, opened 2019-02-12
A significant Refactor of Ledger Query Service Client

### LINKED ISSUE #763: Deployment and funding transactions not always available
We want to store the IDs of the deployment and funding transactions, but currently `LedgerEvents::htlc_deployed` returns `htlc_location` and `LedgerEvents::htlc_funded` an `Option<FundTransaction>`.

LLFourn:
The motivation for these changes is the following:

- The layer that creates a query should consume the results of the query (rather than passing up a query and letting the generic layer consume the results)
- We should be able to have different types of queries for different ledgers

The code previously was overly generic and restrictive (it was not doing "Tell don't ask"). Things kept getting passed back and forth between layers. For example:

1. The lower layer was asked for the "Query" and then the upper (generic) layer got the results
2. With the resulting transaction the upper layer then sent it down again and asked the lower layer for various things through generic traits like `ExtractSecret` and `FindHtlcLocation`.

These traits are all gone 🎉 and we simply have a trait `HtlcEvents<L: Ledger, A: Asset>` that returns a `Future` for each type of event.

The de-generification of this did create some duplication that could be solved but I didn't see the need to address this right now. When we start using the new query types for Ethereum that @bonomat worked on they may diverge more.

Other things:
- Removed most "validation" code -- The LQS already guarantees things for us. We have to trust it. There's no point in duplicating checks like does the contract bytecode match the desired contract bytecode when that's exactly what we queried for! ~IMO We should offload more validation into the query e.g query for an amount greater than or equal to what was requested.~

- Deleted the `QueryCache`. After looking at it I don't think it was the best approach in the end. Now to make duplicate queries after a comit node restart we can just implement `HtlcEvents` for something with an inner `QueryBitcoin` or `QueryEthereum` that is idempotent because it stores the query for a particular event in a database. 

---

Because of the merge, this PR now also:

Resolves #763.


application/comit_node/src/swap_protocols/rfc003/bitcoin/htlc_events.rs
```diff
@@ -0,0 +1,126 @@
...
+            state_machine::HtlcParams,
+        },
+    },
+};
+use bitcoin_support::{BitcoinQuantity, OutPoint};
+use futures::{
+    future::{self, Either},
+    Future,
+};
+
```
LLFourn: Remove this space.

application/comit_node/src/ledger_query_service/timer_poll_future.rs
```diff
@@ -0,0 +1,217 @@
...
+                is_new
+            }),
+    )
+}
+
+/// Keep polling a future that returns a vec and puts the results
+/// deduplicated results into a stream.
+// Here in case we want to use it later
+#[allow(dead_code)]
+pub fn poll_future_into_stream<
```
LLFourn: Note: this is here in case we want to be able to return query results as a stream (before implementing websockets).

application/comit_node/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -0,0 +1,223 @@
...
+            query_ethereum
+                .transaction_first_result(&query_id)
+                .map_err(rfc003::Error::LedgerQueryService)
+        });
+
+        let htlc_location = transaction.and_then(move |tx| {
+            let actual_value = EtherQuantity::from_wei(tx.value);
+            let required_value = htlc_params.asset;
+
+            if actual_value < required_value {
```
thomaseizinger: :x: I really don't like that functionality like this is not abstracted away by a trait but implemented next to some other boilerplate code that glues stuff together.

Let's discuss this.

D4nte:
> Here are some initial comments. Need to think about this a bit more.
> 
> I am not really happy with all the traits + generic implementation of LqsEvents going away.
> The reason is that now, actually crucial logic that should be shared across multiple ledgers is duplicated. In fact, you have to do many things again in exactly the same way if you want to implement an new ledger.
> 
> Regarding the QueryCache: Did you verify that even without a restart, only 1 query is created under all circumstances? The reason it existed is not to facilitate restarts. It is to guarantee that we don't miss a transaction by accidentally creating a 2nd query in a 2nd invocation of the `poll` where the query is created.

@LLFourn I agree with both these comments. Not sure worth reviewing until this is settled.

thomaseizinger:
> Regarding the QueryCache: Did you verify that even without a restart, only 1 query is created under all circumstances? The reason it existed is not to facilitate restarts. It is to guarantee that we don't miss a transaction by accidentally creating a 2nd query in a 2nd invocation of the `poll` where the query is created.

Not that the current (my) solution is super ugly and I'd rather replace it with something much more explicit. I just didn't know how :)

thomaseizinger:
> We should offload more validation into the query e.g query for an amount greater than or equal to what was requested.

I disagree. I think LQS should return us the transaction that is very likely the *attempt* of the other person to do what they meant to do. The COMIT node can then verify if it actually has the correct amount.

If we offload the amount into the query, we will miss such an attempt and the state machine will never move forward but also the other party will never send a 2nd transaction.

application/comit_node/src/swap_protocols/rfc003/events/ledger_event_futures.rs
```diff
@@ -0,0 +1,62 @@
...
+        Ledger,
+    },
+};
+
+// This is an adaptor struct that exists because our current state
+// machine implementation requires that we return &mut Futures. This
+// is not the actual API we want to implement so we have HtlcEvents
+// where the methods return plain Futures and we save them here so we
+// don't duplicate them at runtime.
+// TODO: Make the state machine save the futures
```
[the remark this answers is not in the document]
LLFourn: TODO: Write the TODO so that it has a nice issue body/title.

application/comit_node/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -0,0 +1,223 @@
...
+use std::sync::Arc;
+
+impl HtlcEvents<Ethereum, EtherQuantity> for Arc<dyn QueryEthereum + Send + Sync + 'static> {
+    fn htlc_deployed(
+        &self,
+        htlc_params: HtlcParams<Ethereum, EtherQuantity>,
+    ) -> Box<Deployed<Ethereum>> {
+        let query_ethereum = Arc::clone(&self);
+        let query = query_ethereum
+            .create(EthereumQuery::Transaction {
```
LLFourn: Create method on EthereumQuery to create a contract creation query given the bytecode.

application/comit_node/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -0,0 +1,223 @@
...
+    fn htlc_redeemed_or_refunded(
+        &self,
+        htlc_params: HtlcParams<Ethereum, EtherQuantity>,
+        htlc_location: &Address,
+    ) -> Box<RedeemedOrRefunded<Ethereum>> {
+        htlc_redeemed_or_refunded(Arc::clone(&self), htlc_params, htlc_location)
+    }
+}
+
+use crate::swap_protocols::rfc003::ethereum::TRANSFER_LOG_MSG;
```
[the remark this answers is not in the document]
LLFourn: I moved this const definitions into this file.

application/comit_node/src/swap_protocols/rfc003/events/ledger_event_futures.rs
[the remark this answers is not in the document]
LLFourn: I ended up deleting the TODO in favour of creating a proper issue.

application/comit_node/src/swap_protocols/rfc003/events/ledger_event_futures.rs
```diff
@@ -0,0 +1,62 @@
...
+#[allow(missing_debug_implementations)]
+pub struct LedgerEventFutures<L: Ledger, A: Asset> {
+    htlc_events: Box<dyn HtlcEvents<L, A>>,
+    htlc_deployed: Option<Box<Deployed<L>>>,
```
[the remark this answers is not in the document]
LLFourn: As discussed, IMO creating your own futures should be reserved for when you need to poll things in certain orders at certain times that the inbuilt future's composition functions don't allow. (e.g what @D4nte had to do with the noise handshake).

application/comit_node/src/swap_protocols/rfc003/ethereum/htlc_events.rs
LLFourn: I moved the responsibility fo checking the asset into the state machine itself -- i.e. the state machine shouldn't transition to Funded unless the asset that the HTLC was funded with is of greater or equal value to the asset that was agreed upon.

LLFourn:
> I am not really happy with all the traits + generic implementation of LqsEvents going away.
> The reason is that now, actually crucial logic that should be shared across multiple ledgers is duplicated. In fact, you have to do many things again in exactly the same way if you want to implement an new ledger.

With your feedback I've constructed it so the query making processing code no longer has to check that the HTLC was funded with the correct asset. The way I did it seems to have worked out well because now you have to return (in the future) the asset that was funded so it's very hard accidentally avoid the check (we have to fudge the ERC20 one until we have the new receipt api).

> Regarding the QueryCache: Did you verify that even without a restart, only 1 query is created under all circumstances? The reason it existed is not to facilitate restarts. It is to guarantee that we don't miss a transaction by accidentally creating a 2nd query in a 2nd invocation of the `poll` where the query is created.

Yes only one query is created under all circumstances as the thing that is polled only asks for the future once. Hopefully we can re-architect the state machine itself to guarantee this for us.

application/comit_node/src/swap_protocols/rfc003/ethereum/htlc_events.rs
thomaseizinger: That is a brilliant idea!

Issue #691, opened 2019-01-23
Transfer omnicoins via HTLC or Multi-input Tx


D4nte:
Follow-up of #666. See [notes](https://github.com/coblox/docs/wiki/Omnilayer)

We need to understand whether Omnilayer wallets[1][2] recognise transfer of ownership through bitcoin scripts (HTLC) or through multi-input transactions.

DoD:
- [x] Create a transaction spending from multiple inputs, 1 for Alice (USDT) and 1 for Bob (BTC) and swaps the ownership, i.e. spends to Alice (BTC) and spends to Bob (USDT). 
- [x] Create P2WSH HTLC on testnet and verify that OL wallets recognize transfer of ownerships
- [x] If P2WSH fails, try with P2SH
- [x] If either succeed, Open ticket to write RFC comit-network/RFCs#33

Child of #663 
Blocked by #724
[1]https://www.omniwallet.org/
[2]https://github.com/OmniLayer/omnicore

D4nte:
Transaction would look like (Alice sells omnicoins to Bob for BTC)
Inputs:
- Bob BTC `bob_address`
- Alice Omni `alice_address`
Outputs:
- Bob Change BTC `bob_address` (must be same than input to avoid omnis linked to it)
- Alice BTC `alice_address` (must be same than input to avoid omnis linked to it)
- Bob Omni `bob_address2` (new address, hopefully omnis will be linked to this address)
- OP_RETURN Omni

D4nte:
All PoC are now demonstrated in #772 

What we learned is that by adding omnilayer payload in an `OP_RETURN` output it is possible to lock omnilayer tokens in an HTLC (P2SH or P2SH(P2WSH)) and then redeem them to a normal P2PKH.

What we can also ponder is that I chose to use our testing suite and bitcoinjs-lib to proceed with this PoC as I thought that updating comit-rs and bsieve would be too much overhead.
However, we created COMIT exactly for this kind of experience so maybe (next time?) we need to do our PoC in COMIT and discover/remove all obstacle that makes such work hard (e.g., being able to not have to override bsieve for a PoC?)

thomaseizinger:
> What we can also ponder is that I chose to use our testing suite and bitcoinjs-lib to proceed with this PoC as I thought that updating comit-rs and bsieve would be too much overhead.
[...]

PR #757, opened 2019-02-14
Bitcoin for Omnilayer Token swap PoC [do not merge]

D4nte:
Proof of Concept for Bitcoin to Omnilayer Token (USDT-like) swap.
Using **one** transaction (using various inputs and outputs to proceed with the exchange).

This is a proof of concept so I aim not to merge but keep the branch open as a partial resolution to #691.

PR can be closed once people are happy with the result.

Added findings in [this comment](https://github.com/comit-network/comit-rs/issues/691#issuecomment-462998340)

I strongly recommend **not to** review the codestyle but only the exactitude of the PoC in term of bitcoin/omnicore concepts.

### DIFF
--- a/api_tests/e2e/rfc003/btc_eth/test.js
+++ b/api_tests/e2e/rfc003/btc_eth/test.js
@@ -18,7 +18,7 @@ const bob = actor.create("bob", {});
 
 const alice_final_address = "0x03a329c0248369a73afac7f9381e02fb43d2ea72";
 const bob_final_address =
-    "bcrt1qs2aderg3whgu0m8uadn6dwxjf7j3wx97kk2qqtrum89pmfcxknhsf89pj0";
+    "mzNFGtxdTSTJ1Lh6fq5N5oUgbhwA7Nm7cA";
 const bob_comit_node_address = bob.config.comit.comit_listen;
 
 const alpha_asset = 100000000;
@@ -242,6 +242,7 @@ describe("RFC003: Bitcoin for Ether", () => {
     });
 
     it("[Alice] Can execute the funding action", async () => {
+        console.log(alice_funding_action);
         alice_funding_action.payload.should.include.all.keys(
             "to",
             "amount",
--- a/api_tests/e2e/rfc003/btc_eth/test.js
+++ b/api_tests/e2e/rfc003/btc_eth/test.js
@@ -17,8 +17,7 @@ const alice = actor.create("alice", {});
 const bob = actor.create("bob", {});
 
 const alice_final_address = "0x03a329c0248369a73afac7f9381e02fb43d2ea72";
-const bob_final_address =
-    "mzNFGtxdTSTJ1Lh6fq5N5oUgbhwA7Nm7cA";
+const bob_final_address = "mzNFGtxdTSTJ1Lh6fq5N5oUgbhwA7Nm7cA";
 const bob_comit_node_address = bob.config.comit.comit_listen;
 
 const alpha_asset = 100000000;
--- a/api_tests/e2e/rfc003/btc_usdt/test.js
+++ b/api_tests/e2e/rfc003/btc_usdt/test.js
@@ -4,6 +4,7 @@ const omnilayer = require("../../../lib/omnilayer.js");
 const actor = require("../../../lib/actor.js");
 const should = chai.should();
 const logger = global.harness.logger;
+const bitcoin = require("bitcoinjs-lib");
 
 const omni_rpc_client = omnilayer.create_client();
 
@@ -25,18 +26,34 @@ describe("RFC003: Bitcoin for USD Tether (Omnilayer)", () => {
         this.timeout(50000);
         //await omnilayer.activate_segwit();
         await alice.wallet.omni().btcFund(1);
+        await bob.wallet.omni().btcFund(1);
         await omnilayer.omni_generate();
     });
 
     let tokenId;
     it("Create RegtestOmniCoin", async function() {
         const res = await alice.wallet.omni().createOmniToken();
-        console.log("Property: " + res.propertyid);
         tokenId = res.propertyid;
     });
 
+    let aliceOmniUTXO;
     it("Grant RegtestOmniCoin", async function() {
-        const res = await alice.wallet.omni().grantOmniToken(tokenId);
-        console.log("balance", res);
+        aliceOmniUTXO = await alice.wallet.omni().grantOmniToken(tokenId);
+    });
+
+    it("Swaperoo it", async function() {
+        const aliceDetails = {
+            alice_keypair: null, //TODO
+            alice_omni_utxo: aliceOmniUTXO,
+            alice_final_address: alice_final_address
+        };
+        const bobDetails = {
+            bob_keypair: null, //TODO
+            bob_btc_utxo: bob.wallet.omni().bitcoin_utxos.shift(),
+            bob_btc_output: bob.wallet.omni().identity().output, //TODO
+            bob_final_address: bob_final_address
+        };
+
+        const res = await omnilayer.swaperoo(aliceDetails, bobDetails, tokenId, 550, 2);
     });
 });
--- a/api_tests/e2e/rfc003/btc_usdt/test.js
+++ b/api_tests/e2e/rfc003/btc_usdt/test.js
@@ -9,15 +9,14 @@ const bob = actor.create("bob", {});
 
 const bob_final_address = "mzNFGtxdTSTJ1Lh6fq5N5oUgbhwA7Nm7cA";
 
-const alpha_asset = 3;
-const beta_asset = 4200;
+const beta_asset = 3;
+const alpha_asset = 4200;
 
 describe("RFC003: Bitcoin for Reg Test Omni Token (USD Tether style)", () => {
     before(async function() {
         this.timeout(50000);
         await omnilayer.activateSegwit();
         await alice.wallet.omni().btcFund(1);
-        await omnilayer.generate();
         await bob.wallet.omni().btcFund(1);
         await omnilayer.generate();
     });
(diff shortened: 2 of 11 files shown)


api_tests/e2e/rfc003/btc_eth/test.js
```diff
@@ -17,8 +17,7 @@ const alice = actor.create("alice", {});
 const bob = actor.create("bob", {});
 
 const alice_final_address = "0x03a329c0248369a73afac7f9381e02fb43d2ea72";
-const bob_final_address =
-    "bcrt1qs2aderg3whgu0m8uadn6dwxjf7j3wx97kk2qqtrum89pmfcxknhsf89pj0";
+const bob_final_address = "mzNFGtxdTSTJ1Lh6fq5N5oUgbhwA7Nm7cA";
```
bonomat: ❓would it work with a bech32 address as well?

api_tests/e2e/rfc003/btc_usdt/test.js
```diff
@@ -0,0 +1,58 @@
...
+        res.propertyid.should.be.a("number");
+        tokenId = res.propertyid;
+    });
+
+    let aliceOmniUTXO;
+    it("Grant Regtest Omni Token", async function() {
+        const grantAmount = beta_asset * 3;
+        aliceOmniUTXO = await alice.wallet.omni().grantOmniToken(tokenId, alice.wallet.omni().identity().output, grantAmount);
+        const balance = await omnilayer.getBalance(tokenId, alice.wallet.omni().identity().address);
+        balance.should.equal(grantAmount.toString());
```
bonomat: Let's add this to ensure Bob has no tokens
```suggestion
        balance.should.equal(grantAmount.toString());
        
        const bob_token_balance = await omnilayer.getBalance(tokenId, bob.wallet.omni().identity().address);
        bob_token_balance.should.equal("0");
```

api_tests/e2e/rfc003/btc_usdt/test.js
```diff
@@ -0,0 +1,58 @@
...
+            bob_keypair: bob.wallet.omni().keypair,
+            bob_btc_utxo: bob.wallet.omni().bitcoin_utxos.shift(),
+            bob_btc_output: bob.wallet.omni().identity().output,
+            bob_final_address: bob_final_address,
+        };
+
+        await omnilayer.swaperoo(aliceDetails, bobDetails, tokenId, beta_asset, alpha_asset);
+
+        const bob_omni_balance = await omnilayer.getBalance(tokenId, bob_final_address);
+        bob_omni_balance.should.equal(beta_asset.toString());
```
bonomat: Something is odd here: 
`await omnilayer.swaperoo(aliceDetails, bobDetails, tokenId, beta_asset, alpha_asset);`
in the function: 
`beta_asset` is  `omni_value`
`alpha_asset` is `btc_value` 

and here you check that Bob holds the `beta_asset`

Shouldn't `alpha_asset` and `beta_asset` be the other way around, i.e. 
Alice holds `alpha_asset` and receives `beta_asset` and Bob holds `beta_asset` at the beginning and then receives `alpha_asset`.

I think the checks are correct, just the naming is a bit off.

api_tests/lib/omnilayer.js
```diff
@@ -0,0 +1,232 @@
...
+    txb.addInput(alice_omni_utxo.txid, alice_omni_utxo.vout, null, alice_output);
+    // Bob BTC input
+    txb.addInput(bob_btc_utxo.txid, bob_btc_utxo.vout, null, bob_btc_output);
+
+    // Add BTC change back to Bob
+    txb.addOutput(bob_btc_output, bob_btc);
+    // Add BTC output to Alice
+    txb.addOutput(alice_output, alice_btc);
+    // Add Omni output to Bob (it's Omni because it's different from the inputs)
+    txb.addOutput(bob_omni_output, dust);
```
bonomat: ❓this can be/should be `0` right?
Your test run with 0 at least.

api_tests/lib/omnilayer.js
```diff
@@ -0,0 +1,232 @@
...
+
+    // Add Omni instructions
+    const payload = await createOmniRpcClient().command([
+        {
+            method: "omni_createpayload_simplesend",
+            parameters: [tokenId, omni_value.toString()],
+        },
+    ]);
+    const embed = payloadToEmbed(payload[0]);
+    txb.addOutput(embed.output, 500);
```
bonomat: ❓I thought this value should be `5460` as this is the dust level... or did you change the dust level to 500 somewhere?

bonomat:
Awesome, good job. I've added some questions which were not clear to me.  

When seeing these lines: 
```
txb.sign(0, alice_keypair, null, null, alice_omni_utxo.value);
txb.sign(1, bob_keypair, null, null, bob_btc_utxo.value);
```
I realized that the approach of having two inputs and swapping in one transaction is actually not the most suitable solution. The problem is that one party has to sign first and send its signed transaction to the counter party. This one now holds a beautiful option and can just wait and don't do anything. If the other party wants to cancel, he has to spend her input to another address. If we would want to do this, we should come up with some kind of penalty transaction construct similar to the one used in LN. 

Otherwise, this approach is actually worse than using HTLCs because one party can literally just wait and sit back and monitor how the rates evolve. If using HTLCs for same chain swaps we can make the time intervals really small and make this unprofitable/or the risk to high. 

Do you want to have a look at using HTLCs for same chain swaps aswell or should we create a new ticket for this? I think we should explore this option as well.

api_tests/e2e/rfc003/btc_eth/test.js
D4nte: I don't think it works but I moved that at the start so I'll test more and provide exact findings regarding bech32

api_tests/e2e/rfc003/btc_usdt/test.js
```diff
@@ -0,0 +1,58 @@
...
+
+const alpha_asset = 3;
+const beta_asset = 4200;
+
+describe("RFC003: Bitcoin for Reg Test Omni Token (USD Tether style)", () => {
+    before(async function() {
+        this.timeout(50000);
+        await omnilayer.activateSegwit();
+        await alice.wallet.omni().btcFund(1);
+        await omnilayer.generate();
```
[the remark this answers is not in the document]
D4nte: Yes, don't get too hang up on the code, the main purpose is to provide a work Omni Token -> BTC transaction and show it works.

api_tests/e2e/rfc003/btc_usdt/test.js
D4nte: yes indeed :)

api_tests/lib/omnilayer.js
D4nte: Ah yes indeed, we do not need dust for the `OP_RETURN` output. I got mistaken. Thanks

D4nte: Hum, I expect a minimum dust amount to be present on this UTXO or the transaction would be rejected.
Indeed, it did not get rejected when I changed it. This is unexpected, I'll do more research.

api_tests/lib/omnilayer.js
D4nte: Yes, I did not remember the dust level but as per my odd comment, the behaviour is unexpected, let me do more research.

D4nte:
Hi @bonomat,

Thanks.

Yes, I came to the same conclusion than you on the fact that the signed transaction must be exchanged between parties over the wire.

As you mentioned, there is a possibility for the first party to spend the UTXO before the swap transaction is broadcast to the blockchain and hence cancel the swap.
This gives actually more flexibility to "Bob" (let's assume Bob signs first) as he could literally say to Alice "I'll wait 1 minute and if you don't send me the fully sign transaction by then I'll spend my UTXO).

That is one way.
The other way as you mentioned would be to include an option fee that goes to Bob if Alice decides not to sign.

I will clarify your comment and then proceed with the HTLC part as per the issue DoD.

LLFourn:
Nice work.

With regards to the above conversation, In order to prove that there is no "option problem" you have to show that the protocol offers `MakeOffer`, `TakeOffer` and `CancelOffer` functionality. To me it looks like this is satisfied:

- MakeOffer: Bob signs his bitcoin input
- TakeOffer: Alice signs her Omni input
- CancelOffer: Bob double spends the his bitcoin input

`MakeOffer` is instant while TakeOffer and CancelOffer happen as fast as it takes the participant to confirm a transaction (which is optimal). Even though this protocol is optimal, one party can still try and play the exchange rate if it can confirm transactions faster than the other one (by speeding themselves up or slowing the other one down). Note: we should always make Alice the person going first!!!

### Non-interactive mutli-party asset exchange
This code demonstrates an interactive protocol (both parties need to share upfront their addresses etc so they can both construct the transaction). It is like the fair swap except that you don't have to put any collateral.

The question is can this be done where Alice can make `MakeOffer` to anyone in the world who has Bitcoin (by pasting a partially signed tx in a chat channel for example) and have anyone in the world be able to `TakeOffer` without them exchanging messages first?

I'm not sure. The question I have is how does omni handle "change" of the omni tokens? I made a diagram to clarify my understanding:

https://docs.google.com/drawings/d/1prjZB6A2oVkNY5LVWFvCeNrorijuOe7CBOXX_rG-9q0/edit?usp=sharing

It looks like you could do this where Alice does `MakeOffer` uses `SIGHASH_SINGLE` to sign the input and output in green and then Bob fills in the blue bits to `TakeOffer`. But I don't know how change works in omni so I guess Bob always take the entirety of the tokens in the omni output?

D4nte:
@llfourn there is no "change" in OmniLayer.
If the UTXO has 100 Omni Tokens and you do a simple send of 10 (you set the amount in the `OP_RETURN` payload) then the output has 10 and the UTXO that had 100 now has 90.

thomaseizinger:
One question:

Would there be a way to design the protocol such as that both parties end up with a transaction already signed by the other party?

This would allow both parties to go through with the trade at any time, similar to the LN channel closing TX.

api_tests/lib/omnilayer.js
D4nte: Ok, the issue is that dust checks are deactivated on regtest by default. Now sorted, see https://github.com/comit-network/comit-rs/issues/691#issuecomment-462998340 for details

api_tests/lib/omnilayer.js
D4nte: See https://github.com/comit-network/comit-rs/issues/691#issuecomment-462998340 for details

D4nte:
@thomaseizinger I don't think so.

api_tests/e2e/rfc003/btc_eth/test.js
D4nte: omnicored does not support bech32 https://github.com/OmniLayer/omnicore/blob/master/src/omnicore/doc/release-notes.md#internal-preperations-for-native-segregated-witness-support

So yes, we must use base58 addresses.

D4nte:
All comments are now addressed. PR can be closed.

LLFourn:
@thomaseizinger https://crypto.stackexchange.com/questions/35828/is-it-possible-for-alice-and-bob-to-both-sign-a-message-simultaneously/36860#36860

PR #758, opened 2019-02-15
[WIP] Reject non standard transactions

D4nte:
Just realised that in regtest/testnet `-acceptnonstdtxn` is activated by default.
Best to deactivate it to ensure our code would work in mainnet. 
Need to look at doing the same in testcontainers.


D4nte:
It's failing so investigation work can be done with #762

PR #774, opened 2019-02-19
Rename Ledger Query Service to btsieve

### LINKED ISSUE #749: Rename ledger_query_service to btsieve
Code
GitHub

luckysori:
Fixes #749


api_tests/btsieve/config.toml
```diff
@@ -0,0 +1,6 @@
+ledgers = ["bitcoin", "ethereum"]
```
D4nte: can you `git mv` this file instead?

luckysori: As discussed, I tried but the problem is that both the name of the folder and the contents of the file changed, and `git` seems to only be able to represent this by deleting and adding, rather than by modifying.

PR #782, opened 2019-02-21
Minor name changes in get swap/list swaps http api responses

### LINKED ISSUE #775: Polish swap response in HTTP API
I came across the spec (https://github.com/comit-network/comit-rs/issues/497#issuecomment-455029878) of the swap response again and noticed two minor mis-alignments.

1. The name `state.outcome` property doesn't really make sense for the variant `IN_PROGRESS`.
I propose renaming it to `status` and actually *moving* it one layer up. This makes more sense IMO since it describes an overall property of the swap, similar to `role`.

2. The property `swap` is kind of redundant. Since the whole response is about a single swap. Different names could be:
- `details`
- `parameters`
- ???

D4nte:
Resolves #775

### DIFF
--- a/api_tests/dry/multiple_peers.js
+++ b/api_tests/dry/multiple_peers.js
@@ -104,7 +104,7 @@ it("[Alice] Should be IN_PROGRESS and SENT after sending the swap request to Cha
         chai,
         alice_swap_with_charlie_href,
         body =>
-            body.state.outcome == "IN_PROGRESS" &&
+            body.status == "IN_PROGRESS" &&
             body.state.communication.current_state == "SENT"
     );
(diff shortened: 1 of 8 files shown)


luckysori:
Should have mentioned this before, but how about also replacing the properties `state.communication.current_state` and `state.xxx_ledger.current_state` with `state.communication.status` and `state.xxx_ledger.status` respectively. It's shorter, less redundant and consistent.

thomaseizinger:
Is description really the best word for this? I would favor something like _parameters_. A description is something textual to me.

D4nte:
Will change to parameters.

PR #783, opened 2019-02-22
Add actions to GET swaps

### LINKED ISSUE #776: Extend `EmbeddedSwapResource` for GET /swaps with swap parameters and available actions
Currently, the `EmbeddedSwapResource` only contains the current state and the protocol.

https://github.com/comit-network/comit-rs/blob/c797f951726b6fc032c1e426cfbb8686aa05ffda/application/comit_node/src/http_api/rfc003/handlers/get_swaps.rs#L17-L21

However, if applications want to provide a list of current swaps (with their parameters inlined), the need to do n+1 requests to the COMIT node.

I propose to add the parameters of the swap + the available actions to the `EmbeddedSwapResource`.

D4nte:
Add parameters and actions to `GET swapS` response.
Also "fix" tests to test for type (using `===`).

Resolves #776

### DIFF
--- a/api_tests/dry/rfc003_swap_reject.js
+++ b/api_tests/dry/rfc003_swap_reject.js
@@ -235,8 +235,8 @@ it("[Bob] Shows the swaps as Start in /swaps", async () => {
     for (let swap of swaps) {
         swap.protocol.should.equal("rfc003");
         swap.status.should.equal("IN_PROGRESS");
-        swap._links.accept.href.should.be.a("string");
-        swap._links.decline.href.should.be.a("string");
+        swap._links.accept.should.be.a("object");
+        swap._links.decline.should.be.a("object");
     }
 
     let swap_1_link = swaps[0]._links.self;
--- a/api_tests/dry/rfc003_swap_reject.js
+++ b/api_tests/dry/rfc003_swap_reject.js
@@ -227,7 +227,7 @@ it("[Bob] Shows the swaps as Start in /swaps", async () => {
     let body = await bob.poll_comit_node_until(
         chai,
         "/swaps",
-        body => body._embedded.swaps.length == 2
+        body => body._embedded.swaps.length === 2
     );
 
     let swaps = body._embedded.swaps;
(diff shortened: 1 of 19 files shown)


api_tests/dry/rfc003_swap_reject.js
```diff
@@ -191,14 +227,16 @@ it("[Bob] Shows the swaps as Start in /swaps", async () => {
...
-        body => body._embedded.swaps.length == 2
+        body => body._embedded.swaps.length === 2
     );
 
     let swaps = body._embedded.swaps;
 
     for (let swap of swaps) {
         swap.protocol.should.equal("rfc003");
         swap.status.should.equal("IN_PROGRESS");
+        swap._links.accept.href.should.be.a("string");
```
thomaseizinger: :x: I think we should just test that the properties "accept" and "decline" exist. No need to test if rustic-hal serializes correctly.

application/comit_node/src/http_api/rfc003/handlers/get_swaps.rs
```diff
@@ -37,17 +38,27 @@ pub fn handle_get_swaps<T: MetadataStore<SwapId>, S: StateStore>(
...
                             status,
                             protocol: RFC003.into(),
+                            parameters,
                         };
 
                         let mut hal_resource = HalResource::new(swap);
                         hal_resource.with_link("self", swap_path(id));
+
+                        for action in actions {
+                            let route = format!("{}/{}", swap_path(id), action);
```
thomaseizinger: :question: Do we not have a method somewhere that generates these? We shouldn't duplicate that ...

application/comit_node/src/http_api/rfc003/routes.rs
```diff
@@ -48,9 +51,10 @@ pub fn get_swap<T: MetadataStore<SwapId>, S: StateStore>(
         .map(|(swap_resource, actions)| {
             let mut response = HalResource::new(swap_resource);
             for action in actions {
-                let route = format!("{}/{}", swap_path(id), action);
-                response.with_link(action, route);
+                let link = new_hal_link(&id, action);
+                response.with_link(action.to_string(), link);
```
thomaseizinger: :wrench: I think `with_link` takes anything that goes `Into<String>`. You could try omitting the `to_string()` here.

PR #784, opened 2019-02-22
Extract Secret and Asset from contract logs for Ethereum HTLCs

### LINKED ISSUE #638: ERC20/ETH Htlc log needs to emit secret
While doing #412 a security issue was found: 

We cannot rely on the fact that the transaction redeeming an HTLC:
a) was sent to our HTLC
b) nor contains the secret in the data field

For example, one could create another contract (contract A), which `calls` the HTLC contract (contract B) in order to redeem it. The transaction invoking contract A might not have the secret as data, and the `call()` to contract B is not tracked in the transaction receipt. 

Hence, our HTLCs need to log the secret when it was successfully redeemed.

Blocked by #756

thomaseizinger:
With this PR, we are now properly extracting the secret from the logs of our HTLCs instead of looking that the calldata for the found transaction.

This fixes #638.

We are also using the transaction receipt to extract the ERC20 token amount that was transferred. This allows us to properly validate that enough tokens were transferred to the HTLC.


thomaseizinger:
> Nice work. Would prefer to rename `Transform` to `ToHttpPayload`.

That conflicts with Rust's naming convention to only prefix methods with `to_` that take `&self` and convert it to an owned value.

But I agree that something with `HttpPayload` would be a better name. What do you think of `CreateHttpPayload`?

thomaseizinger:
I renamed it to `ResultToHttpPayload`. That should express very clearly what the trait is about :)

LLFourn:
What about just implementing `ToHttpPayload` on QueryResult and changing the trait to take &self as the first parameter?

thomaseizinger:
I had to get out my best Rust-fu but I managed to do it :D 

Tell me what you think.

LLFourn:
It reads better now. Well done. Not sure exactly why you had to add the Query parameter but I'm sure it has something to do with not being able to implement the same trait twice for the same type :)

application/btsieve/src/route_factory.rs
```diff
@@ -51,16 +47,23 @@ impl RouteFactory {
...
         &self,
         query_repository: Arc<QR>,
         query_result_repository: Arc<QRR>,
-        client: Option<Arc<<Q as ExpandResult>::Client>>,
+        client: Arc<C>,
         ledger_name: &'static str,
-    ) -> BoxedFilter<(impl Reply,)> {
+    ) -> BoxedFilter<(impl Reply,)>
+    where
+        for<'de> R: Deserialize<'de>,
```
LLFourn: ❓ `where for` -- what is this?

thomaseizinger:
> Not sure exactly why you had to add the Query parameter

Without it, rustc is not able to infer what goes together :D 

> `where for` -- what is this?

Has something to do with lifetimes and allows you do specify `'de` for the `Deserialize` trait without declaring it for the local fn. I never write this myself but just let rustc tell me to do it and copy it from the error :D

application/btsieve/src/bitcoin/queries/transaction.rs
```diff
@@ -22,31 +21,55 @@ impl QueryType for TransactionQuery {
...
+}
 
-            let transaction = client
-                .get_raw_transaction_verbose(&tx_id)
-                .map_err(Error::BitcoinRpcConnection)?
-                .map_err(Error::BitcoinRpcResponse)?;
-            expanded_result.push(transaction.into());
-        }
-        Ok(expanded_result)
+fn convert_to_payload(
```
luckysori: :upside_down_face: Unless I'm missing something, this could be called `to_payload` to keep it consistent with what you do in `ethereum/queries/transaction.rs`

application/btsieve/src/ethereum/queries/event.rs
```diff
@@ -114,49 +113,94 @@ impl QueryType for EventQuery {
...
             .wait()
     }
 }
 
+fn to_payload(
+    client: &Web3<Http>,
+    transaction_id: H256,
+    return_as: &ReturnAs,
+) -> Box<dyn Future<Item = Option<PayloadKind>, Error = Error>> {
+    let transaction_future = client
```
luckysori: :+1: This works because these are evaluated lazily, right?

application/btsieve/src/ethereum/queries/event.rs
```diff
@@ -114,49 +113,94 @@ impl QueryType for EventQuery {
...
+        ReturnAs::TransactionAndReceipt => Box::new(transaction_future.join(receipt_future).map(
+            move |maybe| match maybe {
+                (Some(transaction), Some(receipt)) => Some(PayloadKind::TransactionAndReceipt {
+                    transaction,
+                    receipt,
+                }),
+                // TODO: Investigate under which circumstances the ethereum node fails to return a
+                // transaction/receipt
+                (None, _) => {
+                    warn!(
```
luckysori: :question: Why do you log the absence of a return element in the `ReturnAs::TransactionAndReceipt` match arm and not for the other ones?

application/btsieve/src/ethereum/queries/transaction.rs
```diff
@@ -66,45 +65,58 @@ impl QueryType for TransactionQuery {
...
-        result: &QueryResult,
-        client: Arc<Web3<Http>>,
+    fn to_http_payload(
+        &self,
+        return_as: &ReturnAs,
+        client: &Web3<Http>,
     ) -> Result<Vec<Self::Item>, Error> {
-        let futures: Vec<_> = result
-            .0
+        let to_payload = |transaction_id: H256| to_payload(client, transaction_id, return_as);
```
luckysori: :+1: Makes it so much clearer to read when you extract it like this.

application/btsieve/src/route_factory.rs
```diff
@@ -25,24 +24,21 @@ pub trait QueryType {
     fn route() -> &'static str;
 }
 
-pub trait ExpandResult {
+pub trait ToHttpPayload<Q, R> {
```
luckysori: :question: Where is the `Q` type parameter used?

application/comit_node/src/btsieve/client.rs
```diff
@@ -228,6 +264,46 @@ mod ethereum {
...
+            query: &QueryId<Ethereum>,
+        ) -> Box<dyn Future<Item = TransactionAndReceipt, Error = Error> + Send> {
+            let poll_client = self.client.clone();
+            let query = query.clone();
+
+            poll_until_item(self.ethereum_poll_interval, move || {
+                let mut url = query.as_ref().clone();
+                url.set_query(Some("return_as=transaction_and_receipt"));
+
+                let transactions = poll_client
```
luckysori: :question: Isn't this variable name misleading if we're only returning the first result?

application/comit_node/src/swap_protocols/rfc003/ethereum/contract_templates/erc20_contract.asm
```diff
@@ -84,7 +84,7 @@ exit:
 
 */
 redeem:
-    log1(0, 0, 0xB8CAC300E37F03AD332E581DEA21B2F0B84EAAADC184A295FEF71E81F44A7413) // log keccak256(Redeemed())
+    log1(0, 32, 0xB8CAC300E37F03AD332E581DEA21B2F0B84EAAADC184A295FEF71E81F44A7413) // log keccak256(Redeemed())
```
luckysori: :wrench: Modify the comment to state that the log includes the secret and where?

application/comit_node/src/swap_protocols/rfc003/ethereum/contract_templates/ether_contract.asm
```diff
@@ -67,7 +67,7 @@ exit:
     return(0, 0)
 
 redeem:
-    log1(0, 0, 0xB8CAC300E37F03AD332E581DEA21B2F0B84EAAADC184A295FEF71E81F44A7413) // log keccak256(Redeemed())
+    log1(0, 32, 0xB8CAC300E37F03AD332E581DEA21B2F0B84EAAADC184A295FEF71E81F44A7413) // log keccak256(Redeemed())
```
luckysori: :wrench: Modify the comment to state that the log includes the secret and where?

application/comit_node/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -109,18 +110,26 @@ fn htlc_redeemed_or_refunded<A: Asset>(
...
-                    extract_secret(&transaction, &htlc_params.secret_hash).ok_or_else(|| {
-                        rfc003::Error::Internal(
-                            "Redeem transaction didn't have the secret in it".into(),
-                        )
-                    })?;
-                Ok(Redeemed {
-                    transaction,
-                    secret,
+            .and_then(move |TransactionAndReceipt {transaction, receipt}| {
+
```
luckysori: :question: Is this newline intentional?

application/comit_node/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -109,18 +110,26 @@ fn htlc_redeemed_or_refunded<A: Asset>(
...
+                    .into_iter()
+                    .find(|log| log.topics.contains(&REDEEM_LOG_MSG.into()))
+                    .ok_or_else(|| {
+                        rfc003::Error::Internal(format!("transaction receipt {:?} did not contain a REDEEM log", transaction.hash))
+                    }).and_then(|log| {
+
+                    let log_data = log.data.0.as_ref();
+                    let secret = Secret::from_vec(log_data)
+                        .map_err(|e| rfc003::Error::Internal(format!("failed to construct secret from data in transaction receipt {:?}: {:?}", transaction.hash, e)))?;
+
```
luckysori: :question: Is this newline intentional?

application/btsieve/src/ethereum/queries/event.rs
thomaseizinger: Yes correct.

application/btsieve/src/route_factory.rs
thomaseizinger: It is not used at all but without it, rustc cannot infer how this method is used in the code because the implementation is not bound to the query :(

application/btsieve/src/bitcoin/queries/transaction.rs
```diff
@@ -22,31 +21,55 @@ impl QueryType for TransactionQuery {
...
+    return_as: &ReturnAs,
+    id: TransactionId,
+) -> Result<PayloadKind, Error> {
+    match return_as {
+        ReturnAs::TransactionId => Ok(PayloadKind::Id { id }),
+        ReturnAs::Transaction => match client.get_raw_transaction_verbose(&id) {
+            Ok(Ok(transaction)) => Ok(PayloadKind::Transaction {
+                transaction: transaction.into(),
+            }),
+            Ok(Err(e)) => Err(Error::BitcoinRpcResponse(e)),
```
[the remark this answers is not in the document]
thomaseizinger: Do you mean instead of erroring out? I don't think so since whoever calls this endpoint has expectations of what is returned because they set the query parameter. Returning something different will break them at runtime ...

application/btsieve/src/ethereum/queries/event.rs
thomaseizinger: Good question :D 
Should probably log it for all of them :)

application/btsieve/src/route_factory.rs
luckysori: Thanks for explaining! I think the type parameter is fine :grinning:

application/comit_node/src/btsieve/client.rs
thomaseizinger: This one will return all results so I would say the variable name is ok. The whole thing is in a closure that only returns the first item eventually.

application/comit_node/src/swap_protocols/rfc003/ethereum/htlc_events.rs
thomaseizinger: No.

application/comit_node/src/swap_protocols/rfc003/ethereum/htlc_events.rs
thomaseizinger: Yes.

application/btsieve/src/route_factory.rs
thomaseizinger: Has something to do with lifetimes and allows you do specify `'de` for the `Deserialize` trait without declaring it for the local fn. I never write this myself but just let rustc tell me to do it and copy it from the error :D

thomaseizinger:
@comit-network/rust-devs please vote with thumbs-up/down where you would like it better without a query-param:

https://github.com/comit-network/comit-rs/compare/638-emit-secret-in-htlc...without-query-param

application/btsieve/src/ethereum/queries/mod.rs
```diff
@@ -1,7 +1,82 @@
...
+    Receipt {
+        receipt: Box<TransactionReceipt>,
+    },
+    TransactionAndReceipt {
+        transaction: Box<Transaction>,
+        receipt: Box<TransactionReceipt>,
+    },
+}
+
+// TODO: Investigate under which circumstances the ethereum node fails to return
```
D4nte: ❓ would such investigation be worth it? What would you do with the result.
I feel like we cannot really answer this and hence questioning the need for a TODO

application/btsieve/src/route_factory.rs
D4nte: I thought you were better than that @thomaseizinger  😄

application/btsieve/src/ethereum/queries/mod.rs
thomaseizinger: Resolved!

PR #785, opened 2019-02-25
Take direct control of serialization in HTTP API

### LINKED ISSUE #765: Don't unwrap in constructor of `SwapDescription`
Currently, we are unwrapping the constructor of `SwapDescription` after the `to_http_ledger` and `to_http_asset` call:

https://github.com/comit-network/comit-rs/blob/6cc6ea0b1eb9018c256cfb7d5c498098ff53c2ed/application/comit_node/src/http_api/rfc003/handlers/get_swap.rs#L136-L147

Even though the error is quite rare, we should not do this.

Passing on the error is tricky since there are different ones for `HttpLedger` and `HttpAsset`.
Some investigation on serialization uncovered that there might actually be a better way to achieve contextual serialization, which is the reason why `HttpLedger` and `HttpAsset` were introduced.

A different approach would be to create a generic wrapper struct `Http<T>` that is used in all payloads where a different serialization for a type is needed, for example:

```rust
#[derive(Clone, Debug, Deserialize, PartialEq)]
#[serde(deny_unknown_fields)]
pub struct SwapRequestBody<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset, PartialIdentities> {
    alpha_asset: Http<AA>,
    beta_asset: Http<BA>,
    alpha_ledger: Http<AL>,
    beta_ledger: Http<BL>,
    alpha_expiry: Timestamp,
    beta_expiry: Timestamp,
    #[serde(flatten)]
    partial_identities: PartialIdentities,
    #[serde(with = "http_api::rfc003::socket_addr")]
    peer: SocketAddr,
}
```

Since `Http<T>` is a local type, one can `impl Serialize for Http<Bitcoin>` to define a serialization for this context. Doing so would remove the need for `HttpLedger` and `HttpAsset` which would also simply get rid of the code that contains the `unwrap` that is mentioned at the beginning of this ticket.

As part of doing this, you will probably need [custom trait bounds for serde](https://serde.rs/container-attrs.html#bound).

### LINKED ISSUE #777: GET /swaps/rfc003/:id does not return transaction IDs
`GET /swaps/rfc003/:id` should return transactions IDs but is instead returning transaction objects. This could be solved in several ways:
1. Storing `Ledger::TxId`s instead of `Ledger::Transaction`s in the storage layer.
2. Implementing the serialization of these transactions so that it returns the transaction ID.
3. Implementing `Into<TxId> for Transaction` for all (`Transaction`, `TxId`) pairs and adding the corresponding trait bound on the `Ledger` trait.

luckysori:
A new struct `Http<I>(I)` is introduced to control the serialization of type `I` at HTTP API level. This is used for `Ledger`, `Asset`, `Identity`, `HtlcLocation` and `Transaction`.

Specifically, the `Transaction` type for both Bitcoin and Ethereum now serializes to a transaction ID, which resolves #777.

Also resolves #765.


application/comit_node/src/http_api/impl_serialize_http.rs
```diff
@@ -0,0 +1,30 @@
+macro_rules! _count {
+    () => (0usize);
+    ($x:tt $($xs:tt)*) => (1usize + _count!($($xs)*));
+}
```
LLFourn: ❗️ this looks like you had some fun. Explain this to me tomorrow :)

application/comit_node/src/http_api/rfc003/handlers/get_swap.rs
```diff
@@ -61,28 +64,40 @@ pub fn handle_get_swap<T: MetadataStore<SwapId>, S: StateStore>(
 }
 
 #[derive(Debug, Serialize)]
-#[serde(bound = "AL: Ledger, BL: Ledger")]
-pub struct GetSwapResource<AL: Ledger, BL: Ledger> {
-    parameters: SwapParameters,
+#[serde(
+    bound = "Http<AL>: Serialize, Http<BL>: Serialize, Http<AA>: Serialize, Http<BA>: Serialize,\
```
LLFourn: ❗️ also this

application/comit_node/src/http_api/impl_serialize_http.rs
luckysori: I'd love to take credit, but I took it from [here](https://stackoverflow.com/a/34324856).

application/comit_node/src/http_api/rfc003/handlers/get_swap.rs
luckysori: And this I figured out with @thomaseizinger's help :stuck_out_tongue:

PR #772, opened 2019-02-19
Omnilayer tokens through multi-input tx,  P2SH & P2SH(P2WSH) HTLCs [Do Not Merge]

D4nte:
This contains all PoCs for Omnilayer protocol:
1. BTC/Omnilayer Token swap using multi-input transaction (already showcased with #757)
2. Omnilayer transfered from Alice to Bob throught P2SH HTLC
3. Omnilayer transfered from Alice to Bob throught P2SH(P2WSH) HTLC (please not the wrapping in P2SH due to lack of bech32 support from omnicored)

Resolves #691

Issue #767, opened 2019-02-18
Consolidate e2e tests


luckysori:
There is excessive code repetition in e2e tests. These tests should be generalised since for every pair of ledgers we follow pretty much the same steps. This should make it much easier to modify them in the future or add more ledger pair combinations.

Additionally, consider adding test logs.

D4nte:
Can you provide some code block examples on the repetition and what you are aiming to delete?

luckysori:
> Can you provide some code block examples on the repetition and what you are aiming to delete?

I think @thomaseizinger may be able to better explain his vision, since he suggested this ticket.

In terms of examples, putting [btc_eth/test.js](https://github.com/comit-network/comit-rs/blob/c797f951726b6fc032c1e426cfbb8686aa05ffda/api_tests/e2e/rfc003/btc_eth/test.js) and  [eth_btc/test.js](https://github.com/comit-network/comit-rs/blob/master/api_tests/e2e/rfc003/eth_btc/test.js) side by side makes this issue quite obvious, because they're pretty much identical except for things like Bob's response, which could be stored in variables.

thomaseizinger:
> In terms of examples, putting [btc_eth/test.js](https://github.com/comit-network/comit-rs/blob/c797f951726b6fc032c1e426cfbb8686aa05ffda/api_tests/e2e/rfc003/btc_eth/test.js) and [eth_btc/test.js](https://github.com/comit-network/comit-rs/blob/master/api_tests/e2e/rfc003/eth_btc/test.js) side by side makes this issue quite obvious, because they're pretty much identical except for things like Bob's response, which could be stored in variables.

Yes this is basically it.
The tests only test the workflow of doing a swap, which is the same for all swaps since
- the flow is defined by `rfc003`
- the actions are self-descriptive and the tests don't care
- the parameters are set in the request by Alice and in the response of Bob, both could be dynamically fetched from the _instance_ of the actors (Alice and bob)

LLFourn:
quick note about what this looks like in my mind. I want to be able to specify a test generically:

```
alice_choices = ["fund", "redeem"];
bob_choices = ["fund", "redeem"];

let result = await test_executor.execute(alice_choices, bob_choices);
// check the correct balances at the end
```

There area few ways we could arrange this but the important part is that we can declare Alice and Bob's behaviour and test the result. We would also probably want a callback for each action so you could do some particular tests that are run at that particular stage.

thomaseizinger:
[...]

PR #787, opened 2019-02-26
Verify Bitcoin redeem transaction using transaction ID

### LINKED ISSUE #681: Stop using import_address in e2e
Having the bitcoin_rpc have some state from our tests is a bad idea. We should remove this before trying to run our tests without restarting docker.

When #497 is implemented we should simply verify that the transaction that the state claims to exist does exist and the output has the assets as described in the swap request

DOD:
- No more `importAddress` calls in e2e

Blocked by #693.

### LINKED ISSUE #698: Add it back to UTXOs after transaction is successful...
https://github.com/comit-network/comit-rs/blob/6dac5ba4f1e54aee4d138716d02fd438c83402c7/api_tests/lib/bitcoin.js#L103-L104

* Found on master

luckysori:
When verifying that a redeem transaction has been successful, instead of watching the recipient's Bitcoin address using a Bitcoin RPC client we now look at the transaction itself using `getRawTransaction(txid)`.

Also test suite logs have been removed because no one is looking at them anymore.

Resolves #698.
Resolves #681.

### DIFF
--- a/api_tests/e2e/rfc003/btc_eth-erc20/test.js
+++ b/api_tests/e2e/rfc003/btc_eth-erc20/test.js
@@ -8,7 +8,6 @@ const ethereum = require("../../../lib/ethereum.js");
 const should = chai.should();
 const wallet = require("../../../lib/wallet.js");
 const sb = require("satoshi-bitcoin");
-const logger = global.harness.logger;
 
 const toby_wallet = wallet.create();
 
@@ -98,7 +97,6 @@ describe("RFC003: Bitcoin for ERC20", () => {
 
         res.should.have.status(201);
         swap_location = res.headers.location;
-        logger.info("Alice created a new swap at %s", swap_location);
         swap_location.should.be.a("string");
         alice_swap_href = swap_location;
     });
@@ -130,7 +128,6 @@ describe("RFC003: Bitcoin for ERC20", () => {
         swap_link.should.be.a("object");
         bob_swap_href = swap_link.self.href;
         bob_swap_href.should.be.a("string");
-        logger.info("Bob discovered a new swap at %s", bob_swap_href);
     });
 
     let bob_accept_href;
@@ -150,12 +147,6 @@ describe("RFC003: Bitcoin for ERC20", () => {
             beta_ledger_refund_identity: bob.wallet.eth().address(),
             alpha_ledger_redeem_identity: null,
         };
-
-        logger.info(
-            "Bob is accepting the swap via %s with the following parameters",
-            bob_accept_href,
-            bob_response
-        );
 
         let accept_res = await chai
             .request(bob.comit_node_url())
@@ -180,11 +171,6 @@ describe("RFC003: Bitcoin for ERC20", () => {
             .get(alice_fund_href);
         res.should.have.status(200);
         alice_fund_action = res.body;
-
-        logger.info(
-            "Alice retrieved the following funding parameters",
-            alice_fund_action
-        );
     });
 
     it("[Alice] Can execute the fund action", async () => {
@@ -209,11 +195,6 @@ describe("RFC003: Bitcoin for ERC20", () => {
         let res = await chai.request(bob.comit_node_url()).get(bob_deploy_href);
         res.should.have.status(200);
         bob_deploy_action = res.body;
-
-        logger.info(
-            "Bob retrieved the following deployment parameters",
-            bob_deploy_action
-        );
     });
 
     it("[Bob] Can execute the deploy action", async () => {
@@ -240,11 +221,6 @@ describe("RFC003: Bitcoin for ERC20", () => {
         let res = await chai.request(bob.comit_node_url()).get(bob_fund_href);
         res.should.have.status(200);
         bob_fund_action = res.body;
-
-        logger.info(
-            "Bob retrieved the following funding parameters",
-            bob_fund_action
-        );
     });
 
     it("[Bob] Can execute the fund action", async () => {
@@ -274,11 +250,6 @@ describe("RFC003: Bitcoin for ERC20", () => {
             .get(alice_redeem_href);
         res.should.have.status(200);
         alice_redeem_action = res.body;
-
-        logger.info(
-            "Alice retrieved the following redeem parameters",
-            alice_redeem_action
-        );
     });
 
     let alice_erc20_balance_before;
@@ -331,11 +302,6 @@ describe("RFC003: Bitcoin for ERC20", () => {
             );
         res.should.have.status(200);
         bob_redeem_action = res.body;
-
-        logger.info(
-            "Bob retrieved the following redeem parameters",
-            bob_redeem_action
-        );
     });
 
     it("[Bob] Can execute the redeem action", async function() {
--- a/api_tests/e2e/rfc003/btc_eth-erc20/test.js
+++ b/api_tests/e2e/rfc003/btc_eth-erc20/test.js
@@ -314,7 +314,7 @@ describe("RFC003: Bitcoin for ERC20", () => {
         let body = await bob.poll_comit_node_until(
             chai,
             bob_swap_href,
-            body => body.state.alpha_ledger.status == "Redeemed"
+            body => body.state.alpha_ledger.status === "Redeemed"
         );
         let bob_redeem_txid = body.state.alpha_ledger.redeem_tx;
 
--- a/api_tests/e2e/rfc003/btc_eth-erc20/test.js
+++ b/api_tests/e2e/rfc003/btc_eth-erc20/test.js
@@ -318,7 +318,7 @@ describe("RFC003: Bitcoin for ERC20", () => {
         );
         let bob_redeem_txid = body.state.alpha_ledger.redeem_tx;
 
-        let bob_satoshi_received = await bitcoin.get_satoshi_transferred_to(
+        let bob_satoshi_received = await bitcoin.get_first_utxo_value_transferred_to(
             bob_redeem_txid,
             bob_final_address
         );
--- a/api_tests/e2e/rfc003/btc_eth-erc20/test.js
+++ b/api_tests/e2e/rfc003/btc_eth-erc20/test.js
@@ -7,9 +7,8 @@ const ethutil = require("ethereumjs-util");
 const ethereum = require("../../../lib/ethereum.js");
 const should = chai.should();
 const wallet = require("../../../lib/wallet.js");
+const sb = require("satoshi-bitcoin");
 const logger = global.harness.logger;
-
-const bitcoin_rpc_client = bitcoin.create_client();
 
 const toby_wallet = wallet.create();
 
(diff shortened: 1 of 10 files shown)


api_tests/e2e/rfc003/eth_btc/test.js
```diff
@@ -302,32 +201,34 @@ describe("RFC003: Ether for Bitcoin", () => {
...
+        let alice_redeem_txid = body.state.beta_ledger.redeem_tx;
+        let alice_redeem_tx = await bitcoin.getRawTransaction(
+            alice_redeem_txid
         );
 
-        const alice_btc_balance_expected =
-            alice_btc_balance_before + beta_asset_amount - beta_max_fee;
-        alice_btc_balance_after.should.be.at.least(alice_btc_balance_expected);
+        let receive_address = alice_redeem_tx.vout[0].scriptPubKey.addresses[0];
+        receive_address.should.be.equal(alice_final_address);
```
thomaseizinger: :x: Instead of _verifying_ this, I would _search_ for the UTXO with the receive address and compare the value against the expected one.

You don't want to test that the COMIT node gives you the right TX (at least not primarily). You want to verify that this TX pays Alice enough money. If you can't find a UTXO that pays Alice money, it is not enough.

There also seems to be some repetition here that could be extracted into a fn like `get_satoshi_transferred_to(tx_id, address)`.

api_tests/e2e/rfc003/btc_eth-erc20/test.js
```diff
@@ -334,29 +302,29 @@ describe("RFC003: Bitcoin for ERC20", () => {
...
     });
 
     it("[Bob] Should have received the alpha asset after the redeem", async function() {
-        let bob_btc_balance_after = await bitcoin.btc_balance(
-            bob_final_address
+        this.timeout(10000);
+        let body = await bob.poll_comit_node_until(
+            chai,
+            bob_swap_href,
+            body => body.state.alpha_ledger.status == "Redeemed"
```
D4nte:
```suggestion
            body => body.state.alpha_ledger.status === "Redeemed"
```

api_tests/e2e/rfc003/btc_eth-erc20/test.js
```diff
@@ -154,12 +148,6 @@ describe("RFC003: Bitcoin for ERC20", () => {
             alpha_ledger_redeem_identity: null,
         };
 
-        logger.info(
```
D4nte: I am not sure how wise it is to remove all logging

api_tests/e2e/rfc003/eth-erc20_btc/test.js
```diff
@@ -282,32 +256,34 @@ describe("RFC003: ERC20 for Bitcoin", () => {
...
+        );
+
+        let receive_address = alice_redeem_tx.vout[0].scriptPubKey.addresses[0];
+        receive_address.should.be.equal(alice_final_address);
 
-        const alice_btc_balance_expected =
-            alice_btc_balance_before + beta_asset_amount - beta_max_fee;
-        alice_btc_balance_after.should.be.at.least(alice_btc_balance_expected);
+        const alice_satoshi_expected = beta_asset_amount - beta_max_fee;
+        let alice_satoshi_received = sb.toSatoshi(
```
D4nte: 🔧 you are doing that a lot, maybe worth having a function for it.

D4nte:
❌  I don't understand how it resolves #698

api_tests/e2e/rfc003/btc_eth-erc20/test.js
luckysori: I am happy to add them back in, but I believe they're not being used all that much. @comit-network/rust-devs: thoughts?

thomaseizinger:
> x I don't understand how it resolves #698

Good call. How does it resolve that TODO?

api_tests/e2e/rfc003/btc_eth-erc20/test.js
thomaseizinger: I think we no longer need it. Maybe we can add some logging once the tests are consolidated with #767. That would allow to do it without a lots of repetition.

luckysori:
>  I don't understand how it resolves #698

Discussing it with @LLFourn, we came to the conclusion that we don't care if this happens or not, since we're not checking the balance of the wallet at any point.

api_tests/e2e/rfc003/btc_eth-erc20/test.js
D4nte: ok, please update #767

D4nte:
@luckysori I had issues when testing omnilayer but I did not have enough UTXO to spend to do several transactions.
If you look at #772 I resolved the issue but doing many `fund`.
I was thinking that the resolution of #698 would help with this matter so that instead of doing several funds to have enough UTXOs, I could just use the change UTXO.
Does the introduction of `satoshi-bitcoin` help with this matter?

luckysori:
> @luckysori I had issues when testing omnilayer but I did not have enough UTXO to spend to do several transactions.
> If you look at #772 I resolved the issue but doing many `fund`.
> I was thinking that the resolution of #698 would help with this matter so that instead of doing several funds to have enough UTXOs, I could just use the change UTXO.

Fair enough, in that case I'm happy to complete the original 'TODO`.

> Does the introduction of `satoshi-bitcoin` help with this matter?

That library is only to convert between Bitcoin and Satoshi.

thomaseizinger:
> @luckysori I had issues when testing omnilayer but I did not have enough UTXO to spend to do several transactions.
> If you look at #772 I resolved the issue but doing many `fund`.
> I was thinking that the resolution of #698 would help with this matter so that instead of doing several funds to have enough UTXOs, I could just use the change UTXO.
> Does the introduction of `satoshi-bitcoin` help with this matter?

Thoughts on that @LLFourn ?

api_tests/e2e/rfc003/btc_eth-erc20/test.js
```diff
@@ -334,29 +301,29 @@ describe("RFC003: Bitcoin for ERC20", () => {
...
-        let bob_btc_balance_after = await bitcoin.btc_balance(
+        this.timeout(10000);
+        let body = await bob.poll_comit_node_until(
+            chai,
+            bob_swap_href,
+            body => body.state.alpha_ledger.status === "Redeemed"
+        );
+        let bob_redeem_txid = body.state.alpha_ledger.redeem_tx;
+
+        let bob_satoshi_received = await bitcoin.getSatoshiTransferredTo(
```
LLFourn: 🔧 The case of `getSatoshiTransferredTo` is inconsistent with the_way_we_normally_write_things

api_tests/e2e/rfc003/btc_eth-erc20/test.js
```diff
@@ -334,29 +301,29 @@ describe("RFC003: Bitcoin for ERC20", () => {
...
+            bob_redeem_txid,
             bob_final_address
         );
-        const bob_btc_balance_expected =
-            bob_btc_balance_before + alpha_asset_amount - alpha_max_fee;
-        bob_btc_balance_after.should.be.at.least(bob_btc_balance_expected);
+        const bob_satoshi_expected = alpha_asset_amount - alpha_max_fee;
+
+        bob_satoshi_received.should.be.at.least(bob_satoshi_expected);
```
LLFourn: ❓ Any reason we're not exact here? We know the fee right so can't we just do an equality?

api_tests/e2e/rfc003/btc_eth/test.js
```diff
@@ -138,11 +53,11 @@ describe("RFC003: Bitcoin for Ether", () => {
                 },
                 alpha_asset: {
                     name: "Bitcoin",
-                    quantity: alpha_asset.toString(),
+                    quantity: alpha_asset_amount.toString(),
```
LLFourn: 🔧 `alpha_asset_quantity`

api_tests/lib/bitcoin.js
```diff
@@ -47,38 +48,30 @@ module.exports.btc_activate_segwit = async function() {
     return create_bitcoin_rpc_client().generate(432);
 };
 
-module.exports.btc_import_address = async function(address) {
-    return create_bitcoin_rpc_client().importAddress(address);
-};
+async function getSatoshiTransferredTo(tx_id, address) {
```
LLFourn: ❌ This is a halfway implementation. To do what this says it should iterate over all utxos going to the address and add up the total.

api_tests/lib/bitcoin.js
```diff
@@ -47,38 +48,30 @@ module.exports.btc_activate_segwit = async function() {
...
-) {
-    global.harness.logger.info(
-        "%s the swap, %s has %s satoshis at the %s address %s",
-        when,
-        player,
-        await btc_balance(address),
-        address_type,
-        address
-    );
+module.exports.getSatoshiTransferredTo = async function(tx_id, address) {
```
LLFourn: 🔧 Just set it to the function `= GetSatoshiTransferredTo`

api_tests/lib/bitcoin.js
luckysori: It is intentional, because we only want to consider transactions that have one utxo, built using the action that we return in the HTTP API.

api_tests/e2e/rfc003/btc_eth-erc20/test.js
luckysori: Where can we get the fee from?

api_tests/lib/bitcoin.js
```diff
@@ -47,38 +48,30 @@ module.exports.btc_activate_segwit = async function() {
...
-        address
-    );
+module.exports.get_satoshi_transferred_to = async function(tx_id, address) {
+    return get_satoshi_transferred_to(tx_id, address);
 };
 
 class BitcoinWallet {
     constructor() {
         this.keypair = bitcoin.ECPair.makeRandom({ rng: util.test_rng });
+        // TODO: Use an actual wallet instead of an array for keeping track of
```
thomaseizinger: :x: Can we have to comment on a single line somehow? Maybe by rephrasing it?

api_tests/lib/bitcoin.js
```diff
@@ -47,38 +48,30 @@ module.exports.btc_activate_segwit = async function() {
...
-) {
-    global.harness.logger.info(
-        "%s the swap, %s has %s satoshis at the %s address %s",
-        when,
-        player,
-        await btc_balance(address),
-        address_type,
-        address
-    );
+module.exports.get_satoshi_transferred_to = async function(tx_id, address) {
```
thomaseizinger: You should be able to just assign the fn here:

```suggestion
module.exports.get_satoshi_transferred_to = get_satoshi_transferred_to;
```

api_tests/e2e/rfc003/btc_eth-erc20/test.js
thomaseizinger: @LLFourn Please provide further input on this.

api_tests/e2e/rfc003/btc_eth-erc20/test.js
thomaseizinger: Well, Rust code style != JS code style ...

But in order to be consistent now, it is better to have snake_case.

api_tests/lib/bitcoin.js
thomaseizinger: (This will make the issue title nicer!)

PR #794, opened 2019-02-28
Deactivate bitcoind `-acceptnonstdtxn` in tests

### LINKED ISSUE #762: Ensure all Bitcoin transactions are accepted on mainnet (-acceptnonstdtxn=0)
On regtest and testnet, bitcoind is configured to activate `-acceptnonstdtxn`.

This means that our transactions may not work on mainnet!

With this issue, we should pass `-acceptnonstdtxn=0` to the various bitcoind instances used (e2e, testcontainers) and verify that `comit-rs` would work on mainnet.

An attempt was done with #758 so investigation is needed.

D4nte:
Resolves #762

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,7 +205,7 @@ version = "0.1.0"
 dependencies = [
  "bitcoin_rpc_client 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "bitcoin_support 0.1.0",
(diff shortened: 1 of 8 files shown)


thomaseizinger:
For historical reasons:

This PR resolves #762 because https://github.com/testcontainers/testcontainers-rs/pull/69 added this as a flag to `BitcoinCoreImageArgs` and the default value is `true`!

PR #803, opened 2019-03-04
Make sure we always run against node 11.10

### LINKED ISSUE #802: Fix Master Build
some node issue to fix

thomaseizinger:
This PR removes any usage of `nvm` from the project.

Instead, we fail the build if the available node version is not `11.10`. This means we now have to install that version in Travis.

This also means that *you* (@comit-network/rust-devs) have to setup your environment so that node version 11.10 is used inside the project directory. How you do that is left up to you :)

This hopefully fixes #802.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: rust
+language: minimal
 cache:
   yarn: true
   directories:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,25 +1,26 @@
-language: rust
+language: minimal
```
thomaseizinger: I wasn't able to install node in a way so that the existing installation is no longer present. That is why I switched to the `minimal` image that doesn't have any language-bias.

.travis.yml
```diff
@@ -1,25 +1,26 @@
...
 cache:
   yarn: true
   directories:
   - "$HOME/.cargo"
   - "$HOME/.cache/sccache"
   - "$TRAVIS_BUILD_DIR/api_tests/node_modules"
 sudo: required
-rust: nightly-2018-12-06
```
thomaseizinger: This has the nice side-effect that we are no longer duplicating the used Rust version :)

.travis.yml
```diff
@@ -1,25 +1,26 @@
...
     - libzmq3-dev
     - yarn
-install: true
+    - nodejs=11.10*
 services: docker
-before_script:
-- which cargo-make || cargo install --debug cargo-make
-- nvm install 11
+install:
+  - curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain $(< rust-toolchain) && source $HOME/.cargo/env
```
thomaseizinger: Slightly more inconvenient but in exchange non-redundant.

api_tests/package.json
```diff
@@ -6,6 +6,9 @@
     "scripts": {
         "test": "mocha --bail"
     },
+    "engines": {
+        "node": "11.10"
```
thomaseizinger: This makes `yarn` fail if it would execute against a node version that is `< 11.10` or `> 11.11`

PR #805, opened 2019-03-04
Don't restart nodes between tests

### LINKED ISSUE #467: Do not restart blockchain dockers between 2 e2e
Note:
`harness.js` now handles start of all dockers
However, the trigger is done from the Makefile, it decides what tests to run. May need to make `harness.js` decide on the tests (btc_eth, eth_btc) to run.

Blocked by #446
Blocked by #681

LLFourn:
Make harness.js take the tests it should run as args. Now you can just do this:

```
# run all happy rfc003 happy path tests
./harness.js ./e2e/rfc003/*/happy.js
```

- Now we look at each test individually and manage its dependencies
  independently.
- Dependencies are started up on demand through idempotent calls to objects
  that manage each type of dependency
- Everything is just in general much less brittle
- Logs are simply written to `api_tests/log` rather than log directories in the tests

Now we are able to keep the bitcoin/ethereum nodes running inbetween
tests.
We still have to restart the comit nodes because our tests rely on there
not being other tests interfering with the node's state.

Resolves #467.


api_tests/dry/config.toml
```diff
@@ -1,3 +1,5 @@
+tests = [ "sanity", "multiple_peers", "rfc003_swap_reject"]
```
LLFourn: Remove this (it's no longer used in the final version).

api_tests/dry/config.toml
```diff
@@ -9,3 +11,4 @@ config_dir = "./regtest/bob/"
 [comit_node.charlie]
 host = "127.0.0.1"
 config_dir = "./regtest/charlie/"
+
```
D4nte:
```suggestion
```

api_tests/dry/config.toml
```diff
@@ -1,3 +1,5 @@
+tests = [ "sanity", "multiple_peers", "rfc003_swap_reject"]
+
```
D4nte:
```suggestion
```

api_tests/dry/rfc003_swap_reject.js
```diff
@@ -12,8 +12,8 @@ const beta_ledger_name = "Ethereum";
 const beta_ledger_network = "regtest";
 
 const alpha_asset_name = "Bitcoin";
-const alpha_asset_reasonable_quantity = "100000000";
-const alpha_asset_stingy_quantity = "100";
+const alpha_asset_reasonable_quantity = "400000002";
```
LLFourn: Undo this. This is a side effect of me trying to make comit-nodes not need to be restarted between each dry test.

api_tests/harness.js
```diff
@@ -16,202 +20,267 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
+        this.block_timers = {};
+    }
+
+    async ensure_ledgers_running(ledgers) {
+        let wait_times = [0];
+        let running_ledgers = this.running_ledgers;
+
+        let to_be_started = ledgers.filter(name => !running_ledgers[name]);
+
+        if (to_be_started.length > 0) {
```
thomaseizinger: :question: Do you really need to handle this explicitly? Iterating over a 0-length array should be fine :)

api_tests/harness.js
```diff
@@ -16,202 +20,267 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
+            for (let ledger of to_be_started) {
+                let ledger_config = ledgers_config[ledger];
+
+                await execSync(
```
thomaseizinger: :question: If it is called `execSync`, why do you have to await it?

api_tests/npm_wrapper.sh
```diff
@@ -7,5 +7,5 @@ export TEST_DIR="$1";
 cd "${ROOT}/api_tests";
 yarn install;
 
-npm test "./harness.js" \
+node "./harness.js" \
```
thomaseizinger: :wrench: You should be able to call it directly right?

```suggestion
"./harness.js" \
```

api_tests/harness.js
```diff
@@ -16,202 +20,267 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
-            "ignore",
-            fs.openSync(log_dir + "/docker-compose.log", "w"),
-            fs.openSync(log_dir + "/docker-compose-err.log", "w"),
-        ],
-    });
+    stop_ledgers() {
+        let names = Object.keys(this.running_ledgers);
+        if (names.length > 0) {
+            console.log("Stopping ledgers: " + names.join(", "));
+            for (let [name, block_timer] of Object.entries(this.block_timers)) {
```
thomaseizinger: :wrench: This is a bit more concise:

```js
Object.values(this.block_timers).forEach(clearInterval)
```

api_tests/npm_wrapper.sh
LLFourn: err wait didn't you delete this thing? Do we still need this for anything?

api_tests/e2e/rfc003/btc_eth/happy.js
```diff
@@ -151,7 +151,8 @@ describe("RFC003: Bitcoin for Ether", () => {
             "amount",
             "network"
         );
-        await alice.do(alice_fund_action);
+        let resp = await alice.do(alice_fund_action);
```
luckysori: :x: I guess you were using it for debugging.

api_tests/e2e/rfc003/btc_eth/happy.js
```diff
@@ -151,7 +151,8 @@ describe("RFC003: Bitcoin for Ether", () => {
...
+
         let res = await chai
```
luckysori: :x: Not that you added it, but this is not used and makes no sense here.

api_tests/e2e/rfc003/btc_eth/happy.js
```diff
@@ -160,11 +161,13 @@ describe("RFC003: Bitcoin for Ether", () => {
...
 
     it("[Bob] Can get the fund action after Alice funds", async function() {
-        this.timeout(10000);
+        this.timeout(20000);
         let body = await bob.poll_comit_node_until(
             chai,
             bob_swap_href,
-            body => body._links.fund
+            body => {
+                return body._links.fund;
```
luckysori: :question: Why this return?

api_tests/e2e/rfc003/btc_eth/happy.js
bonomat: ❓Is this variable required? 
```suggestion
        await alice.do(alice_fund_action);
```

api_tests/harness.js
D4nte: ❌ please don't use that as it is not TypeScript friendly :)

Makefile.toml
```diff
@@ -251,16 +245,14 @@ private = true
...
-for GLOB_DIR in ${PARAMS}; do
-    cd "$ROOT/api_tests/e2e/rfc003/"
-    for DIR in ${GLOB_DIR}; do
-        cd "$ROOT";
-        [[ -d "./api_tests/e2e/rfc003/${DIR}" ]] && ./api_tests/npm_wrapper.sh "./e2e/rfc003/${DIR}"
-    done;
+glob="${1:-*}";
+for dir in api_tests/e2e/rfc003/$glob; do
+    if test -d "$dir"; then
+        files="$files "$dir/*.js;
```
D4nte: ❓ does this still work fine if there is an empty directory in `rfc003` ? Or do you get an error "no file found" because no file matches the glob `$dir/*.js`?

api_tests/harness.js
```diff
@@ -16,202 +20,267 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
 
-            if (config.btsieve) {
-                Object.keys(config.btsieve).forEach(async function(name) {
-                    const subprocess = await startBtsieve(
-                        name,
-                        config.btsieve[name]
-                    );
-                    subprocesses.push(subprocess);
-                });
+    async ensure_btsieve_running(btseives) {
```
luckysori: :x: The word "btseive" is used in several places instead of the correct "btsieve".

api_tests/e2e/rfc003/btc_eth/happy.js
D4nte: ❌ no need to assign a value if it's not used.
Also confusing as you know have a `resp` and `res` variable.

api_tests/e2e/rfc003/btc_eth/happy.js
```diff
@@ -160,11 +161,13 @@ describe("RFC003: Bitcoin for Ether", () => {
     let bob_fund_action;
 
     it("[Bob] Can get the fund action after Alice funds", async function() {
-        this.timeout(10000);
+        this.timeout(20000);
```
D4nte: ❓ did you change something that made this change needed?

api_tests/harness.js
```diff
@@ -16,202 +20,267 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
+            console.log("Stopping comit nodes: " + names.join(", "));
+            for (let [_, process] of Object.entries(this.running_nodes)) {
+                process.kill();
             }
+            this.running_nodes = {};
+        }
+    }
+}
+
+class BtsieveRunner {
```
luckysori: :upside_down_face: Could/should `BtsieveRunner` and `ComitRunner` be generalised further since they are almost identical?

api_tests/e2e/rfc003/btc_eth/happy.js
D4nte: ❓  is that really needed?

api_tests/harness.js
```diff
@@ -16,202 +20,267 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
-    cleanUp();
-});
-
 // ********************** //
 // Start services helpers //
 // ********************** //
+class LedgerRunner {
+    constructor(ledgers_config) {
+        this.running_ledgers = {};
+        this.ledgers_config = ledgers_config;
```
bonomat: ❌ There is a global variable called `ledgers_config` already. Do you want to rename this one so that it gets more clear which one is meant?

api_tests/harness.js
```diff
@@ -16,202 +20,267 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
+        this.block_timers = {};
+    }
+
+    async ensure_ledgers_running(ledgers) {
```
D4nte:
```suggestion
    async ensureLedgersRunning(ledgers) {
```
Let's use JS appropriate naming

api_tests/e2e/rfc003/btc_eth/happy.js
LLFourn: No. Random debugging things.

api_tests/e2e/rfc003/btc_eth/happy.js
LLFourn: Random debugging things.

LLFourn: No!

api_tests/harness.js
LLFourn: They are the same though? 😕=

api_tests/harness.js
LLFourn: Are you asking me to change everything from snake case to camel case? Or just not to make things worse?

If we're doing this we should probably use a linter to enforce this.

api_tests/harness.js
LLFourn: Yeah. We need to log when we start waiting for a node to start up. If we're not waiting for any of them we shouldn't log anything.

api_tests/harness.js
LLFourn: You don't :)

api_tests/harness.js
LLFourn: Yes. But I thought it was premature to generalise them into some kind of process runner.

LLFourn:
@bonomat Having multiple commands for setting up, running and tearing down the test environment pushes the complexity of setting up dependencies up to the user. This is what I would call cumbersome. 

I didn't get this part:
>I could think of something like this:

>harness.js initialise -- implicitly takes a config file, e.g. ./config.toml (this could be as simple as a docker-compose file)
>test-runner.sh api_tests/e2e/rfc003/*/*.js -- or similar
>harness.js shutdown

How is the `config.toml` going to know which dependencies is needed by `test-runner.sh api_tests/e2e/rfc003/*/*.js `. What if you did `test-runner.sh api_tests/e2e/rfc003/btc_*/*.js`? If we had omni we wouldn't need to start the omin node then because you don't do rfc003 swaps to omni.

IMO, It is more flexible to manage dependencies in the harness file for other reasons:
1. We may want to have tests that shutdown nodes to see how the system handles it. With this approach whatever you shutdown can be repaired by the harness before the next test.
2. You can still do everything you can do with your suggested approach by adding cli flags that tell the harness what the harness doesn't have to deal with.

api_tests/harness.js
LLFourn: I did the thing @thomaseizinger suggested will that work?

api_tests/harness.js
D4nte: use camelCase for new variables/functions.
No need to refactor everything.
Good idea for the linter, we can do that later.

api_tests/harness.js
bonomat: do you mean they have the same value? 
They are not the same variable though, i.e. they are accessible in different contexts. 
Have a look at the other definition, it's a constant: 
```
const ledgers_config = Toml.parse(
    fs.readFileSync(project_root + "/api_tests/regtest/ledgers.toml", "utf8")	   
);
```
mea

Makefile.toml
LLFourn: It works fine.

Makefile.toml
```diff
@@ -204,24 +204,22 @@ description = "Defines the E2E-tests flow."
...
-export ROOT=$(pwd);
-./api_tests/npm_wrapper.sh ./dry/
-'''
-]
+install_script = [ "( cd ./api_tests && yarn install; )" ]
+script = [ "./api_tests/harness.js ./api_tests/dry/*.js" ]
 
-[tasks.pre-dry-tests]
-alias = "pre-e2e-tests"
+[tasks.install_test_dependencies]
```
D4nte:
```suggestion
[tasks.install-test-dependencies]
```
We use kebab-case for task names

Makefile.toml
```diff
@@ -230,17 +228,10 @@ description = "Defines the E2E-tests flow."
 workspace = false
 dependencies = [
     "build",
-    "pre-e2e-tests",
+    "install_test_dependencies",
```
D4nte:
```suggestion
    "pre-e2e-tests",
```
❌

Makefile.toml
D4nte: :)

api_tests/npm_wrapper.sh
LLFourn: It's deleted.

Makefile.toml
LLFourn: But it's not just used in e2e tests?

api_tests/harness.js
```diff
@@ -16,202 +25,275 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
+                }
+            );
+        }
+    }
+
+    async stopBtsieves() {
+        let names = Object.keys(this.running_btsieves);
+
+        if (names.length > 0) {
+            console.log("Stopping Btsieves: " + names.join(", "));
```
D4nte:
```suggestion
            console.log("Stopping Btsieve(s): " + names.join(", "));
```

api_tests/harness.js
```diff
@@ -16,202 +25,275 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
+        for (let item of test_files) {
+            mocha.addFile(test_file);
+        }
+
+        if (config.ledgers) {
+            await ledger_runner.ensureLedgersRunning(config.ledgers);
+        }
+
+        if (config.btsieve) {
+            await btsieve_runner.ensureBtsieveRunning(
```
D4nte: 🙃 there are some discrepancies with btsieve some times plural, sometimes singular.

api_tests/harness.js
```diff
@@ -1,8 +1,17 @@
+#!/usr/bin/env node
+
 const bitcoin = require("./lib/bitcoin.js");
 const execSync = require("child_process").execSync;
 const spawn = require("child_process").spawn;
 const Toml = require("toml");
 const fs = require("fs");
+const Mocha = require("mocha");
+const path = require("path");
+const program = require("commander");
```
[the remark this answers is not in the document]
LLFourn: copy pasting from examples.

Makefile.toml
D4nte: `pre-api-tests` ?

api_tests/harness.js
```diff
@@ -16,202 +25,274 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
+                    fs.openSync(`${log_dir}/docker-compose.log`, "w"),
+                    "inherit",
+                ],
+            });
+
+            for (let ledger of to_be_started) {
+                let ledger_config = ledgers_config[ledger];
+                this.running_ledgers[ledger] = true;
+                wait_times.push(
+                    process.env.CARGO_MAKE_CI == "TRUE"
```
bonomat: ❌ 
```suggestion
                    process.env.CARGO_MAKE_CI === "TRUE"
```

api_tests/harness.js
```diff
@@ -16,202 +25,274 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
+            mocha.run(async failures => {
+                res(failures);
+            });
+        });
+
+        let failures = await test_finish;
+
+        if (failures) {
+            process.exitCode = 1;
+            if (commander.dumpLogs || process.env.CARGO_MAKE_CI == "TRUE") {
```
bonomat: ❌ 
```suggestion
            if (commander.dumpLogs || process.env.CARGO_MAKE_CI === "TRUE") {
```

api_tests/harness.js
```diff
@@ -16,202 +25,274 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
+        const mocha = new Mocha({
+            bail: true,
+            ui: "bdd",
+        });
 
-for (let item of items) {
-    if (item.endsWith(".js")) {
-        run_tests(test_dir + "/" + item);
+        for (let item of test_files) {
+            mocha.addFile(test_file);
```
bonomat: ❓I'm not sure what is happening here and why, but do you intend to add `test_file` and not `item` ?

LLFourn: WOOPS

api_tests/harness.js
```diff
@@ -16,202 +25,272 @@ const project_root = execSync("git rev-parse --show-toplevel", {
...
+                    cwd: test_root,
+                    encoding: "utf-8",
+                    env: { COMIT_NODE_CONFIG_PATH: comit_config.config_dir },
+                    stdio: [
+                        "ignore",
+                        fs.openSync(
+                            log_dir + "/comit_node-" + name + ".log",
+                            "w"
+                        ),
+                        fs.openSync(
```
bonomat: ❓copy pasta or do you want to open the log file twice on purpose?

api_tests/package.json
```diff
@@ -4,10 +4,10 @@
     "description": "COMIT node end-to-end testing suite",
     "main": "index.js",
     "scripts": {
-        "test": "mocha --bail"
+        "test": "harness.js"
     },
     "engines": {
-        "node": "11.10"
+        "node": ">=10.14 <11"
```
thomaseizinger: :wrench: 
This should do the same thing:
```suggestion
        "node": "^10.14"
```

LLFourn: Good to know!

thomaseizinger: FYI: https://jubianchi.github.io/semver-check/#/^11.14/11.30.0

api_tests/harness.js
LLFourn: Good question. I think it does work but it's not guaranteed to look pretty. When stderr and stdout are written at the same time then some lines might be a bit mangled.  We could write stdout and stderr to different log files but I think its better to keep them in one so we don't accidentally miss something.

PR #806, opened 2019-03-04
Use Network::Mainnet instead of Network::Bitcoin

### LINKED ISSUE #644: Add Network support as per RFC-002
As per [RFC-002](https://github.com/coblox/COMIT-RFCs/blob/master/RFC-002-SWAP.md) and [COMIT-registry](https://github.com/coblox/COMIT-RFCs/blob/master/COMIT-registry.md), the following changes should be done to COMIT-rs:
- [x] Add support of `network` parameter for Ethereum (Done with #687)
- [x] Use `mainnet` for Bitcoin mainnet (instead of `Bitcoin`) (Done with #806)

Related to #376

luckysori:
Replace `rust-bitcoin::Network` with own `Network` enum so that mainnet is called mainnet and not bitcoin...

Resolves #644.

### DIFF
--- a/vendor/bitcoin_support/src/network.rs
+++ b/vendor/bitcoin_support/src/network.rs
@@ -1,5 +1,4 @@
-use bitcoin::network::constants::Network as ExternNetwork;
-use serde::Deserialize;
+use bitcoin;
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Deserialize, Serialize, Hash)]
 #[serde(rename_all = "lowercase")]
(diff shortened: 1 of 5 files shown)


thomaseizinger:
@luckysori FYI, I linked the ticket to this PR.

vendor/bitcoin_support/Cargo.toml
```diff
@@ -14,6 +14,7 @@ lazy_static = "1"
 secp256k1_support = { path = "../secp256k1_support" }
 serde = "1"
 serde_derive = "1"
+toml = "0.4"
```
thomaseizinger: :x: Do we still actually use that?

vendor/bitcoin_support/src/network.rs
```diff
@@ -0,0 +1,30 @@
+use bitcoin::network::constants::Network as ExternNetwork;
```
thomaseizinger: :x: I'd prefer if you don't call this `ExternNetwork` but just write the `From`-impl with the FQN.

vendor/bitcoin_support/src/pubkey.rs
```diff
@@ -32,12 +32,12 @@ impl IntoP2wpkhAddress for PubkeyHash {
                     match network {
```
thomaseizinger: :x: We should now be able to write a `From` impl for this.

thomaseizinger:
> I don't see where it is serialised in BAM! Can you help me understand?

It is the combination of this:

https://github.com/comit-network/comit-rs/blob/3276d46f63aedf755136d1c061aa21fe3496130a/vendor/bitcoin_support/src/network.rs#L4-L10

and this:

https://github.com/comit-network/comit-rs/blob/b298171c3b07cfdd539d7e184f1cc544b9e82a27/application/comit_node/src/bam_api/mod.rs#L33-L35

and this:

https://github.com/comit-network/comit-rs/blob/b298171c3b07cfdd539d7e184f1cc544b9e82a27/vendor/bam/src/json/header.rs#L124-L140

vendor/bitcoin_support/src/network.rs
luckysori: I also always prefer that. I don't know why I did it like this lol.

D4nte:
@thomaseizinger thanks!

PR #807, opened 2019-03-05
Convert bitcoin.js and ethereum.js to TypeScript

### LINKED ISSUE #804: Convert tests to TypeScript
For #767 it is likely that a number of objects (Alice, Bob, addresses) need to be passed done to a `rfc003` "framework" that execute the tests.

Introducing TS before that seems like a good opportunity as it would assist with #767

D4nte:
Only convert `bitcoin.js`, `ethereum.js` and `wallet.js` for now. The rest will be done in a subsequent pull request.

Related to #804


Makefile.toml
```diff
@@ -126,7 +126,7 @@ workspace = false
 install_script = ["(cd ./api_tests; yarn install;)"]
 script = [
 '''
-(cd api_tests; yarn run prettier --write '**/*.js')
+(cd api_tests; yarn run prettier --write '**/*.[tj]s')
```
thomaseizinger: I think we can go for duplication here for the benefit of clarity :D

```suggestion
(cd api_tests; yarn run prettier --write '**/*.{ts,js}')
```

Makefile.toml
```diff
@@ -136,7 +136,7 @@ workspace = false
 install_script = ["(cd ./api_tests; yarn install;)"]
 script = [
 '''
-(cd api_tests; yarn run prettier --check '**/*.js')
+(cd api_tests; yarn run prettier --check '**/*.[tj]s')
```
thomaseizinger:
```suggestion
(cd api_tests; yarn run prettier --check '**/*.{ts,js}')
```

api_tests/dry/multiple_peers.js
```diff
@@ -1,5 +1,6 @@
 const chai = require("chai");
 const Web3 = require("web3");
+const Utils = require("web3-utils");
```
thomaseizinger: :thinking: I think in JS, only classes are named Uppercase. Modules should be lowercase.

api_tests/lib/bitcoin.ts
```diff
@@ -0,0 +1,156 @@
...
+        signature: Buffer;
+        input: Buffer;
+        witness: Buffer[];
+    };
+
+    constructor() {
+        this.keypair = ECPair.makeRandom({ rng: util.test_rng });
+        // TODO: Use wallet instead of array to track Bitcoin UTXOs
+        this.bitcoinUtxos = [];
+        this._identity = payments.p2wpkh({
```
thomaseizinger: :question: Is there no proper return type for `p2wpkh`?

api_tests/package.json
```diff
@@ -4,6 +4,7 @@
     "description": "COMIT node end-to-end testing suite",
     "main": "index.js",
     "scripts": {
+        "build": "./node_modules/typescript/bin/tsc",
```
thomaseizinger: :wrench: 

`tsc` should also be in `node_modules/.bin/tsc`. If that is the case, you can just refer to `tsc` here :)

api_tests/npm_wrapper.sh
```diff
@@ -12,5 +12,6 @@ nvm use 11.10.0 || (nvm install 11.10.0 && nvm use 11.10.0);
 cd "${ROOT}/api_tests";
 yarn install;
 
-npm test "./harness.js" \
+npm run build \
```
thomaseizinger: :wrench: Should actually both be `yarn`.
```suggestion
yarn run build \
```

api_tests/npm_wrapper.sh
```diff
@@ -12,5 +12,6 @@ nvm use 11.10.0 || (nvm install 11.10.0 && nvm use 11.10.0);
...
+&& npm test "./harness.js" \
```
thomaseizinger:
```suggestion
&& yarn test "./harness.js" \
```

api_tests/lib/bitcoin.ts
D4nte: Nope.

api_tests/e2e/rfc003/btc_eth/test.js
```diff
@@ -28,11 +32,12 @@ const beta_expiry = new Date("2080-06-11T13:00:00Z").getTime() / 1000;
 describe("RFC003: Bitcoin for Ether", () => {
     before(async function() {
         this.timeout(5000);
-        await bitcoin.btc_activate_segwit();
+        await bitcoin.activateSegwit();
+        // Initialize the client
```
thomaseizinger: :question: Why is this comment here? It doesn't seem to be there in other tests :D 
```suggestion
```

api_tests/dry/multiple_peers.js
```diff
@@ -15,15 +16,21 @@ const alpha_asset_bob_quantity = "100000000";
...
+const beta_asset_charlie_quantity = utils.toWei("20", "ether");
 
 const alpha_expiry = new Date("2080-06-11T23:00:00Z").getTime() / 1000;
 const beta_expiry = new Date("2080-06-11T13:00:00Z").getTime() / 1000;
 
-const alice = actor.create("alice");
-const bob = actor.create("bob");
-const charlie = actor.create("charlie");
+const alice = actor.create("alice", {
+    ethConfig: global.harness.ledgers_config.ethereum,
```
thomaseizinger: :question: Why do the actors not need a `bitcoin` config?

api_tests/e2e/rfc003/btc_eth/test.js
D4nte: Let's remove for now, it's a future clean up.

api_tests/dry/multiple_peers.js
D4nte: Check the new todo in `actor.js`.

This will be cleaned up with a future PR when migrating actor to TypeScript

LLFourn: What's the reason not to use global configs in the bitcoin and ethereum files?
What's your "clean up" plan with regards to the rpc config?

api_tests/e2e/rfc003/btc_eth/happy.js
```diff
@@ -28,11 +32,11 @@ const beta_expiry = new Date("2080-06-11T13:00:00Z").getTime() / 1000;
 describe("RFC003: Bitcoin for Ether", () => {
     before(async function() {
         this.timeout(5000);
-        await bitcoin.btc_activate_segwit();
+        await bitcoin.activateSegwit();
```
LLFourn: 🤡 Since we're not restarting nodes maybe we can just do this once in harness.js when the bitcoin node starts up and remove it from the tests. This would remove a lot of spam from the bitcoin node logs when running multiple tests.

api_tests/lib/bitcoin.ts
```diff
@@ -0,0 +1,165 @@
...
+        txId: string,
+        verbose?: boolean,
+        blockHash?: string
+    ): Promise<any>;
+    sendToAddress(address: string, amount: number | string): Promise<any>;
+    sendRawTransaction(hexString: string): Promise<any>;
+}
+
+interface IUtxo {
+    // TODO: declare transactionId type
```
LLFourn: ❓  The bitcoin typescript didn't have a type for this?

api_tests/e2e/rfc003/eth-erc20_btc/happy.js
```diff
@@ -1,26 +1,32 @@
...
-const toby_wallet = wallet.create("toby");
+const toby_wallet = wallet.create("toby", {
+    ethConfig: global.harness.ledgers_config.ethereum,
+});
 
-const toby_initial_eth = "10";
-const alice_initial_eth = "5";
-const alice_initial_erc20 = Web3.utils.toWei("10000", "ether");
+const toby_initial_eth = utils.toBN(10);
+const alice_initial_eth = utils.toBN(5);
```
LLFourn: So we're using BN now everywhere rather than BigInt. What's the reason? Ethereum ts only accepts BN?

Makefile.toml
```diff
@@ -283,3 +276,25 @@ script = [
...
+]
+
+[tasks.api-tests]
+description = "Runs all API tests."
+workspace = false
+private = true
+script_runner = "bash"
+script = [
+'''
+./api_tests/harness.js ./api_tests/btsieve/*.js ./api_tests/btsieve/*.js api_tests/e2e/rfc003/*/*.js
```
thomaseizinger: Two times btsieve?

LLFourn: How could you be against more testing?

api_tests/lib/bitcoin.ts
```diff
@@ -0,0 +1,171 @@
...
+        vout.scriptPubKey.addresses.length === 1 &&
+        vout.scriptPubKey.addresses[0] === address
+    ) {
+        satoshi = sb.toSatoshi(vout.value);
+    }
+
+    return satoshi;
+}
+
+module.exports.get_first_utxo_value_transferred_to = getFirstUtxoValueTransferredTo;
```
luckysori: :question: Why do you export it in a different case?

api_tests/lib/bitcoin.ts
```diff
@@ -0,0 +1,165 @@
...
+const sb = require("satoshi-bitcoin");
+const util = require("./util.js");
+
+// TODO: Remove regtest
+// Once this is merged: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33571
+
+const networksAny: any = networks;
+const regtest: any = networksAny.regtest;
+
+interface IBitcoinConfig {
```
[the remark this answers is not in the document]
D4nte: Such is life.

api_tests/lib/bitcoin.ts
D4nte: Not my code 😉 will be corrected once all users of this method are TypeScript (ie, when I covert the `happy.js` files to TypeScript)

Makefile.toml
D4nte: oops, forgot `dry` actually..

api_tests/e2e/rfc003/eth-erc20_btc/happy.js
D4nte: Indeed, Ethereum support BN not BigInt.
Better keep uniform and only use one "Big Number" type.

api_tests/e2e/rfc003/btc_eth/happy.js
D4nte: look at my latest commit ;)

api_tests/dry/multiple_peers.js
D4nte: I believe that the libraries should be passed the values instead of using a global variable.

api_tests/lib/bitcoin.ts
D4nte: They only have that:
```
export interface Out {
    script: Buffer;
    value: number;
}
```

api_tests/e2e/rfc003/btc_eth/happy.js
D4nte: https://github.com/comit-network/comit-rs/pull/807/commits/9fc03e510cd0f7745c6a87089ffb5b28780560af

api_tests/lib/bitcoin.ts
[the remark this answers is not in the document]
thomaseizinger: > This is how I have seen it done. However looks like in TypeScript doc they just do Capitalisation.

@D4nte It is originally a C# thing and TS as well as C# are made by the same company, Microsoft. That is why there are similar conventions. I don't like it either because IMO, it clutters names with a meaningless character that are only in your way when doing type searches etc.

api_tests/lib/bitcoin.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    async fund(btcValue: number) {
+        let txId = await _bitcoinRpcClient.sendToAddress(
+            this.identity().address,
+            btcValue
+        );
+        let raw_transaction = await _bitcoinRpcClient.getRawTransaction(txId);
+        let transaction = Transaction.fromHex(raw_transaction);
+        let entries: Out[] = transaction.outs;
+        for (let i in entries) {
```
bonomat: 🙃 You could make this more sexy  
`this.bitcoinUtxos. concat(entries.filter....map..)`

api_tests/btsieve/test.js
```diff
@@ -7,12 +7,22 @@ const actor = require("../lib/actor.js");
 const ethereum = require("../lib/ethereum.js");
 const wallet = require("../lib/wallet.js");
 const btsieve_conf = require("../lib/btsieve.js");
-
-const bitcoin_rpc_client = bitcoin.create_client();
+const utils = require("web3-utils");
+
+// TODO: Do not expose BitcoinRpcClient
```
thomaseizinger: :question: Is this a personal TODO or meant to be done afterwards?

api_tests/e2e/rfc003/btc_eth/happy.js
thomaseizinger: :wrench: Maybe rename it to `ensureSegwit` then?

api_tests/lib/actor.js
```diff
@@ -4,10 +4,13 @@ const fs = require("fs");
 const bitcoin = require("./bitcoin.js");
 const ethutil = require("ethereumjs-util");
 
-const bitcoin_rpc_client = bitcoin.create_client();
+// TODO: pass value instead of using global config
```
thomaseizinger: :upside_down_face: Would that be so hard to do? :)

api_tests/lib/bitcoin.ts
thomaseizinger: If you were to do this, you would have to overwrite `this.bitcoinUtxos` because `concat` [does not override the original arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat#Syntax)

api_tests/btsieve/test.js
D4nte: Meant to be done afterwards as it's easier done if the `happy.js` are TypeScript.

api_tests/e2e/rfc003/btc_eth/happy.js
```diff
@@ -207,23 +211,24 @@ describe("RFC003: Bitcoin for Ether", () => {
...
-        let alice_eth_balance_expected =
-            alice_eth_balance_before + beta_asset_quantity;
+        let alice_eth_balance_expected = alice_eth_balance_before.add(
+            beta_asset_quantity
+        );
 
         alice_eth_balance_after
-            .toString()
-            .should.be.equal(alice_eth_balance_expected.toString());
+            .eq(alice_eth_balance_expected)
```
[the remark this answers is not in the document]
D4nte: Nope
```
      AssertionError: expected <BN: 10f0cf064dd59200000> to equal <BN: 10f0cf064dd59200000>
      + expected - actual
```

api_tests/lib/bitcoin.ts
D4nte: Can you provide the full gist of what you are suggesting?

api_tests/lib/actor.js
D4nte: I just want to draw a line in the sand now so that we can get this one merge and I can then do it straight away.

This is already a `+1,002 −758` PR. The number is likely to triple if I don't stop now.

api_tests/lib/bitcoin.ts
bonomat: I just tried this and this seems to be working :)
```
this.bitcoinUtxos.push(
            ...entries
                .filter(entry => entry.script.equals(this.identity().output))
                .map((entry, index) => {
                    return {
                        txId: txId,
                        vout: index,
                        value: entry.value,
                    };
                })
        );
```

D4nte: Are you sure it works? did you run `cargo make api`

```
RpcError: bad-txns-in-belowout, value in (2.499963) < value out (9.999975) (code 16)
```

PR #809, opened 2019-03-07
Add CORS headers to comit_node HTTP API

thomaseizinger:
This allows any application to use the HTTP API of a COMIT node.

### DIFF
--- a/api_tests/dry/sanity.js
+++ b/api_tests/dry/sanity.js
@@ -69,6 +69,16 @@ describe("Sanity tests", () => {
             });
     });
 
+    it("[Alice] Sets appropriate CORS headers", async () => {
+        let res = await chai
+            .request(alice.comit_node_url())
+            .get("/swaps")
+            .set("Origin", "localhost:3000");
+
+        res.should.have.status(200);
+        res.should.have.header("access-control-allow-origin", "localhost:3000");
+    });
+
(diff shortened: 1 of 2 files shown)


D4nte:
Do you have a link to some doc so we can verify this is correct?

thomaseizinger:
> Do you have a link to some doc so we can verify this is correct?

Here you go: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

I also verified locally that it works with comit-i.

thomaseizinger:
> Do you have a link to some doc so we can verify this is correct?

I should maybe note that `warp` has an interesting behavior here. When saying `allow_any_origin`, the `Origin` that was sent by the client is returned in the header. Most other frameworks just send back `*` as the value, which means any origin is allowed to access it.

api_tests/dry/sanity.js
```diff
@@ -69,6 +69,16 @@ describe("Sanity tests", () => {
             });
     });
 
+    it("[Alice] Sets appropriate CORS headers", async () => {
+        let res = await chai
```
bonomat: 🔧 Should we verify that `alice.comit_node_url() !== "localhost:3000"`? 
```suggestion
        alice.comit_node_url().should.not.equal("localhost:3000");
        let res = await chai
```

PR #816, opened 2019-03-11
Finish Migration to TypeScript

### LINKED ISSUE #804: Convert tests to TypeScript

D4nte:
Full conversion to TypeScript.

Review can be done. Few TODOs remaining.

Resolves #804


.gitignore
```diff
@@ -7,6 +7,7 @@
 
 # User-specific stuff:
 .idea/**/workspace.xml
+.idea/workspace.xml
```
bonomat: Shouldn't that be covered with the entry above?

api_tests/btsieve/test.ts
```diff
@@ -73,7 +68,7 @@ describe("Test btsieve API", () => {
                     })
                     .then(res => {
                         res.should.have.status(201);
-                        location = res.headers.location;
+                        location = res.header.location;
                         location.should.be.a("string");
```
bonomat: 🤡 Probably obsolete as a few lines up it is introduced as string.
Maybe check for not empty or a concrete location?
```suggestion
                         location.should.not.be.empty;
```

api_tests/btsieve/test.ts
```diff
@@ -158,7 +153,7 @@ describe("Test btsieve API", () => {
...
```
bonomat: 🤡Same as above. Should be obsolete. 
```suggestion
                         location.should.not.be.empty;
```

api_tests/btsieve/test.ts
```diff
@@ -238,7 +233,7 @@ describe("Test btsieve API", () => {
...
```
bonomat: 🤡 one more.  Maybe check for not empty or a concrete location?
```suggestion
                         location.should.not.be.empty;
```

api_tests/btsieve/test.ts
```diff
@@ -419,7 +414,7 @@ describe("Test btsieve API", () => {
...
```
bonomat: 🤡 one more.  Maybe check for not empty or a concrete location?
```suggestion
                        location.should.not.be.empty;
```

api_tests/lib/bitcoin.ts
```diff
@@ -54,11 +49,15 @@ interface Utxo {
...
+
+export function init(btcConfig: BtcConfig) {
+    createBitcoinRpcClient(btcConfig);
+}
 
-function createBitcoinRpcClient(btcConfig: BitcoinConfig) {
+function createBitcoinRpcClient(btcConfig: BtcConfig) {
     if (!btcConfig && !_bitcoinConfig) {
-        throw new Error("bitcoin configuration is needed");
+        throw new Error("bitcoin configuration is needed}");
```
bonomat: 🙃 
```suggestion
        throw new Error("bitcoin configuration is needed");
```

api_tests/lib/ethereum.ts
```diff
@@ -81,31 +84,36 @@ export class EthereumWallet {
...
             parity_dev_password
         );
     }
 
-    async send_eth_transaction_to(
+    async sendEthTransactionTo(
         to: string,
         data: string = "0x0",
-        value: string = "0",
+        value: BN | string | number = utils.toBN(0),
```
bonomat: Does this mean this function accepts either `BN` or `string` or defaults to `number = ...` ?

api_tests/dry/rfc003_swap_reject.ts
```diff
@@ -268,7 +277,7 @@ describe("SWAP request REJECTED", () => {
         }
```
bonomat: Not your code, but above, line 266 
`let swap_2 = await chai.request(bob.comit_node_url()).get(swap_2_href);`
is unused.

.gitignore
D4nte: it's a mistake thanks.

api_tests/lib/ethereum.ts
D4nte: accepts `BN` or `string` or `number` and default to `utils.toBN(0)` which is a `BN`

api_tests/dry/rfc003_swap_reject.ts
D4nte: Yep, already removed :)

api_tests/dry/rfc003_swap_reject.ts
```diff
@@ -21,18 +27,18 @@ const beta_asset_quantity = utils.toWei("10", "ether");
...
 });
-const bob = actor.create("bob", {
-    ethConfig: global.harness.ledgers_config.ethereum,
+const bob = new Actor("bob", global.config, global.test_root, {
+    ethConfig: global.ledgers_config.ethereum,
 });
 const alice_final_address = "0x00a329c0648769a73afac7f9381e08fb43dbea72";
-const alice_comit_node_address = alice.config.comit.comit_listen;
-const bob_comit_node_address = bob.config.comit.comit_listen;
+//const alice_comit_node_address = alice.comitNodeConfig.comit.comit_listen;
```
LLFourn: 🔧 delete if not necessary

api_tests/lib/actor.ts
```diff
@@ -1,39 +1,72 @@
...
-                global.harness.test_root +
-                    "/" +
-                    node_config.config_dir +
-                    "/default.toml",
-                "utf8"
-            )
-        );
-        this.wallet = wallet.create(name, config);
+        if (testConfig) {
+            const metaComitNodeConfig = testConfig.comit_node[name];
```
bonomat: ❓ Why **meta**ComitNodeConfig?

api_tests/dry/rfc003_swap_reject.ts
```diff
@@ -313,55 +318,54 @@ describe("SWAP request REJECTED", () => {
...
 
     it("[Bob] Can execute a decline action, without providing a reason", async () => {
-        let bob_decline_href_2;
+        let bob_decline_href_2: string;
 
         await chai
             .request(bob.comit_node_url())
             .get(bob_reasonable_swap_href)
-            .then(res => {
+            .then(async function(res) {
```
thomaseizinger: I really dislike the way we mix `async/await` and `.then` in these tests. Since these things are going away, there is no point in changing it now but in your other PR, you could think about changing everything to `async/await` to make things more readable and consistent.

api_tests/harness.ts
```diff
@@ -1,10 +1,14 @@
-#!/usr/bin/env node
+#!/usr/bin/env ./api_tests/node_modules/.bin/ts-node --project api_tests/tsconfig.json
```
thomaseizinger: Nice! I didn't know you could do that :D

api_tests/harness.ts
```diff
@@ -176,11 +185,13 @@ class ComitRunner {
 }
 
 class BtsieveRunner {
+    running_btsieves: { [key: string]: any };
```
thomaseizinger: :wrench: I guess this `any` is also a `ChildProcess`?

api_tests/lib/actor.ts
```diff
@@ -1,39 +1,72 @@
...
+        }
     }
 
     comit_node_url() {
-        return "http://" + this.host + ":" + this.config.http_api.port;
+        return "http://" + this.host + ":" + this.comitNodeConfig.http_api.port;
     }
 
-    poll_comit_node_until(chai, location, predicate) {
+    // `chai` used to be passed down. Not sure why?
```
thomaseizinger: :x: This comment seems to be pretty useless :)

Regarding the actual comment: It is kind of dirty that this lib depends on `chai` since that means it knows it will be executed from within a test.

api_tests/lib/btsieve.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export interface BitcoinResponse {
+    query: any;
+    matches: IdMatch[];
+}
+
+export interface EthereumResponse {
+    query: any;
+    matches: EthereumMatch[];
```
thomaseizinger: :x: This does not accurately represent the API. You can choose the type of response for each ledger so `EthereumResponse` might also have `IdMatch`s, depending on which query parameter you return.

api_tests/lib/comit.ts
```diff
@@ -0,0 +1,59 @@
...
+    status: string;
+}
+
+export interface Action {
+    type: string;
+    payload: any;
+}
+
+export interface SwapsResponse {
+    _link: any;
```
thomaseizinger: :x: Should be `_links`.

You could also simply create a type called `HalResource` that contains these two properties to accurately represent, why we have these fields.

api_tests/lib/ethereum.ts
```diff
@@ -81,31 +86,38 @@ export class EthereumWallet {
         return this._address;
     }
 
-    async fund(ethAmount: BN) {
+    async fund(ethAmount: string) {
         const parity_dev_account = "0x00a329c0648769a73afac7f9381e08fb43dbea72";
         const parity_dev_password = "";
-        const weiAmount = _web3Client.utils.toWei(ethAmount, "ether");
+
+        const weiAmount = utils.toWei(ethAmount, "ether");
```
thomaseizinger: :wrench: If you would use `let` you could overwrite the variable.

api_tests/package.json
```diff
@@ -27,18 +28,20 @@
         "prettier": "^1.15.2",
         "satoshi-bitcoin": "^1.0.4",
         "toml": "^2.3.3",
+        "ts-node": "^8.0.3",
```
thomaseizinger: :x: Should be a dev dependency.

api_tests/dry/rfc003_swap_reject.ts
```diff
@@ -313,55 +318,54 @@ describe("SWAP request REJECTED", () => {
...
                 res.should.have.status(200);
                 bob_decline_href_2 = res.body._links.decline.href;
                 bob_decline_href_2.should.equal(
                     bob_reasonable_swap_href + "/decline"
                 );
-            });
+                let decline_res = await chai
```
LLFourn: ❌ I think it was better before not inside the `then`. Why change it?

api_tests/lib/util.js
```diff
@@ -1,20 +0,0 @@
-{
```
LLFourn: Good night sweet prince.

api_tests/lib/actor.ts
LLFourn: I think @D4nte got rid of this in his consolidation branch anyway.

api_tests/harness.ts
LLFourn: Last chance to remove `harness.js` from the typescript stuff :D.

api_tests/dry/rfc003_swap_reject.ts
LLFourn: (oh thomas made a comment wrt to removing all then stuff which we should consider).

api_tests/lib/actor.ts
D4nte: Because it's not the actual comit node config.
This refers to the `[comit_node.alice]` section in the `config.toml` file that is with the test.

The section look like that:
```
[comit_node.bob]
host = "127.0.0.1"
config_dir = "./regtest/bob/"
```

The `config_dir` contains the path to the directory that contains the actual configuration that comit-rs uses.

api_tests/dry/rfc003_swap_reject.ts
D4nte: I was trying to adopt a more functional style.

You reckon all `async/await` is a good way to go?

api_tests/lib/actor.ts
D4nte: The comment should be removed and now is.

I do not think there is much choice here. The request needs to be done with `chai` in any case.

api_tests/lib/btsieve.ts
D4nte: I updated, tell me what you think and/or propose alternative.

api_tests/lib/comit.ts
D4nte: Ok. Do you think `Swap` (few lines below) should also extend `HalResource`? Because it never contains `_embedded`. Is is it considered optional in `HalResource`?

api_tests/lib/ethereum.ts
D4nte: They have different meaning as one is in wei and other in Ether so happy not to override.

api_tests/dry/rfc003_swap_reject.ts
D4nte: because the variable may not be initialised as it is initialised inside the `then`.

api_tests/lib/comit.ts
thomaseizinger: Yes to both questions!

api_tests/dry/rfc003_swap_reject.ts
thomaseizinger: I would prefer that yes.

D4nte: Ok, I'll do that in the other PR then.

api_tests/lib/actor.ts
bonomat: Thanks for the clarification.

PR #819, opened 2019-03-18
Consolidate e2e tests

### LINKED ISSUE #767: Consolidate e2e tests

D4nte:
Resolves #767


api_tests/e2e/rfc003/btc_eth/happy.ts
```diff
@@ -2,283 +2,153 @@ import * as bitcoin from "../../../lib/bitcoin";
...
-        )) as SwapsResponse;
-        let swap_embedded = body._embedded.swaps[0];
-        swap_embedded.protocol.should.equal("rfc003");
-        swap_embedded.status.should.equal("IN_PROGRESS");
-        let swap_link = swap_embedded._links;
-        swap_link.should.be.a("object");
-        bob_swap_href = swap_link.self.href;
-        bob_swap_href.should.be.a("string");
+    describe("RFC003: Bitcoin for Ether", async () => {
+        execute(alice, bob, actions, initialUrl, listUrl, swapRequest);
```
thomaseizinger: :wrench: This method really just creates the test steps right? Can we name it according to that? :)

api_tests/e2e/rfc003/btc_eth/happy.ts
```diff
@@ -2,283 +2,153 @@ import * as bitcoin from "../../../lib/bitcoin";
...
-        let bob_satoshi_received = await bitcoin.getFirstUtxoValueTransferredTo(
-            bob_redeem_txid,
-            bob_final_address
-        );
-        const bob_satoshi_expected = alpha_asset_quantity - alpha_max_fee;
-
-        bob_satoshi_received.should.be.at.least(bob_satoshi_expected);
-    });
-});
+test();
```
thomaseizinger: :wrench: Instead of having to name to this function, you could use the [IIFE pattern](https://developer.mozilla.org/en-US/docs/Glossary/IIFE).

D4nte:
Point 2. 

> I am not convinced that things are better with this PR than they were before.
> 
> The actual logic seems be scattered among places and it is really hard to see, where things are happening.
> I was hoping that with this ticket, we can actually create a test-framework _independent_ library that can use the COMIT node API and only use the test framework to provide the execution environment.
> 
> Such a library could:
> 
>     1. potentially reused in applications talking to comit-rs
> 
>     2. give us a clear picture of how good our API is and where we need to invest in terms of self-descriptivness

Not sure where you want to go from there.
1. This is the first time I hear of it.
2. Yes, it gives a picture. Actions are not self-descriptive enough (GET/POST methods) + list swaps link should be available at protocol level (ie, `swaps/rfc003` should return a list of swaps).

Please clarify your expectations as I believe this is done inline with what was asked issue #767.

thomaseizinger:
> Not sure where you want to go from there.

Me neither that is why I want to discuss it with the team.

Regarding 1.: This was just me thinking ahead a bit. Since our tests are clients of the HTTP API, they very likely contain code that can potentially re-used by other clients.

My main concern is that with this PR, the layer between our test code and the actual API that we are testing becomes too thick. There too much non-production code that actually doesn't have any tests on its own.

Creating a library that contains all that code allows it to be tested by itself (if necessary) but more importantly, cleanly separates the code.

> Please clarify your expectations as I believe this is done inline with what was discussed for issue #767. If you had something else in mind then maybe best to write it clearly in an issue?

I think my concern is orthogonal to what is described in #767. The amount of actual test code should always be kept to a minimum so that your tests have more value. The more code you write just for testing, the higher the potential that the test code has a bug that maybe hides an actual bug in the code you are trying to test.
But that is just one argument for why I would like the code to be more separated. The other is that I genuinely don't understand what is going on without thinking a lot about certain specifics, for example why there is a seemingly random call to a function `run()` just after the test declaration.

As I said yesterday, I think we are bending the test framework too much here. This adds more complexity IMO than what was present with the duplication before.

D4nte:
After more thinking.

Yes I agree that more code to test means more potential for a bug in the testing code.

I do not really agree with the bending of the testing framework. We use standard options to allow us to run the code.
The `run()` call is not random if you read the Mocha documentation. This is not a hidden feature of the framework I pulled out of my hat.
I feel like you are saying we should only half-use a framework, otherwise we should write our own.

If you are saying that we should make a client library out of this then I would be happy to oblige.

At the end of the day, there is no code really new, most of it has just been factorised in a function.

api_tests/e2e/rfc003/btc_eth/happy.ts
```diff
@@ -2,283 +2,153 @@ import * as bitcoin from "../../../lib/bitcoin";
...
+            timeout: 10000,
+        }),
+        new ActionTrigger({
+            actor: alice,
+            action: Action.Redeem,
+            timeout: 10000,
+            afterTest: new AfterTest(
+                "[alice] Should have received the beta asset after the redeem",
+                async function(swapLocations: { [key: string]: string }) {
+                    let body = (await alice.pollComitNodeUntil(
```
LLFourn: 🔧 All these `AfterTest` can be deleted in favour of state based assertions.

```js
{
  actor: alice,
  state: state => state.beta_ledger.status === "Redeemed",
  timeout: 10000,
  name: "Should have received the beta asset after the redeem",
  test: async function(swap) {
    const aliceEthBalanceAfter = await ethereum.ethBalance(
      aliceFinalAddress
    );
    const aliceEthBalanceExpected = aliceEthBalanceBefore.add(
      betaAssetQuantity
    );
    aliceEthBalanceAfter
      .eq(aliceEthBalanceExpected)
      .should.be.equal(true);
  }
}
```

api_tests/e2e/rfc003/eth_btc/happy.ts
```diff
@@ -2,271 +2,145 @@ import * as bitcoin from "../../../lib/bitcoin";
...
+        alpha_expiry: alphaExpiry,
+        beta_expiry: betaExpiry,
+        peer: bobComitNodeAddress,
+    };
+
+    const actions: ActionTrigger[] = [
+        new ActionTrigger({
+            actor: bob,
+            action: Action.Accept,
+            timeout: 10000,
```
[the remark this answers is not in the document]
D4nte: Ah yes, I need to do a pass and re-adjust the timeouts as they were originally, I did not do so as I wanted to know what would happen with this PR.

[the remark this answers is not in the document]
D4nte: Actually it's always 10000...

D4nte:
@LLFourn all your comments have been implemented.

thomaseizinger:
> I do not really agree with the bending of the testing framework. We use standard options to allow us to run the code.

My former boss once said: "Frameworks: once you hit the frame, they don't work". The usecase mocha's APIs are designed for is writing statically defined tests. 
We are doing quite the opposite. In addition, our tests involve async setup code which doesn't make it simpler to follow the execution path when you are reading the code.

> The run() call is not random if you read the Mocha documentation. This is not a hidden feature of the framework I pulled out of my hat.

Things unfamiliar to you (when reading through a new code base) should be obvious or easily discoverable. In that regard, `run()` is a pretty hidden feature (where is the globally defined function coming from?). I'd almost call it a hack in the mocha library because they could not figure out how to wait for an async before or something like that.

> I feel like you are saying we should only half-use a framework, otherwise we should write our own.

A framework should help you and not be in your way. The fact that you have to use `run()` is only one sign that the framework was not designed to cover your usecase.

Another is your comment about the `setTimeout` call:
```
// the `setTimeout` forces it to be added on the event loop
// This is needed because there is no async call in the test
// And hence it does not get run without this `setTimeout`
```

I really appreciate the work you've put into this and it is definitely a step in the right direction. I'd just hope that we are all on the same page that using the test framework in there is quite some technical debt that should be cleaned up rather sooner than later.

api_tests/e2e/test_creator.ts
```diff
@@ -0,0 +1,210 @@
...
+        const swapLocation: string = swapLink.self.href;
+        swapLocation.should.not.be.empty;
+        swapLocations["bob"] = swapLocation;
+    });
+
+    while (actions.length !== 0) {
+        let action = actions.shift();
+        let actionHref: string = null;
+        let actionDirective: ActionDirective = null;
+        let actionExecutionResult: any = null;
```
LLFourn: 🔧 This variable is never read. You could delete it.

api_tests/e2e/test_creator.ts
```diff
@@ -0,0 +1,210 @@
...
+                    actionHref,
+                    actionDirective
+                );
+            }
+        );
+
+        let body: any = null;
+        if (action.state) {
+            it(
+                "[" + action.actor.name + "] state is as expected",
```
LLFourn: Nice. I actually didn't realise when I suggested it that a state assertion and an action execution could be rolled into the same object (with the state assertion happening after).

🔧 I'd prefer the text:
```
`[  ${action.actor.name}  ] transitions to correct state`
```

🤔It would be nice if we could say which state we are waiting for. We could do this by using something like: https://github.com/dchester/jsonpath#readme and then specifying the predicate as `"$.beta_ledger.status", "Redeemed"`.

api_tests/e2e/rfc003/eth_btc/happy.ts
```diff
@@ -2,271 +2,140 @@ import * as bitcoin from "../../../lib/bitcoin";
...
+        },
+        {
+            actor: alice,
+            action: ActionKind.Redeem,
+            uriQuery: { address: aliceFinalAddress, fee_per_byte: 20 },
+            state: (state: any) => state.beta_ledger.status === "Redeemed",
+            test: {
+                description:
+                    "[alice] Should have received the beta asset after the redeem",
+                callback: async (body: any) => {
```
LLFourn: 🔧 I'd prefer:
```
test: [ "[alice] Should have received the beta asset after the redeem" , async () => {
....
}, 10000]
```
Unless it doesn't format nicely or something.

thomaseizinger:
Since this is merged without moving away from the test framework, can we please have a ticket that tracks that?

D4nte:
Go for it, best if you open it so you can clarify your expectations.

Issue #643, opened 2019-01-14
Align COMIT-rs with RFC-002 (lower-case)


D4nte:
As per [RFC-002](https://github.com/coblox/COMIT-RFCs/blob/master/RFC-002-SWAP.md) and [COMIT-registry](https://github.com/coblox/COMIT-RFCs/blob/master/COMIT-registry.md), the following changes should be done to COMIT-rs:
- [ ] [Use`protocol` instead of `swap_protocol`](https://github.com/coblox/COMIT-RFCs/blob/master/RFC-002-SWAP.md#protocol)
- ~Sort out reasons: tracked with comit-network/comit-rs/#640~
- [ ] Use ledger/asset `bitcoin`, `ether`, `erc20`

D4nte:
This is not done, only the HTTP API was updated.

PR #821, opened 2019-03-19
Allow preflight requests with OPTIONS method

luckysori:
This is useful because some browsers will send a preflight OPTIONS request as part of [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), so that the server can respond whether it is acceptable to send the request with certain parameters.

Specifically, a preflight request to the comit node would use the `Access-Control-Request-Headers` and the `Access-Control-Request-Method` headers.

Additionally, related to #643, because it forces ledgers and assets to be serialized/deserialized using lowercase.

Requests to the comit node are only allowed from origin `localhost:3000` now. Issue #825 tracks allowing requests from `localhost` plus any port to be permitted.


application/comit_node/src/http_api/route_factory.rs
```diff
@@ -73,14 +73,21 @@ pub fn create<T: MetadataStore<SwapId>, S: state_store::StateStore>(
         .and(warp::path::end())
         .and_then(http_api::routes::peers::get_peers);
 
-    rfc003_get_swap
+    warp::options()
```
thomaseizinger: :x: This should be its own route similar to any others.

thomaseizinger: By its own route I mean extract the two lines into a variable.

luckysori:
> Thank you for this merge gift

It's not merged yet, you can merge yours first!!!!

D4nte:
> > Thank you for this merge gift
> 
> It's not merged yet, you can merge yours first!!!!

Let's hope so then :)

thomaseizinger:
I would hesitate to add a config field for now however, I would be okay with restricting the origin header to only allow `localhost` but with any port.

PR #827, opened 2019-03-25
Test and fix Refund

### LINKED ISSUE #411: When refund is possible...
.. the http API will return the `refund` actions. See #384 and #410.
Alice & Bob should be able to proceed with the refund.

Dod:
- [x] Implement the refund action route `GET swaps/rfc003/:id/refund`
- [x] Returns details as soon as technically possible (ie, funding has happened)
- [x] Return data to allow user to proceed with refund
- [x] Remove `#allow[dead_code]` here: https://github.com/comit-network/comit-rs/blob/6cc6ea0b1eb9018c256cfb7d5c498098ff53c2ed/application/comit_node/src/http_api/rfc003/handlers/get_swap.rs#L126-L133

Pre-work:
- Decide on the data to be returned per blockchain
- Decide on how to acquire data to build response from the state
- Maybe do #467 first

Child of #405.
Blocked by #767.
Blocked by #576.
Blocked by #679
Blocked by #693
Blocked by coblox/COMIT-RFCs#7

D4nte:
Resolves #411


api_tests/e2e/rfc003/eth-erc20_btc/refund.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    const actions = [
+        {
+            actor: bob,
+            action: ActionKind.Accept,
+            requestBody: {
+                beta_ledger_refund_identity: bob.wallet.eth().address(),
+                alpha_ledger_redeem_identity: bobFinalAddress,
+            },
+            state: (state: any) => state.communication.status === "ACCEPTED",
```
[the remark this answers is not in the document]
D4nte: Please open an issue to get it fixed in `application/comit_node` code.

application/comit_node/src/swap_protocols/rfc003/alice/actions/erc20_btc.rs
```diff
@@ -99,21 +99,32 @@ impl Actions for alice::State<Ethereum, Bitcoin, Erc20Token, BitcoinQuantity> {
...
             _ => vec![],
+        };
+
+        if let Funded { htlc_location, .. } = beta_state {
+            actions.push(alice::ActionKind::Redeem(redeem_action(
+                &request,
+                &response,
+                *htlc_location,
+                self.secret_source.as_ref(),
+                self.secret_source.secret(),
```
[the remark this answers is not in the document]
D4nte: This is existing code, indeed this function API can be better. Will try to fix.

application/comit_node/Cargo.toml
```diff
@@ -80,10 +88,10 @@ lazy_static = "1"
 maplit = "1"
 memsocket = "0.1"
 pretty_env_logger = "0.2"
+serde_urlencoded = "0.5.4"
 spectral = "0.6"
-tiny-keccak = "1.4"
 testcontainers = "0.6"
-serde_urlencoded = "0.5.4"
+tiny-keccak = "1.4"
```
LLFourn: wat

application/comit_node/src/swap_protocols/rfc003/bob/actions/btc_erc20.rs
```diff
@@ -127,12 +127,18 @@ impl Actions for bob::State<Bitcoin, Ethereum, BitcoinQuantity, Erc20Token> {
...
             }
+
             (Funded { .. }, NotDeployed, _) => {
                 vec![bob::ActionKind::Deploy(deploy_action(&request, &response))]
             }
+
             (Funded { .. }, Deployed { htlc_location, .. }, _) => vec![bob::ActionKind::Fund(
                 fund_action(&request, &response, *htlc_location),
             )],
+            (_, Funded { htlc_location, .. }, _) => vec![bob::ActionKind::Refund(refund_action(
```
LLFourn: ❌  This doesn't show the refund action `alpha_ledger` is in `funded` state. It should always return refund when beta is in funded state.

api_tests/e2e/rfc003/btc_eth-erc20/refund_beta.ts
```diff
@@ -0,0 +1,183 @@
...
+                    bobErc20BalanceAfter
+                        .lt(bobInitialErc20)
+                        .should.be.equal(true);
+                },
+            },
+        },
+        {
+            actor: bob,
+            test: {
+                description: "[bob] Is waiting for beta htlc to expire",
```
LLFourn: 🤡 since we've already said it's `bob`, maybe the test framework can handle the `[Bob]` for us?

api_tests/e2e/rfc003/eth-erc20_btc/refund.ts
```diff
@@ -0,0 +1,210 @@
...
+        },
+        {
+            actor: alice,
+            test: {
+                description: "[alice] Is waiting for alpha htlc to expire",
+                callback: async () => {
+                    while (Date.now() / 1000 < alphaExpiry) {
+                        await sleep(200);
+                    }
+                },
```
LLFourn: ❌ This thing should be unnecessary. The action should be self descriptive so that when you say you want to do `ActionKind.Refund`, the framework should wait until the time specified in the action and then execute it.

application/comit_node/Cargo.toml
D4nte: Because I used `cargo add` to add a dependency it reformarted the `Cargo.toml` file.
Hopefully most dependencies are unchanged.

D4nte:
> AFAICT this PR doesn't include a way of telling the user when they should execute the action i.e. after or close to the timeout. The ticket didn't say this explicitly but I think it is implied. My suggestion for this would be that we change the API a little so that this is outside the actual payload of the action. i.e.
> 
> ```
> {
>    "type" : "ethereum-deploy-contract",
>    "payload" : {...},
>    "valid_at": 1554078956
> }
> ```

I am happy to proceed. Any objection from @comit-network/rust-devs ?

api_tests/e2e/rfc003/eth-erc20_btc/refund.ts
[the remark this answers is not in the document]
D4nte: tracked with #845

application/comit_node/src/http_api/routes/rfc003/handlers/get_action.rs
```diff
@@ -89,6 +91,8 @@ pub enum ActionResponseBody {
         amount: EtherQuantity,
         gas_limit: ethereum_support::U256,
         network: ethereum_support::Network,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        valid_from: Option<Timestamp>,
```
thomaseizinger: :x: I have some feedback on the field name here. I'd propose naming it `unavailable_until` because:

1. We usually use the inverse language when we talk about the refund action:

- "not yet able to do the refund"
- "have to wait for the refund action"

2. The field is optional. Does a missing `valid_from` mean it is not valid at all?

IMO, an optional negative constraint is more intuitive. If it is not there, you simply exclude it from the evaluation (that is what you did in the test).

application/comit_node/src/swap_protocols/rfc003/bitcoin/actions.rs
```diff
@@ -13,6 +14,7 @@ pub struct SpendOutput {
     // Remember: One man's input is another man's output!
     pub output: PrimedInput,
     pub network: Network,
+    pub invalid_until: Option<Timestamp>,
```
LLFourn: 👃Putting `invalid_until` into the field of the action was not what I had in mind. I was thinking you would have to wrap it in a struct with an inner action and the timeout. Since you managed to get the API right I'm not going to block this PR here.

thomaseizinger: I think we should do it right away in this PR. Nothing is really blocked and this PR introduces refund functionality so I think it should do it without any technical debt :)

D4nte: Starting to do the changes I am not sure I agree. `invalid_until` is part of the action, why would the internals be different than the API on this point?

For this is a parameter of the action, it is a parameter of the `SpendOutput`, the `nlocktime` needs to have a specific value related to `invalid_until` for this `SpendOutput` transaction to be as per Bitcoin consensus.

LLFourn: @D4nte that's an interesting point and I'm tempted to agree with it but I think not being able to do execute an action until a time is something that potentially any action could have. Should Ethereum Deploy contract have a `invalid_until`? At first glance there is no particular reason it should, but then what if the constructor code checks the time and if it's too early something bad happens (note I have an idea I've shared with @bonomat that does exactly this).

LLFourn:
I implemented what I had in mind with my comment.

I also did some rebasing and had to re-create some commits (many conflicts in cargo.lock). I also removed the re-formatting of the cargo.toml.

application/comit_node/src/swap_protocols/rfc003/bitcoin/actions.rs
D4nte: The work is done but to me it is an unnecessary to make it generic pre-emptively.

application/comit_node/Cargo.toml
```diff
@@ -37,6 +37,8 @@ rustic_hal = "0.2"
 serde = "1"
 serde_derive = "1"
 serde_json = "1"
+strum = "*"
```
thomaseizinger: :x: Please don't use `*` as version identifier :)

api_tests/lib/actor.ts
```diff
@@ -111,6 +112,14 @@ export class Actor {
             case "bitcoin-broadcast-signed-transaction": {
                 let { hex } = action.payload;
 
+                let expiry = parseInt(action.payload.invalid_until);
```
thomaseizinger: :thinking: Does this even work? Shouldn't it be `action.invalid_until`?

application/comit_node/src/lib.rs
```diff
@@ -21,6 +19,8 @@ extern crate bam;
 extern crate futures;
 #[macro_use]
 extern crate binary_macros;
+#[macro_use]
```
thomaseizinger: :x: Can we `use` the macro here?

api_tests/lib/actor.ts
LLFourn: yes well spotted. We now do the waiting in the actual test framework.

application/comit_node/src/lib.rs
LLFourn: yes.

api_tests/e2e/rfc003/btc_eth-erc20/refund_beta.ts
```diff
@@ -0,0 +1,171 @@
...
+    const bobFinalAddress =
+        "bcrt1qs2aderg3whgu0m8uadn6dwxjf7j3wx97kk2qqtrum89pmfcxknhsf89pj0";
+    const bobComitNodeAddress = bob.comitNodeConfig.comit.comit_listen;
+
+    const alphaAssetQuantity = 100000000;
+    const betaAssetQuantity = toBN(toWei("5000", "ether"));
+    const alphaMaxFee = 5000; // Max 5000 satoshis fee
+
+    const alphaExpiry: number =
+        new Date("2080-06-11T13:00:00Z").getTime() / 1000;
```
luckysori: :question: Sorry to get to this so late, but why do we only refund beta in some of these end to end tests?

LLFourn: Just thinking about this now, wouldn't it make way more sense to just do the refund test (one for each currency) where it is the `alpha_ledger`. i.e. Alice funds then Alice refunds. We'd put these tests in a `refund` directory;
- `refund/ethereum_eth.ts`
- `refund/bitcoin_bitcoin.ts`
- `refund/ethereum_erc20`
etc

And then maybe one full refund test (where we also refund `beta_ledger`).

D4nte: Actually, it would.

PR #837, opened 2019-04-01
Add ability to specify private key as hex as argument to keygen

LLFourn:
Add ability to specify private key as hex as argument to keygen

And:

- Stop logging the output
- Remove a bunch of unused stuff


LLFourn:
@thomaseizinger I will lazily generate the cli when I need it ;)

PR #844, opened 2019-04-04
Build HttpApiProblems consistently

### LINKED ISSUE #818: Use HttpApiProblem library properly
We discovered that we are actually using the `HttpApiProblem` library wrongly.

The `new` constructor set the *title* of the problem instance instead of the `type` field as we thought.

Ultimately, we should be setting proper `type_url`s on the problem instead that lead to documentation pages for our errors. Until then, I would recommend moving to the `with_title_and_type_from_status` constructor for now.

The information that is currently encoded in the `title` should be encoded as more human-readable in the form of a better written title and using the `detail` property.

For more information, I recommend reading the RFC: https://tools.ietf.org/html/rfc7807

luckysori:
Resolves #818.

### DIFF
--- a/application/comit_node/src/http_api/problem.rs
+++ b/application/comit_node/src/http_api/problem.rs
@@ -15,6 +15,12 @@ pub struct HttpApiProblemStdError {
     inner: HttpApiProblem,
 }
 
+#[derive(Debug, Serialize)]
+pub struct MissingQueryParameter {
+    pub data_type: &'static str,
+    pub description: &'static str,
+}
+
 impl From<HttpApiProblem> for HttpApiProblemStdError {
     fn from(problem: HttpApiProblem) -> Self {
         Self { inner: problem }
@@ -85,7 +91,7 @@ pub fn unexpected_query_parameters(
     parameters: Vec<String>,
 ) -> Result<HttpApiProblem, String> {
     error!(
-        "Unexpected GET parameters {:?} for an {} action type. Expected: None",
+        "Unexpected GET parameters {:?} for a {} action type. Expected: none",
         parameters, action
     );
     let mut problem = HttpApiProblem::with_title_and_type_from_status(StatusCode::BAD_REQUEST)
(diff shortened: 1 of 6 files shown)


application/comit_node/src/http_api/problem.rs
```diff
@@ -60,17 +63,31 @@ pub fn serialize(e: serde_json::Error) -> HttpApiProblem {
...
     error!("{} not yet implemented", feature);
     HttpApiProblem::with_title_and_type_from_status(StatusCode::INTERNAL_SERVER_ERROR)
+        .set_title("Feature not yet implemented.")
         .set_detail(format!("{} is not yet implemented! Sorry :(", feature))
 }
 
 pub fn action_already_done(action: Action) -> HttpApiProblem {
     error!("{} action has already been done", action);
-    HttpApiProblem::new("action-already-done").set_status(StatusCode::INTERNAL_SERVER_ERROR)
+    HttpApiProblem::with_title_and_type_from_status(StatusCode::INTERNAL_SERVER_ERROR)
```
thomaseizinger: :clown_face: This shouldn't be a 500, since it is clearly the clients fault. Retrying will not make this error go away :)

410 GONE seems to be very appropriate for this one:

> 410 Gone
> Indicates that the resource requested is no longer available and will not be available again. This should be used when a resource has been intentionally removed and the resource should be purged. Upon receiving a 410 status code, the client should not request the resource in the future. Clients such as search engines should remove the resource from their indices.[43] Most use cases do not require clients and search engines to purge the resource, and a "404 Not Found" may be used instead.

application/comit_node/src/http_api/problem.rs
```diff
@@ -60,17 +63,31 @@ pub fn serialize(e: serde_json::Error) -> HttpApiProblem {
...
 
 pub fn invalid_action(action: Action) -> HttpApiProblem {
     error!("{} action is invalid for this swap", action);
-    HttpApiProblem::new("invalid-action").set_status(StatusCode::CONFLICT)
+    HttpApiProblem::with_title_and_type_from_status(StatusCode::CONFLICT)
+        .set_title("Invalid action.")
+        .set_detail("Cannot perform requested action for this swap.")
+}
+
+pub fn unexpected_query_parameters(action: &str) -> HttpApiProblem {
```
thomaseizinger: :x: Instead of passing only the action name, it would be helpful to pass the unexpected query parameters as well.
A custom field in the `HttpApiProblem` seems to be appropriate for this.

https://github.com/chridou/http-api-problem/blob/c8cd8f68af4d5e193011d171104bb00d38db234d/src/lib.rs#L396-L402

Something like:

```rust
set_value("unexpected_parameters", vec![
  ...  
])
```

application/comit_node/src/http_api/routes/rfc003/handlers/get_action.rs
```diff
@@ -166,30 +166,32 @@ impl IntoResponseBody for bitcoin::SpendOutput {
...
+                .set_title("Invalid query parameter.")
+                .set_detail("Query parameter fee-per-byte is not a valid float.")),
             },
             _ => {
-                error!("Unexpected GET parameters for a bitcoin::SpendOutput action type. Expected: address and fee-per-byte.");
+                error!("Unexpected GET parameters for a bitcoin::SpendOutput action type. Expected: address and fee-per-byte");
                 let mut problem =
                     HttpApiProblem::with_title_and_type_from_status(StatusCode::BAD_REQUEST)
-                        .set_detail("This action requires additional query parameters");
+                        .set_title("Missing query parameter(s).")
```
thomaseizinger: :wrench: This one could also be extracted into its own function, similar to `unexpected_query_parameters`.

However, instead of flattening out the parameters, it is would probably be good to have them under a single property like `missing_query_parameters`.

Issue #487, opened 2018-11-26
Come up with logging strategy/guidelines


D4nte:
At the moment, there is no consistency in what is logged where and with what level of detail.

This spike is about investigating, how the current logs could be improved and proposing an overall logging strategy that makes the logs actually useful.
This includes looking into the following topics:

- Logs should include context (SwapId)
- Log level of logs should make sense (balance between noise and useful information)
- Formatting of variables in log messages should make sense (no {:?} of [u8] !!!) (maybe be tested?)
- How to control log level from the outside
- Identify high value logging areas and share them with the team
- Making use of the `target` parameter for log-messages to have consistent modules for log messages (https://docs.rs/log/0.4.6/log/macro.error.html)

Potentially useful links:

- https://github.com/rustasync/async-log
- https://github.com/tokio-rs/tracing

Related issue: https://github.com/comit-network/comit-rs/issues/1582

thomaseizinger:
I propose to ditch this spike and use `tracing` directly. Here is a talk that is very convincing IMO:

https://www.youtube.com/watch?v=JjItsfqFIdo

D4nte:
Include tracing as part of the scope of the spike.

thomaseizinger:
> Include tracing as part of the scope of the spike.

It was always there ;)
I am just suggesting that the need for a spike is probably very small and hence the barrier for starting to adopt tracing is also smaller.
Our log files are practically useless, that is why I would like to push for this :)

D4nte:
I don't see the harm of a small spike that explains what tracing is and how it is relevant to us.

tcharding:
Some useful links on log levels
- https://reflectoring.io/logging-levels/

PR #856, opened 2019-04-08
Remove unused dependencies and get rid of `macro_use`

thomaseizinger:
This PR gets rid of some unused dependencies. Since we now start to use dependabot, we should not bump dependencies that we are actually not using.

At the same time, I made an initial effort to get rid of the legacy way of importing macros (`#[macro_use]`) and convert it to the regular `use ....` syntax.

More on this here: https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macro_rules-style-macros

Tagging everyone so that we are all aware.
I think we should, step-by-step convert the whole codebase to the new way of importing macros.


application/comit_node/src/bam_api/rfc003/swap.rs
```diff
@@ -14,6 +14,7 @@ use bam::{
     Status,
 };
 use futures::future::{self, Future};
+use log::{error, warn};
```
D4nte: ❌ While I understand the risk of "poisoning the global space" I really think that for log macros, having to import them individually in each files is a pain.

application/comit_node/src/comit_client.rs
```diff
@@ -5,6 +5,8 @@ use crate::{
 use futures::Future;
 use std::io;
```
D4nte:
```suggestion
```

application/comit_node/src/settings/mod.rs
```diff
@@ -1,9 +1,9 @@
 mod serde;
 
 use crate::seed::Seed;
+use ::serde::Deserialize;
```
D4nte: ❌ can we avoid relative path please

luckysori: I think that just indicates that it's an external crate. Otherwise the name clashes with the local module `serde`. We have this same situation in several places.

application/comit_node/src/bam_api/rfc003/swap.rs
LLFourn: I'd say we could decide this in #487. For now just let them be global. If we are able to design things so that things are only logged at key places then we wouldn't have to import them in many places. i.e. You should be forced to think about actually logging things in random places.

application/comit_node/src/settings/mod.rs
thomaseizinger: Addressed :)

application/comit_node/src/bam_api/rfc003/swap.rs
thomaseizinger: > While I understand the risk of "poisoning the global space" I really think that for log macros, having to import them individually in each files is a pain.

I understand your point and would like to provide some data that might shed a different light onto this situation:

Here are some numbers for the comit_node crate:

- `use log::`: 22
- `use serde::`: 33
- `use std::`: 63
- `use futures::`: 18

I think, it is just a question of what we are used to. Log macros where "just" available. That is why it seems like a pain to import them now. Overall though, I think they are just symbols like any other. Also the numbers show that there is symbols that we have to import today so I don't think they deserve a special treatment.

thomaseizinger: @D4nte @LLFourn @luckysori 

We could also refer to those macros through an absolute path: `log::warn!("I am a warning!")`.

Thoughts?

luckysori: > We could also refer to those macros through an absolute path:
> `log::warn!("I am a warning!")`.
> 
> Thoughts?

I like that.

application/comit_node/Cargo.toml
```diff
@@ -18,30 +18,24 @@ directories = "1.0"
 either = "1.5"
 enum-display-derive = "0.1"
 failure = "0.1"
-frunk = "0.2"
```
D4nte: ❓ is there no way to automatise this?

application/comit_node/src/bin/comit_node.rs
```diff
@@ -18,15 +15,15 @@ use comit_node::{
     },
 };
 use directories;
-use ethereum_support::*;
```
D4nte: how was this not caught by rust warnings?

vendor/bam/src/lib.rs
```diff
@@ -1,17 +1,12 @@
...
-extern crate log;
-#[macro_use]
-extern crate debug_stub_derive;
-
 mod api;
 pub mod client;
 pub mod config;
 pub mod connection;
+
+#[macro_use]
```
D4nte: ❓ is that on purpose?

thomaseizinger: Yes, you have to export macros up until the root of the crate to use the `bam::` syntax.

application/comit_node/src/bin/comit_node.rs
thomaseizinger: It imported the re-exported `Future` trait through that, that is why it was not obsolete.

application/comit_node/Cargo.toml
thomaseizinger: Probably a cargo plugin, a quick search didn't bring anything up :(

PR #858, opened 2019-04-09
Return relative location for btsieve queries

### LINKED ISSUE #810: Remove `external_url` from btsieve
This is a relict from ancient times where I didn't know that you could return relative paths in the `Location` header of a 201 response.

Should be removed and returned links should be relative.

luckysori:
Resolves #810.


application/btsieve/config/default.toml
```diff
@@ -11,4 +11,3 @@ poll_interval_secs = 17
 [http_api]
 address_bind="0.0.0.0"
 port_bind=8080
-external_url="http://localhost:8080/"
```
D4nte: ❌  similar clean up needed in regtest folder

luckysori: Good catch!

PR #860, opened 2019-04-09
Embed comit-i in comit_node

### LINKED ISSUE #855: Embed comit-i in comit-rs
Follow-up of coblox/comit-i#17

DoD:
- [x] Research and how exactly to do it
- [x] Discuss release strategy
- [x] Implement release strategy

D4nte:
Resolves #855


api_tests/regtest/alice/default.toml
```diff
@@ -5,10 +5,13 @@ secret_seed = "f87165e305b0f7c4824d3806434f9d0909610a25641ab8773cf92a48c9d77670"
 [http_api]
 address="0.0.0.0"
 port=8000
-logging=true
+
+[comit_i]
```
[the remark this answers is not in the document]
luckysori: Or maybe add it to Bob's and Charlie's configs if you want these to be complete config files.

vendor/comit_i/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+authors = ["Franck Royer <franck@coblox.tech>"]
```
luckysori: Rude.

api_tests/regtest/alice/default.toml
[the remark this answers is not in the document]
D4nte: 1. We do for the cross-origin auth
2. It's actually good to have it, it allows you to quickly start a node with comit-i using `COMIT_NODE_CONFIG_PATH=./api_tests/regtest/alice/ cargo run --package comit_node`

vendor/comit_i/Cargo.toml
D4nte: Do `cargo init` locally and you'll see.

api_tests/regtest/alice/default.toml
thomaseizinger: :wrench: Should we just call this `web_gui`?

vendor/comit_i/Cargo.toml
luckysori: Of course you'd want to claim comit-i yours.

application/comit_node/src/comit_i_routes.rs
```diff
@@ -0,0 +1,98 @@
...
+            Error::PathConversionFail => "Conversion of the path to PathAndQuery failed",
+        }
+    }
+
+    fn cause(&self) -> Option<&dyn StdError> {
+        None
+    }
+}
+
+pub fn create() -> BoxedFilter<(impl Reply,)> {
```
thomaseizinger: :wrench: I liked it better when this was part of the `comit_i` crate. Why did you remove it?

application/comit_node/src/comit_i_routes.rs
```diff
@@ -0,0 +1,98 @@
...
+            Asset::get("index.html")
+        })
+        .ok_or_else(|| Error::IndexHtmlMissing)?;
+
+    Ok(Response::builder()
+        .header("content-type", mime.to_string())
+        .body(file))
+}
+
+pub fn unpack_problem(rejection: Rejection) -> Result<impl Reply, Rejection> {
```
thomaseizinger: :confused: This is not unpacking a `problem` (you are not using HttpApiProblem) - so I think the function name should not indicate that.

application/comit_node/src/comit_i_routes.rs
```diff
@@ -0,0 +1,98 @@
...
+    if let Some(&err) = rejection.find_cause::<Error>() {
+        let code = match err {
+            Error::IndexHtmlMissing => StatusCode::INTERNAL_SERVER_ERROR,
+            Error::PathConversionFail => StatusCode::BAD_REQUEST,
```
thomaseizinger: :question: Why is a BAD REQUEST? Sounds like an internal error to me.

api_tests/regtest/alice/default.toml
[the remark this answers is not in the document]
luckysori: I see now! I like that then. Please add it to the other configs though :wink:

application/comit_node/src/comit_i_routes.rs
```diff
@@ -0,0 +1,98 @@
...
+        .boxed()
+}
+
+fn serve(path: Tail) -> Result<impl Reply, Rejection> {
+    let path = path.as_str();
+    let path_and_query = PathAndQuery::from_str(path).map_err(|e| {
+        error!("Could not convert path {} to PathAndQuery: {:?}", path, e);
+        Error::PathConversionFail
+    })?;
+    let path = path_and_query.path();
```
thomaseizinger: :confused: How many times do you want to convert back and forth between `Path`s? :see_no_evil: 

Is there no easier way to do this?

The function `guess_mime_type_opt` [takes anything that is `AsRef<Path>`](https://docs.rs/mime_guess/2.0.0-alpha.6/mime_guess/fn.guess_mime_type_opt.html), which is implemented for `&str` if I remember correctly. You should be able to simply pass `path.as_str()` to it.

Also, it looks like using [`warp::path::full()`](https://docs.rs/warp/0.1.15/warp/filters/path/struct.FullPath.html) might be more appropriate than `warp::path::tail()` because you actually want to match against the full thing and not the remaining, unmatched part.
In your concrete case, it is semantically equivalent because you don't do any matching before.

application/comit_node/src/comit_i_routes.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    let mut mime = mime_guess::guess_mime_type_opt(path)
```
thomaseizinger: :wrench: Maybe there is a way to avoid the mutability + double parsing here:

1. `let asset: Option<Cow<'static, [u8]>> = Asset::get(path);`
2. `map` asset to a response + guess the mimetype ([`guess_mime_type`](https://docs.rs/mime_guess/2.0.0-alpha.6/mime_guess/fn.guess_mime_type.html) seems to more appropriate here instead of defaulting to `text/html`)
3. `unwrap_or_else` the option to serve the `index.html` asset by constructing a `Response` that statically sets the `content-type` header to "text/html" (atm you are parsing first and a `to_string` afterwards.

You can fetch the `index.html` asset upfront, since nothing works without that one anyway :)

```rust
let index = Asset::get("index.html").ok_or_else(|| Error::IndexHtmlMissing)?
```

thomaseizinger: This causes a bit of duplication in constructing the `Response`, however, I'd argue that it is overall preferable since you keep the special-case of serving the `index.html` in case the asset is not found separate.

You could even extract it in its own function to not clutter up this one.

vendor/comit_i/build.rs
```diff
@@ -0,0 +1,16 @@
+use std::fs::File;
+
+fn main() {
+    let version = include_str!("./version").trim();
```
thomaseizinger: It might be useful to do some logging here like
- which version of comit-i we are downloading
- where the tmp file is located

Reference: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

application/comit_node/src/comit_i_routes.rs
LLFourn: I guess because it means the client has requested something that doesn't exist (it should probably be 404).

LLFourn:
This seems to need something in gitignore:

```
On branch 855-embed-comit-i
Your branch is up to date with 'origin/855-embed-comit-i'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	vendor/comit_i/build/

nothing added to commit but untracked files present (use "git add" to track)
```

application/comit_node/src/comit_i_routes.rs
D4nte: I must say I find the handling of error with warp extremely confusing. Feel free to 🔧 something cleaner/simpler.

application/comit_node/src/comit_i_routes.rs
D4nte: What was in the comit_i crate was just an example to use rust-embed with warp. I have now moved it directly to rust-embed repo.

To me, the comit_i crate just provides comit_i. The lib consumer should have the choice on how to serve it (warp, rocket, etc).
I did not want to make this choice for the consumer that is why it is in comit_node crate.

application/comit_node/src/comit_i_routes.rs
D4nte: this path is coming from the client and if the conversion to a "path" fails it means it's a bad request.

application/comit_node/src/comit_i_routes.rs
D4nte: > 😕 How many times do you want to convert back and forth between Paths? 🙈
I am sure you can find a better way to phrase your point.

I'll tackle the rest of the comment.

D4nte: Please note that unfortunately warp hides the "PathAndQuery" from the `Tail` struct making this cumbersome.

I'll see if I can make it better with `warp::path::full()`

application/comit_node/src/comit_i_routes.rs
```diff
@@ -0,0 +1,98 @@
...
+        .unwrap_or_else(|| "text/html".parse::<Mime>().unwrap());
```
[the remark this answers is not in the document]
D4nte: haha no I cannot. The `mime` is version `0.3` whereas `mine_guess` internally uses `0.2`. It re-exports the `Mime` struct from the `mime` crate but that's it. Hence I cannot use the correct `mime::TEXT_HTML`.

I'll do a PR on `mime_guess`.

D4nte:
> This seems to need something in gitignore:

Weird, I'll see if I can correct but the `.gitignore` is there.

application/comit_node/src/comit_i_routes.rs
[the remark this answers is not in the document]
D4nte: https://github.com/abonander/mime_guess/pull/44

api_tests/regtest/alice/default.toml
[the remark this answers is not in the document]
LLFourn: After discussion can we:

1. Remove this from here
2. Have COMIT-i tests in its own directory and have a `config.toml` that sets the config dir to something specific to COMIT-i tests (i.e. not `./api_tests/regtest/alice/`)

To make the above work you'll need to change the harness so it doesn't run tests from `api_tests` as it's cwd. This will cause some collateral damage in various places.

application/comit_node/src/comit_i_routes.rs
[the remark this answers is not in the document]
LLFourn: Heh. You can still do it better in the old version:
`mime::mime!(Text/Html)`

There's an example in the docs:
```rust
#[macro_use] extern crate mime;
fn main() {
     let json = mime!(Application/Json);
     let plain = mime!(Text/Plain; Charset=Utf8);
     let text = mime!(Text/Html; Charset=("bar"), ("baz")=("quux"));
     let img = mime!(Image/_);
}
```

application/comit_node/src/comit_i_routes.rs
LLFourn: Ah good point. If the file doesn't exist what happens?

application/comit_node/src/comit_i_routes.rs
D4nte: sorted now

application/comit_node/src/comit_i_routes.rs
D4nte: This has been removed :)

application/comit_node/src/comit_i_routes.rs
thomaseizinger: Hmm, comit_i by itself is provided in its own repository. Everything that is in the `comit_i` crate can be trivially built by everyone (no real gain of re-use here).

For now though, only the functionality matters :)

application/comit_node/src/comit_i_routes.rs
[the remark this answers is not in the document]
D4nte: I don't have access to the correct version of `mime`. Code has been removed anyway.

vendor/comit_i/Cargo.toml
D4nte: God no!

api_tests/regtest/alice/default.toml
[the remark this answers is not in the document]
D4nte: The 🚀 is here to show that the scope creep is heading to the moon.

vendor/comit_i/build.rs
```diff
@@ -0,0 +1,17 @@
...
+
+    let url = format!(
+        "https://github.com/comit-network/comit-i/releases/download/{}/bundle.zip",
+        version
+    );
+
+    println!("Fetching {}", url);
+    let mut tmpfile: File = tempfile::tempfile().unwrap();
+    reqwest::get(&url).unwrap().copy_to(&mut tmpfile).unwrap();
+    unzip::Unzipper::new(tmpfile, "./").unzip().unwrap();
```
LLFourn: 🔧  Use expect for these.

Issue #865, opened 2019-04-11
Generate Bitcoin HTLC from RFC


D4nte:
Use the bytes from the RFC for Bitcoin HTLC.

Blocks #606.
Blocked by coblox/blockchain_contracts#3

bonomat:
Duplicate and/or solved with #933

PR #866, opened 2019-04-11
Use Ethereum HTLC from blockchain_contracts crate

### LINKED ISSUE #864: Generate/test Ethereum HTLCs and provide offset in a vendor/crate
create vendors/blockchain_contracts crate:
- [x] move eth htlc tests in it
- [x] copy eth htlc to it
- [x] add offset definition feature to it

D4nte:
It is a bit odd because Bitcoin HTLC uses concept from blockchain_crates but is still in comit_node.

This is normal and will be cleaned up with #865.

Resolves #864


D4nte:
Not what is asked

PR #868, opened 2019-04-11
Update to match GitHub rules

### LINKED ISSUE #793: Reconfigure mergify to not run if there are requested changes
Currently, mergify will try and merge a PR if it has an approval AND requested changes but it will run into GitHub's branch protection.

We should fix that by adding an additional rule to the mergify configuration.

D4nte:
Resolves #793

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -6,6 +6,7 @@ pull_request_rules:
       - label!=work-in-progress
       - label!=blocked
       - label=automagically-mergify-this
+      - "#changes-requested-reviews-by=0"
     actions:
       merge:
         method: merge


.mergify.yml
```diff
@@ -2,10 +2,11 @@ pull_request_rules:
   - name: merge on CI success and one approval
     conditions:
       - "status-success=Travis CI - Pull Request"
-      - "#approved-reviews-by>=1"
+      - "#approved-reviews-by>=2"
```
thomaseizinger: :question: Why all of the sudden? :O

D4nte: I thought it was the current configuration but it is not indeed.

PR #874, opened 2019-04-15
Btsieve: (n)one ledger support

### LINKED ISSUE #557: btsieve to support one ledger a time
Make ledger configuration optional. For example btsieve should be able to start with only having bitcoin configured.

In any case, all routes of all ledgers should be registered but return errors if they are invoked and the ledger is not configured.

bonomat:
This PR should allow to startup btsieve with just one/or none ledger.

Resolves #557


api_tests/lib/ledger_runner.ts
```diff
@@ -0,0 +1,93 @@
+import * as bitcoin from "./bitcoin";
+import * as toml from "toml";
+import * as fs from "fs";
+import { sleep } from "./util";
+import { execSync, spawn } from "child_process";
+
+const project_root: string = execSync("git rev-parse --show-toplevel", {
```
LLFourn: ❌ Pass in `working_dir` and `log_dir` into the `LedgerRunner` so you won't need to execute this line of code in each file. Liekwise, pass in `log_dir` into btsieve runner.

application/btsieve/src/bin/btsieve.rs
```diff
@@ -52,18 +52,21 @@ fn main() -> Result<(), failure::Error> {
...
     let bitcoin_routes = create_bitcoin_routes(&mut runtime, settings.bitcoin)?;
 
-    let (ethereum_routes, _event_loop_handle) =
-        create_ethereum_routes(&mut runtime, settings.ethereum)?;
+    let (ethereum_routes, _event_loop) = create_ethereum_routes(&mut runtime, settings.ethereum)?;
 
-    let routes = bitcoin_routes.or(ethereum_routes);
+    let log = warp::log("btsieve::api");
+    let ping_200 = warp::path("ping").map(warp::reply);
+    let ping_route = warp::get2().and(ping_200);
```
LLFourn: ❓ What's this for?

api_tests/btsieve/ethereum_only/test.ts
```diff
@@ -29,12 +28,10 @@ const aliceWallet = new Wallet("alice", {
...
 
 setTimeout(async function() {
-    describe("Test btsieve API", () => {
+    describe("Test btsieve API - ethereum", () => {
         let token_contract_address: string;
         before(async function() {
-            this.timeout(5000);
-            await bitcoin.ensureSegwit();
-            await tobyWallet.btc().fund(5);
+            this.timeout(10000);
```
LLFourn: 😄  This does less stuff but takes longer?

bonomat: This just means that the timeout is longer, in the best case this should not do any harm, in the worst case if something crashes, than this will only timeout after 10 sec

application/btsieve/src/bin/btsieve.rs
bonomat: It's an extra endpoint to test if btsieve is up and running.

LLFourn: What's it for in regards to completing this issue?

🔧 If nothing I think remove it.

api_tests/regtest/ledgers.toml
```diff
@@ -4,12 +4,12 @@ rpc_host = "localhost"
 rpc_port = "18443"
 rpc_username = "bitcoin"
 rpc_password = "54pLR_f7-G6is32LP-7nbhzZSbJs_2zSATtZV_r05yg="
-local_docker_wait = 4000
```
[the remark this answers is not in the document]
bonomat: Unfortunately yes. Bitcoin randomly just does not start up in time on my mac.
Happy to consider other solutions but we probably want to takle this in another refactoring ticket.

application/btsieve/src/bin/btsieve.rs
bonomat: I'm running a test where neither bitcoin nor ethereum is connected and I've added this route in order to be able check that btsieve started up correctly.

api_tests/lib/ledger_runner.ts
```diff
@@ -0,0 +1,103 @@
...
+    encoding: "utf8",
+}).trim();
+const docker_cwd = project_root + "/api_tests/regtest";
```
LLFourn: It looks like you forgot to delete these lines :)

api_tests/lib/ledger_runner.ts
```diff
@@ -0,0 +1,103 @@
...
+export class LedgerRunner {
+    private running_ledgers: { [key: string]: boolean };
+    private readonly block_timers: { [key: string]: NodeJS.Timeout };
+    private readonly docker_compose_options: { cwd: string; encoding: string };
+    private readonly docker_cwd: string;
+    private readonly log_dir: string;
+    private readonly ledgers_config: any;
+
+    constructor(
+        project_root: string,
```
LLFourn: I'd just pass in `docker_cwd` instead of this.
It may be nice to rename it to something like `docker_compose_dir` since that's what it really means.
Even better would be to make it `docker_compose_file` and have it be the path to the `docker-compose.yml`. Then we could stop setting the `cwd` everywhere and just use the `-f` flag to `docker-compose` to set the config file to use.

api_tests/lib/ledger_runner.ts
bonomat: 😇

api_tests/lib/ledger_runner.ts
bonomat: I like that idea of having passing just docker-compose file. Will do so.

api_tests/btsieve/bitcoin_only/test.ts
```diff
@@ -0,0 +1,233 @@
...
+                });
+
+                const to_address =
+                    "bcrt1qcqslz7lfn34dl096t5uwurff9spen5h4v2pmap";
+                let location: string;
+
+                it("btsieve should respond not found when creating a bitcoin transaction query for an invalid network", async function() {
+                    return chai
+                        .request(btsieve.url())
+                        .post("/queries/bitcoin/banananet/transactions")
```
D4nte: 🍌

api_tests/btsieve/no_ledger/test.ts
```diff
@@ -0,0 +1,85 @@
...
+                        .then(res => {
+                            res.should.have.status(200);
+                        });
+                });
+            });
+        });
+
+        describe("Bitcoin", () => {
+            describe("Transactions", () => {
+                it("btsieve should respond `SERVICE UNAVAILABLE` as ledger is not connected if queried for transaction", async function() {
```
D4nte: 🙃 Could we have this test in the `ethereum_only` tests?

api_tests/lib/btsieve_runner.ts
```diff
@@ -0,0 +1,56 @@
+import { ChildProcess, spawn } from "child_process";
+import { MetaBtsieveConfig } from "./btsieve";
+import * as fs from "fs";
+
+export class BtsieveRunner {
+    running_btsieves: { [key: string]: ChildProcess };
+    private readonly log_dir: string;
+    private readonly btsive_bin: string;
```
D4nte:
```suggestion
    private readonly btsieve_bin: string;
```

api_tests/lib/btsieve_runner.ts
```diff
@@ -0,0 +1,56 @@
+import { ChildProcess, spawn } from "child_process";
+import { MetaBtsieveConfig } from "./btsieve";
+import * as fs from "fs";
+
+export class BtsieveRunner {
+    running_btsieves: { [key: string]: ChildProcess };
```
D4nte: 🙃 Should we go camelCase?

api_tests/lib/btsieve_runner.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    async ensureBtsievesRunning(btsieves: [string, MetaBtsieveConfig][]) {
+        for (let [name, btsieve_config] of btsieves) {
+            console.log("Starting Btsieve: " + name);
+
+            if (this.running_btsieves[name]) {
+                continue;
+            }
+
+            this.running_btsieves[name] = await spawn(this.btsive_bin, [], {
```
D4nte: ❌ 
```suggestion
            this.running_btsieves[name] = await spawn(this.btsieve_bin, [], {
```

application/btsieve/config/bitcoin_missing.toml
```diff
@@ -0,0 +1,8 @@
+[ethereum]
```
D4nte: 🙃 I'd suggest to use the same terminology than in the api tests "ethereum_only" to be more consistent.

application/btsieve/src/bin/btsieve.rs
D4nte: 🔧 I am going to be the devil(@thomaseizinger)'s advocate here and suggest you replace this endpoint with `/health`. Looking at the RFC https://tools.ietf.org/id/draft-inadarei-api-health-check-01.html the response body only requires a `status` field so I think it'd be worth just implementing that and leaving the rest (`details`) to when someone needs it (as part of a future API improvement).

application/btsieve/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![warn(unused_extern_crates, rust_2018_idioms)]
```
D4nte: ❓ why removing this warn?

api_tests/btsieve/no_ledger/test.ts
bonomat: My idea was to have _clean_ end-to-end tests in the `bitcoin/ethereum_only` tests and these edge cases outside. 
Do you see an issue with having these tests here?

application/btsieve/src/lib.rs
bonomat: It is already included in `rust_2018_idioms` (check the commit message: https://github.com/comit-network/comit-rs/pull/874/commits/5d9bf694db170f84074667a80ce4d3f1a80f9fb7 ->
`#[warn(unused_extern_crates)] implied by #[warn(rust_2018_idioms)]`

api_tests/lib/btsieve_runner.ts
bonomat: I would rather not for now as we use a mix in between camel and snake_case... maybe a good point to bring up for a new linter?

api_tests/btsieve/no_ledger/test.ts
D4nte: I was proposing to duplicate to confirm that indeed there is bitcoin service unavailable when only ethereum is available. Feel free to not action the comment.

D4nte:
ok just need to return a body for `/health` and we're good.

application/btsieve/src/bin/btsieve.rs
bonomat: As discussed offline: this endpoint currently only serves as a small helper for our tests and we implement the linked  [RFC](https://tools.ietf.org/id/draft-inadarei-api-health-check-01.html). While it sounds indeed interesting, we should discuss this feature a bit more in detail, e.g. I can imagine returning information about the currently connected blockchains, database (if we have one), etc.

PR #878, opened 2019-04-16
Move network communication to libp2p

### LINKED ISSUE #800: Migrate network layer to libp2p
Based on the outcome of #779, the network layer of COMIT should be replaced with libp2p.

Since libp2p is a modular network stack, I'd propose the following configuration to begin with:

- Only TCP transport
- Discovery through WebRTC-star (or should be go with mDNS for now since we are always on the same network?)
- no encryption

This should allow us to run our tests again.

The following things will need to be done (not necessarily in this order):

- Throw away `comit_server` and `connection_pool`
- Figure out away on how to put BAM ontop of whichever abstractions libp2p exposes. This will involve registering a protocol with libp2p like `/bam/1.0.0`.
- Change the http_api to accept a libp2p node id instead of an IP address
- Figure out how to establish a connection to the given ID so that the swap request can be sent to the other guy. Not sure if libp2p will handle the discovery transparently for us or we have to do this manually before sending the request
- Derive the keypair used for the libp2p id from the seed specified in the configuration file
- Put COMIT on the list of notable users: https://github.com/libp2p/rust-libp2p#notable-users

Note:
- aim for multiple PRs
- create issues per PR
- Pair programming

thomaseizinger:
resolves #800


application/comit_node/src/bin/comit_node.rs
```diff
@@ -2,60 +2,106 @@
...
+    let local_peer_id = PeerId::from(local_key_pair.clone().public());
+
+    let transport = libp2p::build_development_transport(local_key_pair);
+    let behaviour = network::Behaviour::new(bob_protocol_dependencies, runtime.executor())?;
+
+    let mut swarm = Swarm::new(transport, behaviour, local_peer_id.clone());
+
+    Swarm::listen_on(
+        &mut swarm,
+        format!("/ip4/0.0.0.0/tcp/{}", settings.comit.comit_listen.port()).parse()?,
```
D4nte: Shall we consider ip6 from day1?

D4nte: ❓  What port does it listen on?

application/comit_node/src/libp2p_bam/protocol.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl<TSubstream> OutboundUpgrade<TSubstream> for BamConfig
+where
+    TSubstream: AsyncRead + AsyncWrite,
+{
+    type Output = BamStream<TSubstream>;
+    type Error = Infallible;
+    type Future = FutureResult<Self::Output, Self::Error>;
+
+    #[inline]
```
D4nte: ❓  can you justify please?

application/comit_node/src/swap_protocols/rfc003/alice/spawner.rs
```diff
@@ -79,7 +83,7 @@ impl<T: MetadataStore<SwapId>, S: StateStore> AliceSpawner for ProtocolDependenc
             alice.new_state_machine(
                 ledger_events.create_ledger_events(),
                 ledger_events.create_ledger_events(),
-                self.clone(),
+                self.client.clone(),
```
D4nte: it's an `Arc` if you fancy the `Arc::clone`

application/comit_node/src/bin/comit_node.rs
thomaseizinger: At the moment, it is listing on the comit-port (`settings.comit.comit_listen.port()`). But you are raising a good point. That config variable should be renamed to `network` or something like that since it is no longer exclusive to `bam`.

application/comit_node/src/libp2p_bam/protocol.rs
thomaseizinger: Justification: Copied it from the libp2p example :D

thomaseizinger:
> Can you please organise a walk through of the code?

Can do one next week so that Daniel benefits directly aswell :)

application/comit_node/src/bin/comit_node.rs
D4nte: lol I did not see the second argument... 🤦‍♂

application/comit_node/src/libp2p_bam/protocol.rs
D4nte: > Justification: Copied it from the libp2p example :D

I guessed as much :)

application/comit_node/src/libp2p_bam/behaviour.rs
```diff
@@ -0,0 +1,184 @@
...
+                            "not yet connected to {}, cannot send message",
+                            peer_id.clone()
+                        );
+
+                        self.events.push_back(event);
+
+                        self.current_task = Some(task::current());
+                        return Async::NotReady;
+                    }
+                }
```
LLFourn: This returning `NotReady` here is unsound IMO.
You have events in the queue that may be ready. You should emit them.

Doing what I suggested above would fix this.
Also if you do that, you can change `self.events` into a `mpsc::unbounded` and just poll that here. This would remove all the error prone storing of `task::current`.

application/comit_node/src/libp2p_bam/handler.rs
```diff
@@ -0,0 +1,383 @@
...
+
+fn unknown_mandatory_headers(unknown_headers: UnknownMandatoryHeaders) -> Response {
+    Response::new(Status::SE(1)).with_header(
+        "Unsupported-Headers",
+        Header::with_value(unknown_headers)
+            .expect("list of strings should serialize to serde_json::Value"),
+    )
+}
+
+fn build_outgoing_log_message(frame: &Frame) -> String {
```
LLFourn: ❌ delet this

application/comit_node/src/libp2p_bam/handler.rs
```diff
@@ -0,0 +1,383 @@
...
+
+    fn poll(&mut self) -> Poll<BamHandlerEvent, Self::Error> {
+        self.poll_response_futures();
+
+        match &mut self.network_stream {
+            SubstreamState::Connected(bam_stream) => {
+                futures::try_ready!(bam_stream.poll_complete());
+
+                while let Some(frame) = self.pending_frames.pop_front() {
+                    let message = build_outgoing_log_message(&frame);
```
LLFourn: 🔧 Just log the the frame here, then below when you actually are able to send it say `"sent frame ${id}" to ${peer}"` or whatever.

application/comit_node/src/network.rs
```diff
@@ -107,30 +219,33 @@ pub fn swap_config<B: BobSpawner>(bob_spawner: B) -> Config<ValidatedIncomingReq
...
+fn spawn<AL: rfc003::Ledger, BL: rfc003::Ledger, AA: Asset, BA: Asset, B: BobSpawner>(
     bob_spawner: &B,
     swap_id: SwapId,
     alpha_ledger: AL,
     beta_ledger: BL,
     alpha_asset: AA,
     beta_asset: BA,
     body: rfc003::messages::RequestBody<AL, BL>,
-) -> Box<dyn Future<Item = Response, Error = ()> + Send + 'static>
+) -> Box<dyn Future<Item = Response, Error = Infallible> + Send + 'static>
```
LLFourn: 😎 Infallible!

D4nte:
@thomaseizinger I am surprised this is merged with one approval. I thought we were aiming for 2 approvals per PR. Especially knowing that we were waiting for the walk through to review properly.

PR #885, opened 2019-04-23
Making actions dry 'n wild

### LINKED ISSUE #838: Actions implementation are currently quite wet, make them DRY
We do a lot of repeating ourselves when we `impl Actions` like `impl Actions for alice::State<Bitcoin, Ethereum, BitcoinQuantity, EtherQuantity>`. We do this for every Ledger/Asset pair. We actually only need to specify (`deploy`, `fund`,`redeem`, `reufund`) for each Ledger and Asset combination. Then we could impl them all in one go:

```rust
impl<AL,BL,BA,AA> Actions for alice::State<AL,BL,BA,AA>
  where (AL,AA)::DoesSomeTrait, (BL,BB)::DoesSomeTrait {
    // ... then when you need to generate the action for alpha funding:
   alice::ActionKind::Fund((AL,AA)::fund_action(&request, &response))
}

### PR DESCRIPTION
**Pull request by bonomat**
Resolves #838 : 

Reuse actions between trading pairs and between Alice and Bob.

### CODE REVIEW

File: application/comit_node/src/swap_protocols/rfc003/actions/non_erc20.rs [hunk 1]
```diff
@@ -0,0 +1,131 @@
+use crate::swap_protocols::{
```
D4nte: ❓  Shouldn't the filename be `ether.rs`?

application/comit_node/src/swap_protocols/rfc003/actions/erc20.rs
```diff
@@ -0,0 +1,64 @@
...
+pub fn deploy_action(htlc_params: HtlcParams<Ethereum, Erc20Token>) -> ethereum::ContractDeploy {
+    htlc_params.into()
+}
+
+pub fn fund_action(
+    htlc_params: HtlcParams<Ethereum, Erc20Token>,
+    to_erc20_contract: ethereum_support::Address,
+    beta_htlc_location: ethereum_support::Address,
+) -> ethereum::SendTransaction {
+    let htlc = Erc20Htlc::from(htlc_params.clone());
```
D4nte:
```suggestion
    let network = htlc_params.ledger.network;
    let htlc = Erc20Htlc::from(htlc_params);
    let gas_limit = Erc20Htlc::fund_tx_gas_limit();
```

I wonder if this would work.

PR #887, opened 2019-04-24
New blockchain_contracts crate

### LINKED ISSUE #864: Generate/test Ethereum HTLCs and provide offset in a vendor/crate

D4nte:
For RFC003 Ether and ERC20 HTLCs:
* Tests them against full node
* Printout offsets

Output of `cargo run --package blockchain_contracts --bin print_offsets` is:

```
### RFC003 ###
** Ether on Ethereum **
Contract template:
 6100dc61000f6000396100dc6000f336156051576020361415605c57602060006000376020602160206000600060026048f17f1000000000000000000000000000000000000000000000000000000000000001602151141660625760006000f35b42632000000210609f575b60006000f35b7fb8cac300e37f03ad332e581dea21b2f0b84eaaadc184a295fef71e81f44a741360206000a1733000000000000000000000000000000000000003ff5b7f5d26862916391bf49478b2f5103b0720a842b45ef145a268f2cd1fb2aed5517860006000a1734000000000000000000000000000000000000004ff
| Data | Byte Range | Length (bytes) |
|:--- |:--- |:--- |
| `secret_hash` | 51..83 | 32 |
| `expiry` | 99..103 | 4 |
| `redeem_identity` | 153..173 | 20 |
| `refund_identity` | 214..234 | 20 |
** ERC20 on Ethereum **
Contract template:
 61014461000f6000396101446000f3361561005457602036141561006057602060006000376020602160206000600060026048f17f100000000000000000000000000000000000000000000000000000000000000160215114166100665760006000f35b426320000002106100a9575b60006000f35b7fb8cac300e37f03ad332e581dea21b2f0b84eaaadc184a295fef71e81f44a741360206000a17330000000000000000000000000000000000000036020526100ec565b7f5d26862916391bf49478b2f5103b0720a842b45ef145a268f2cd1fb2aed5517860006000a17340000000000000000000000000000000000000046020526100ec565b63a9059cbb6000527f5000000000000000000000000000000000000000000000000000000000000005604052602060606044601c6000736000000000000000000000000000000000000006620186a05a03f150602051ff
| Data | Byte Range | Length (bytes) |
|:--- |:--- |:--- |
| `secret_hash` | 53..85 | 32 |
| `expiry` | 102..106 | 4 |
| `redeem_identity` | 157..177 | 20 |
| `refund_identity` | 224..244 | 20 |
| `token_quantity` | 261..293 | 32 |
| `token_contract` | 307..327 | 20 |
```

Resolves #864


D4nte:
> I assume you did not make changes to the actual htlc logic, so I did not have a close look at this.
> 
> In general I'm fine with the new structure, but I added a comment regarding the table generation. I think we can get rid of the `END` field.
> I think it's redundant and can actually be a source of errors, e.g. look at your example:
> 
> > Refund timestamp	99		103	4
> 
> If the length is 4, and `START` at 99, shouldn't be the `END` be 102? Meaning, is `END` inclusive or exclusive?
> I know, we have this currently in our RFCs as well, but I think it's about time to get fixed 😬.

As per the initial issue's description, I changed the `end` from _inclusive_ to _exclusive_.
Happy to remove it. @LLFourn what do you think?

vendor/blockchain_contracts/build.rs
```diff
@@ -0,0 +1,81 @@
...
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .spawn()?,
+        Err(_) => {
+            check_bin_in_path("docker");
+            Command::new("docker")
+                .arg("run")
+                .arg("--rm")
+                .arg("-i")
+                .arg("ethereum/solc:0.4.24")
```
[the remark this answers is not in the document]
D4nte: I added a checkbox to https://github.com/comit-network/comit-rs/issues/606 (which is starting to look like an Epic...)

vendor/blockchain_contracts/src/ethereum/rfc003/ether_htlc.rs
```diff
@@ -0,0 +1,167 @@
+use crate::{
```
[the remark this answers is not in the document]
D4nte: I did. Please note that I did not touch the contract code in comit_node so you may want to check out the branch and just do a diff between both `ether_htlc` files.

The change I did is that I created `compile_template_to_hex` that is used to calculate the offset and re-used it in `compile_to_hex` to avoid code duplication.

Sorry it's a pain to review, I am open to suggestions...

vendor/blockchain_contracts/src/ethereum/rfc003/templates/erc20_contract.asm
```diff
@@ -0,0 +1,113 @@
+{
```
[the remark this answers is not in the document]
D4nte: No changes there, feel free to do a local diff with the same name file in comit_node

vendor/blockchain_contracts/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+        write!(
+            f,
+            "{}\t{}\t\t{}\t{}",
+            self.data, self.start, self.end, self.length
+        )
+    }
+}
+
+pub fn format_table(mut offsets: Vec<Offset>) -> String {
+    let mut res = String::from("Data\t\t\t\tStart\tEnd\tLength");
```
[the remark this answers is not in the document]
D4nte: I think it's a fair idea. @LLFourn had some opinion on the matter so I tagged him in your other comment.

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
```diff
@@ -0,0 +1,219 @@
...
+            .replace(
+                Self::CONTRACT_START_POSITION_PLACEHOLDER,
+                &header_length_as_hex,
+            )
+            .replace(Self::CONTRACT_LENGTH_PLACEHOLDER, &code_length_as_hex);
+
+        deploy_header + Self::CONTRACT_CODE_TEMPLATE
+    }
+
+    fn get_offset(name: &str, placeholder: &str) -> Offset {
```
thomaseizinger: :wrench: Usually, getters are not prefixed with `get_` in Rust.

vendor/blockchain_contracts/src/rfc003/secret.rs
```diff
@@ -0,0 +1,171 @@
+use crate::rfc003::secret_hash::SecretHash;
```
thomaseizinger: :x: I think this module should stay inside `comit_node`. The blockchain contracts only need a secret hash and that one can be a `String` or very simple new-type.

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
```diff
@@ -0,0 +1,219 @@
...
+            redeem_address,
+            refund_address,
+            amount,
+            token_contract_address,
+        ]
+    }
+}
+
+impl Htlc for Erc20Htlc {
+    fn compile_to_hex(&self) -> ByteCode {
```
thomaseizinger: :thinking: 

As soon as this code moves to a different repository, this function should only be used by the tests. From there on, the `comit_node` crate will have to construct the HTLCs by making use of the information from the RFC, e.g. a template + offsets.

Should we therefore remove this functionality right away and test the contracts in the way other people (including us in the `comit_node` crate) will have to construct the HTLCs?

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    fn get_offset(name: &str, placeholder: &str) -> Offset {
+        let contract = Self::compile_template_to_hex();
+
+        let re_match = Regex::new(placeholder)
+            .expect("Could not create regex")
+            .find(contract.as_str())
+            .expect("Could not find placeholder in hex code");
+        Offset::new(
+            name.into(),
```
thomaseizinger: :wrench: 

This `name` is a bit dodgy, especially because it contains formatting information (tabs). Maybe you could make use of an enum to express the different placeholders + the information that is associated with them?

vendor/blockchain_contracts/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+            start,
+            end,
+            length,
+        }
+    }
+}
+
+impl PartialEq for Offset {
+    fn eq(&self, other: &Offset) -> bool {
+        self.start == other.start
```
thomaseizinger: :thinking: Why can you not just derive PartialEq if you need it? This implementations looks error prone. It is not used for anything critical but nevertheless, it should be correct.

vendor/blockchain_contracts/src/bin/print_offsets.rs
```diff
@@ -0,0 +1,25 @@
...
+#[allow(clippy::print_stdout)]
+fn main() {
+    println!("### RFC003 ###");
+
+    {
+        println!("** Ether on Ethereum **");
+        let contract = EtherHtlc::compile_template_to_hex();
+        println!("Contract template:\n {}", contract);
+        let offsets = EtherHtlc::get_all_offsets();
+        println!("{}", format_table(offsets));
```
thomaseizinger: :wrench: It is not a super critical part of the software but from a responsibility point of view, the formatting should IMO be done in here and not in the library.

If you want to go for _formatted_ text output, I would prefer MD so that we can just copy it into the RFC (to avoid mistakes).
Alternatively, you could just `Debug` print it.

vendor/blockchain_contracts/src/rfc003/secret.rs
D4nte: The tests of such contracts still need to handle the secret and its hash. Because we wants the contracts to be tested live in this crate then I believe that full `SecretHash` struct needs to be copied across and cannot be replaced with a dummy simple `String` type.

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
D4nte: I see what you mean. This functionality is still needed in any case to calculate the offsets. this new crate has 2 purposes:
1. Calculate offsets/hex based on assembly code
2. Test contracts

The issue you are raising is that 2. is done with contracted generated from assembly, instead of contracts generated via the offsetting.
This means that in call case we need to keep `compile_to_hex` to do 1.

However, we may want to do 2. based on offsetting as it is the way the contract are constructed in comit_node.
If we were to do 2. with offsets in the `blockchain_contracts` crate then we are duplicating functionality (generation of contract by offset) in `comit_node` and `blockchain_contracts`.

I believe that the following is fine:
`blockchain_contracts`:
- generates offset from assembly
- test contracts built from assembly
`comit_node`:
- build contracts from offset

I am happy to consider the following addition:
`comit_node`:
- tests contracts built from offset

Meaning (some) tests "go back" to `comit_node`.

Let me know what you think

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
D4nte: I knew you were going to say that but haven't thought of something better. I'll look into the Enum.

vendor/blockchain_contracts/src/lib.rs
D4nte: I want `Ord` so I need to derive `PartialEq`, `Eq` and `PartialOrd`.
Looks correct to me, please point out the in-correctness.

vendor/blockchain_contracts/src/bin/print_offsets.rs
D4nte: Happy to action this, please answer the other comment regarding your enum idea.

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
D4nte: Ok agreed, https://github.com/comit-network/comit-rs/issues/896 created

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
LLFourn: (see my comment below. I don't think we need an enum).

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
LLFourn: ❌ This trait shouldn't be implemented here (it should not be imported anywhere in this crate).

When the time comes we simply re-implement this in `comit-node` to compile from the offsets.

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub fn tx_gas_limit() -> U256 {
+        U256::from(100_000)
+    }
+
+    pub fn fund_tx_gas_limit() -> U256 {
+        U256::from(100_000)
+    }
+
+    pub fn compile_template_to_hex() -> String {
+        let code_length = Self::CONTRACT_CODE_TEMPLATE.len() / 2; // In hex, each byte is two chars
```
LLFourn: 🔧 Would be nicer if we just had `Vec<u8>` here so we don't have to divide stuff by 2 everywhere.

vendor/blockchain_contracts/src/lib.rs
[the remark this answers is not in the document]
LLFourn: I think we should have start and end and just say start is **included** while end is **excluded**, rather than first byte, last byte so you can translate it directly into a rust (and many languages) range: eg. 42..62.

vendor/blockchain_contracts/src/lib.rs
LLFourn: But he's asking why you don't just `#[derive]` it.

vendor/blockchain_contracts/src/rfc003/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod seconds;
+pub mod secret;
+pub mod secret_hash;
+pub mod timestamp;
```
LLFourn: ❓ shouldn't this stuff be private? Why is it `pub`.

vendor/blockchain_contracts/src/rfc003/secret.rs
LLFourn: Note I think it's fine if a copy pasta of this is here but I note it should not be exported.
Actually this crate shouldn't export anything right? It's just the cli tool.

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
```diff
@@ -0,0 +1,218 @@
...
+            re_match.start() / 2,
+            re_match.end() / 2,
+            placeholder.len() / 2,
+        )
+    }
+
+    pub fn all_offsets() -> Vec<Offset> {
+        let refund_timestamp = Self::offset("Refund timestamp", Self::REFUND_TIMESTAMP_PLACEHOLDER);
+        let redeem_address = Self::offset("Redeem Address\t", Self::REDEEM_ADDRESS_PLACEHOLDER);
+        let refund_address = Self::offset("Refund Address\t", Self::REFUND_ADDRESS_PLACEHOLDER);
```
LLFourn: ❌  Rename this stuff to what it's called in RFC003 e.g. `redeem_address` (and remove tabs).
❌ Output in markdown table format (doesn't matter how nice it looks).

vendor/blockchain_contracts/src/bin/print_offsets.rs
LLFourn: ❌ Yes do this. IMO the goal for this thing is to export nothing as a library and just be a cli to generate hex.

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
D4nte: already done

vendor/blockchain_contracts/src/lib.rs
D4nte: Because "Implementations of PartialEq, PartialOrd, and Ord must agree with each other."
https://doc.rust-lang.org/std/cmp/trait.Ord.html

vendor/blockchain_contracts/src/rfc003/mod.rs
D4nte: can't help it

vendor/blockchain_contracts/src/lib.rs
thomaseizinger: I think if you just make `start` the first field in the struct, you will get the desired behaviour through `#[derive]`.

vendor/blockchain_contracts/src/rfc003/secret.rs
D4nte: @LLFourn hum, do you understand that `crate::` refers to `blockchain_contracts`? It's not imported from `comit_node`. In any case, it's now fixed.

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
D4nte: done

vendor/blockchain_contracts/src/lib.rs
D4nte: Now that I use an enum I need to manually implement some trait on the enum if I were to derive and put `start` as first field.
So I decided to implement the traits directly on the struct that needs to be `Ord`: `Offset`.

vendor/blockchain_contracts/src/bin/print_offsets.rs
D4nte: Please check latest changes and review.

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
D4nte: It does not look like I could do something as easy as I do with regex to search a `Vec<u8>` in a `Vec<u8>` so I prefer to keep `String` and `regex` except if you have a better idea.

LLFourn: There's actually a regex::bytes if you were interested: https://docs.rs/regex/1.1.6/regex/bytes/index.html

vendor/blockchain_contracts/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+    Expiry,
+    RedeemIdentity,
+    RefundIdentity,
+    TokenQuantity,
+    TokenContract,
+}
+
+impl DataName {
+    fn to_markdown(&self) -> String {
+        serde_json::to_string(&self)
```
LLFourn: ❌ It's a bit too weird to use`serde_json` to do this just to get a `snake_case` name for it. Use https://docs.rs/strum/0.15.0/strum/ instead.

vendor/blockchain_contracts/src/rfc003/secret.rs
thomaseizinger: Meh, I had hoped you would get rid of all the unused code. It was created for very different purposes and IMO shouldn't be used here.

thomaseizinger: Meh, I had hoped you would get rid of all the unused code. It was created for very different purposes and IMO shouldn't be used here.

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
thomaseizinger: The idea with the enum would have been to express _all_ the information associated with those placeholders:
- placeholder name
- a function for replacing the placeholder
- the name of the placeholder etc

At the moment, there is a lot of duplication and expanded "iterations" in regards to the placeholders and their use. I voiced the idea with the enum because it seemed you preferred actual iteration over the current expanded one (`Vec<Offset>`). I'd prefer if we are not mixing styles here:

Either use Iterators to go through all the placeholders and perform the logic or always flatten them out. I'd actually prefer the 2nd option because
1. the rest of the code is done like this
2. the list is not expected to grow and explicitly referencing the placeholders makes it more visible, what we are doing here

vendor/blockchain_contracts/src/ethereum/rfc003/erc20_htlc.rs
D4nte: done!

D4nte:
> I am gonna approve this for now so that we can move on with the extraction of the HTLCs. However, the whole crate evolved in a way that in my opinion no longer fits its purpose:
> 
>     * Secret and SecretHash are copied in an provide _way_ more functionality than is actually used now.
> 
>     * What was once a rather simple template mechanism grew into a a rather big struct with over ten string constants
> 
>     * Some code is probably unused aswell (f.e. `deployment_gas_limit()`?)
> 
>     * This is not an exhaustive list.
> 
> 
> In my opinion, this needs a general overhaul where we rethink what the current requirements for this crate are and adapt the code accordingly. This should be done before or as part of moving it to its own repository.

I think it should be done as part of #896

PR #898, opened 2019-04-29
Refactor actions to embed validity information

thomaseizinger:
Instead of a generic invalid_until field, we embed the mediantime timestamp for Bitcoin and the block timestamp for Ethereum directly into the actions.

This allows us to implement customized "wait" strategies depending on the ledger.

- For Bitcoin, we check the mediantime from the getblockchaininfo RPC call and wait until it passes the locktime threshold.

- For Ethereum, it is enough to wait until our local systemtime passes the given timestamp.

### DIFF
--- a/api_tests/e2e/test_creator.ts
+++ b/api_tests/e2e/test_creator.ts
@@ -11,7 +11,7 @@ import * as chai from "chai";
 import { Actor } from "../lib/actor";
 import * as URI from "urijs";
 
-const should = chai.should();
+chai.should();
 
 interface Test {
     /**
--- a/api_tests/e2e/test_creator.ts
+++ b/api_tests/e2e/test_creator.ts
@@ -1,7 +1,7 @@
 import {
     AcceptRequestBody,
-    ActionKind,
     Action,
+    ActionKind,
     getMethod,
     HalResource,
     Method,
@@ -9,7 +9,6 @@ import {
 } from "../lib/comit";
 import * as chai from "chai";
 import { Actor } from "../lib/actor";
-import { sleep } from "../lib/util";
 import * as URI from "urijs";
 
 const should = chai.should();
@@ -44,7 +43,6 @@ interface ActionTrigger {
     action?: ActionKind;
     requestBody?: AcceptRequestBody;
     uriQuery?: object;
-    timeout?: number;
     state?: (state: any) => boolean;
     test?: Test;
 }
@@ -78,16 +76,6 @@ async function getAction(
         return [href, payload];
     }
     return [href, null];
-}
-
-function seconds_until(time: number): number {
-    const diff = time - Math.floor(Date.now() / 1000);
-
-    if (diff > 0) {
-        return diff;
-    } else {
-        return 0;
-    }
 }
 
 async function executeAction(
@@ -162,66 +150,46 @@ export async function createTests(
         let action = actions.shift();
         let actionHref: string = null;
         let actionDirective: Action = null;
-        const timeout = action.timeout || 10000;
         if (action.action) {
-            it(
-                "[" +
-                    action.actor.name +
-                    "] Can get the " +
-                    action.action +
-                    " action",
-                async function() {
-                    this.timeout(timeout);
-                    [actionHref, actionDirective] = await getAction(
-                        swapLocations[action.actor.name],
-                        action
-                    );
-                }
-            );
+            it(`[${action.actor.name}] Can get the ${
+                action.action
+            } action`, async function() {
+                this.timeout(5000);
+                [actionHref, actionDirective] = await getAction(
+                    swapLocations[action.actor.name],
+                    action
+                );
+            });
 
-            it(
-                "[" +
-                    action.actor.name +
-                    "] Can execute the " +
-                    action.action +
-                    " action",
-                async function() {
-                    if (actionDirective && actionDirective.invalid_until) {
-                        const to_wait =
-                            seconds_until(actionDirective.invalid_until) *
-                                1000 +
-                            1000; // Add an extra second for good measure
-                        console.log(
-                            `Waiting ${to_wait}ms for ${
-                                action.actor.name
-                            }'s  ‘${action.action}’ action to be ready`
-                        );
-                        this.timeout(to_wait + timeout);
-                        await sleep(to_wait);
-                    }
-
-                    await executeAction(
-                        action.actor,
-                        action,
-                        actionHref,
-                        actionDirective
-                    );
+            it(`[${action.actor.name}] Can execute the ${
+                action.action
+            } action`, async function() {
+                if (action.action == ActionKind.Refund) {
+                    this.timeout(30000);
+                } else {
+                    this.timeout(5000);
                 }
-            );
+
+                await executeAction(
+                    action.actor,
+                    action,
+                    actionHref,
+                    actionDirective
+                );
+            });
         }
 
         let body: any = null;
         if (action.state) {
-            it(
-                "[" + action.actor.name + "] transitions to correct state",
-                async function() {
-                    this.timeout(timeout);
-                    body = (await action.actor.pollComitNodeUntil(
-                        swapLocations[action.actor.name],
-                        body => action.state(body.state)
-                    )) as HalResource;
-                }
-            );
+            it(`[${
+                action.actor.name
+            }] transitions to correct state`, async function() {
+                this.timeout(10000);
+                body = (await action.actor.pollComitNodeUntil(
+                    swapLocations[action.actor.name],
+                    body => action.state(body.state)
+                )) as HalResource;
+            });
         }
 
         const test = action.test;
--- a/api_tests/lib/actor.ts
+++ b/api_tests/lib/actor.ts
@@ -114,7 +114,10 @@ export class Actor {
                 return this.wallet.btc().sendToAddress(to, parseInt(amount));
             }
             case "bitcoin-broadcast-signed-transaction": {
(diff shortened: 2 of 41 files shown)


api_tests/lib/actor.ts
```diff
@@ -106,17 +106,44 @@ export class Actor {
...
+
+                let { hex, locktime } = action.payload;
+
+                // If it is equal, it is rejected by bitcoin-core :(
+                locktime += 1;
+
+                let medianTime = await fetchMedianTime();
+                let diff = locktime - medianTime;
+
+                if (locktime && diff > 0) {
```
D4nte: you just did `locktime += 1;` so I think `locktime` will always be "true".

api_tests/lib/actor.ts
```diff
@@ -106,17 +106,44 @@ export class Actor {
...
+                    console.log(`Waiting for median time to pass %d`, locktime);
+
+                    while (diff > 0) {
+                        await sleep(1000);
+
+                        medianTime = await fetchMedianTime();
+                        diff = locktime - medianTime;
+
+                        console.log(
```
D4nte: ❌  to be removed

api_tests/lib/actor.ts
```diff
@@ -131,16 +158,32 @@ export class Actor {
...
+                    min_block_timestamp,
                 } = action.payload;
 
+                if (seconds_until(min_block_timestamp) > 0) {
+                    // Ethereum needs a buffer, otherwise the contract code is run but doesn't transfer any funds,
+                    // see https://github.com/comit-network/RFCs/issues/62
+                    let buffer = 2;
+                    let delay = seconds_until(min_block_timestamp) + buffer;
+
+                    console.log(
```
D4nte: ❌ remove

api_tests/lib/actor.ts
```diff
@@ -106,17 +106,44 @@ export class Actor {
...
+
+                let fetchMedianTime = async () => {
+                    let blockchainInfo = await bitcoin.getBlockchainInfo();
+                    return blockchainInfo.mediantime;
+                };
+
+                let { hex, locktime } = action.payload;
+
+                // If it is equal, it is rejected by bitcoin-core :(
+                locktime += 1;
```
LLFourn: ❓ Is bitcoin-core wrong to reject it? I can't remember wether the locktime had to be strictly greater than or could be equal in consensus. If bitcoin-core is right then the action we create should already have +1.

api_tests/lib/actor.ts
thomaseizinger: Why? Without these logs, your test suite hangs for an unknown amount of time and you don't know what is going on.

api_tests/lib/util.ts
```diff
@@ -15,6 +15,16 @@ export async function sleep(time: number) {
     });
 }
 
+export function seconds_until(time: number): number {
+    const diff = time - Math.floor(Date.now() / 1000);
+
+    if (diff > 0) {
```
bonomat: 🙃 *not tested but should work.
```suggestion
    return diff > 0? diff : 0;
```

api_tests/e2e/test_creator.ts
```diff
@@ -162,66 +150,46 @@ export async function createTests(
         let action = actions.shift();
         let actionHref: string = null;
         let actionDirective: Action = null;
-        const timeout = action.timeout || 10000;
```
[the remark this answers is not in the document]
thomaseizinger: Because I need to set a much bigger timeout for refund actions and it felt inappropriate to set a global one first if we override afterwards anyway.

api_tests/lib/util.ts
thomaseizinger: I just moved the code :upside_down_face:

api_tests/lib/actor.ts
thomaseizinger: What is the difference other than taste?

bonomat: > What is the difference other than taste?
Better coding style? I think the log vs. loop behavior makes more sense that way.

Currently:
* `sleep`
* `fetchMedianTime();`
* log `Current median time:          XX+~sleep time` 
* continue with/after the loop

While I would propose to: 

* log `Current median time:          XX`  --> you still have this value from before
* `sleep`
* `fetchMedianTime();`
* continue with/after the loop

i.e..current behavior will be

* _current_median_time=8_
* `Waiting for median time to pass 10.`
* *sleeping randomly unknown time...*
* `Current median time:                   9`
* *sleeping randomly unknown time...*
* .... continue with after loop

vs. 

* _current_median_time=8_
* `Waiting for median time to pass 10.`
* `Current median time:                   8`
* *sleeping randomly unknown time...*
* `Current median time:                   9`
* *sleeping randomly unknown time...*
* .... continue with after loop

api_tests/lib/actor.ts
```diff
@@ -106,17 +106,46 @@ export class Actor {
...
+        switch (action.type) {
             case "bitcoin-send-amount-to-address": {
                 action.payload.should.include.all.keys("to", "amount");
                 let { to, amount } = action.payload;
 
                 return this.wallet.btc().sendToAddress(to, parseInt(amount));
             }
             case "bitcoin-broadcast-signed-transaction": {
-                let { hex } = action.payload;
+                action.payload.should.include.all.keys("hex", "locktime");
```
D4nte: isn't it `min_median_time` ?

D4nte: so... is this test useless if it was passing despite being wrong?

api_tests/lib/actor.ts
```diff
@@ -106,17 +106,46 @@ export class Actor {
...
+                let fetchMedianTime = async () => {
+                    let blockchainInfo = await bitcoin.getBlockchainInfo();
+                    return blockchainInfo.mediantime;
+                };
+
+                let { hex, min_median_time } = action.payload;
+
+                if (min_median_time) {
+                    let medianTime = await fetchMedianTime();
+                    let diff = min_median_time - medianTime;
```
D4nte: 🔧 maybe worth doing the diff calculation directly in a function that also fetches the median time.

api_tests/lib/actor.ts
D4nte: Don't we use log4js?

application/comit_node/src/http_api/routes/rfc003/handlers/get_action.rs
```diff
@@ -101,15 +96,27 @@ pub enum ActionResponsePayload {
...
-    pub payload: ActionResponsePayload,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub invalid_until: Option<Timestamp>,
+impl ActionResponseBody {
+    fn bitcoin_broadcast_signed_transaction(transaction: &Transaction, network: Network) -> Self {
+        let min_median_time = if transaction.lock_time == 0 {
+            None
+        } else {
+            // The first time a tx with lock_time can be broadcasted is when
+            // mediantime == locktime + 1
```
D4nte:
```suggestion
            // mediantime > locktime
```
This is a weird way to phrase it.

api_tests/lib/actor.ts
thomaseizinger: Fair point, looks pretty useless then :D

application/comit_node/src/http_api/routes/rfc003/handlers/get_action.rs
thomaseizinger: What is so weird about it?

The comment (both lines!) explains why we are doing `+ 1` to the `lock_time` and call it `min_median_time` afterwards.
I don't see how using different operations in the comments vs the actual code makes things easier to understand?

api_tests/lib/actor.ts
thomaseizinger: Nope.

api_tests/lib/actor.ts
thomaseizinger: Then I have to return two values, which is kind of awkward in JavaScript. I am aware of the duplication, I just preferred being explicit over making it too complicated.

api_tests/lib/actor.ts
thomaseizinger: It actually breaks :D

application/comit_node/src/http_api/routes/rfc003/handlers/get_action.rs
D4nte: It is weird because the consensus is that nLockTime must be strictly less than the median time for the transaction to be valid, not that it needs to be `+1`.
You translated that to `+1` in the code which mathematically correct but it is not what the consensus says, it is a consequence of the rule, not the rule.

D4nte: The consensus for info:
```
    if ((int64_t)tx.nLockTime < ((int64_t)tx.nLockTime < LOCKTIME_THRESHOLD ? (int64_t)nBlockHeight : nBlockTime))
        return true;
```

api_tests/lib/actor.ts
D4nte: why 2 values? just return the diff.

thomaseizinger: Need the median time for logging, unless I do it in the function.

D4nte: 🙃 I was against the logging

D4nte:
A feels good PR! 👼 🧐

D4nte:
@thomaseizinger I approve the last 3 commits

PR #900, opened 2019-04-30
Remove waffle.io badge

bonomat:
Bye bye waffle badge.
ZenHub's badge does not work currently, so no replacement for now. 

see: 
https://www.zenhub.com/faq
i.e. the following image cannot be found: 

<a href="https://zenhub.com"></a>


D4nte:
> ZenHub's badge does not work currently, so no replacement for now.

I dropped a message to support

PR #909, opened 2019-05-01
Strum

D4nte:
I had some weird compilation error with clippy (but not when just building) with latest strum.
Upgrading the nightly helped with that so here we are.


D4nte:
Blocked on Peternator7/strum#49

D4nte:
To do once Peternator7/strum#49 is fixed

PR #937, opened 2019-05-06
Renames SendTransaction to CallContract

### LINKED ISSUE #835: Ethereum `SendTransaction` action should be `CallContract`
Our `SendTransaction` action should actually be called `CallContract`.

https://github.com/comit-network/comit-rs/blob/f85842f5ccbfd6348d04275490d838655c416b3a/application/comit_node/src/swap_protocols/rfc003/ethereum/actions.rs#L18-L25

https://github.com/comit-network/comit-rs/blob/f85842f5ccbfd6348d04275490d838655c416b3a/application/comit_node/src/http_api/routes/rfc003/handlers/get_action.rs#L234-L261

As part of that, the `amount` field should be removed since it is always set to zero and then unnecessarily returned to the user. The user should never send any Ether to an HTLC when invoking them. We should prevent that by not even suggesting that there might be an amount.

DoD:
- [ ] rename `SendTransaction` to `CallContract`
- [ ] remove `amount` field from `CallContract`
- [ ] update http API action type from `ethereum-invoke-contract` to `ethereum-call-contract`

da-kami:
Resolves #835

### DIFF
--- a/api_tests/lib/actor.ts
+++ b/api_tests/lib/actor.ts
@@ -164,11 +164,10 @@ export class Actor {
 
                 return this.wallet.eth().deploy_contract(data, amount);
             }
-            case "ethereum-invoke-contract": {
+            case "ethereum-call-contract": {
                 action.payload.should.include.all.keys(
                     "contract_address",
                     "data",
-                    "amount",
                     "gas_limit",
                     "min_block_timestamp"
                 );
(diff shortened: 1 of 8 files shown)


da-kami:
> Looks good! Well done :)
> 
> Can you open an issue on the comit-i repository (https://github.com/comit-network/comit-i) to remove handling of the `amount` field?

see: https://github.com/comit-network/comit-i/issues/43

PR #938, opened 2019-05-06
Generate ERC20/Ether HTLCs from offsets

### LINKED ISSUE #896: Test Ethereum contracts built on RFC offset
In `blockchain_contracts` crate, the contract hex are generated from assembly code and then tested against full nodes.

Instead, such contracts should be generated using the offset described in the RFC.

See https://github.com/comit-network/comit-rs/pull/887#discussion_r279236004

One crate, soon to be in own repo, (blockchain_contracts) will contain:
- bin to generate offsets (for rfc) dynamically based on placeholders
- lib to generate htlcs from offsets (to be used by comit_node). Offsets are static in the code.
- tests that tests the lib

DoD:
- [ ] Do it for Ether
- [ ] Do it for ERC20

D4nte:
Resolves #896 

- Generate ERC20 and Ether HTLC from offset.
- Test new HTLC against parity
- restructure blockchain_contracts crate to separate lib (htlc generation) to bin (generate offsets)


vendor/blockchain_contracts/src/rfc003/timestamp.rs
```diff
@@ -20,6 +27,29 @@ impl Timestamp {
     pub fn plus(self, seconds: u32) -> Self {
         Self(self.0.checked_add(seconds).unwrap_or(std::u32::MAX))
     }
+
+    pub fn to_vec(self, len: usize) -> Result<Vec<u8>, ToVecError> {
+        let mut vec = Vec::with_capacity(len);
+        vec.write_u32::<BigEndian>(self.0).map_err(ToVecError::Io)?;
```
LLFourn: As mentioned above this should be implemented as a trait `EncodeEthereum` and this whole function can just be:

```rust
pub fn to_vec(&self) -> Vec<u8> {
  let mut wtr = vec![];
  wtr.write_u32::<BigEndian>(self.0).expect("writing to Vec can't fail");
  return wtr;
}
```

vendor/blockchain_contracts/build.rs
```diff
@@ -10,9 +10,9 @@ use std::{
 };
 
 fn main() -> std::io::Result<()> {
-    compile("./src/ethereum/rfc003/templates/ether_contract.asm")?;
-    compile("./src/ethereum/rfc003/templates/deploy_header.asm")?;
-    compile("./src/ethereum/rfc003/templates/erc20_contract.asm")?;
+    compile("./src/bin/calculate_offsets/ethereum/rfc003/templates/ether_contract.asm")?;
```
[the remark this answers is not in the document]
D4nte: Because this code is only to be used by the `print_offset` bin.
The aim of this crate is also to be imported as a library.

By putting all the `print_offset` code in the `/bin/` folder we ensure that the library consumer does not use this `print_offset` code.

vendor/blockchain_contracts/src/ethereum/mod.rs
```diff
@@ -1,13 +1,58 @@
...
+}
 
-impl Into<Bytes> for ByteCode {
-    fn into(self) -> Bytes {
-        Bytes(hex::decode(self.0).unwrap())
+impl EncodeToEvm for Timestamp {
+    fn encode_to_evm(&self) -> Result<Vec<u8>, ToEvmError> {
+        let mut vec = vec![];
+        vec.write_u32::<BigEndian>(self.clone().into())
+            .map_err(ToEvmError::Io)?;
```
LLFourn: If you  don't want to do an `expect` here, you can get rid of the error by doing:

```
let mut buf = [0; 4];
BigEndian::write_u32(&mut buf, 1_000_000);
buf.to_vec()
```
This looks better anyway.

:x: This stuff shouldn't return a `Result`.

vendor/blockchain_contracts/src/ethereum/rfc003/ether.rs
```diff
@@ -0,0 +1,92 @@
...
+        secret_hash: SecretHash,
+    ) -> Result<EtherHtlc, Error> {
+        let offsets = vec![
+            OffsetParameter::new(expiry, EXPIRY_RANGE)?,
+            OffsetParameter::new(refund_identity, REFUND_IDENTITY_RANGE)?,
+            OffsetParameter::new(redeem_identity, REDEEM_IDENTITY_RANGE)?,
+            OffsetParameter::new(secret_hash, SECRET_HASH_RANGE)?,
+        ];
+
+        let data = apply_offsets(CONTRACT_TEMPLATE, offsets)?;
```
LLFourn: :sunglasses:

vendor/blockchain_contracts/src/ethereum/rfc003/erc20.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub fn new(
+        expiry: Timestamp,
+        refund_identity: Address,
+        redeem_identity: Address,
+        secret_hash: SecretHash,
+        token_contract_address: Address,
+        token_quantity: U256,
+    ) -> Result<Erc20Htlc, Error> {
+        let offsets = vec![
+            OffsetParameter::new(expiry, EXPIRY_RANGE)?,
```
thomaseizinger: :upside_down_face: It might actually be more readable (when you compare it to the code) if you _don't_ extract the range into a variable.

D4nte: I think it's more readable if values coming from the RFC are in constant at the beginning of the file.

thomaseizinger:
> Approved. Nice one. I do agree with that comment from @thomaseizinger though. I forgot to mention the same thing.

Mergify was too fast :D

PR #942, opened 2019-05-08
Initial circle ci setup

bonomat:
Simple circle ci setup.
Last build duration: 17:48 minutes

Currently we are only able to go for machine executors.
The reason is, we would need docker in docker. This is generally supported by circle ci, however, they only provide a remote docker host, which is currently not supported by testcontainer-rs. 
If we fix this in testcontainer-rs, we can probably optimize the build time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,17 +24,28 @@ jobs:
           name: Check formatting
           command: cargo make check-rs-format
       - restore_cache:
-          key: dependency-cache-{{ checksum "Cargo.lock" }}
+          key: dependency-rs-sccache
+      - restore_cache:
+          key: dependency-rs-cache-{{ checksum "Cargo.lock" }}
+      - restore_cache:
+          key: dependency-js-cache-{{ checksum "./api_tests/package.json" }}
       - run:
           name: Build using cargo make
           command: |
             cargo --version
             cargo make travis
       - save_cache:
-          key: dependency-cache-{{ checksum "Cargo.lock" }}
+          key: dependency-rs-sccache
+          paths:
+            - "~/.cache/sccache"
+      - save_cache:
+          key: dependency-rs-cache-{{ checksum "Cargo.lock" }}
           paths:
             - "~/.cargo"
             - "./target"
+      - save_cache:
+          key: dependency-js-cache-{{ checksum "./api_tests/package.json" }}
+          paths:
             - "./api_tests/node_modules"
 
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,6 @@ jobs:
       - run:
           name: Build using cargo make
           command: |
-            source $HOME/.cargo/env
             cargo --version
             cargo make travis
       - save_cache:
@@ -78,7 +77,6 @@ commands:
       - run:
           name: "Prints version for rust, etc"
           command: |
-             source $HOME/.cargo/env
              echo 'node --version' $(node --version)
              echo 'yarn --version' $(yarn --version)
              echo 'rustc --version' $(rustc --version)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,76 @@
+version: 2.1
+jobs:
+  build:
+    working_directory: ~/comit
+    machine:
+      image: ubuntu-1604:201903-01
+      docker_layer_caching: true
+    steps:
+      - checkout
+      - restore_cache:
+          key: dependency-cache-{{ checksum "Cargo.lock" }}
+      - setup_system
+      - setup_nodejs
+      - setup_rust
+      - print_current_versions
+      - run:
+          name: Build using cargo make
+          command: |
+            source $HOME/.cargo/env
+            cargo --version
+            cargo make ci
+      - save_cache:
+          key: dependency-cache-{{ checksum "Cargo.lock" }}
+          paths:
+            - "~/.cargo"
+            - "./target"
+
+
+commands:
+  setup_system:
+    steps:
+       - run:
+           name:  "Setup system environment"
+           command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,76 @@
...
+      - setup_system
+      - setup_nodejs
+      - setup_rust
+      - print_current_versions
+      - run:
+          name: Build using cargo make
+          command: |
+            source $HOME/.cargo/env
+            cargo --version
+            cargo make ci
```
thomaseizinger: :thinking: I think we are using `cargo make travis` in the travis file? That one runs `ci-flow` where you are only running `ci`. I think there might be some `pre-*` steps missing then.

.circleci/config.yml
```diff
@@ -0,0 +1,76 @@
...
+      - restore_cache:
+          key: dependency-cache-{{ checksum "Cargo.lock" }}
+      - setup_system
+      - setup_nodejs
+      - setup_rust
+      - print_current_versions
+      - run:
+          name: Build using cargo make
+          command: |
+            source $HOME/.cargo/env
```
thomaseizinger: :wrench: Can this be not part of `setup_rust`?

thomaseizinger: It actually is, should be redundant here then?

.circleci/config.yml
```diff
@@ -0,0 +1,76 @@
...
+            cargo --version
+            cargo make ci
+      - save_cache:
+          key: dependency-cache-{{ checksum "Cargo.lock" }}
+          paths:
+            - "~/.cargo"
+            - "./target"
```
thomaseizinger: Why not cache the `sccache` folder like in travis?
What about `node_modules`?

.circleci/config.yml
```diff
@@ -0,0 +1,76 @@
...
+             echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
+             sudo apt-get update
+             sudo apt-get install -y libzmq3-dev
+  setup_nodejs:
+    steps:
+       - run:
+           name:  "Install nodejs"
+           command: |
+             set +e
+             curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.5/install.sh | bash
```
thomaseizinger: :x: Why can't we just the same thing as in `travis` and install the correct node version from `apt`?

.circleci/config.yml
thomaseizinger: :x: I saw you deleted the `sccache` setup? This will massively slow down our build times on travis. As long as that one is still our primary build pipeline (PR check, slack notifications) that one should continue as it is IMO.

Why don't you use sccache on circle-ci?

D4nte: ❌ please use `cargo make travis` (happy to rename it) so it sets the threads and setup for sscache:
`env = { "RUST_TEST_THREADS" = "8", "RUSTC_WRAPPER" = "${HOME}/.cargo/bin/sccache"  }`

.circleci/config.yml
```diff
@@ -0,0 +1,76 @@
+version: 2.1
+jobs:
+  build:
+    working_directory: ~/comit
```
D4nte: ❓ Where does this come from?

.circleci/config.yml
```diff
@@ -0,0 +1,76 @@
...
+    machine:
+      image: ubuntu-1604:201903-01
```
D4nte: ❓ Where is the osx setup?

.circleci/config.yml
D4nte: ❓ why caching `./target`? We did not use to do that for Travis.

bonomat:
> I believe sccache should stay

why? There is no need for that in circleCI

.circleci/config.yml
bonomat: Unfortunately I was not able to set this globally, i.e. for each new `command` cargo was not in the path anymore :S

.circleci/config.yml
bonomat: I was unfortunately not able to do this and did not invest more time. I always ended up having node 6.x in the path and a quick google recommended using nvm on CircleCI

.circleci/config.yml
bonomat: What are you talking about?

.circleci/config.yml
bonomat: What do you mean? This is just to set the path where to build things

bonomat:
@D4nte 
> ❓ why caching ./target? We did not use to do that for Travis.

we might not need to build everything all the time, hence I cache `./target`

.circleci/config.yml
bonomat: @thomaseizinger : 
yes, should cache `node_modules` as well
sccache: I don't think we need it, at least with the current setup.

@D4nte :
we might not need to build everything all the time, hence I cache ./target

.circleci/config.yml
D4nte: We are migrating to Circle CI to be able to build and test on linux **and** mac os x. Hence my question :)

bonomat: I would not set MacOS a hard requirement but a nice to have. 
Worth a shot if Docker is (or can be easily be) installed.

D4nte: I think we need to be clear on the gain of this effort if it;s not for mac os

bonomat: Yes, the main goal would be performance improvements.

bonomat:
Once merged, we can track our build here: 
https://circleci.com/gh/comit-network/comit-rs/

.circleci/config.yml
bonomat: Readded sccache

.circleci/config.yml
bonomat: Keep this in mind for future improvements

PR #952, opened 2019-05-10
Migrate comit-node HTTP API to siren!

### LINKED ISSUE #904: Move /swaps and /swaps/rfc003/:id from hal to siren
https://github.com/kevinswiber/siren

DoD:
- [ ] Move `/swaps` from hal to siren and minimally change the tests
- [ ] Move `/swaps/rfc003/:id` from hal to siren and minimally change the tests
- [ ] Write a test that automatically verify the API against siren schema

Note:
Do not embed the actions in the new API just yet

thomaseizinger:
Fixes #904.

I recommend reviewing the code commit-wise, especially the last two since one of them is a refactoring and the 2nd one actually adapts the tests to the new API.


.idea/compiler.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
thomaseizinger: Gonna remove this after lunch. Don't bother!

vendor/blockchain_contracts/src/bin/calculate_offsets/ethereum/rfc003/templates/out/deploy_header.asm.hex
```diff
@@ -0,0 +1 @@
+6120026110016000396120026000f3
```
LLFourn: :question: Why is this being moved here?

application/comit_node/src/http_api/swap_resource.rs
```diff
@@ -64,12 +64,12 @@ pub enum IncludeState {
     No,
 }
 
-pub fn new_rfc003_hal_swap_resource<S: StateStore>(
+pub fn new_rfc003_siren_entity<S: StateStore>(
```
LLFourn: :wrench: I think I missed the PR that restructured this stuff but a better name for this would be `retrieve_rfc003_siren_entity`.
:question: Can't we just make From implementation for (alice/bob)::State -> siren::Entity

vendor/blockchain_contracts/src/bin/calculate_offsets/ethereum/rfc003/templates/out/deploy_header.asm.hex
thomaseizinger: Switching too much between branches ...

PR #965, opened 2019-05-16
Configure logging for btsieve in settings file and print startup messages

### LINKED ISSUE #358: [BTSieve] Configure logging levels in settings file
DoD:
- [ ] Logging levels are set in a config file with possible override via en var

### LINKED ISSUE #950: Add startup message to btsieve
There is no message that would show that the `btsieve` service started correctly.

Add a message upon startup: "btsieve successfully started. listening on localhost:XXXX".

bonomat:
I started doing #950: 
However, I was wondering why no log message was shown first and it was because of the wrongly set log_level. 
So I did #358 at the same time and allow to configure the log level in the settings file. 
For now, only the default btsieve log level and warp is set.

Note: 
Instead of printing:
```"btsieve successfully started. listening on localhost:XXXX"```
btsieve will just print the default message when warp is loaded: 
```[2019-05-16 16:22:20.358][INFO][...] warp drive engaged: listening on http://0.0.0.0:8080```

In my opinion, this PR: 
Resolves #950 
and 
Resolves #358

### CODE REVIEW

**Comment by bonomat** [comment]
NOTE: I'm rebasing first as I was working on an old branch

File: application/btsieve/src/settings/mod.rs [hunk 1]
```diff
@@ -87,4 +95,15 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn can_deserialize_log_level() -> Result<(), failure::Error> {
+        let settings = Settings::create("./config/default.toml");
```
D4nte: 📣 Please update this config file with the new configuration.

Also, it'd be good to update the regtest config file accordingly.

bonomat: > 📣 Please update this config file with the new configuration.

but it is already updated? Or what do you mean?

but yes, can add it to the regtest configs as well.

D4nte: > > 📣 Please update this config file with the new configuration.
> 
> but it is already updated? Or what do you mean?
> 
> but yes, can add it to the regtest configs as well.

My apologies, looks like I did not see it!

PR #974, opened 2019-05-20
Comit_node settings injections in Comit-i endpoint

### LINKED ISSUE #945: comit-i API port injection when served by the comit node
**Problem:** 
When comit-i is served by the comit_node, the port for the comit_node API within comit-i does not reflect the configuration of the comit_node but is always set to the default port `8000`.

**Solution (comit-rs part)**
Fetch a script at a new configuration endpoint that contains a function that returns the API port as set in the comit node configuration. 

1. add endpoint that serves the script (Content-Type `application/javascript`)
2. template for script (code for the function)
3. inject the API port from the configuration into the function code upon startup of the comit node.

prerequisite for: https://github.com/comit-network/comit-i/issues/52

D4nte:
New endpoint on the same warp that serves comit-i `/config/comitNode.js?callback=callbackFunctionName`

the `callback` parameter allows you to specify the function name.
This will serve a javascript function definition. The function returns a JSON objected when invoked.

The returned JSON object is as follow:
```json
{
  "host": "127.0.0.1",
  "port": 8000
}
```

These are the details to connect to the comit_node http api.

Note that `host` is included because comit_node could be configured to only listen on a specific ip which is not the local loopback.

Note: `chai-http` does not give you a text/body for `application/json` hence I had to go with an alternative library. See https://github.com/chaijs/chai-http/issues/249

Resolves #945


api_tests/webgui/webgui.ts
```diff
@@ -19,61 +19,110 @@ const david = new Actor("david", global.config, global.project_root);
...
     describe("Web GUI tests", () => {
         it("Returns 403 'Forbidden for invalid origins or headers' for invalid preflight OPTIONS /swaps request for GET", async () => {
-            let res = await request(david.comit_node_url())
-                .options("/swaps")
-                .set("Origin", "http://localhost:4000")
-                .set("Access-Control-Request-Headers", "content-type")
-                .set("Access-Control-Request-Method", "GET");
-
-            res.should.have.status(403);
+            await expect(
```
D4nte: Revert this commit and only use "request-promise-native" for the last test.

application/comit_node/src/comit_i_routes.rs
```diff
@@ -1,32 +1,101 @@
...
         .and(warp::path::tail())
-        .and_then(|path: Tail| serve(path.as_str()))
-        .boxed()
+        .and_then(|path: Tail| serve_comit_i_file(path.as_str()));
+
+    comit_node_config.or(comit_i).boxed()
 }
 
-fn serve(path: &str) -> Result<impl Reply, Rejection> {
+fn serve_comit_i_file(path: &str) -> Result<impl Reply, Rejection> {
```
bonomat: 🙃Why the rename?

D4nte: Because now it can serve comit-i files or the comit node conn details config file.

PR #976, opened 2019-05-21
Config loading is optional

bonomat:
If the parameter `--config` is not passed in, a config file with default values will be created.


application/comit_node/src/settings/mod.rs
```diff
@@ -5,76 +5,165 @@ use crate::seed::Seed;
...
+
+        ComitNodeSettings {
+            comit: Comit {
+                secret_seed: Seed::from(data),
+            },
+            network: Network {
+                listen: vec![listen],
+            },
+            http_api: HttpSocket {
+                address: IpAddr::V4(Ipv4Addr::LOCALHOST),
```
D4nte: 📣 
```suggestion
                address: IpAddr::V4(Ipv4Addr::UNSPECIFIED),
```

application/comit_node/src/settings/mod.rs
```diff
@@ -5,76 +5,165 @@ use crate::seed::Seed;
...
+                    poll_interval_secs: Duration::from_secs(300),
+                    network: "regtest".into(),
+                },
+                ethereum: PollParameters {
+                    poll_interval_secs: Duration::from_secs(20),
+                    network: "regtest".into(),
+                },
+            },
+            web_gui: Some(HttpSocket {
+                address: IpAddr::V4(Ipv4Addr::LOCALHOST),
```
D4nte: 📣 
```suggestion
                address: IpAddr::V4(Ipv4Addr::UNSPECIFIED),
```

application/comit_node/src/settings/mod.rs
```diff
@@ -5,76 +5,165 @@ use crate::seed::Seed;
...
+                ConfigError::Message(format!("Could not create folders: {:?}: {:?}", path, error))
+            })?;
+        }
+
+        let default_config = path.join(file_name);
+
+        if default_config.exists() {
+            if default_config.is_file() {
+                log::warn!(
+                    "Config files exists, loading config from: {:?}",
```
D4nte: 📣  It means that the `create_with_default` was called but no creation is happening because a file already exist. I think we should error out here.

application/comit_node/src/settings/mod.rs
```diff
@@ -110,4 +199,26 @@ mod tests {
...
+            if config_path.clone().join(config_file).exists() {
+                let default_config_file = config_path.clone().join(config_file);
+                fs::remove_file(default_config_file).unwrap();
+            }
+            fs::remove_dir(config_path.clone()).unwrap();
+        }
+
+        let default_settings =
+            ComitNodeSettings::create_with_default(config_path.clone(), config_file.clone());
+        let settings = ComitNodeSettings::read(config_path.join(config_file));
```
D4nte: 📣 I think we should delete after to not endup with a million files in `/tmp` on our machines.

application/comit_node/config/default.toml
```diff
@@ -1,6 +1,9 @@
 [comit]
 secret_seed = "c1fd6fc5bde7fee2c2fb6d868dc0f40368051fede1d83f814839d562c210aa27"
 
+[log_levels]
+comit_node="DEBUG"
```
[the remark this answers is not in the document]
bonomat: Yes, for now it is. The reason why I introduced the section `[log_levels]` was because of serde, i.e. it needs a section to being able to serialize.

application/comit_node/src/settings/mod.rs
```diff
@@ -5,76 +5,165 @@ use crate::seed::Seed;
...
-    #[serde(with = "self::serde_log", default = "default_log")]
-    pub log_level: LevelFilter,
     pub web_gui: Option<HttpSocket>,
+    #[serde(default = "default_log_levels")]
+    pub log_levels: LogLevels,
+}
+
+impl Default for ComitNodeSettings {
+    fn default() -> Self {
+        let listen = "/ip4/0.0.0.0/tcp/8011".parse().unwrap();
```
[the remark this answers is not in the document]
bonomat: I took it from the default.toml file.

application/comit_node/src/settings/mod.rs
```diff
@@ -5,76 +5,165 @@ use crate::seed::Seed;
...
+        let url = Url::parse("http://localhost:8001").unwrap();
```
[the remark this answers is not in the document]
bonomat: > 🤡 let's default to a more original port for btsieve :)

Which port do you have in mind? I took the stuff from `./config/default.toml`.

application/comit_node/src/settings/mod.rs
```diff
@@ -5,76 +5,165 @@ use crate::seed::Seed;
...
     pub poll_interval_secs: Duration,
     pub network: String,
 }
 
 impl ComitNodeSettings {
-    pub fn create<D: AsRef<OsStr>, R: AsRef<OsStr>>(
-        default_config: D,
-        run_mode_config: R,
-    ) -> Result<Self, ConfigError> {
+    pub fn create_with_default(path: PathBuf, file_name: &str) -> Result<Self, ConfigError> {
```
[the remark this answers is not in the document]
bonomat: That's what I tried before, but the problem is, that if the folder does not exist, creating that file will fail, that's why I split it up.

Check a few lines down: 
* `if !path.exists()` --> the folders will be created
* `if default_config.exists()` --> the config file will be created

application/comit_node/src/settings/mod.rs
bonomat: Ok, can do so, any other opinions on this?

application/comit_node/src/settings/mod.rs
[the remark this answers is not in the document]
thomaseizinger: You could do `.ends_with` and check if the passed path is actually the config file and then use `.parent` to get the folder path.

application/comit_node/src/settings/mod.rs
```diff
@@ -5,76 +5,165 @@ use crate::seed::Seed;
...
+            } else {
+                return Err(ConfigError::Message(format!(
+                    "Config file exists but is not a file: {:?}",
+                    default_config
+                )));
+            }
+        }
+
+        let default_settings = ComitNodeSettings::default();
+        let toml_string = toml::to_string(&default_settings).unwrap();
```
thomaseizinger: :mega: No `unwrap`s please.

application/comit_node/src/settings/mod.rs
```diff
@@ -5,76 +5,165 @@ use crate::seed::Seed;
...
 }
 
 impl ComitNodeSettings {
-    pub fn create<D: AsRef<OsStr>, R: AsRef<OsStr>>(
-        default_config: D,
-        run_mode_config: R,
-    ) -> Result<Self, ConfigError> {
+    pub fn create_with_default(path: PathBuf, file_name: &str) -> Result<Self, ConfigError> {
+        if !path.exists() {
+            log::warn!(
```
thomaseizinger: :mega: This should not be a `warn` but a `debug` at most.

application/comit_node/src/settings/mod.rs
thomaseizinger: Yeah I was thinking the same.

I might make sense to change the API to be:

```rs
fn write_to(&self, path: PathBuf);
```

and combine it with `Config::default()` with something like:

High-level flow:

1. If --config is passed, try to load it
2. If --config is not passed, try to load from default location
3. If not found, create new instance of config with `Config::default()` and write it to the default location with `config.write_to(DEFAULT_LOCATION)`.

application/comit_node/src/settings/mod.rs
```diff
@@ -5,76 +5,165 @@ use crate::seed::Seed;
...
         config.merge(File::from(default_config_file))?;
 
-        // Add in the current environment file
-        // Default to 'development' env
-        // Note that this file is _optional, in our case this holds all the keys
-        config.merge(File::from(environment_config_file).required(false))?;
-
         // Add in a local configuration file
         // This file shouldn't be checked in to git
         config.merge(File::with_name("config/local").required(false))?;
```
thomaseizinger: Do we still need this?

application/comit_node/src/settings/mod.rs
bonomat: Done

application/comit_node/src/settings/mod.rs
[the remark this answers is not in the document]
LLFourn: It seems like it should be possible to get directory a file is in?

application/comit_node/src/settings/mod.rs
bonomat: @thomaseizinger : that's what I am doing right now. Do think of controlling this from the outside though?

application/comit_node/src/settings/mod.rs
LLFourn: :question: delete this?

application/comit_node/src/settings/mod.rs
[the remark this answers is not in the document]
bonomat: I like the current approach more unless we are fine with this function making assumptions about either config file name+type (this should be passed in from the outside). Otherwise you would need to pass in the file name+path anyways leaving us with the same situation.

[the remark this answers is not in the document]
bonomat: @LLFourn 
> It seems like it should be possible to get directory a file is in?

I would not know how.

application/comit_node/src/settings/mod.rs
bonomat: updated

application/comit_node/src/settings/mod.rs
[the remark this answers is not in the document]
bonomat: updated

application/comit_node/src/settings/mod.rs
```diff
@@ -5,93 +5,188 @@ use crate::seed::Seed;
...
     pub web_gui: Option<HttpSocket>,
+    #[serde(default = "default_log_levels")]
+    pub log_levels: LogLevels,
+}
+
+impl Default for ComitNodeSettings {
+    fn default() -> Self {
+        let comit_listen = "/ip4/0.0.0.0/tcp/8011".parse().unwrap();
+        let btsieve_url = Url::parse("http://localhost:8181").unwrap();
+        let seed = Seed::new_random().unwrap();
```
LLFourn: :wrench: Maybe don't unwrap this and give a good error message by matching the result

application/comit_node/src/settings/mod.rs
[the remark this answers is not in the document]
D4nte: `8181` sounds good :)

application/comit_node/src/seed.rs
```diff
@@ -27,6 +28,13 @@ impl Seed {
         sha.result(&mut result);
         result
     }
+
+    pub fn new_random() -> Result<Seed, rand::Error> {
+        let mut rng = OsRng::new()?;
+        let mut arr = [0u8; 32];
```
thomaseizinger: https://www.youtube.com/watch?v=osimhaV4JsA

application/comit_node/src/settings/mod.rs
```diff
@@ -5,93 +5,191 @@ use crate::seed::Seed;
...
     pub network: String,
 }
 
 impl ComitNodeSettings {
-    pub fn create<D: AsRef<OsStr>, R: AsRef<OsStr>>(
-        default_config: D,
-        run_mode_config: R,
-    ) -> Result<Self, ConfigError> {
+    pub fn write_to(self, config_file: PathBuf) -> Result<Self, ConfigError> {
+        ComitNodeSettings::verify_directory_exists(&config_file)?;
```
thomaseizinger: :upside_down_face: To make the side-effect clear.

```suggestion
        ComitNodeSettings::ensure_directory_exists(&config_file)?;
```

application/comit_node/src/settings/mod.rs
```diff
@@ -110,4 +208,35 @@ mod tests {
...
+        let default_settings = ComitNodeSettings::default();
+
+        let default_settings = default_settings.write_to(config_file_incl_path.clone());
+        let settings = ComitNodeSettings::read(config_file_incl_path.clone());
+
+        delete_tmp_files(&config_path, &config_file);
+
+        assert_that(&default_settings).is_ok();
+        assert_that(&settings).is_ok();
+        assert_that(&default_settings.unwrap()).is_equal_to(&settings.unwrap());
```
thomaseizinger: :wrench: You verified that their are `Ok` already, no need for the unwrap. You could just compare the results.
```suggestion
        assert_that(&default_settings).is_equal_to(&settings);
```

application/comit_node/src/load_settings.rs
```diff
@@ -11,20 +11,26 @@ pub struct Opt {
...
             None => Err(ConfigError::Message(
                 "Unable to determine user's home directory".to_string(),
             )),
-            Some(dirs) => Ok(Path::join(dirs.home_dir(), ".config/comit_node")),
+            Some(dirs) => {
+                let config_path = Path::join(dirs.home_dir(), ".config/comit_node/default.toml");
+                if config_path.exists() {
+                    ComitNodeSettings::read(config_path)
+                } else {
+                    ComitNodeSettings::default().write_to(config_path)
```
thomaseizinger: :mega: I think we should tell the user here what we are doing.

```rust
log::info("config file was neither provided nor found at default location, generated default config at {}", config_path);
```

application/comit_node/src/seed.rs
bonomat: hahaha :D 🏴‍☠️

application/comit_node/src/settings/mod.rs
bonomat: Unfortunately not:
```
    = note: the method `is_equal_to` exists but the following trait bounds were not satisfied:
            `std::result::Result<settings::ComitNodeSettings, config::error::ConfigError> : std::cmp::PartialEq`
```

PR #996, opened 2019-05-28
947 cleanup config management

### LINKED ISSUE #947: Improve the comit-rs configuration handling
**Current situation:**
Currently the configuration for the `comit_node` and `btsieve`  has to be placed in `~/.config/comit_node/default.toml`.

Additionally the path to the `default.toml` file can be changed through the `COMIT_NODE_CONFIG_PATH` and `BTSIEVE_CONFIG_PATH` environment variable. However, currently only the path can be set there. The file name always has to be `default.toml` which is weird. 

Additionally the error messages should be improved.

**Improvement Proposal:**
1. _Command line parameter for config:_ Add command line parameter `--config` that accepts a file-path to a configuration file. 
1. _Treat `default` as fall back:_ Try to fall back to a `default.toml` file in a directory where a specific configuration file cannot be found. 
2. _Allow file-path instead of just folder-path:_ Change the environment variable handling so one can pass a complete file-path (not only folder). If there is only a folder provided the fallback mechanism should try yo load the `default.toml`.

**Improvement Proposal for error message**:

current (when configuration file cannot be found):
```
Error: configuration file "/Users/dakami/.config/comit_node/default" not found
```
change (add file extension):
```
Error: configuration file "/Users/dakami/.config/comit_node/default.toml" not found
```

current (when configuration parameter is missing in the specified configuration file):
```
Error: missing field `comit`
``` 
change (add configuration file)
```
Error: missing configuration entry `comit` in [name_of_loaded_config_file].toml
``` 

Related to: https://github.com/comit-network/comit-rs/issues/919

---

Tasks:

- comit_node: Change the path of the config file to: "$HOME/.config/comit/comit_node.toml"
- btsieve: Same thing but btsieve.toml
- try to improve the error message as per Daniel's suggestion (full path + extension) above
- prevent config-rs to look for a configuration file named "development.toml" based on the run environment
- there should also be no fallback mechanism for config files

bonomat:
This is a simple PR that

* rename config files from `default.toml` to `btsieve.toml` and `comit_node.toml`
* removes unneeded merging of `development.toml` etc
* allows to pass-in the exact config-file-path and not just the config-file-directory

Resolves #947


application/btsieve/src/load_settings.rs
```diff
@@ -1,43 +1,22 @@
...
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use structopt::StructOpt;
 
 #[derive(StructOpt, Debug)]
 pub struct Opt {
     /// Path to configuration folder
     #[structopt(short = "c", long = "config", parse(from_os_str))]
-    config_path: Option<PathBuf>,
+    config_file: PathBuf,
```
thomaseizinger: :mega: So there is no more default location for `btsieve` now? Why not?
I think we should have the default location feature but no auto-generation.

application/btsieve/src/settings/mod.rs
```diff
@@ -43,17 +43,12 @@ pub struct Ethereum {
 }
 
 impl Settings {
-    pub fn create<D: AsRef<OsStr>>(default_config: D) -> Result<Self, ConfigError> {
+    pub fn read<D: AsRef<OsStr>>(config_file: D) -> Result<Self, ConfigError> {
```
thomaseizinger: :wrench: `from_file`?

application/comit_node/src/settings/mod.rs
```diff
@@ -169,13 +169,13 @@ impl ComitNodeSettings {
         }
     }
 
-    pub fn read<D: AsRef<OsStr>>(default_config: D) -> Result<Self, ConfigError> {
+    pub fn read<D: AsRef<OsStr>>(config_file: D) -> Result<Self, ConfigError> {
```
thomaseizinger: :wrench: `from_file`?

application/comit_node/src/load_settings.rs
```diff
@@ -5,54 +5,40 @@ use structopt::StructOpt;
...
         }
         None => match directories::UserDirs::new() {
             None => Err(ConfigError::Message(
                 "Unable to determine user's home directory".to_string(),
             )),
             Some(dirs) => {
-                let config_path = Path::join(dirs.home_dir(), ".config/comit_node/default.toml");
-                if config_path.exists() {
-                    ComitNodeSettings::read(config_path)
+                let config_file = Path::join(dirs.home_dir(), ".config/comit_node/comit_node.toml");
```
thomaseizinger: :mega: The ticket says we want the location to be: `$HOME/.config/comit/comit_node.toml`

application/btsieve/src/settings/mod.rs
```diff
@@ -43,17 +43,12 @@ pub struct Ethereum {
...
-        let default_config_file = Path::new(&default_config);
+        let config_file = Path::new(&config_file);
 
-        // Start off by merging in the "default" configuration file
-        config.merge(File::from(default_config_file))?;
-
-        // Add in a local configuration file
-        // This file shouldn't be checked in to git
-        config.merge(File::with_name("config/local").required(false))?;
+        config.merge(File::from(config_file))?;
```
D4nte: ❓ do you still need this `.merge` call? Can't you create a `Config` directly from the file?

application/comit_node/src/settings/mod.rs
```diff
@@ -169,13 +169,13 @@ impl ComitNodeSettings {
...
         let mut config = Config::new();
 
-        let default_config_file = Path::new(&default_config);
+        let config_file = Path::new(&config_file);
 
         // Start off by merging in the "default" configuration file
-        config.merge(File::from(default_config_file))?;
+        config.merge(File::from(config_file))?;
```
D4nte: ❓ same question related to `.merge`

D4nte:
Also I'd like to see less commit/better commit text

application/btsieve/src/load_settings.rs
bonomat: Sure, I re-added it.

application/comit_node/src/load_settings.rs
bonomat: Updated.

bonomat:
> Please update PR title.
> 
> I don't think this is correct, please clarify:
> 
> > remove optional from btsieve ops as it does not make sense to generate a default file right now
> 
> One non-blocking comment.

ups, this was made redundant after Thomas' comment.

application/btsieve/src/settings/mod.rs
bonomat: Unfortunately we still need it as config-rs does not support different parsing. 

I thought about moving to toml-rs instead - which would allow deserialization from a toml string - but we decided to stay with config-rs, so I left it. 
Also, toml would require us to parse the file by ourselves.

bonomat:
> Also I'd like to see less commit/better commit text

This was actually done on purpose to have a clean commit history instead of 100 small commits :)

application/btsieve/src/settings/mod.rs
LLFourn: :question: What do you mean "parse the file by ourselves"?  https://docs.rs/toml/0.5.1/toml/ supports serde which should cover everything we want to do with it unless I'm missing something.

bonomat: I meant we would need to read the file into a string first and then let toml deserialize it into our Config.

thomaseizinger: Serde can do `from_reader` which you can get from a file I think. No need for the string inbetween.

PR #1001, opened 2019-05-31
Set CORS headers on error responses

### LINKED ISSUE #935: comit-rs does not set CORS headers for error responses
I just got an unexpected error in comit-i:
When sending an invalid request to the comit-node, I received: 

`Access to XMLHttpRequest at 'http://localhost:8000/swaps/rfc003' from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.` 

This means, the comit-node did not include the corresponding CORS header in the error response.

Since we are using the cors-module from `warp`, this is very likely an issue with `warp` and has to be fixed upstream.

Tasks:
- figure out if it is an upstream bug -> open a warp issue
- try to fix it in warp otherwise patch comit-rs
- make a dry-test

thomaseizinger:
Fixes #935.

### DIFF
--- a/api_tests/webgui/webgui.ts
+++ b/api_tests/webgui/webgui.ts
@@ -70,6 +70,18 @@ setTimeout(async function() {
             );
         });
 
+        it("[David] Sets appropriate CORS headers on error responses", async () => {
+            let res = await request(david.comit_node_url())
+                .get("/swaps/rfc003/deadbeef-dead-beef-dead-deadbeefdead")
+                .set("Origin", "http://localhost:8080");
+
+            res.should.have.status(404);
+            res.should.have.header(
+                "access-control-allow-origin",
+                "http://localhost:8080"
+            );
+        });
+
         it("[David] comit-i returns 200 OK", async () => {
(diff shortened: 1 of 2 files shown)


api_tests/webgui/webgui.ts
```diff
@@ -70,6 +70,18 @@ setTimeout(async function() {
             );
         });
 
+        it("[David] Sets appropriate CORS headers on error responses", async () => {
+            let res = await request(david.comit_node_url())
+                .get("/swaps/rfc003/deadbeef-dead-beef-dead-deadbeefdead")
+                .set("Origin", "http://localhost:8080");
+
+            res.should.have.status(404);
```
D4nte: 🤡  @thomaseizinger would comment "please use expect here".

thomaseizinger: Fair :D

thomaseizinger: I rewrote all assertions in `webgui` to use `expect`.

PR #1010, opened 2019-06-04
Run tsc to catch TypeScript syntax error early on

D4nte:
Would have saved me some time :)

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -142,7 +142,7 @@ script = [
 [tasks.check-ts]
 description = "Runs tsc to validate TypeScript syntax."
 workspace = false
-install_script = ["(cd ./api_tests; yarn install;)"]
+install_script = ["(cd ./api_tests; yarn install; yarn run gen;)"]
 script = ["./api_tests/node_modules/.bin/tsc --project ./api_tests/tsconfig.json"]
 
 [tasks.ci]
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -142,6 +142,7 @@ script = [
 [tasks.check-ts]
 description = "Runs tsc to validate TypeScript syntax."
 workspace = false
+install_script = ["(cd ./api_tests; yarn install;)"]
 script = ["./api_tests/node_modules/.bin/tsc --project ./api_tests/tsconfig.json"]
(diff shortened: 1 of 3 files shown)


api_tests/tsconfig.json
```diff
@@ -10,5 +10,6 @@
         "noUnusedParameters": true,
         "resolveJsonModule": true,
         "esModuleInterop": true
-    }
+    },
+    "include": ["dry", "btsieve", "e2e", "webgui"]
```
thomaseizinger: :question: Shouldn't need this?

> If the "files" and "include" are both left unspecified, the compiler defaults to including all TypeScript (.ts, .d.ts and .tsx) files in the containing directory and subdirectories

https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

D4nte: Indeed I now realised I previously used `tsc` incorrectly.

Makefile.toml
```diff
@@ -138,6 +139,12 @@ script = [
 '''
 ]
 
+[tasks.check-ts]
+description = "Runs tsc to validate TypeScript syntax."
+workspace = false
+install_script = ["(cd ./api_tests; yarn install; yarn run gen;)"]
```
thomaseizinger: :mega: 
`yarn run gen` should be in the `postinstall` script I think. Why do we need to repeat it here?

D4nte: Oh, indeed, it was all messy before! Now cleaned up

Makefile.toml
```diff
@@ -148,7 +155,7 @@ run_task = "ci-flow"
 description = "Install/build dependencies for api tests"
 workspace = false
 private = true
-script = [ "( cd ./api_tests && yarn install; yarn run gen; )" ]
+script = [ "( cd ./api_tests && yarn install;)" ]
```
da-kami: ❓ The `gen` script was renamed to `postinstall` - but as far as I could see `postinall` is never triggered. Is it triggered implicitly somehow, am I missing something?

D4nte: It is indeed triggered implicitly after an install

PR #1015, opened 2019-06-04
Upgrade embedded comit-i to v0.2.0-alpha

D4nte:
- Uses new siren API
- Support address hint use

### DIFF
--- a/vendor/comit_i/version
+++ b/vendor/comit_i/version
@@ -1,1 +1,1 @@
-v0.1.1-alpha
+v0.2.0-alpha


vendor/comit_i/version
```diff
@@ -1 +1 @@
-v0.1.1-alpha
+v0.1.2-alpha
```
da-kami:
```suggestion
v0.2.0-alpha
```

🙃 Not that it matters too much for a pre-release but I think it should be a minor version release rather than a patch. Maybe even a major version (first digit) because we did changes that are not backwards compatible.

D4nte: ok!

PR #1017, opened 2019-06-04
Fix the SwapStatus::new function

### LINKED ISSUE #994: Swap marked as in progress when Alice refunds before Bob funds
If Alice funds the alpha ledger HTLC and then refunds before Bob funds the beta ledger HTLC, there are no actions for either party to perform and the swap status should be `NOT_SWAPPED`. This is not the case, since the swap status remains as `IN_PROGRESS`.

thomaseizinger:
Resolves #994.

Please get it in before the sprint ends :upside_down_face:


application/comit_node/src/http_api/routes/rfc003/swap_state.rs
```diff
@@ -194,38 +194,120 @@ impl<L: Ledger> From<rfc003::LedgerState<L>> for LedgerState<L::HtlcLocation, L:
...
-                    "Internal failure with swap communication state {:?},\
-                     alpha ledger state {:?}, beta ledger state {:?} and error {:?}",
-                    swap_communication_state,
-                    alpha_ledger,
-                    beta_ledger,
-                    error
-                );
-                SwapStatus::InternalFailure
-            }
+
```
luckysori: :+1: So much cleaner than my original implementation :sparkles:

application/comit_node/src/http_api/routes/rfc003/swap_state.rs
```diff
@@ -194,38 +194,120 @@ impl<L: Ledger> From<rfc003::LedgerState<L>> for LedgerState<L::HtlcLocation, L:
...
+        match (alpha_ledger, beta_ledger) {
+            (Redeemed, Redeemed) => SwapStatus::Swapped,
+            (Refunded, _) | (_, Refunded) => SwapStatus::NotSwapped,
+            _ => SwapStatus::InProgress,
         }
     }
 }
+
+#[cfg(test)]
+impl quickcheck::Arbitrary for SwapCommunicationState {
```
luckysori: :+1: Omg yes quickcheck

application/comit_node/src/http_api/swap_resource.rs
```diff
@@ -82,16 +86,19 @@ pub fn build_rfc003_siren_entity<S: StateStore>(
                 .get::<ROLE>(&id)?
                 .ok_or_else(problem::state_store)?;
 
-            // TODO: Implement From<actor::State> for SwapOutcome
```
luckysori: :question: Did you mean to delete this?

luckysori:
> Can you explain the "Radical optimisation" a bit more?

It is definitely much better implemented than what I had done before haha.

application/comit_node/src/http_api/swap_resource.rs
thomaseizinger: Yes I deleted it on purpose because I think it is unnecessary and the TODO ticket that is flying around is annoying :see_no_evil:

application/comit_node/src/http_api/routes/rfc003/swap_state.rs
thomaseizinger: Only possible because I wrote a lot of tests around it :)
The quickcheck implementation is now actually a bit less useful because there is no return paths for `SwapStatus::InteralFailure` other than providing and error. I had an inbetween version where that was not the case.

However, I was so proud for including it that I didn't want to remove it :D

application/comit_node/src/http_api/routes/rfc003/swap_state.rs
thomaseizinger: We should definitely make more use of quickcheck!

PR #1020, opened 2019-06-05
Return Listen addresses

### LINKED ISSUE #1009: comit_node should return listen addresses on GET /
In addition to the peer id, `GET /` should return the multiaddress(es) on which the comit_node listen.

This would allow the replacement of the multiaddress manipulation in the e2e tests. (link to come)

D4nte:
Resolves #1009


application/comit_node/src/bin/comit_node.rs
```diff
@@ -80,6 +83,7 @@ fn main() -> Result<(), failure::Error> {
         alice_protocol_dependencies,
         Arc::clone(&swarm),
         local_peer_id,
+        Arc::clone(&swarm),
```
thomaseizinger: :wrench: Creating a new trait for each piece of functionality doesn't really scale since we are passing the `Swarm` now already two times.

What do you think about unifying `BamPeers` and `ListenAddresses` to something like `SwarmInfo` that provides us convenient access to all kinds of data from the `Swarm`?

PR #1023, opened 2019-06-07
Add SwapId to SwapResource

luckysori:


api_tests/swap.schema.json
```diff
@@ -6,6 +6,13 @@
     "title": "Properties of a SWAP",
     "required": ["role", "protocol", "status", "parameters", "counterparty"],
     "properties": {
+        "swap_id": {
```
thomaseizinger: Can we have this just be named `id`?

thomaseizinger: Also, I'd say it should be a required field!

PR #1034, opened 2019-06-08
Cargo update

D4nte:


bonomat:
I think this won't work. Let's just discard this PR and let the version pumper do the job step by step.

PR #1035, opened 2019-06-08
pin-tokio

D4nte:
To avoid error BitcoinRpcConnection(Transport(Error(Executor(EnterError { reason: "attempted to run an executor while another executor is already running" }), "http://127.0.0.1:18443/")))

See https://github.com/mimblewimble/grin/pull/2128


D4nte:
Issue fixed with #1038

Issue #1046, opened 2019-06-20
Investigate how to get rid of cargo-make


thomaseizinger:
I propose to find a way for getting rid of cargo-make for our build setup. Here is why:

- **The Makefile is hard to understand because it is only the _diff_ to the default Makefile.** Hence, it is hard to reason about how this whole thing works. This makes it especially hard for new-comers to the project to get started.
- **It does not capture everything.** cargo-make does not take care of node, yarn and other dependencies. It is not a one-stop shop that takes care of everything.
- **It causes issues from time to time.** We have had several occasions where updates to cargo-make broke our build or where changes in the ecosystem (clippy stabilized) did not go very smooth.
- **The design is not very good.** This is more of a personal note but I just dislike the design and working with it. I find it cumbersome to achieve the tasks I want to do.

The TenX people use `nix` for their builds. There may also be other solutions.

This spike is about evaluating which tools may replace cargo-make to ease the build and setup process.

Cargo by itself actually supports custom commands in the form of aliases similarly to git: https://doc.rust-lang.org/cargo/reference/config.html#configuration-keys

thomaseizinger:
Removed dependency on #985 because of the work in #1100.

thomaseizinger:
I propose doing a very quick spike on this that outlines, how a Makefile as done by @D4nte in https://github.com/comit-network/create-comit-app.

PR #1047, opened 2019-06-20
Compile on stable Rust!

thomaseizinger:
This PR fixes the Rust version we use for compiling to stable version 1.35.0.

Unfortunately, we also require a nightly toolchain to be installed because we depend on nightly features of rustfmt.

Depends on #1045 because `noise_handshake` still requires nightly to compile.


Makefile.toml
```diff
@@ -108,16 +108,17 @@ extend = "api-tests-flow"
...
 ################
 # Custom tasks #
 
 [tasks.check-rs-format]
 description = "Runs cargo fmt to check appropriate code format."
 workspace = false
-install_crate = { crate_name = "rustfmt-nightly", rustup_component_name = "rustfmt-preview", binary = "rustfmt", test_arg = "--help" }
-command = "cargo"
-args = ["fmt", "--", "--check"]
+script = [
```
D4nte: Are you sure this work? Your build is broken, probably related to that.

README.md
```diff
@@ -24,10 +24,13 @@ Contains crates specific to our application. Can depend on libraries located in
 ## Setup build environment
 
 1. Install `rustup`: `curl https://sh.rustup.rs -sSf | sh`
-2. Install libzmq:
+2. Install the following versions of Rust
+    - stable: `rustup install stable`
+    - nightly-2019-04-30: `rustup install nightly-2019-04-30` (this one is only used for `rustfmt`)
```
D4nte: 📣 move that to the `## Setup testing/dev environment` section

rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2019-04-30
+1.35.0
```
D4nte: ⏎

Makefile.toml
thomaseizinger: Yes I tested it locally.
I also opened #1046 because cargo-make is so annoying :sweat_smile: 

The build failed because in this branch, `noise_handshake` is still included. That is why I added the line in the PR description that it depends on #1045. I rebased now so the build should be green after that.

README.md
thomaseizinger: Good idea, will do that!

.circleci/config.yml
```diff
@@ -42,6 +42,8 @@ commands:
           name:  "Setup rust environment"
           command: |
             curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain $(< rust-toolchain) && source $HOME/.cargo/env
+            rustup install nightly-2019-04-30
```
D4nte: 🔧 anyway you can put the version in a variable so that we don't miss one when updating?

Makefile.toml
```diff
@@ -79,6 +80,14 @@ args = [
     "-D", "warnings"
 ]
 
+# Overridden because: workspace = false and custom toolchain
+[tasks.format]
+workspace = false
+clear = true
+script = [
+  "cargo +nightly-2019-04-30 fmt"
```
D4nte: 🔧 same than above. maybe have a file called `rust-toolchain-for-fmt` containing `nightly-2019-04-30` (with no end of line :) )

README.md
```diff
@@ -48,16 +48,17 @@ In order to do a swap you will have to start two comit nodes.
 
 1. Install `docker` & `docker-compose`
 2. Install `node` (check the version required in package.json) & `yarn`
-3. Install `cargo-make`: `cargo install cargo-make`
-4. Run `cargo make` in the root folder of the repository, this will install various crates & tools such as rustfmt & clippy
+3. Install Rust nightly-2019-04-30: `rustup install nightly-2019-04-30` (this one is only used for `rustfmt`)
+4. Install `rustfmt` for `nightly-2019-04-30`: `rustup component add rustfmt --toolchain nightly-2019-04-30`
```
D4nte: 📣 sometimes you put `nightly-2019-04-30` within backticks and sometimes not

thomaseizinger: What I tried to do was having a _sentence_ that describes what you should do, followed by a command that you can copy and paste into the terminal.

Do you want me to change that.

Makefile.toml
thomaseizinger: If the toolchain where to change often, yeah maybe. But I don't see a reason why it would change ever (apart from moving to stable, which would make the file obsolete or new formatting rules/changes that are only present in a newer nightly version). The only reason we have to pin it is because not every nightly has `rustfmt`, otherwise we could just go for the latest nightly.

Given that, I actually like having one file _less_ in the repository and some minor duplication within several configuration files. If any of that is out of sync, you immediately notice it anyway because the build on CI fails.

.circleci/config.yml
thomaseizinger: You can't really miss one because the build fails if they are out of sync.

README.md
```diff
@@ -48,16 +48,17 @@ In order to do a swap you will have to start two comit nodes.
 
 1. Install `docker` & `docker-compose`
 2. Install `node` (check the version required in package.json) & `yarn`
-3. Install `cargo-make`: `cargo install cargo-make`
-4. Run `cargo make` in the root folder of the repository, this will install various crates & tools such as rustfmt & clippy
+3. Install Rust nightly-2019-04-30: `rustup install nightly-2019-04-30` (this one is only used for `rustfmt`)
```
thomaseizinger:
```suggestion
3. Install Rust `nightly-2019-04-30`: `rustup install nightly-2019-04-30` (this one is only used for `rustfmt`)
```

README.md
D4nte: I am referring to point `3.` and `4.`

thomaseizinger: Did you check the latest state if your comment is resolved now?

PR #1052, opened 2019-06-25
Add unit tests for HTTP API primitive types

### LINKED ISSUE #766: Test the serialization of primitive types used in HTTP API
Test the format of primitive types used in HTTP API.
E.g: Test that an Ethereum address is serialized as expected against a string.

DoD:
- [ ] Format test (ie, serialization of specific types) should go to Rust tests.

tcharding:
We would like to test that the HTTP API primitive types serialize as expected.  Here we add unit tests that test the serialization as it is now (i.e. passing tests).  This makes the format on the wire explicit and if the serialization code changes ever these tests will catch the change.

Fixes #766

### DIFF
--- a/application/comit_node/src/http_api/mod.rs
+++ b/application/comit_node/src/http_api/mod.rs
@@ -217,6 +217,7 @@ mod tests {
     use ethereum_support::{
         self, Address, Bytes, Erc20Quantity, Erc20Token, EtherQuantity, H160, H256, U256,
     };
+    use libp2p::PeerId;
     use std::{convert::TryFrom, str::FromStr};
 
     #[test]
@@ -375,5 +376,17 @@ mod tests {
             swap_id_serialized,
             r#""ad2652ca-ecf2-4cc6-b35c-b4351ac28a34""#
         )
+    }
+
+    #[test]
+    fn http_peer_id_serializes_correctly_to_json() {
+        let peer_id = PeerId::from_str("QmfUfpC2frwFvcDzpspnfZitHt5wct6n4kpG5jzgRdsxkY").unwrap();
+        let peer_id = Http(peer_id);
+
+        let serialized = serde_json::to_string(&peer_id).unwrap();
+        assert_eq!(
+            serialized,
+            r#""QmfUfpC2frwFvcDzpspnfZitHt5wct6n4kpG5jzgRdsxkY""#
+        );
     }
 }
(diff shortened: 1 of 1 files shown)


tcharding:
Creating draft PR to solicit basic Rust help on this one.
```
$ git grep -IPn "What goes here"
mod.rs:285:            foo: "What goes here?",
mod.rs:457:            multihash: "What goes here?",
```

application/comit_node/src/http_api/mod.rs
```diff
@@ -247,8 +247,51 @@ mod tests {
...
+        assert_eq!(
+            &serialized,
+            r#""c33258694320aa8f976e88574d01b15f129f402be3e48d86c01a35a736134e03""#, /* This is a SHA-256 hash. */
+        );
+    }
+
+    #[test]
+    fn http_bitcoin_support_pubkeyhash_serializes_to_json() {
+        let vec = vec![1; 20]; // This is represents a ripemd160 hash.
+        match bitcoin_support::PubkeyHash::try_from(vec.as_ref()) {
```
D4nte:
```suggestion
        let hash = bitcoin_support::PubkeyHash::try_from(vec.as_ref()).expect("failed to create pubkey hash: {}", err);
        let hash = Http(hash);
        let serialized = serde_json::to_string(&hash).unwrap();
        assert_eq!(&serialized, r#""0101010101010101010101010101010101010101""#,);
```

Use `expect` :)

tcharding: Sweet, thanks.

application/comit_node/src/http_api/mod.rs
```diff
@@ -398,4 +407,15 @@ mod tests {
             r#""ad2652ca-ecf2-4cc6-b35c-b4351ac28a34""#
         )
     }
+
+    #[test]
+    fn http_peer_id_serializes_correctly_to_json() {
+        let peer_id = PeerId::random();
+        let want = format!("\"{}\"", peer_id);
```
D4nte: Hum, I don't think is a great test because if `serde_json::to_string` uses the `Display` trait underneath then there may a bug and you would not see it.

Can you initialize a `PeerId` from a `String` for example? I'd like to see that the test is actually done against the string slice `"QmfUfpC2frwFvcDzpspnfZitHt5wct6n4kpG5jzgRdsxkY"`

tcharding: Implemented as suggested.

PR #1053, opened 2019-06-25
Print offsets for Bitcoin Basic HTLC script

### LINKED ISSUE #910: Generate Bitcoin HTLCs and print offset
A bit similar to #864 (Ethereum) but with a restricted scope.
Generate Bitcoin HTLC hex & offset used by the RFC from Bitcoin script code.

DoD:
- [x] blockchain_contracts crate print the offsets and the hex for bitcoin htlc
- [x] Issue is opened to correct Bitcoin RFC if needed #82

Note:
- No test are moved from comit_node
- Brand new code to generate the offset
- comit_node stays untouched

D4nte:
Resolves #910


vendor/blockchain_contracts/src/bin/calculate_offsets/bitcoin/rfc003/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+impl From<crate::calculate_offsets::Error> for Error {
+    fn from(err: crate::calculate_offsets::Error) -> Self {
+        match err {
+            crate::calculate_offsets::Error::PlaceholderNotFound => Error::PlaceholderNotFound,
+            crate::calculate_offsets::Error::Hex(err) => Error::Hex(err),
+        }
+    }
+}
+
+impl From<crate::calculate_offsets::placeholder_config::Error> for Error {
```
tcharding: Why do we do this instead of bringing `crate::calculate_offsets::placeholder_config` into scope and using
`impl From<placeholder_config::Error> for Error {`?

vendor/blockchain_contracts/src/bin/calculate_offsets/mod.rs
```diff
@@ -1 +1,74 @@
...
+    contract_template
+        .windows(placeholder.len())
+        .position(|window| window == placeholder)
+}
+
+fn check_bin_in_path(bin: &str) {
+    let output = Command::new("which").arg(bin).output().unwrap();
+    if output.stdout.is_empty() {
+        let mut msg = format!("`{}` cannot be found, check your path", bin);
+        msg = format!("{}\nPATH: {:?}", msg, std::env::var("PATH"));
```
bonomat: 🔧 How about combining these two lines? 
```suggestion
        let msg = format!("`{}` cannot be found, check your path \n {}", bin, std::env::var("PATH"));
```

PR #1055, opened 2019-06-26
Check format based on file suffix

tcharding:
Currently our git pre-commit hook runs the Typescript format checker and the Rust format checker.  The Typescript checker is slow because it invokes `yarn`.  If a commit only modifies Rust source files there is no need to run the Typescript checker.  Also when modifying only Typescript it is unnecessary to check the format of the Rust code.

Add clause to the pre-commit hook shell script to only run language specific format checkers if source files of that language were modified.

Only tested on Linux `GNU bash, version 4.4.19(1)-release (x86_64-pc-linux-gnu)`

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -20,7 +20,7 @@ done
 if (( $check_rs != 0 )); then
     cargo make check-rs-format
     ret=$?
-    if [ $? -ne 0 ]; then
+    if [ $ret -ne 0 ]; then
        exit $ret
     fi
 fi
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -10,7 +10,7 @@ if [ $? -eq 0 ]; then
     fi
 fi
 
-# Run check-rs-format if Typescript source files were modified.
+# Run check-ts-format if Typescript source files were modified.
 git status | egrep "*.{ts,json,yml}"
 if [ $? -eq 0 ]; then
     cargo make check-ts-format
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -19,10 +19,18 @@ done
 
 if (( $check_rs != 0 )); then
     cargo make check-rs-format
+    ret=$?
+    if [ $? -ne 0 ]; then
+       return $ret
+    fi
 fi
 
 if (( $check_ts != 0 )); then
     cargo make check-ts-format
+    ret=$?
+    if [ $? -ne 0 ]; then
+       return $ret
+    fi
 fi
 
 exit 0
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,32 +1,40 @@
 #!/usr/bin/env bash
 
-check_rs=0
-check_ts=0
+debug=0                         # Set to 1 to enable debugging, 0 to disable.
+do_check_rs=0
+do_check_ts=0
+
+check_rs_cmd="cargo make check-rs-format"
+check_ts_cmd="cargo make check-ts-format"
+if (( debug == 1 )); then
+    check_rs_cmd="echo check_rs ran"
+    check_ts_cmd="echo check_ts ran"
+fi
 
 declare -a extra_flags=("--" "--cached")
 
 for flag in ${extra_flags[@]}; do
     git diff --name-only $flag | grep '.rs$' > /dev/null 2>&1
     if [ $? -eq 0 ]; then
-        check_rs=1
+        do_check_rs=1
     fi
 
     git diff --name-only $flag | grep -E '.ts$|.json$|.yml$' > /dev/null 2>&1
     if [ $? -eq 0 ]; then
-        check_ts=1
+        do_check_ts=1
     fi
 done
 
-if (( $check_rs != 0 )); then
-    cargo make check-rs-format
+if (( $do_check_rs != 0 )); then
+    eval "$check_rs_cmd"
     ret=$?
     if [ $ret -ne 0 ]; then
        exit $ret
     fi
 fi
 
-if (( $check_ts != 0 )); then
-    cargo make check-ts-format
+if (( $do_check_ts != 0 )); then
+    eval "$check_ts_cmd"
     ret=$?
     if [ $ret -ne 0 ]; then
        exit $ret
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -11,19 +11,25 @@ if (( debug == 1 )); then
     check_ts_cmd="echo check_ts ran"
 fi
 
-declare -a extra_flags=("--" "--cached")
+git diff --name-only | grep '.rs$' > /dev/null 2>&1
+if [ $? -eq 0 ]; then
+    do_check_rs=1
+fi
 
-for flag in ${extra_flags[@]}; do
-    git diff --name-only $flag | grep '.rs$' > /dev/null 2>&1
-    if [ $? -eq 0 ]; then
-        do_check_rs=1
-    fi
+git diff --name-only --cached | grep '.rs$' > /dev/null 2>&1
+if [ $? -eq 0 ]; then
+    do_check_rs=1
+fi
 
-    git diff --name-only $flag | grep -E '.ts$|.json$|.yml$' > /dev/null 2>&1
-    if [ $? -eq 0 ]; then
-        do_check_ts=1
-    fi
-done
+git diff --name-only | grep -E '.ts$|.json$|.yml$' > /dev/null 2>&1
+if [ $? -eq 0 ]; then
+    do_check_ts=1
+fi
+
+git diff --name-only --cached | grep -E '.ts$|.json$|.yml$' > /dev/null 2>&1
+if [ $? -eq 0 ]; then
+    do_check_ts=1
+fi
 
 if (( $do_check_rs != 0 )); then
     eval "$check_rs_cmd"
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -28,7 +28,9 @@ fi
 if (( $check_ts != 0 )); then
     cargo make check-ts-format
     ret=$?
-    exit $ret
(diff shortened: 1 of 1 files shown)


tcharding:
Note: Please wait for review by an OSX user _and_ a Linux user before merging.

.githooks/pre-commit
```diff
@@ -1,4 +1,23 @@
...
+if [ $? -eq 0 ]; then
+    cargo make check-rs-format
+    ret=$?
+    if [ $ret -ne 0 ]; then
+        exit $ret
+    fi
+fi
+
+# Run check-rs-format if Typescript source files were modified.
+git status | egrep "*.ts"
```
D4nte: 📣 `check-ts-format` actually checks few other formats. Maybe we should rename it to `check-prettier`.
```suggestion
git status | egrep "*.{ts,json,yml}"
```

.githooks/pre-commit
```diff
@@ -1,4 +1,23 @@
...
+git status | egrep "*.rs"
+if [ $? -eq 0 ]; then
+    cargo make check-rs-format
+    ret=$?
+    if [ $ret -ne 0 ]; then
+        exit $ret
+    fi
+fi
+
+# Run check-rs-format if Typescript source files were modified.
```
D4nte:
```suggestion
# Run check-ts-format if Typescript source files were modified.
```
Actually I think you can remove the comment, the code is self-explanatory.

.githooks/pre-commit
```diff
@@ -1,4 +1,23 @@
...
+git status | egrep "*.ts"
+if [ $? -eq 0 ]; then
+    cargo make check-ts-format
+    ret=$?
+    if [ $ret -ne 0 ]; then
+        exit $ret
```
D4nte: 🙃 I don't think this `if` clause is needed because it's the last check you do.

D4nte:
Not valid on Mac. Why do you use `egrep` instead of `grep`?
```
▶ ./.githooks/pre-commit 
egrep: repetition-operator operand invalid
egrep: repetition-operator operand invalid
```

We can check together tomorrow.

.githooks/pre-commit
```diff
@@ -1,4 +1,23 @@
 #!/usr/bin/env bash
 
-cargo make check-rs-format && \
-cargo make check-ts-format
+# Run check-rs-format if Rust source files were modified.
+git status | egrep "*.rs"
```
thomaseizinger: :wrench: There is actually a nice command for listing only the filenames of all staged files: `git diff --name-only --cached`

Source: http://stackoverflow.com/a/33610683

This way, we even get a feature for free with this change: Checking only the files that you actually want to commit :)

.githooks/pre-commit
thomaseizinger: Hmm, now we have one more duplication of the file extensions we want to check for formatting / format automatically.

Any ideas on how those could be unified to a single place? Currently, the other two usages are in `package.json`.

.githooks/pre-commit
```diff
@@ -1,4 +1,23 @@
...
+if [ $? -eq 0 ]; then
+    cargo make check-rs-format
+    ret=$?
+    if [ $ret -ne 0 ]; then
```
thomaseizinger: :wrench: I think you can avoid this if you do `set -e` at the beginning of the script.

.githooks/pre-commit
tcharding: You are correct it is not needed.  I like the verbose format though, I feel it makes the code easier to read since the return value is so explicit there is no wasted thought spent considering what is returned.  I tend to do this in C code also for the same reason.

tcharding:
> Not valid on Mac. Why do you use `egrep` instead of `grep`?
> 
> ```
> ▶ ./.githooks/pre-commit 
> egrep: repetition-operator operand invalid
> egrep: repetition-operator operand invalid
> ```

`grep` does not work here.  I'm not sure why it doesn't find it.  Needs further investigation.

tcharding:
Total re-write, please review from scratch.  Thanks.

**requires:** `grep -E foo` on OSX.  Do you guys have this @bonomat @D4nte @da-kami ?

.githooks/pre-commit
tcharding: We can't use `set -e` here because `grep` failure is part of the normal code path.

.githooks/pre-commit
tcharding: I like it, implemented as suggested.  Thanks.

tcharding: I use a loop and run with and without `--cached` though for those of use who commit without staging.

thomaseizinger:
It is probably personal preference but I find the usage if loops here harder to understand than duplicated commands.

The reason is because I have to unroll the loop in my head to understand, why it is there (to avoid the duplication of calling the command twice with different flags). Now, the loops might come in handy if we expect those list of parameters to change over time. A change would then only require to update the list vs calling the same command again. However, I find that to be very unlikely and would those deem the complexity added by the loop (manual mental unrolling + state variables) to be not worth it. 

On the other hand, the functionality is there and I totally understand if you don't want to invest more time in such a low priority element of the code base :)

I would only have one question: What is the other workflow you mentioned in the commit message for why we need to call `git status` without `--cached`?

D4nte:
> I would only have one question: What is the other workflow you mentioned in the commit message for why we need to call `git status` without `--cached`?

CLion does not stage changes before running pre-hooks so if the pre-hook only call `git diff` with `--cached` then I will not a commit failure for forgetting to format my files.

Also, non-CLion users in the team prefer to have non-staged files checked too.

Considering what we check (formatting) and how this is fixed (running one command for the whole repo) I don't think it makes sense to check staged files but not non-staged files.

D4nte:
Current version seems to work when calling manually but does not stop CLion to do a commit with bad formatting. Need more investigation on my side.

D4nte:
Ok, the issue was not CLion specific, I pushed a fix. `return` only returns for the function. Even with a test failed `exit 0` was executed so the commit did not fail in CLion nor on the command line.

thomaseizinger:
> Considering what we check (formatting) and how this is fixed (running one command for the whole repo) I don't think it makes sense to check staged files but not non-staged files.

Yes , I think that makes sense.

The only reason I suggested `--cached` was because I sometimes had problems with new types of files being checked in one branch and not the other. Those are then untracked but the pre-commit hook would still complain. That is probably a very rare problem though.

tcharding:
> > Considering what we check (formatting) and how this is fixed (running one command for the whole repo) I don't think it makes sense to check staged files but not non-staged files.
> 
> Yes , I think that makes sense.
> 
> The only reason I suggested `--cached` was because I sometimes had problems with new types of files being checked in one branch and not the other. Those are then untracked but the pre-commit hook would still complain. That is probably a very rare problem though.

This is a valid comment, non-tracked files should not be parsed by the pre-commit hook. I'll work on this some more.

tcharding:
> On the other hand, the functionality is there and I totally understand if you don't want to invest more time in such a low priority element of the code base :)

Will fix, simplicity is even more important in code that is not expected to be read much.  Thanks.

tcharding:
Patch ping-pong going on here @D4nte, I changed your exit code back to how I had it (conditional exit instead of direct exit).  Thanks for the `exit` instead of `return` fix though!

.githooks/pre-commit
```diff
@@ -1,4 +1,50 @@
...
+    check_rs_cmd="echo check_rs ran"
+    check_ts_cmd="echo check_ts ran"
+fi
+
+git diff --name-only | grep '.rs$' > /dev/null 2>&1
+if [ $? -eq 0 ]; then
+    do_check_rs=1
+fi
+
+git diff --name-only --cached | grep '.rs$' > /dev/null 2>&1
```
luckysori: :question: This may be obvious, but bear with me :stuck_out_tongue: Why are these `git diff`s done twice, but the second time with the `--cached` flag? Isn't this redundant?

D4nte: `man git` :)

.githooks/pre-commit
```diff
@@ -1,34 +1,50 @@
...
-check_ts=0
+debug=0                         # Set to 1 to enable debugging, 0 to disable.
+do_check_rs=0
+do_check_ts=0
 
-declare -a extra_flags=("--" "--cached")
+check_rs_cmd="cargo make check-rs-format"
+check_ts_cmd="cargo make check-ts-format"
+if (( debug == 1 )); then
+    check_rs_cmd="echo check_rs ran"
```
D4nte: I don't think you want to keep that in here, I guess it's for testing?

D4nte:
> Patch ping-pong going on here @D4nte, I changed your exit code back to how I had it (conditional exit instead of direct exit). Thanks for the `exit` instead of `return` fix though!

Feel free to rebase to remove my discarded code ;)

.githooks/pre-commit
tcharding: `--cached` acts _only_ on staged changes.  Without it acts on unstaged changes (i.e. files that are in the index but not staged).

.githooks/pre-commit
tcharding: We want this in IMO because next time someone wants to modify this script the testing capability is in there ready.  (I added this because of all the iterations so far).

tcharding:
@D4nte epic simplification - check it out.

.githooks/pre-commit
```diff
@@ -1,4 +1,26 @@
 #!/usr/bin/env bash
 
-cargo make check-rs-format && \
-cargo make check-ts-format
+set -e                          # grep failure does not trigger exit.
+
+debug=0                         # Set to 1 to enable debugging, 0 to disable.
```
D4nte:
```suggestion
dry=0
```
It's more of a _dry_ option than a _debug_ one.

tcharding: Implemented as suggested.

PR #1059, opened 2019-06-29
Add automagically-mergify-this label by default

D4nte:
Because if the tests pass and it is approved then there is no
reasons to not let mergify handle it.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -3,6 +3,9 @@ update_configs:
   - package_manager: "rust:cargo"
     directory: "/"
     update_schedule: "weekly"
+    default_labels:
+      - "dependencies"
+      - "automagically-mergify-this"
     allowed_updates:
       - match:
           update_type: "all"


thomaseizinger:
Hopefully only a temporary fix until https://github.com/dependabot/feedback/issues/190 is implemented :)

bonomat:
If I understand this correctly: this adds the default label `automagically-mergify-this` which allows mergify to automatically merge it right? But we still need two approvals for the PR?

D4nte:
> If I understand this correctly: this adds the default label `automagically-mergify-this` which allows mergify to automatically merge it right? But we still need two approvals for the PR?

Yes it just adds the default label but you still need two approvals.

thomaseizinger:
Seems like this is no longer needed. There is checkbox now in the account settings of dependabot: 
Screenshot_20190701-182912

There is one that says "Treat PR approval as request to merge"

D4nte:
> Seems like this is no longer needed. There is checkbox now in the account settings of dependabot:
> 
> There is one that says "Treat PR approval as request to merge"

Hum, I am not sure it is wise to have 2 different bot trying to automerge things in the repo. Especially that mergify queues the PRs for merging.

PR #1069, opened 2019-07-04
Simplify root Cargo.toml through catch-all * syntax

thomaseizinger:
Saw this on another project and thought it was cool.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,18 +1,6 @@
 [workspace]
 
 members = [
-    "application/btsieve",
-    "application/comit_node",
-    "vendor/bam",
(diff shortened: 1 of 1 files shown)


D4nte:
We may want to reconsider using `toml-format` now.

thomaseizinger:
> We may want to reconsider using `toml-format` now.

Ah yeah I remember!
We only dropped it because it formatted the `members` array weirdly right?

D4nte:
> > We may want to reconsider using `toml-format` now.
> 
> Ah yeah I remember!
> We only dropped it because it formatted the `members` array weirdly right?

Yes, it was putting them on one line.

Issue #999, opened 2019-05-29
Set ethereum refund action data field to None


luckysori:
The `data` field in `ethereum::CallContract` is set to `0x` when creating a refund action. This value is deemed "garbage" by @D4nte, and I tend to agree.

https://github.com/comit-network/comit-rs/blob/6afb4f37d5e23d7c9e0d200e732203f71855e56a/application/comit_node/src/swap_protocols/actions.rs#L55-L62

Suggested change:

- Modify `data` field in `ethereum::CallContract` to be `Option<Bytes>` and set it to `None` when creating the refund action.
- Update comit-i story where this is used.

PR #1076, opened 2019-07-10
Use Option for CallContract member 'data'

tcharding:
Currently we use `bytes.default()` to generate an empty value for byte array `data` member of the `CallContract` structure.  This results in the value `0x` being used (I do not know the exact reason why or where this `0x` is found, please see note below on testing).  Rust has options for such a case.

This needs testing, I have no idea what will now be expressed by the `None` in place of the original `0x`.

Fixes: #999

### DIFF
--- a/api_tests/lib/ethereum.ts
+++ b/api_tests/lib/ethereum.ts
@@ -147,7 +147,7 @@ export class EthereumWallet {
 
     public async sendEthTransactionTo(
         to: string,
-        data: string = "0x0",
+        data: string,
         value: BN | string | number = utils.toBN(0),
         gas_limit: string = "0x100000"
     ) {
--- a/comit_node/src/swap_protocols/actions.rs
+++ b/comit_node/src/swap_protocols/actions.rs
@@ -61,3 +61,27 @@ pub mod ethereum {
         pub min_block_timestamp: Option<Timestamp>,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ethereum_support::{web3::types::U256, Address, Network};
+    use std::str::FromStr;
+
+    #[test]
+    fn call_contract_serializes_correctly_to_json_with_none() {
+        let addr = Address::from_str("0A81e8be41b21f651a71aaB1A85c6813b8bBcCf8").unwrap();
+        let contract = ethereum::CallContract {
+            to: addr,
+            data: None,
+            gas_limit: U256::from(1),
+            network: Network::Ropsten,
+            min_block_timestamp: None,
+        };
+        let serialized = serde_json::to_string(&contract).unwrap();
+        assert_eq!(
+            serialized,
+            r#"{"to":"0x0a81e8be41b21f651a71aab1a85c6813b8bbccf8","gas_limit":"0x1","network":"ropsten"}"#,
+        );
+    }
+}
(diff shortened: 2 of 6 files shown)


thomaseizinger:
`CallContract` gets converted to `ActionResponseBody::EthereumCallContract`. 

What do you think @D4nte and @luckysori in terms of the API?
I think we want a test that the `data` field just does not exist if we serialize `ActionResponseBody::EthereumCallContract` to JSON and the `Option` is `None`.

D4nte:
> `CallContract` gets converted to `ActionResponseBody::EthereumCallContract`.
> 
> What do you think @D4nte and @luckysori in terms of the API?
> I think we want a test that the `data` field just does not exist if we serialize `ActionResponseBody::EthereumCallContract` to JSON and the `Option` is `None`.

Yes a serialisation test when `data` is `None` to confirm it's not present in the JSON would be great please.

thomaseizinger:
> > `CallContract` gets converted to `ActionResponseBody::EthereumCallContract`.
> > What do you think @D4nte and @luckysori in terms of the API?
> > I think we want a test that the `data` field just does not exist if we serialize `ActionResponseBody::EthereumCallContract` to JSON and the `Option` is `None`.
> 
> Yes a serialisation test when `data` is `None` to confirm it's not present in the JSON would be great please.

@tcharding Go for that then :)

tcharding:
The enum variant `ActionResponseBody::EthereumCallContract` uses a struct that has fields that mirror the `ethereum::CallContract` struct.  This is duplication of knowledge, we can just use the `ethereum::CallContract` directly in the enum variant instead of a struct.  This allows us to rely on the serialization testing done for `CallContract`.  The bitcoin variants are probably able to be done the same way but its Friday afternoon so this is all for this PR :)  Can be merged without doing the bitcoin variants IMO.

D4nte:
Taking over, I will just remove commit 601aa41391d7360b3cc4a2a54593f9af64353df0

comit_node/src/swap_protocols/actions.rs
```diff
@@ -52,12 +55,57 @@ pub mod ethereum {
...
     pub struct CallContract {
         pub to: Address,
-        pub data: Bytes,
+        pub data: Option<Bytes>,
         pub gas_limit: U256,
         pub network: Network,
         pub min_block_timestamp: Option<Timestamp>,
     }
+
+    impl Serialize for CallContract {
```
thomaseizinger: This implementation should not exist.

comit_node/src/swap_protocols/actions.rs
```diff
@@ -52,12 +55,57 @@ pub mod ethereum {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ethereum_support::{web3::types::U256, Address, Network};
+    use std::str::FromStr;
+
+    #[test]
+    fn call_contract_serializes_correctly_to_json_with_none() {
```
thomaseizinger: This test should be testing what is now actually serialized to JSON, which is `ActionResponsePayload::EthereumCallContract`.

D4nte:
Looks like I messed up the commits by rebasing master after Tobin merged master. Will do fresh PR shortly.

api_tests/lib/ethereum.ts
```diff
@@ -147,7 +147,7 @@ export class EthereumWallet {
 
     public async sendEthTransactionTo(
         to: string,
-        data: string = "0x0",
```
thomaseizinger: This stupid `EthereumWallet` has bitten us so many times now, on `bobtimus` as well as here in the tests.
I think we should extract what we have here into an NPM package and publish it. Basically an idiot-proof `EthereumWallet` that cannot be used wrongly :D

D4nte: hahahahaha, yes I think it's worth considering and to give back to the community.

thomaseizinger: Created ticket here: https://github.com/coblox/bobtimus/issues/78

PR #1078, opened 2019-07-10
Re-name package executables

tcharding:
Currently we name the executables after the program they represent.  Rust convention [1] is to name the main entry point for packages producing executables `main.rs`.  Also main files should be directly under the `src` directory.

This may seem anal but language conventions are useful to have and important to follow because they improve project accessibility and abide by the principle of least surprise.

Re-name package executables from `<package>/src/bin/<package>.rs` to `<package>/src/main.rs`

[1] https://doc.rust-lang.org/cargo/reference/manifest.html


tcharding:
> Fine with me!
> 
> The `/bin` folder is actually also a convention :)
> If you have more than one binary per crate, it allows you to have all of them be named after the `.rs` file instead of mangeling with entries in the `Cargo.toml`

Correct.  However we only have a single binary per package so we don't need bin/.  (A pedantic point)

PR #1079, opened 2019-07-10
Add Cargo.lock to version control

tcharding:
As stated in the comment Cargo.lock should be under version control when a package produces a binary.  We are producing binaries so we should have `Cargo.lock` under version control.  For further information please see the rust-lang [FAQ](https://doc.rust-lang.org/cargo/faq.html?highlight=Cargo.lock,version#why-do-binaries-have--cargolock--in-version-control-but-not-libraries)


D4nte:
Build failed?

I think we removed Cargo.lock because of rocket but now that we do not use rocket we should be fine.

tcharding:
Epic fail.  This PR does the opposite of what it says - current `.gitignore` correctly **includes** `Cargo.lock` in version control.

PR #1073, opened 2019-07-09
Construct BitcoinHtlc based on hex from RFC

thomaseizinger:
This PR changes the BitcoinHtlc to be constructed from the hex code from the RFC with the parameters replaced by offset.

The PR is done apart from the fact that for some reason the `refund` tests are not passing.

Fixes #932.
Fixes #933.

PR #1080, opened 2019-07-11
Remove hard coded paths

tcharding:
We should not use hard coded paths.  This makes assumptions about the operating
system that COMIT runs on.  Rust provides `PathBuf` construction mechanisms so as
to be OS independent, we should use them.

Remove hard coded paths i.e. "path/to/foo" and construct the path using a
`PathBuf`.

### DIFF
--- a/btsieve/src/load_settings.rs
+++ b/btsieve/src/load_settings.rs
@@ -27,7 +27,9 @@ pub fn load_settings(opt: Opt) -> Result<Settings, ConfigError> {
                 "Unable to determine user's home directory".to_string(),
             )),
             Some(dirs) => {
-                let config_file = Path::join(dirs.home_dir(), ".config/comit/btsieve.toml");
+                let user_path_components: PathBuf =
+                    [".config", "comit", "btsieve.toml"].iter().collect();
+                let config_file = Path::join(dirs.home_dir(), user_path_components);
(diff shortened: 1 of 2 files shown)


thomaseizinger:
Would be good to have build on Windows and MacOS to not break this kind of stuff by accident :)

D4nte:
> Would be good to have build on Windows and MacOS to not break this kind of stuff by accident :)

If we remove ZMQ and we do the "no compilation" KR then I think Windows will be for free.

tcharding:
> I think there are more changes to do in the `blockchain_contracts` crate and it would be better to wait for #1073 to be merged first

Are you talking about path changes like this PR or other changes?

thomaseizinger:
> I think there are more changes to do in the `blockchain_contracts` crate and it would be better to wait for #1073 to be merged first

Yeah totally. I was just randomly voicing thoughts, nothing to action now :)

D4nte:
> > I think there are more changes to do in the `blockchain_contracts` crate and it would be better to wait for #1073 to be merged first
> 
> Are you talking about path changes like this PR or other changes?

I am talking about path changes as we read local files to generate the offsets.
Actually, it is all in `blockchain_crates/bin` so it does not depend on #1073.
Please have a look and do comment on the structure of the `blockchain_crates` if it does not match what you learnt recently :)

Issue #1026, opened 2019-06-07
beta-ledger is reported as `NotDeployed` if alpha-ledger is refunded even though it was previously funded


thomaseizinger:
The comit_node REST API returns `NotDeployed` for BetaLedger when AlphaLedger is funded and then refunded. This was observed for Alice (need to check Bob)

1. Alpha is Funded
2. Beta is Funded
3. Refund alpha -> Beta shows as `NotDeployed`, it should show as `Funded`

Beta should be refunded first.

DoD:
- [ ] Create failing test for Alice (checks that beta is `Funded` after alpha refund)
- [ ] Create test for Bob to confirm that beta is `Funded` after alpha refund and ensure refund action is returned
- [ ] Fix the tests

tcharding:
Looks like this issue is invalid.  A PR wast still opened in response to this issue:  #1088.

PR #1088, opened 2019-07-11
Add test where Alice refunds first

tcharding:
While investigating #1026 (closed, found to be invalid) it was discovered that the e2e test runner did not correctly handle running tests due to an asynchronous call.   Commit 1 fixes this issue.  Commit 2 adds a passing test that verifies that Alice can refund first (even though this is _not_ the expected order of refunds).

### DIFF
--- a/api_tests/e2e/rfc003/btc_eth/inverted-refund.ts
+++ b/api_tests/e2e/rfc003/btc_eth/inverted-refund.ts
@@ -84,7 +84,7 @@ declare var global: HarnessGlobal;
         {
             actor: bob,
             test: {
-                description: "Should should see that Alice refunded",
+                description: "Should see that Alice refunded",
                 callback: async body => {
                     let status = body.properties.state.alpha_ledger.status;
 
--- a/api_tests/e2e/rfc003/btc_eth/inverted-refund.ts
+++ b/api_tests/e2e/rfc003/btc_eth/inverted-refund.ts
@@ -1,0 +1,139 @@
+import * as bitcoin from "../../../lib/bitcoin";
+import { Actor } from "../../../lib/actor";
+import { ActionKind, SwapRequest } from "../../../lib/comit";
+import { toBN, toWei } from "web3-utils";
+import { HarnessGlobal } from "../../../lib/util";
+import { Step, createTests } from "../../test_creator";
+import { expect } from "chai";
+import "chai/register-should";
+import "../../../lib/setupChai";
+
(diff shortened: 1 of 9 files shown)


api_tests/e2e/rfc003/btc_eth/inverted-refund.ts
```diff
@@ -0,0 +1,139 @@
...
+            waitUntil: state =>
+                state.alpha_ledger.status === "Funded" &&
+                state.beta_ledger.status === "Funded",
+        },
+        {
+            actor: alice,
+            action: ActionKind.Refund,
+            waitUntil: state => state.alpha_ledger.status === "Refunded",
+        },
+        {
```
thomaseizinger: We have a possible race-condition here that could lead to flacky tests.

In short, it is not guaranteed that Bob's comit node sees the refund at the same time as Alice's node does, however, the last action only does `waitUntil` for Alice.

I think this step should be changed to mimic the step above, e.g.:

```js
{
   actor: bob,
   waitUntil: state => state.alpha_ledger.status === "Refunded",
},
```

tcharding: Yes I agree, thanks.

tcharding: Implemented as suggested.

api_tests/e2e/rfc003/btc_eth/inverted-refund.ts
```diff
@@ -0,0 +1,132 @@
...
+            actor: bob,
+            test: {
+                description: "Should should see that Alice refunded",
```
D4nte:
```suggestion
                description: "Should see that Alice refunded",
```

tcharding: Done, thanks.

tcharding: Re-based on top master, can no longer re-request review by @D4nte.

api_tests/e2e/rfc003/btc_eth/inverted-refund.ts
thomaseizinger: > Implemented as suggested.

Where?
The `step` I made this comment on still asserts that `bob` sees the `alpha_ledger.status` as `Refunded` even though he doesn't wait for it.

thomaseizinger:
I'll take this over to get it merged.

PR #1100, opened 2019-07-16
Add TS Linter and allow e2e tests to be run with yarn

D4nte:
The dev env for e2e tests is actually not great. Being able to run the e2e tests
with yarn allows more flexibility.

Linter and prettier allows better standard for the TypeScript code too.


api_tests/harness.ts
```diff
@@ -204,39 +205,31 @@ function validTestPath(path: string): boolean {
...
 
-function expandPath(
-    paths: string[],
-    parentDir: string = "",
-    depth: number = 5
-): string[] {
-    if (!depth) {
-        return [];
-    }
+function expandGlob(paths: string[]): string[] {
```
thomaseizinger: Honestly, I'd vote for removing this "feature" because I think it is quite unexpected. It is not like you are typing it everyday and we have aliases in cargo-make anyway.

thomaseizinger: I think we are better of with less complexity in here compared to a (for now) slightly more inconvenient way of running several tests.
Also, you made it much better already with this PR!

D4nte: This feature is useful to me. I don't think these 10 lines of code will stop anyone to update the code. It's not even that complex.

thomaseizinger: Well, if you insist on it I can't stop you anyway.

I am still surprised that you even took the time to write it given the marginal utility it provides over just using plain glob patterns.

api_tests/lib/ethereum.ts
```diff
@@ -1,14 +1,14 @@
...
-import * as util from "./util";
-import { HttpProvider } from "web3-providers";
-import * as utils from "web3-utils";
 import BN from "bn.js";
-import * as fs from "fs";
 import EthereumTx = require("ethereumjs-tx");
+import * as fs from "fs";
+import { HttpProvider } from "web3-providers";
+import * as utils from "web3-utils";
+import * as util from "./util";
```
bonomat: Is this new line here on purpose?
```suggestion
```

PR #1101, opened 2019-07-16
Return 400 when fee is too high

### LINKED ISSUE #1000: Cannot redeem bitcoin swaps with very small amounts
I just experienced an error when the comit-node is not able to generate a redeem transaction for bitcoin. 

How to reproduce:
1) Swap request from Alice to Bob with 10 satoshi to 10 wei
2) Continue normally with accepting everything and funding everything
3) ...
4) When Bob tries to receive the redeem transaction, the comit node response with an error: `(node:14698) UnhandledPromiseRejectionWarning: RequestError: Error: socket hang up` 

comit_node log: 
```
error message: 
thread 'tokio-runtime-worker-0' panicked at 'attempt to subtract with overflow', /Users/bonomat/.cargo/git/checkouts/bitcoin-quantity-be308e6dbd3e534b/d5e9726/src/lib.rs:47:25
```

Assumption:
When calling the refund action on the comit-node, we try to create the redeem transaction. This involves subtracting a transaction fee from the HTLC amount. Since the HTLC amount was only 10 satoshi and the fee higher, the operation failed. 

----

Solution: Return a 400 error (with a message) if the calculated fee is higher than the UTXO value on GET refund/redeem for bitcoin.

Use the "new" amounts module from rust-bitcoin to replace our custom calculations.

DoD:
- [ ] comit-rs does not depends on bitcoin-quantity repo
- [ ] comit_node http API returns 400 "fee too high" if the calculated fee is higher than the HTLC value
- [ ] bitcoin-quantity repository is archived

D4nte:
Resolves #1000

### DIFF
--- a/api_tests/e2e/rfc003/eth_btc/bitcoin_high_fee.ts
+++ b/api_tests/e2e/rfc003/eth_btc/bitcoin_high_fee.ts
@@ -20,6 +20,8 @@ declare var global: HarnessGlobal;
     const bob = new Actor("bob", global.config, global.project_root, {
         ethereumNodeConfig: global.ledgers_config.ethereum,
         bitcoinNodeConfig: global.ledgers_config.bitcoin,
+        addressForIncomingBitcoinPayments:
+            "bcrt1qs2aderg3whgu0m8uadn6dwxjf7j3wx97kk2qqtrum89pmfcxknhsf89pj0",
     });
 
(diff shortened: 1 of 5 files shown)


thomaseizinger:
I removed the reviews since it is not yet ready to review and waiting for another PR.

PR #1125, opened 2019-07-22
Remove executable permissions

tcharding:
There are various files in the project that have executable permissions set.  Following the security principle of least permission for job required a file should not have permissions that it does not need to function.


D4nte:
Please remove exec bit on all files including those under api_tests

thomaseizinger:
> Please remove exec bit on all files including those under api_tests

Be careful on those inside `node_modules` though. The subfolder `.bin` for example needs them and they often point to files inside the modules. I would not touch anything inside `api_tests/node_modules`.

D4nte:
> > Please remove exec bit on all files including those under api_tests
> 
> Be careful on those inside `node_modules` though. The subfolder `.bin` for example needs them and they often point to files inside the modules. I would not touch anything inside `api_tests/node_modules`.

Hopefully `api_tests/node_modules` is in `.gitignore` but yeah, don't break your local install @tcharding

D4nte:
Not sure why my review is still required, I'd expect to see the `harness.ts` in there please (and any other api_tests files).

thomaseizinger:
> Not sure why my review is still required, I'd expect to see the `harness.ts` in there please (and any other api_tests files).

Why should the `harness.ts` not(!) be executable? I thought that was the whole point when it was initially created?

tcharding:
For future reference; using this command to find all files with executable permissions `find . -perm /+x -type f | grep -v git | grep -v node_modules`

D4nte:
> > Not sure why my review is still required, I'd expect to see the `harness.ts` in there please (and any other api_tests files).
> 
> Why should the `harness.ts` not(!) be executable? I thought that was the whole point when it was initially created?

It's changed now :) `harness.ts` should not be executed from the shell anymore but by using `yarn`.

PR #1126, opened 2019-07-23
Fix btsieve-client network bug

bonomat:
Fixes the bug that btsieve client tried to get queries for bitcoin fro
`"queries/bitcoin/{}/transactions", ethereum_network` and ethereum from 
`"queries/ethereum/{}/transactions", bitcoin_network`. 

This wasn't an issue for regtest, but becomes critical for ropsten, i.e. we just had : get bitcoin transaction from `queries/bitcoin/ropsten/transactions` ... which obviously didn't work

### DIFF
--- a/comit_node/src/btsieve/client.rs
+++ b/comit_node/src/btsieve/client.rs
@@ -55,10 +55,10 @@ struct QueryResponse<T> {
 impl BtsieveHttpClient {
     pub fn new(
         endpoint: &Url,
-        ethereum_poll_interval: Duration,
-        ethereum_network: &str,
         bitcoin_poll_interval: Duration,
-        bitcoin_network: &str,
+        bitcoin_network: bitcoin_support::Network,
+        ethereum_poll_interval: Duration,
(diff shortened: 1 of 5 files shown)


comit_node/src/btsieve/client.rs
```diff
@@ -55,10 +55,10 @@ struct QueryResponse<T> {
 impl BtsieveHttpClient {
     pub fn new(
         endpoint: &Url,
-        ethereum_poll_interval: Duration,
-        ethereum_network: &str,
         bitcoin_poll_interval: Duration,
-        bitcoin_network: &str,
```
D4nte: 📣 I would suggest to type the `bitcoin_network` and `ethereum_network` with an enum to avoid this bug in the future.

bonomat: That's exactly what this PR was about ;)

D4nte: somehow I missed that indeed

PR #1128, opened 2019-07-23
Revert "Second fix of network serialization for serde"

bonomat:
Reverts comit-network/comit-rs#1127

Because comit-i is not yet updated and we don't want to start the breaking of apis just now. 

Hence, start with updating the RFCs: https://github.com/comit-network/RFCs/issues/101
and continue with creating sub-sequential tickets for cnd and comit-i

### DIFF
--- a/vendor/bitcoin_support/src/network.rs
+++ b/vendor/bitcoin_support/src/network.rs
@@ -18,19 +18,19 @@ use serde::{Deserialize, Serialize};
 #[serde(rename_all = "lowercase")]
 pub enum Network {
     #[strum(serialize = "main")]
-    Main,
+    Mainnet,
     #[strum(serialize = "regtest")]
     Regtest,
     #[strum(serialize = "test")]
-    Test,
+    Testnet,
 }
 
 impl From<bitcoin::network::constants::Network> for Network {
     fn from(item: bitcoin::network::constants::Network) -> Network {
         match item {
-            bitcoin::network::constants::Network::Bitcoin => Network::Main,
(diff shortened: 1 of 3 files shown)


D4nte:
Please note I think it's fine to return `mainnet` on the cnd HTTP API. Bobtimus can handle conversion to `main` when talking with bitcoind.
All this is very bitcoind specific. This is part of the JSON RPC and not the consensus. It is ok for us to make up our own mind.

e.g. btcd uses neither: https://github.com/btcsuite/btcd/blob/master/docs/json_rpc_api.md#getcurrentnet

thomaseizinger:
> Please note I think it's fine to return `mainnet` on the cnd HTTP API. Bobtimus can handle conversion to `main` when talking with bitcoind.
> All this is very bitcoind specific. This is part of the JSON RPC and not the consensus. It is ok for us to make up our own mind.

That is a good point actually. Should use whatever the consensus encoding is?

thomaseizinger:
> e.g. btcd uses neither: [btcsuite/btcd:docs/json_rpc_api.md@`master`#getcurrentnet](https://github.com/btcsuite/btcd/blob/master/docs/json_rpc_api.md#getcurrentnet) [](/btcsuite/btcd/blob/HEAD@{2019-07-23T07:12:11Z}/docs/json_rpc_api.md#getcurrentnet)

Just checked it. Those numbers are actually the values used on the consensus layer.

image

Those hex values, when represented in decimal, are what `btcd` returns. I am sold. I think similar to Ethereum, we should use whatever is used on the consensus layer.

PR #1150, opened 2019-07-29
FIX CIRCLE-CI: Run `restore_caches` and `setup_rust` steps before `setup_system`

thomaseizinger:
We want to delay all the `apt` stuff as long as possible so that we don't run into issues with other parts of the system having a lock on it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,10 +9,10 @@ jobs:
       RUST_TEST_THREADS: "8"
     steps:
       - checkout
-      - setup_system
-      - print_current_versions
       - restore_caches
       - setup_rust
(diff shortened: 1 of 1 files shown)


thomaseizinger:
@D4nte please check this as it is blocking all the other PRs from passing.

thomaseizinger:
> Realistically I am approving blind because I do not know what was conflicting with the `apt` calls in setup_system. What steps get a lock on `apt` except for our custom `setup_system`.

`systemd` is apparently doing all kinds of stuff when the VM first gets created:
 Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)

Issue #1102, opened 2019-07-16
Implement structured logging


bonomat:
We want the possibility to emit structured logs from comit-rs. At the same time, the default log output should be lines to make debugging easier. Structured logging is only useful for environments that need monitoring.

This is to be implemented as follows:

- Change the config table `log_levels` to `logging`
- Add an optional boolean `structured` that defaults to false
- rename the field `comit_node` under `log_levels` to `level`
- Inside `logging.rs`, if `structured` is true, emit a JSON object instead of a single line

The keys of the JSON object should be the one from the `log::Record` (line, path, etc).
We will use it together with `fluentd`, so if there is a recommended format, check it out to see if it makes our life easier if we slightly adapt the JSON that we are emitting.

PR #1151, opened 2019-07-29
Remove side-effect from `load_settings` and test thoroughly

thomaseizinger:
While doing #1102, I realized that the current tests for `load_settings` actually have side-effects on the system they are running on.

In preparation to reworking the settings this PR:

- Removes this side-effect from those tests (they are actually re-written and now reside in `tests/`)
- Makes the settings parsing actually testable by bubbling up the dependency on `Rng` of `Seed`
- Actually makes the log messages during the setup visible by converting them to `println`

Reviewers:
- Tagging @bonomat because he wrote that code initially
- @D4nte and @tcharding are the main reviewers


cnd/Cargo.toml
```diff
@@ -45,6 +45,7 @@ state_machine_future = { version = "0.2" }
 structopt = "0.2"
 strum = "0.15"
 strum_macros = "0.15"
+tempfile = "3.1.0"
```
D4nte: 📣  This can go in dev-dependencies.

Issue #1012, opened 2019-06-04
Change the listening port of the comit protocol to a better one


D4nte:
The default port for comit_node is `8011` which is just "can't use 8000 so let's add 10 and then 1".

https://github.com/comit-network/comit-rs/blob/ea24d1747f2c706c69780c4f5cf35bc2c7063a3c/application/comit_node/src/settings/mod.rs#L32

Change the default port to `9939` which is the value for the utf-8 "chains" emoji: ⛓ 
Reason: COMIT is here to facilitate cross-chains applications.

Tasks:
- Change the default config (`Settings::default()` impl)
- Change the config files for the e2e
- Open issue on the RFC repository for adapting RFC001

tcharding:
> Open issue on the RFC repository for adapting RFC001

RFC001 does not appear to mention 8011 or the word 'port'.  Does the RFC really need changing?  Please confirm @D4nte.

tcharding:
Whole RFC directory also contains 0 mentions of 8011
```
cd github.com/comit-network/RFCs
git grep 8011
```

D4nte:
> > Open issue on the RFC repository for adapting RFC001
> 
> RFC001 does not appear to mention 8011 or the word 'port'. Does the RFC really need changing? Please confirm @D4nte.

Yes, 8011 was a random port, now we say that 9939 is the default port.

PR #1158, opened 2019-07-31
Use port 9939 for cnd

tcharding:
The default port for comit_node is 8011 which is just "can't use 8000 so let's
add 10 and then 1".

Change the default port to 9939 which is the value for the utf-8 "chains" emoji:
chains Reason: COMIT is here to facilitate cross-chains applications.

Fixes: #1012


tcharding:
> Fine with me but the issue also states to open an issue on the RFC repository, please don't forget about that!

Damn, I forgot that - there goes my good-feeling.

PR #1159, opened 2019-08-01
Allow users to configure structured logging which outputs all log messages in JSON format

### LINKED ISSUE #1102: Implement structured logging

thomaseizinger:
We removed the log statement about the settings from `main.rs`, please have a look at the relevant commit here ec587fc76e5adb44bcd4be642178773355a8cdb6 for an explanation.

Fixes #1102.


cnd/Cargo.toml
```diff
@@ -63,6 +63,7 @@ memsocket = "0.1"
 pretty_env_logger = "0.3"
 quickcheck = "0.8.5"
 quickcheck_macros = "0.8.0"
+regex = "1.2.0"
```
D4nte: Didn't we say we prefer no to specify the patch version?

thomaseizinger: Hmm, that was pre-dependabot though. Does it still matter now?

D4nte: I don't know, you're the one who came up with the previous rule...

cnd/src/logging/initialize.rs
```diff
@@ -0,0 +1,214 @@
...
+/// 1. `out.finish` only accepts `std::fmt::Arguments` and hence we can only use
+/// `format_args!`
+///
+/// 2. (1) is very likely due to the fact that we don't want to
+/// unnecessarily allocate strings in this function for calls to `Log::log` with
+/// a level that is not even active (think about all the calls to `log::trace`
+/// in all our depenencies that we don't see)
+///
+/// Note that this formatter **CANNOT** handle multiline message as for example
+/// produced by: `log::info!("{:#?}", my_struct);`
```
D4nte: So what's the plan? How do we handle this scenario?

thomaseizinger: I'd propose to just not use these log statements :)

D4nte: Does the compilation fail if we were to use them?

PR #1169, opened 2019-08-05
Detect missing block for bitcoin

bonomat:
This is the first step towards lite clients.
I kept the change as small as possible, so this PR introduces: 

* `Bitcoin` and `Ethereum`-chain in https://github.com/comit-network/comit-rs/blob/a32b6cbf904e68c6bec8dad856fc6a24c98ff763/btsieve/src/lib.rs#L58-L60
* `BlockchainDAG` data-structure to model the blockchain as a _Directed Acylic Graph_ https://github.com/comit-network/comit-rs/blob/a32b6cbf904e68c6bec8dad856fc6a24c98ff763/btsieve/src/lib.rs#L44
* and an implementation to store incoming blocks and check for missed blocks in: https://github.com/comit-network/comit-rs/blob/a32b6cbf904e68c6bec8dad856fc6a24c98ff763/btsieve/src/bitcoin/mod.rs#L12-L43

--> The idea is to add the same abstraction for Ethereum next (TODO create new ticket) and then update the way of how blocks are being processed.

### DIFF
--- a/btsieve/src/bitcoin/mod.rs
+++ b/btsieve/src/bitcoin/mod.rs
@@ -6,27 +6,34 @@ pub use self::{
     block_processor::{check_block_queries, check_transaction_queries},
     queries::{BlockQuery, TransactionQuery},
 };
-use crate::{AddBlocks, Bitcoin};
-use bitcoin_support::MinedBlock;
+use crate::{AddBlock, Bitcoin};
+use bitcoin_support::{MinedBlock, Sha256dHash};
 
-impl AddBlocks<MinedBlock> for Bitcoin {
+impl AddBlock<MinedBlock> for Bitcoin {
     fn add_block(&mut self, block: MinedBlock) {
-        if self.contains_precessor(&block) {
-            log::warn!("Could not find previous block of {:?} ", block);
-        }
         if self.0.nodes.contains(&block) {
-            log::warn!("Block already known {:?} ", block);
-        } else {
-            self.0.nodes.push(block);
+            return log::warn!("Block already known {:?} ", block);
         }
+        match self.find_predecessor(&block) {
+            Some(prev) => {
+                self.0.vertices.push((
+                    prev.block.header.merkle_root,
+                    block.clone().block.header.merkle_root,
+                ));
+            }
+            None => {
+                log::warn!("Could not find previous block for {:?} ", block);
+            }
+        }
+        self.0.nodes.push(block);
     }
 
     fn size(&self) -> usize {
         self.0.nodes.len()
     }
 
-    fn contains_precessor(&self, block: &MinedBlock) -> bool {
-        self.0.nodes.iter().any(|b| {
+    fn find_predecessor(&self, block: &MinedBlock) -> Option<&MinedBlock> {
+        self.0.nodes.iter().find(|b| {
             b.block
                 .header
                 .merkle_root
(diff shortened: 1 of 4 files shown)


btsieve/src/bitcoin/mod.rs
```diff
@@ -6,3 +6,161 @@ pub use self::{
...
+use bitcoin_support::{MinedBlock, Sha256dHash};
+
+impl AddBlock<MinedBlock> for Bitcoin {
+    fn add_block(&mut self, block: MinedBlock) {
+        if self.0.nodes.contains(&block) {
+            return log::warn!("Block already known {:?} ", block);
+        }
+        match self.find_predecessor(&block) {
+            Some(prev) => {
+                self.0.vertices.push((
```
D4nte: You went further than needed, the storage of links between blocks is out of scope :)

btsieve/src/lib.rs
```diff
@@ -38,3 +40,27 @@ impl PartialOrd for QueryMatch {
         self.0.partial_cmp(&other.0)
     }
 }
+
+struct BlockchainDAG<T, V> {
```
D4nte: The Blockchain DAG is not needed for this PR :)

bonomat: I know, but otherwise this would have been too easy :D

btsieve/src/bitcoin/mod.rs
```diff
@@ -6,3 +6,161 @@ pub use self::{
...
+                .merkle_root
+                .eq(&block.block.header.prev_blockhash)
+        })
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::BlockchainDAG;
```
tcharding: In line with Rust convention can we name this `BlockchainDag` please.  (Acronyms not capitalized, source: @luckysori)

btsieve/src/bitcoin/mod.rs
```diff
@@ -6,3 +6,161 @@ pub use self::{
...
+            .unwrap(),
+            Sha256dHash::from_hex(
+                "0000000000000000000000000000000000000000000000000000000000000002",
+            )
+            .unwrap(),
+            1,
+        );
+
+        assert_that(&bitcoin_chain.size()).is_equal_to(&0);
+        bitcoin_chain.add_block(block);
```
tcharding: Perhaps another assert statement here `assert_that(&bitcoin_chain.size()).is_equal_to(&1);`

btsieve/src/lib.rs
```diff
@@ -38,3 +40,27 @@ impl PartialOrd for QueryMatch {
...
+            vertices: Vec::new(),
+        })
+    }
+}
+
+struct Bitcoin(BlockchainDAG<MinedBlock, (Sha256dHash, Sha256dHash)>);
+
+struct Ethereum;
+
+trait AddBlock<T> {
```
tcharding: I do not know typical Rust naming for traits but this name does not seem to match the trait functions. Suggest `Blockchain`.

tcharding: Also perhaps `default` should be in the trait (if that is how Rust likes it)?

btsieve/src/bitcoin/mod.rs
bonomat: Mhmm... must have deleted that one :D

btsieve/src/lib.rs
bonomat: True, at the beginning there was only one function --> I'll update it.

PR #1171, opened 2019-08-05
libp2p_bam: Add comments on enum variants

tcharding:
Recently the substream enum was separated into two separate enums.  The enum
variant comments got lost during the process.  Since these enums represent state
transitions some extra information by way of variant comments is useful.

Add the original variant comments back in to better describe meaning of each
state represented by individual enum variants.

Fixes comit ffb816ca ("Separate SubstreamState enum into two enums")


tcharding:
> Why did the formatting change if you just add a comment?

No idea, I just let my formatter handle the formatting :)

PR #1174, opened 2019-08-07
Track blocks for ethereum

### LINKED ISSUE #1173: Detect missing blocks for ethereum
Similar to #1169 we should track the incoming blocks in ethereum as well.

bonomat:
Resolves #1173


bonomat:
Ignore for now please, need to investigate this build failure.

PR #1189, opened 2019-08-12
Add test-harness for testing substream state transitions

thomaseizinger:
This PR adds a test harness for setting up a substream that can be used in unit tests to verify the transition behaviour of the substream states.

It also adds two sample tests which
a) showcase the usage
b) test useful stuff that is currently only covered implicitly.


thomaseizinger:
> The code in this PR is hot, good effort. The individual patches are a bit confusing.
> 
>     * Why is the showcase test added in the same commit as the test harness? It was confusing to me to see the `derivative` stuff in `outbound.rs` in this commit
> 
>     * Why is there a bug fix in `outbound.rs` in the second patch which adds a unit test to inbound.rs?

@tcharding Feedback addressed by rewriting those tests to use the `ProtocolsHandler` implementation of `BamHandler`.
This should:
a) be more stable because it is the libp2p interface and doesn't depend on the implementation details (our `State`s)
b) removes the need for the `derivative` stuff

cnd/src/libp2p_bam/substream/outbound.rs
```diff
@@ -79,10 +79,11 @@ impl<TSubstream: AsyncRead + AsyncWrite> Advance for State<TSubstream> {
                 response_sender,
                 mut stream,
             } => match stream.poll_complete() {
-                Ok(Async::Ready(_)) => Advanced::transition_to(WaitingAnswer {
+                Ok(Async::Ready(_)) => WaitingAnswer {
                     response_sender,
                     stream,
-                }),
+                }
+                .advance(&known_headers),
```
tcharding: Thanks for splitting the PR up into clear patches.  The whole PR looks clean to me now except for this change.  I cannot see why it is in this patch?  On a higher level, I can't work out why we call `.advance()` for some state transitions in this function but not for others (just returning an`Advanced{...}`.  If, and only if, you have time perhaps you could explain it to me.  Thanks.

FTR I removed the change and can see that the tests hang without it, so that explains *why* its in the patch but not *why* it is needed. In outbound.rs the implementation of `advance` the `WaitingAnswer` match arm we return
```
                    Advanced {
                        new_state: Some(WaitingClose { stream }),
                        event: Some(ProtocolsHandlerEvent::Custom(event)),
                    }
```
How does this advance to `WaitingClose` without a call to `advance()`?

thomaseizinger: To fully explain it, I'd need a whiteboard :D
It is a combination of:
- how `ProtocolsHandler::poll` is implemented (returns `NotReady` if nothing else is available to return)
- `tokio` will not poll your futures again if no underlying future registered a `Task` (like an IO interrupt) that notifies tokio that something changed

Recursing at this particular stage makes us `poll` the underlying stream _at least once_ which registers a `Task` that makes tokio poll the whole thing again.
Without this recursion, the test gets stuck because nothing notifies tokio that the incoming data via IO means we want to be polled again.

Issue #845, opened 2019-04-04
Improve homogenous naming in comit_node http api


D4nte:
~In the HTTP API response we have:~
~- Communication status: `SCREAMING_SNAKE_CASE` (e.g. `ACCEPTED`)~
~- Asset status: `PascalCase` (e.g. `Funded`)~
~- Most field names: `snake_case` (e.g `alpha_asset`)~
~- Action types: `kebab-case` (e.g. `ethereum-invoke-contract`)~

~This is messy I'd suggest to have everything one way: `kebab-case`.~

~Also, may want to consider the GET parameters (e.g. `fee_per_bytes`).~

Change `ledger status` to `SCREAMING_SNAKE_CASE`.

thomaseizinger:
The only inconsistency I can see is in the naming of the status fields. Those should definitely be unified. To me, SCREAMING_SNAKE_CASE is the most appealing for fields that indicate a kind of status.

`snake_case` field names are per Rust convention. We could think of serializing those fields as `camelCase`, which would be more idiomatic for JSON. After all, API consumers should not care which language we are using to develop our COMIT node.

> Action types: kebab-case (e.g. ethereum-invoke-contract)

Changing those I am not so sure about, all other casing looks kinda off IMO.

D4nte:
I love how your comment agrees with this issue while looking like you disagree with the issue.

thomaseizinger:
> I love how your comment agrees with this issue while looking like you disagree with the issue.

😅
I wasn't sure if you were implying that all the different uses of cases should be aligned. It sounded like that, so I expressed my disagreement with that.
I am glad that it just seemed to be a misunderstanding :)

PR #1190, opened 2019-08-12
845 Improve homogenous HTTP API naming

tcharding:
As described in the issue; the ledger status currently uses Pascal case (upper camel case).  This is not uniform with other fields in the HTTP API.  We should use SCREAMING_SNAKE_CASE so as to be uniform.

This PR also adds the recently added status field `IncorrectlyFunded` to the API test scheme (seems to have been omitted when added to cnd).  Next we change the API test scheme to use SCREAMING_SNAKE_CASE, making the dry tests fail.  We add a failing unit test.  Then in patch 4 we fix cnd to use SCREAMING_SNAKE_CASE.  Patch 5 updates the e2e tests to reflect the case change.

Fixes: #845 and feels good!

### DIFF
--- a/api_tests/e2e/rfc003/btc_eth-erc20/happy.ts
+++ b/api_tests/e2e/rfc003/btc_eth-erc20/happy.ts
@@ -93,22 +93,22 @@ declare var global: HarnessGlobal;
         {
             actor: alice,
             action: ActionKind.Fund,
-            waitUntil: state => state.alpha_ledger.status === "Funded",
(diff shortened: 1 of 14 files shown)


tcharding:
This work led to an update to the `strum` `README.md` file documenting the currently undocumented feature of `strum_discriminants` that @D4nte worked out. PR is yet to be raised, waiting on write access to `coblox/strum` repository please @bonomat.

tcharding:
> have another look once the build is fixed.

Build is fixed @bonomat , my bad.  Thanks.

PR #1192, opened 2019-08-13
Introduce `BlockSource` and implement it for `Web3<Http>`

### LINKED ISSUE #1132: Replace web3-polling with `BlockSource` implementation
- Introduce `BlockSource` abstraction: a trait that allows you to 
   - fetch block based on their hash
   - get the latest block hash

Pre-work:
As part of implementing this, make sure you have a good understanding of how full nodes exchange blocks and what the best API is to fetch new blocks.

Goals of the implementation:
1. we no longer use `client.eth_filter().create_blocks_filter()` but `web3.getBlock()` to get blocks
2. continuously fetch the latest block through the `BlockSource` and add it to the "block storage" introduced in #1173.

thomaseizinger:
Resolves #1132.


btsieve/src/main.rs
```diff
@@ -201,48 +202,42 @@ fn create_ethereum_routes(
...
 
             let executor = runtime.executor();
-            let web3_processor = blocks.for_each(move |block| {
-                ethereum_chain.add_block(block.clone());
-                ethereum::check_block_queries(block_query_repository.clone(), block.clone())
-                    .for_each(|QueryMatch(id, block_id)| {
-                        block_query_result_repository.add_result(id.0, block_id);
+            let web3_processor = block_source
+                .blocks()
+                .map_err(|e| log::warn!(target: "ethereum", "error fetching latest block {:?}", e))
```
D4nte: The implication of this design is that if the block_source error bubbles up then it becomes irrecoverable as the polling would stop (I think).

It may be interesting to point that out in the log or in the Trait implementation.

btsieve/src/ethereum/web3_http_blocksource.rs
```diff
@@ -0,0 +1,60 @@
...
+        };
+
+        log::info!(target: "ethereum::blocksource", "polling for new blocks on {} every {} seconds", self.network, poll_interval);
+
+        let stream = Interval::new_interval(Duration::from_secs(poll_interval))
+            .map_err(Error::Timer)
+            .and_then(move |_| {
+                web.eth().block_with_txs(BlockId::Number(BlockNumber::Latest)).map_err(Error::Source)
+            })
+            .filter_map(|maybe_block| maybe_block)
```
luckysori: :upside_down_face: Super nitpick but just to make sure that I understand things properly:

```suggestion
            .filter_map(|maybe_block| block)
```

Would that be more correct?

bonomat: Well, this would not compile :P 
But yes, `filter_map` filters out the `Option::None` elements in this case.

luckysori: Oh lol. Fixed my original suggestion :stuck_out_tongue:

thomaseizinger: Well, the datatype is a `Option<Block>` and we `filter_map` so we make a `Block` out of an `Option<Block`, hence the name `maybe_block`. Makes sense? :)

luckysori: It makes sense, I had it wrong.

btsieve/src/main.rs
```diff
@@ -201,48 +202,42 @@ fn create_ethereum_routes(
...
+                .for_each(move |block| {
+                    ethereum_chain.add_block(block.clone());
```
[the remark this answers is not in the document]
thomaseizinger: Don't really think so, I added a documentation note about it.

btsieve/src/main.rs
thomaseizinger: Not bubbeling up IO errors now and added a test that verifies that.

PR #1196, opened 2019-08-14
Replace configuration option for poll interval with sophisticated algorithm

### LINKED ISSUE #1191: Automatically determine feasible poll interval based on connected network
In order to create a useful abstraction in https://github.com/comit-network/comit-rs/issues/1132, I would like to propose to remove the "poll interval" configuration parameter from the btsieve configuration and instead replace it with an algorithm that determines a reasonable poll interval based on the average blocktime of the connected network.

The current implementation requires the caller to pass in a "ticker" that will trigger a new request to fetch the latest block: 

https://github.com/comit-network/comit-rs/pull/1192/files#diff-713ff15f4b79a4a5f85bd0b190deb65fR18-R21

However, that abstraction does not really make sense. The underlying client can already talk to its source and determine a feasible interval based on the connected network.

thomaseizinger:
Fixes #1191.


thomaseizinger:
> Nit: commit subject is a bit long. Also AFAICT this PR only relates to Ethereum but that is not mentioned anywhere in the PR or commit message.

Addressed!

btsieve/src/main.rs
```diff
@@ -201,8 +201,16 @@ fn create_ethereum_routes(
 
             let web3_client = web3_client.clone();
 
-            let blocks = ethereum_block_listener(web3_client.clone(), settings.poll_interval_secs)
-                .expect("Should return a Web3 block poller");
+            let blocks = ethereum_block_listener(
+                web3_client.clone(),
+                Duration::from_secs(match network {
+                    EthereumNetwork::Mainnet => 5,
```
luckysori: :question: Any reason why the Mainnet value went from 17 to 5?

D4nte: I think 5 is better :)

Issue #1139, opened 2019-07-26
Fix unused variable warning


bonomat:
Unused variable warning:

```
warning: unused variable: `t`
  --> vendor/bitcoin_support/src/network.rs:74:35
   |
74 |     fn assert_display<T: Display>(t: T) {}
   |                                   ^ help: consider prefixing with an underscore: `_t`
   |
   = note: #[warn(unused_variables)] on by default

warning: unused variable: `t`
  --> vendor/ethereum_support/src/network.rs:74:35
   |
74 |     fn assert_display<T: Display>(t: T) {}
   |                                   ^ help: consider prefixing with an underscore: `_t`
   |
   = note: #[warn(unused_variables)] on by default

```

Tasks:
- Investigate why clippy didn't fail the build (probably because this is on `#[cfg(test)]`)
- Fix the warnings

D4nte:
Also investigate why it does not fail the build

PR #1197, opened 2019-08-14
Remove bitcoin_support dependency from blockchain_contracts

D4nte:
Related to coblox/blockchain_contracts#4

Note: Issue to be open to contribute the TODOs back, see coblox/blockchain_contracts#4.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -431,6 +431,8 @@ dependencies = [
 name = "blockchain_contracts"
 version = "0.1.0"
 dependencies = [
+ "bitcoin 0.18.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "bitcoin_hashes 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "bitcoin_rpc_test_helpers 0.1.0",
  "bitcoin_support 0.1.0",
  "bitcoin_witness 0.1.0",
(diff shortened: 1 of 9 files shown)


vendor/bitcoin_support/src/network.rs
```diff
@@ -71,7 +71,7 @@ mod test {
         assert_eq!(testnet, "test");
     }
 
-    fn assert_display<T: Display>(t: T) {}
+    fn assert_display<T: Display>(_t: T) {}
```
luckysori: :mega: I guess you shouldn't fix this unless you want to do #1139 too.

vendor/ethereum_support/src/network.rs
```diff
@@ -71,7 +71,7 @@ mod test {
         );
     }
 
-    fn assert_display<T: Display>(t: T) {}
+    fn assert_display<T: Display>(_t: T) {}
```
luckysori: :mega: How dare you fix Clippy warnings?!!

vendor/bitcoin_support/src/network.rs
thomaseizinger: It sounds like a nitpick but the ticket also includes to investigate why clippy doesn't fail this so it should probably not be done as part of this :)

D4nte:
> Might be confusing to fix those warnings here without dealing with #1139 in full, but approved nevertheless.

Good point, I'll trash this commit, rebase and merge.

PR #1200, opened 2019-08-14
Rename vendor crate to internal

### LINKED ISSUE #1160: Rename the `vendor` directory
"Vendoring" actually has a very distinct meaning and is totally unrelated to what we are doing with the `vendor` folder: https://doc.rust-lang.org/nightly/cargo/commands/cargo-vendor.html

We should rename our folder to `internal` to something else that makes more sense :)

luckysori:
Fixes #1160.


luckysori:
> Checking out this branch left `vendor/` in `comit-rs/`. Did you do `git mv vendor internal`? Otherwise each dev is going to have to notice this patch has merged and manually remove `vendor/`.

I could be wrong but I think what I did is equivalent to `git mv vendor internal`, see https://stackoverflow.com/a/1094392. If you know better, which you probably do, feel free to take over. I just tried to use `git mv` and swap between master and the branch and these untracked folders still sometimes show up.

luckysori:
Closing in favour of #1211.

PR #1201, opened 2019-08-14
Remove ethereum_support dependency from blockchain_contracts

D4nte:
Relates to coblox/blockchain-contracts#4


D4nte:
> Since it is only used in the tests, you may want to extract those `U256` instances into constants declared somewhere globally/statically so that it is more readable of how much it is.

I quickly tried to put the `U256` in a constant but it's not straightforward so proceeded forward without it.

PR #1211, opened 2019-08-15
Rename vendor crate to internal

### LINKED ISSUE #1160: Rename the `vendor` directory

luckysori:
Merging #1200 with `master` after #1024 was merged was quite hard so I have just created this one instead. I ran `git mv vendor internal` as @tcharding suggested too :smiley:

Fix #1160.


tcharding:
> Merging #1200 with `master` after #1024 was merged was quite hard so I have just created this one instead.

Next time please, when/if you get into a mess like this, just reset your local branch, do the changes then force push so we only have one PR.  Thanks

thomaseizinger:
Superseded by https://github.com/comit-network/comit-rs/pull/1214.

PR #1233, opened 2019-08-19
Upgrade libp2p to v0.12

thomaseizinger:
To achieve that, we also needed to upgrade multistream-select which we only use for the tests.
Unfortunately, I also had to run `cargo update` to resolve a compile
error inside a libp2p dependency. That in turn broke the compilation of some downstream dependency of warp, which is why I disabled the websocket feature for warp because we are no using it anyway.


D4nte:
Did you open an issue on the downstream dep?

thomaseizinger:
> Did you open an issue on the downstream dep?

Nope, couldn't be bothered to be honest :D 

I should have probably ...

Issue #1074, opened 2019-07-09
Store config file in a path based on platform conventions


thomaseizinger:
During the workshop, I received the feedback that we are statically assuming the user's desired folder for config files is `.config` whereas at least under (most?) Linux distributions, that is controlled through the `$XDG_CONFIG_HOME` environment variable.

Make use of `directories-rs` which is specifically designed to solve this problem cross-platform: https://github.com/soc/directories-rs#projectdirs

Tasks:
- Use `directories-rs` to resolve the config folder for 
  - `btsieve` - done in: comit-network/comit-rs#1236 
  - `comit_node` - done in: comit-network/comit-rs#1242
- Print a log message of where the config file is loaded from (no matter if path was overridden through `--config` or the default path was used)

D4nte:
As I previously stated I disagree with the use of `.config` folder.
This folder only contains my comit and yarn config files (a strange combination).

On Mac Os, everyone uses the `Library Support` folder, even the Lightning Labs apps and Bitcoin-QT!

I am strongly in favour of using `directories-rs`.

PR #1236, opened 2019-08-19
Use cross platform config directory

tcharding:
Currently we hard code the `.config/` path.  We can be cross platform if we use the `directories` crate instead.

Implements solution to #1074 for btsieve.  cnd coming as a separate PR.

### DIFF
--- a/btsieve/src/load_settings.rs
+++ b/btsieve/src/load_settings.rs
@@ -12,16 +12,7 @@ pub struct Opt {
 
 pub fn load_settings(opt: Opt) -> Result<Settings, ConfigError> {
     match opt.config_file {
-        Some(config_file) => {
-            if config_file.exists() {
-                Settings::read(config_file)
-            } else {
-                Err(ConfigError::Message(format!(
-                    "Could not load config file: {:?}",
-                    config_file
-                )))
-            }
-        }
+        Some(file) => parse_config_file(file),
         None => match directories::UserDirs::new() {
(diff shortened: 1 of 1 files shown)


btsieve/src/load_settings.rs
```diff
@@ -12,34 +13,30 @@ pub struct Opt {
...
-        Some(config_file) => {
-            if config_file.exists() {
-                Settings::read(config_file)
-            } else {
-                Err(ConfigError::Message(format!(
-                    "Could not load config file: {:?}",
-                    config_file
-                )))
+        Some(file) => parse_config_file(file),
+        None => match ProjectDirs::from("tech", "CoBloX", "btsieve") {
```
thomaseizinger: Do we want this to say comit-network instead?
```suggestion
        None => match ProjectDirs::from("network", "comit, "btsieve") {
```

tcharding: Cool, I like it.

bonomat: Please also update the comments below.

bonomat: Without `network` (or `tech`) the path resolves to: 
`"/Users/<username>/Library/Preferences/comit-network.btsieve/btsieve.toml"
`

tcharding: > Please also update the comments below.

I don't understand what you mean @bonomat, can you explain please (if action is still required), thanks.

Issue #1249, opened 2019-08-21
Use deterministic btsieve query IDs in cnd


thomaseizinger:
Currently, we rely on btsieve assigning an ID to our query upon creation.

[Spike 14](https://github.com/comit-network/spikes/blob/master/0014-resume-swaps-after-restart.adoc) defines to switch to deterministic query IDs based on the actual query.

Tasks:
- Create deterministic query ID from HTLC
- Use PUT instead of POST endpoint in cnd
- Delete the POST endpoint and tests in btsieve
- [Remove `rand` dependency](https://github.com/comit-network/comit-rs/pull/1256#discussion_r316453619) in btsieve - no more random query ID generation

PR #1256, opened 2019-08-21
Implement idempotent get-or-update query endpoint for btsieve

### LINKED ISSUE #1248: Implement new idempotent query endpoint for btsieve
[Spike 14](https://github.com/comit-network/spikes/blob/master/0014-resume-swaps-after-restart.adoc) defines a way of adding a idempotent query creation endpoint to btsieve.

This ticket is about:
- adding the endpoint to btsieve
- creating e2e tests for it

luckysori:
Fixes #1248.


btsieve/src/route_factory.rs
```diff
@@ -108,6 +108,17 @@ where
     let query_repository = warp::any().map(move || Arc::clone(&query_repository));
     let query_result_repository = warp::any().map(move || Arc::clone(&query_result_repository));
 
+    // validate query id function
+    let validate_query_id = warp::path::param::<String>().and_then(|id: String| {
+        if id.len() > 40 {
```
tcharding: Can we have a `const` declared instead of a magic number please.  A comment stating why 40 was chosen on the const would make me smile.

btsieve/src/route_factory.rs
```diff
@@ -108,6 +108,17 @@ where
     let query_repository = warp::any().map(move || Arc::clone(&query_repository));
     let query_result_repository = warp::any().map(move || Arc::clone(&query_result_repository));
 
+    // validate query id function
+    let validate_query_id = warp::path::param::<String>().and_then(|id: String| {
```
tcharding: Since the validation is only doing a single check perhaps we could name this `validate_query_id_length`.  `validate_query_id` implies, to me, more extensive validation.  Especially considering the matching error is `QueryIdTooLong`.

btsieve/Cargo.toml
```diff
@@ -26,6 +26,7 @@ hyper = "0.12"
 itertools = "0.8"
 log = "0.4"
 pretty_env_logger = "0.3"
+rand = "0.7"
```
bonomat: Don't forget to remove this again once we go for deterministic id generation

luckysori: Added to #1249.

btsieve/src/route_factory.rs
luckysori: Good point! It should be clear now :smile:

D4nte:
Changelog?

Issue #1243, opened 2019-08-20
Add persistent storage to cnd


tcharding:
cnd persistent storage epic - A.K.A. Sqlite database. 

https://app.zenhub.com/workspaces/coblox-scrum-5be4c2ea4b5806bc2bf2d351/issues/comit-network/comit-rs/1243

## Problem

Currently cnd only stores swaps in memory.  This means we can only expose to cnd clients the swaps done since the current cnd process was started.  ~Also we cannot continue in-progress swaps if the node restarts.~ (this is described by issue #1244).

## Goal

Store all swaps persistently (i.e., on disk) so that cnd exposes all swap history and restart in-progress swaps.

Spike 9 (https://github.com/comit-network/spikes/blob/master/0009-comit-btsieve-db.adoc) describes adding a database to cnd in order to save these states to disk. There seems to be no disagreement from the team on the recommendations within this spike.

tcharding:
Adding a plan of attack here for comments or suggestions.  Intend to first put together a draft PR with scope limited to swap metadata.  This will mean clients can list all previous swaps but not get any information on them.  If I understand correctly, this means only implementing an Sqlite backed version of `MetadataStore`.  For API queries on a swap id that was created by a previous instance of cnd the query will return `problem::state_store`.

thomaseizinger:
> Adding a plan of attack here for comments or suggestions. Intend to first put together a draft PR with scope limited to swap metadata. This will mean clients can list all previous swaps but not get any information on them. If I understand correctly, this means only implementing an Sqlite backed version of `MetadataStore`. For API queries on a swap id that was created by a previous instance of cnd the query will return `problem::state_store`.

Yes that seems reasonable.
I had another thought though and just realized that we may have been to quick in dismissing two of the tasks you added to the list above, sorry for that :(

At the moment, the e2e tests restart cnd and btsieve for each test to ensure an empty state. As soon as we have a database, that is no longer true. I think the e2e tests will actually break because the often assume that the list of swaps is empty in the beginning. Hence, you will need to modify the test harness to delete the database between the tests.
That in turn raises the question about the location of the database if several instances are running (do we want to have a `data_dir` commandline argument to override that?).

I think it would make sense to start with a PR that adds the following functionality:
- create a database on startup if it does not exist yet
- allows users to somehow override where it is stored (`directories-rs` has a `data_dir` so I'd suggest we add a commandline option to override that which we can use in the e2e tests)
[...]

PR #1257, opened 2019-08-22
Add Sqlite database

tcharding:
Add Sqlite database.  This PR only saves `Metadata` in the database.  Most of the comments from @thomaseizinger below are seen to, one notable exception, the migrations are not embedded in the executable.  Currently in order to test this one must copy the migrations directory from `cnd/migrations/` to `~/.local/share/migrations`.

Final commit of the series disables the `SqliteMetadataStore`, this is so we can introduce dormant code without introducing regressions.  To test please checkout the second to last commit.

Works towards resolving #1243


tcharding:
```
= Diesel/Sqlite database notes

For how to get diesel working with Sqlite see Setup below.

Binary to interact with the database is available in `cnd/src/bin/db.rs`

Execute with: `DATABASE_URL=cnd/cnd.sqlite3 target/debug/db show`


== Setup

reference:

https://profectium.blogspot.com/2017/11/rust-setting-up-sqlite-database-with.html
https://diesel.rs/guides/getting-started/

=== Using `diesel`

The database is currently set up using the command line tool `diesel`.
Install with `cargo install diesel --no-default-features --features sqlite`.

=== Initialize the database


Currently we initialize the database using the `diesel` command line tool: `diesel setup`
This requires setting the env var DATABASE_URL, currently this is set in `cnd/.env`

.cnd/.env
----
DATABASE_URL=cnd.sqlite3
----

=== Add migrations

ref: https://diesel.rs/guides/getting-started/

- Add initial migration: `diesel migration generate create_swaps`
- Write SQL for migrations in `up.sql` and `down.sql` under `cnd/migrations/`

.down.sql
[-source,sql]
----
DROP TABLE swaps
----

.up.sql
[-source,sql]
----
CREATE TABLE swaps (
       swap_id TEXT NOT NULL PRIMARY KEY,
       alpha_ledger VARCHAR NOT NULL,
       beta_ledger VARCHAR NOT NULL,
       alpha_asset VARCHAR NOT NULL,
       beta_asset VARCHAR NOT NULL,
       role VARCHAR NOT NULL,
       counterparty INTEGER NOT NULL
)
----
=== Apply the migration


- Configure diesel to use `cnd/src/db` directory

.cnd/diesel.toml
----
[print_schema]
file = "src/db/schema.rs"
----

Apply the migration using `diesel`: `diesel migration run`.
Check `down.sql` is correct by rolling back the migration: `diesel migration redo`.


== TODO

- Initialize the database programatically, currently we initialize the database using the `diesel` command line tool.
This may require setting the env var DATABASE_URL pragmatically so it is cross platform, currently this is set in `cnd/.env`
- .env file is not being used when running the executable?
```

thomaseizinger:
> Currently we initialize the database using the `diesel` command line tool: `diesel setup`
> This requires setting the env var DATABASE_URL, currently this is set in `cnd/.env`

@tcharding I did some quick investigation of what the `diesel` CLI acually does and there is not much magic behind it. The setup command starts here: https://github.com/diesel-rs/diesel/blob/master/diesel_cli/src/main.rs#L68

If you follow the code, it is mainly just creating config files and initial migrations.

I would propose the following (@comit-network/rust-devs, feel free to jump in here and disagree/argue):
- We use the diesel CLI to manage migrations during development
- We only support a single backend (sqlite) and hence, most of the abstractions provided by `diesel-cli` (DATABASE_URL f.e.) are not useful to us
- `cnd` has to create the database by itself at runtime. From looking at the diesel-cli code, this can be done through `SqliteConnection::establish(path)`
- `cnd` has to apply migrations at startup. This means we need to:
   - embed the migrations from the repository into the binary (there is a diesel macro for that)
   - run the migrations at startup against the given database
- We will have to create the schema at compile-time, probably through a buildscript (i.e. `build.rs`)
- We should write a blogpost about this approach :)

cnd/.env
```diff
@@ -0,0 +1 @@
+DATABASE_URL=cnd.sqlite3
```
D4nte: current folder?

cnd/migrations/2019-08-21-185335_create_swaps/up.sql
```diff
@@ -0,0 +1,10 @@
+-- MetadataStore only i.e., no StateStore.
+CREATE TABLE swaps (
+       swap_id TEXT NOT NULL PRIMARY KEY,
+       alpha_ledger VARCHAR NOT NULL,
+       beta_ledger VARCHAR NOT NULL,
+       alpha_asset VARCHAR NOT NULL,
+       beta_asset VARCHAR NOT NULL,
+       role VARCHAR NOT NULL,
+       counterparty INTEGER NOT NULL
```
D4nte: Integer?

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+use self::models::NewSwap;
+use diesel::{prelude::*, sqlite::SqliteConnection};
+use dotenv::dotenv;
+use schema::swaps;
+use std::env;
+
+pub fn establish_connection() -> SqliteConnection {
+    dotenv().ok();
+
+    let database_url = env::var("DATABASE_URL").expect("DATABASE_URL must be set");
```
D4nte: May I suggest to default it to the config path `comit/cnd.sql3` and allow to override it via command line?

cnd/src/db/models.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[derive(Queryable, Debug)]
+pub struct Swap {
+    pub swap_id: String,
+    pub alpha_ledger: String,
+    pub beta_ledger: String,
+    pub alpha_asset: String,
+    pub beta_asset: String,
+    pub role: String,
+    pub counterparty: i32, // FIXME: this should be u64 but cannot work out how to deserialize it?
```
D4nte: need to implement a manual deserialize trait, not sure you can do that with `Queryable`.

cnd/src/db/mod.rs
tcharding: This is done now using directories crate but I have not added a config option (either in file or command line).

cnd/migrations/2019-08-21-185335_create_swaps/up.sql
tcharding: Cargo cult programming from the diesel docs.

cnd/.env
tcharding: > > Currently we initialize the database using the `diesel` command line tool: `diesel setup`
> > This requires setting the env var DATABASE_URL, currently this is set in `cnd/.env`
> 
> @tcharding I did some quick investigation of what the `diesel` CLI acually does and there is not much magic behind it. 

Thanks man, this post is/was really helpful.

> The setup command starts here: [diesel-rs/diesel:diesel_cli/src/main.rs@`master`#L68](https://github.com/diesel-rs/diesel/blob/master/diesel_cli/src/main.rs#L68) [](/diesel-rs/diesel/blob/HEAD@{2019-08-23T00:41:50Z}/diesel_cli/src/main.rs#L68)
> 
> If you follow the code, it is mainly just creating config files and initial migrations.
> 
> I would propose the following (@comit-network/rust-devs, feel free to jump in here and disagree/argue):
> 
>     * We use the diesel CLI to manage migrations during development
> 
>     * We only support a single backend (sqlite) and hence, most of the abstractions provided by `diesel-cli` (DATABASE_URL f.e.) are not useful to us
> 
>     * `cnd` has to create the database by itself at runtime. From looking at the diesel-cli code, this can be done through `SqliteConnection::establish(path)`
> 
>     * `cnd` has to apply migrations at startup. This means we need to:
>       
>       * embed the migrations from the repository into the binary (there is a diesel macro for that)
>       * run the migrations at startup against the given database
> 
>     * We will have to create the schema at compile-time, probably through a buildscript (i.e. `build.rs`)
> 
>     * We should write a blogpost about this approach :)

Ok, will work on these points.

> 
> 
> > Binary to interact with the database is available in `cnd/src/bin/db.rs`
> 
> Since it is SQL, you can just use any SQL-viewer to view the database :)
> If it is helpful to you, feel free to have the CLI but I wouldn't merge it into master and ship it.

Cool, dropping this.

cnd/src/db/mod.rs
thomaseizinger: @D4nte Is the _config_ folder really the best place for this? `directories-rs` also has a `data_dir` method if I remember correctly. Could use that one?

thomaseizinger: Anyway, that is bike-shedding. Let's try and get it working first :)

tcharding: I used cache dir.  Confusingly there is a data directory and a cache directory.  I belive cache is the correct choice.

thomaseizinger: Hmm, according to [this](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), the cache dir is for: 

> user specific non-essential data

In full: 

>  $XDG_CACHE_HOME defines the base directory relative to which user specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used. 

Now, I guess it is a question of whether we consider the swap history to be non-essential. If you are in the middle of a swap, loosing the database could be equivalent to loosing money. I think it would be surprising to the user to loose his swaps if they delete the cache directory. Also, on Windows, the `data_dir` property resolves to the `AppData\Roaming` folder which, for example, is [kept in sync](https://blogs.windows.com/windowsdeveloper/2016/05/03/getting-started-with-roaming-app-data) across all Windows installations of a user for UWP Apps (Universal Windows Platform Apps).

I'd suggest the `data_dir` folder since we are storing that is is not recoverable. Caches sound like you can rebuild them from something whereas loosing your swap database means they are gone.

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        run_migrations(&db, &migrations)?;
+
+        Ok(SqliteMetadataStore {
+            db: db.clone().to_path_buf(),
+        })
+    }
+}
+
+impl MetadataStore for SqliteMetadataStore {
+    fn insert<M: Into<metadata_store::Metadata>>(
```
bonomat: Any particular reason why this needs to be `<M: Into<...>` and not just `Metadata`?

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    fn get(&self, key: SwapId) -> Result<Option<metadata_store::Metadata>, Error> {
+        use self::schema::metadatas::dsl::*;
+
+        let key = key.to_string();
+        let conn = establish_connection(&self.db)?;
+        let records = metadatas
+            .filter(swap_id.eq(key))
+            .load::<Metadata>(&conn)
+            .map_err(|err| Error::Load(err.to_string()))?;
```
bonomat: I was wondering if there is no nicer way and I found this: 

```
metadatas
.filter(swap_id.eq(key))
.first(conn)
.optional()
```

wouldn't that work for you?

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        use self::schema::metadatas::dsl::*;
+
+        let conn = establish_connection(&self.db)?;
+        let records = metadatas
+            .load::<Metadata>(&conn)
+            .map_err(|err| Error::Load(err.to_string()))?;
+
+        let v: Result<Vec<_>, _> = records
+            .iter()
+            .map(move |m| metadata_store::Metadata::try_from(m.clone()))
```
bonomat: 🔧 `move` should be redundant.

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        use self::schema::metadatas::dsl::*;
+
+        let key = key.to_string();
+        let conn = establish_connection(&self.db)?;
+        let records = metadatas
+            .filter(swap_id.eq(key))
+            .load::<Metadata>(&conn)
+            .map_err(|err| Error::Load(err.to_string()))?;
+
+        if records.len() == 0 {
```
bonomat: 🙃 `is_empty` :)

cnd/src/db/mod.rs
tcharding: That was in the code before I touched it.  I'm not sure what is the full benefit of having it or why we added the trait instead of a concrete type.

cnd/src/db/mod.rs
tcharding: Cool, I'll give this a shot.  I copied this from the docs.

cnd/src/db/mod.rs
tcharding: Nice, good reviewing @bonomat!

cnd/src/db/mod.rs
tcharding: +1

cnd/src/db/mod.rs
tcharding: Good investigation on cache vs data @thomaseizinger.  Thanks for this, used data dir as suggested.

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        use schema::metadatas;
+
+        let md = Metadata::new(metadata.into());
+        let new = NewMetadata::new(&md);
+
+        let conn = establish_connection(&self.db)?;
+        diesel::insert_into(metadatas::table)
+            .values(new)
+            .execute(&conn)
+            .map(|_| ()) // Drops the '1' for successful row insert.
```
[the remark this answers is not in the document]
tcharding: > Please have a look at my comment in regards of the migration, not what the consensus is on moving forward with this.

I'm not seeing the comment you reference here @bonomat?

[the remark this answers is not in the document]
tcharding: Good idea, will do.

tcharding:
> Please have a look at my comment in regards of the migration, not what the consensus is on moving forward with this.

I'm not seeing the comment you reference here @bonomat?

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        log::info!("Sqlite database file: {}", db.display());
+
+        let migrations = default_migrations_dir()
+            .ok_or_else(|| Error::Path("Failed to get migrations directory".to_string()))?;
+
+        create_migrations_directory_if_needed(&migrations)?;
+        run_migrations(&db, &migrations)?;
+
+        Ok(SqliteMetadataStore {
+            db: db.clone().to_path_buf(),
```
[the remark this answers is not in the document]
tcharding: We don't need `to_path_buf()` either.

bonomat:
> > Please have a look at my comment in regards of the migration, not what the consensus is on moving forward with this.
> 
> I'm not seeing the comment you reference here @bonomat?

here we go: https://github.com/comit-network/comit-rs/pull/1257/files#r323053146

Not sure why @luckysori marked this comment as *resolved* :D

cnd/src/db/mod.rs
tcharding: I got this
```
        metadatas
            .filter(swap_id.eq(key))
            .first(&conn)
            .optional()
            .map_err(|err| Error::Load(err.to_string()))?
            .map(move |m: Metadata| metadata_store::Metadata::try_from(m.clone()))
            .transpose()
```

bonomat: not sure what `transpose` does ... 
but if it works, cool :)

tcharding:
All  comments seen to @bonomat.  Force pushed branch so that the 'disable' commit stays last to ease testing for those that wish to.

This needs at least one OSX user to test this please.  To do so checkout the second last commit, copy the `cnd/migrations/` directory to the OSX project data directory (should be /Users/<user>/Library/Application Support/comit/) and run e2e tests.  Could I please get a thumbs if someone does this testing @bonomat @da-kami @D4nte, thanks.

luckysori:
> Not sure why @luckysori marked this comment as _resolved_ :D

I either missclicked thrice on my phone or it was already resolved at some point and I left it as is.

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+            .ok_or_else(|| Error::Path("Failed to get database path".to_string()))?;
+
+        create_database_if_needed(&db)?;
+        log::info!("Sqlite database file: {}", db.display());
+
+        let migrations = default_migrations_dir()
+            .ok_or_else(|| Error::Path("Failed to get migrations directory".to_string()))?;
+
+        create_migrations_directory_if_needed(&migrations)?;
+        run_migrations(&db, &migrations)?;
```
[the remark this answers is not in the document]
tcharding: I'm no database expert but I think your concerns are valid if/when we add a second migration.  Currently we have only a single migration so it can be run automatically.  I think a comment it this point in the code to note this could be a good idea.

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+
+        Ok(SqliteMetadataStore { db })
+    }
+}
+
+impl MetadataStore for SqliteMetadataStore {
+    fn insert<M: Into<metadata_store::Metadata>>(
+        &self,
+        metadata: M,
+    ) -> Result<(), metadata_store::Error> {
```
D4nte: What are you talking about in the comment at line 66? This does indeed return a `Result` so the `MetadataStore` trait must return a `Result` or this would not compile.

cnd/src/db/models.rs
```diff
@@ -0,0 +1,54 @@
...
+            role: md.role.to_string(),
+            counterparty: md.counterparty.to_string(),
+        }
+    }
+}
+
+// Diesel docs say to use this second structure for database inserts.
+#[derive(Insertable, Debug)]
+#[table_name = "metadatas"]
+pub struct NewMetadata<'a> {
```
D4nte:
```suggestion
pub struct InsertableMetadata<'a> {
```

cnd/src/swap_protocols/metadata_store.rs
```diff
@@ -85,49 +82,87 @@ impl Metadata {
...
+    Connect(String),
+    Load(String),
+    Insert(String),
+    Parse(String),
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            // FIXME: is this idiomatic [CoBloX] Rust?
```
D4nte:
```suggestion
```
Yes it's fine/expected when implementing `Display` for an `enum`.

cnd/src/swap_protocols/metadata_store.rs
```diff
@@ -85,49 +82,87 @@ impl Metadata {
...
+            Error::Init(msg) => write!(f, "Failed to initialize datastore : {}", msg),
+            Error::Connect(msg) => write!(f, "Failed to connect to datastore: {}", msg),
+            Error::Load(msg) => write!(f, "Failed to load record: {}", msg),
+            Error::Insert(msg) => write!(f, "Failed to insert new record: {}", msg),
+            Error::Parse(msg) => write!(f, "Failed to parse stored record: {}", msg),
+        }
+    }
+}
+
+// FIXME: Do we need this?
```
D4nte: Can you compile without it?

cnd/src/db/mod.rs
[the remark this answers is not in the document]
thomaseizinger: > Do we really want to automatically run migration without a warning?

In the spirit of "it just works TM" I would say yes.

> The user should then have created a back up prior running the migration.

Cnd should probably create a backup of the database automatically before running migrations.

I would not want to burden the user with that. That seems like an unnecessary complication if the software is telling you to do stuff manually that it just could have done automatically.

cnd/src/db/mod.rs
thomaseizinger: `transpose` converts between `Result<Option>` and `Option<Result>` IIRC: https://doc.rust-lang.org/std/result/enum.Result.html#method.transpose

cnd/diesel.toml
```diff
@@ -0,0 +1,5 @@
+# For documentation on how to configure this file,
+# see diesel.rs/guides/configuring-diesel-cli
```
thomaseizinger: We are not using the `diesel-cli`. Do we still need this file?

cnd/src/config/settings.rs
```diff
@@ -12,6 +12,7 @@ pub struct Settings {
     pub comit: Comit,
     pub network: Network,
     pub http_api: HttpSocket,
+    pub database: Option<Database>,
```
thomaseizinger: The `Settings` struct is supposed to have all optional elements applied already, i.e. when converting from `File` to `Settings`, all the defaults are filled in. This also seems to be the case for the `database` config option (we always need a database).

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,162 @@
+mod models;
+mod schema;
+
+pub use crate::db::models::Metadata;
+
+use crate::{
+    db::models::NewMetadata,
```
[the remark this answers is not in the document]
thomaseizinger: Does it though? For me, aliases cause major headache usually because you cannot find the type when you are looking for it ..

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+
+        let conn = establish_connection(&self.db)?;
+        diesel::insert_into(schema::metadatas::table)
+            .values(new)
+            .execute(&conn)
+            .map(|res| {
+                // MetadataStore trait does not return any indication whether the
+                // insert was succesful or not so just log it and drop the return value.
+                // If/when InMemoryMetadataStore goes away we can probably change the trait.
+                if res == 1 {
```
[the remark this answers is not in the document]
thomaseizinger: When are we checking for `1` here?
I would expect this call to return you the created primary key for the row, hence this is only going to run _once_ for every user (there is only one swap with ID 1).

cnd/src/swap_protocols/metadata_store.rs
```diff
@@ -85,49 +82,86 @@ impl Metadata {
     }
 }
 
-#[derive(Debug, Fail)]
+#[derive(Debug)]
 pub enum Error {
```
thomaseizinger: Usually, we only store the source of the error inside the variant. This would potentially leak some information here so I am not quite sure. Would have to investigate in detail.

If you can't pass on the source error, I'd still be in favor of just putting relevant data into the variant and not an already formatted string. Try to delay the formatting for as long as possible (you don't know how your caller wants to handle the error, let alone localization etc).

cnd/src/swap_protocols/metadata_store.rs
```diff
@@ -85,49 +82,86 @@ impl Metadata {
...
-    DuplicateKey,
+    Path(String),
+    Init(String),
+    Connect(String),
+    Load(String),
+    Insert(String),
+    Parse(String),
+}
+
+impl fmt::Display for Error {
```
thomaseizinger: Usually, there is no need for a `Display`-impl. Why do you need it?

cnd/src/swap_protocols/metadata_store.rs
thomaseizinger: I think most of your error enums don't implement `std::error:Error`. I don't have a source now but I think this trait is deemed useless by the community.

cnd/src/db/mod.rs
D4nte: I removed it 😁

cnd/src/db/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+}
+
+impl MetadataStore for SqliteMetadataStore {
+    fn insert<M: Into<metadata_store::Metadata>>(
+        &self,
+        metadata: M,
+    ) -> Result<(), metadata_store::Error> {
+        let md = Metadata::new(metadata.into());
+        let new = NewMetadata::new(&md);
```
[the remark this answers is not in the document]
thomaseizinger: Yes I think it is, @tcharding added a comment to the struct definition explaining that.

cnd/diesel.toml
tcharding: You may be right, since we run the migrations in code i don't think we need this.  Will remove and test.

cnd/src/config/settings.rs
tcharding: Oh, yes.  Will fix, thanks.

cnd/src/swap_protocols/metadata_store.rs
tcharding: The error handling in general needs work, it was best effort at the time - I'd like to spend some time on it.  There are a bunch of things that I'm not happy with in this PR, finding the balance of getting something merged and working towards perfection is not easy here especially considering the complexity of this task compared to my Rust-foo.  I'm not going to merge this right away, open to suggestions on things that you think should be done before merge and things that you think are ok to do after merge.  I'll probably go for 'loose consensus and running code' as they say.

cnd/src/swap_protocols/metadata_store.rs
tcharding: I got tangled working with converting from one type to another using `ToSt`, `FromStr`, `From`, and constructors.  This may be a hang over from that - I'll investigate.

tcharding: > Does it though? For me, aliases cause major headache usually because you cannot find the type when you are looking for it ..

Are you referring to the `pub use ...`?  if so it seems to not need the pub.  Will remove that.

tcharding: > When are we checking for 1 here?
> I would expect this call to return you the created primary key for the row, hence this is only going to run once for every user (there is only one swap with ID 1).

We don't need to return the primary key because we do not create it, its the swap id.

tcharding:
Thanks for the review @thomaseizinger, will re-work using your suggestions.  Onwards and upwards!

cnd/src/swap_protocols/metadata_store.rs
thomaseizinger: Yes, should have probably made this more explicit: this comment was definitely not a blocker :)

cnd/migrations/initial_metadata/up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE metadatas (
+       swap_id VARCHAR(255) NOT NULL PRIMARY KEY,
```
thomaseizinger: Not sure if we want to do this as part of this PR but:

I think it is best practise to have a "stupid" database-only primary key if you have a domain-specific identifier already. (primary keys should be numbers I think)
The primary key would be used for foreign-key things and the domain-identifier only needs to have a `UNIQUE` attribute.

cnd/src/config/settings.rs
thomaseizinger: This will probably bubble up to not accepting an `Option` in your constructor of the SqliteMetadataStore.

cnd/src/db/mod.rs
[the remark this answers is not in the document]
thomaseizinger: > I would expect this call to return you the created primary key

This was targeted at the diesel API, not our function :)

cnd/src/db/mod.rs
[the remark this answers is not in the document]
D4nte: Arg, wrong line, there are 2 types of `Metadata`.
I don't have your headache but I'd be happy to just keep the path so that they can be differentiate in file.

cnd/src/config/settings.rs
thomaseizinger: > Why is web_gui optional in Settings?

Because it is actually optional at runtime.

thomaseizinger: > Alright, a lot of round-a-bout comments here (from me). Eventually InMemoryDataStore will go away. I'm going to leave the config stuff as it is now and fix it then. There are non-trivial complications with config of the database and calling the db constructor iff both memory stores are in use - that is why this PR has the 'disable Sqlite' patch at the end. Admittedly its ugly.

Fair :)
Although the only thing the settings struct will contain is the default path right? What is the difference between setting the default path here if it does not exist vs. having an `Option` in the constructor?

PR #1260, opened 2019-08-23
Remove the noise on stdout from cargo-make when installing clippy

thomaseizinger:


D4nte:
Please explain a bit more.

thomaseizinger:
@D4nte 
> Please explain a bit more or provide link to documentation.

Run `cargo make clippy` locally before and after this patch to see, it is hard to explain :D

PR #1263, opened 2019-08-23
Automatically merge dependabot PRs

thomaseizinger:


D4nte:
Wouldn't it make more sense to use Mergify to have one queue of merging or instead of 2 bots fighting for merges?

thomaseizinger:
> Wouldn't it make more sense to use Mergify to have one queue of merging or instead of 2 bots fighting for merges?

Yes we will do that by EOB today.

PR #1267, opened 2019-08-23
Replace web3 with ether.js

bonomat:


api_tests/lib/ethereum.ts
```diff
@@ -52,26 +44,23 @@ async function erc20Balance(
...
 async function mintErc20Tokens(
     ownerWallet: EthereumWallet,
     contractAddress: string,
     toAddress: string,
-    amount: BN | string | number
+    amount: BigNumber | string | number
 ) {
     const functionIdentifier = "40c10f19";
 
     toAddress = toAddress.replace(/^0x/, "").padStart(64, "0");
```
D4nte: I wonder if the new library gives you tools to make this easier.

bonomat: For this in particular I did not find anything obvious.

PR #1280, opened 2019-08-26
Refactor protocol implementation to align code with rfc-002/rfc-003

### LINKED ISSUE #1187: Update code inline with RFC-002 (remove status code)
Recent amendment to the RFCs removed the status code.  An issue was created that was meant to bring the code inline with RFCs 1-3.  During coding on that issue (https://github.com/comit-network/comit-rs/issues/995) we added a `CLOSE` frame (still in progress).  We (@da-kami and @tcharding) decided to split #995 into separate issues to reduce the scope. Code in cnd that implements RFC-002 (and consecutively RFC-003)  needs to be brought into line with the RFCs.

- Remove status code
- Ensure all error logic is correctly handled by the `decision: declined` header

da-kami:
fixes #1187 

* remove status code from bam-crate
* Replace all `Rejected` states by `Declined`
* Decision header
* Decline response body (reason only, no details atm)


da-kami:
> As discussed, `Response::default()` seems to be an empty response. I'd expect an `accepted` or `declined` header to always be present.
> 
> I'd suggest to change the `Default` trait implementation because I would expect it to give me a valid `Response` but actually both headers and body are empty so it's not functionally valid.
> 
> My proposal would be:
> 
> * A: the headers is set to `decline` for `default()`.
> * B: Don't use Default trait and instead implement a static function `Response::empty()` that returns an empty response.
> 
> Mark "Request changes" as we should not return empty responses.

Initially I thought about going for `accept` and `decline` constructors - but given the current structure that would mean that we have to move the `Decision` into the `bam` crate. Eventually this will be changed to `libp2p-comit` instead of `bam`, then we can reconsider this solution. 
For now I added `empty()` to construct the Response. It is, however, not enforcing he header. It should but I don't have a proper solution for enforcing it without moving the `Decision` into the crate...

PR #1284, opened 2019-08-26
Configure Circle CI to attache artifacts to releases

### LINKED ISSUE #1121: Create `--release` binaries for Linux upon each release (tag) in comit-rs
- Create a `--release` binary on every release build in CircleCI
- Automatically attach it to the release on GitHub

D4nte:
New jobs:
- `debug-build-test`: Debug build and tests, always run and will be the new check to have before a PR merge on master
- `release-build-test`: Release build and tests, only ran on a `release/X.Y.Z` branch
- `publish-github-release`: attach binaries to a GitHub release, only ran on a `X.Y.Z` tag (which must be done on a `release/X.Y.Z` branch).

#1294 tests that `release-build-test` runs on a `release/X.Y.Z` branch and that the releasing process works.

Check out test release: https://github.com/comit-network/comit-rs/releases (it will be deleted once this PR is merged).

Resolves #1121.


.circleci/config.yml
```diff
@@ -1,6 +1,26 @@
...
+            tags:
+              only: /^.*$/
+      - publish-github-release:
+          requires:
+            - release-build-test
+          filters:
+            branches:
+              only: /^release\/\d+\.\d+\.\d+$/
+            tags:
+              only: /^\d+\.\d+\.\d+$/
```
thomaseizinger: I think it is convention to prefix the tag with `v`.
My limited regex knowledge tells me that this regex will _not_ match that. Is that correct?

.circleci/config.yml
```diff
@@ -10,24 +30,82 @@ jobs:
...
+      - install_rust
+      - install_node_devlibs
+      - print_current_versions
+      - run:
+          name: "Package & Publish Release on GitHub"
+          command: |
+            go get github.com/tcnksm/ghr
+            VERSION=$(cargo pkgid -- cnd|cut -d# -f2)
+
+            # Let's abort if the git tag does not match the version in Cargo.toml as it would be fishy
```
thomaseizinger: This is a very good idea!
Should we have that already as part of the release branch?
Like, make sure the version in the release branch matches the Cargo.toml version?
Even if not, I'd this the very first step, right after checkout: we don't need to do anything if those tags don't match :)

Another suggestions: The `pkgid` thingy looks a bit odd to me :sweat_smile: 
I had a go at the `cargo metadata` command and put together the following:

```
cargo metadata --no-deps --format-version 1 | jq '.packages[] | select(.name | contains("cnd")).version' -r
```

Gets the version of the `cnd` package and prints it on stdout.
Up to you if you want to use it :)

.circleci/config.yml
```diff
@@ -10,24 +30,82 @@ jobs:
...
+            test "${VERSION}" = "${CIRCLE_TAG}"
+            mkdir ~/package
+            cd ~/artifacts && tar czvf ~/package/comit-rs_v${VERSION}_$(uname -s)_$(uname -m).tar.gz *
+            ghr -t ${GITHUB_TOKEN_FOR_RELEASES} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -c ${CIRCLE_SHA1} -delete ${VERSION} ~/package/
 
 commands:
-  setup_system:
+  install_node_devlibs:
```
thomaseizinger: What was wrong about `setup_system`?
Your current name doesn't capture that we kill `systemd` for example ..

.circleci/config.yml
D4nte: This is a details that was not agreed in the spike. For me we should out the v everywhere or nowhere.

.circleci/config.yml
D4nte: ... Pkgid is fine and documented in the cargo book along metadata

.circleci/config.yml
thomaseizinger: `comit-i` uses `vX.Y.Z` tags and GitHub suggests that as-well when creating a tag. I'd vote for `v`-style tags.

.circleci/config.yml
D4nte: Setup system does not really gives you much about what it actually does. Also, the systemd is a side effect of achievement the goal of the command: installing node and dev libs

.circleci/config.yml
D4nte: Yes for the release branch

.circleci/config.yml
thomaseizinger: Can we do `install_node_and_devlibs` then? Currently it reads like we are installing node development libraries.

.circleci/config.yml
thomaseizinger: > ... Pkgid is fine and documented in the cargo book along metadata

Hmm, I guess I just don't like this manual parsing of strings with `cut` and friends. As long as it is a documented and stable format, shouldn't matter much :)

D4nte: > > ... Pkgid is fine and documented in the cargo book along metadata
> 
> Hmm, I guess I just don't like this manual parsing of strings with `cut` and friends. As long as it is a documented and stable format, shouldn't matter much :)

Now I see your point.

.circleci/config.yml
D4nte: Ok we agreed on this one :)

.circleci/config.yml
D4nte: - I have added match against version in branch name
- I see your point for `pkgid`, you are welcome to do a follow PR on this.

.circleci/config.yml
D4nte: Please do a follow-up PR please, I am want to merge this :)

thomaseizinger:
Would you mind squashing all those commits together?
Since we are really changing only one thing (the CircleCi config), I don't think the history of trial and error is really useful.

D4nte:
> Would you mind squashing all those commits together please?
> Since we are really changing only one thing (the CircleCi config), I don't think the history of trial and error is really useful.
> Thanks 🤗

I can't squash much as I merged master. I'll squash the most I can.

D4nte:
> Would you mind squashing all those commits together please?
> Since we are really changing only one thing (the CircleCi config), I don't think the history of trial and error is really useful.
> Thanks 🤗

Done!

PR #1285, opened 2019-08-26
Runtime version check between btsieve and cnd

### LINKED ISSUE #1219: Runtime version check between btsieve and cnd
btsieve should return its own version in REST API responses and cnd should check it and confirm it is the same than its own.

Design:
- cnd sends a header with every request, containing the expected version of `btsieve`
- btsieve fails with 400 and descriptive `application/problem+json` error

According to the [release spike](https://github.com/comit-network/spikes/blob/master/0012-release-strategy.adoc), btsieve and cnd are always released together, hence there should be a strict-equality check (even if compatibility might be possible).

In addition, add a check to `cnd` at startup whether btsieve is available. Use the existing `/health` endpoint for that (with the new header, so that the version check is run).
If btsieve is not available, log a warning.
If it is available but version check failed, also log a warning.

luckysori:
- Add mandatory `Expected-Version` header to btsieve requests. If the expected version doesn't match the actual version, the request fails.
- Ping btsieve on cnd startup using a `GET /health` request. 
  - Warn if no btsieve is found on the configured port.
  - Warn if the request fails with a version mismatch.
  - Inform that btsieve was found successfully otherwise.

Fixes #1219.


api_tests/btsieve/bitcoin_only/test.ts
```diff
@@ -25,9 +25,9 @@ setTimeout(async function() {
         describe("Bitcoin", () => {
             describe("Transactions", () => {
                 it("btsieve should respond not found when getting a non-existent bitcoin transaction query", async function() {
-                    const res = await request(btsieve.url()).get(
-                        "/queries/bitcoin/regtest/transactions/1"
-                    );
+                    const res = await request(btsieve.url())
```
D4nte: I'd try to override this request with one that do a set of the header.
```
const request(url: string) {
return Request.request(url).set("Expected-Version", btsieve.expectedVersion);
}
Note: I assume the `.set()` can get done before `.get()` call, if not, it's a bit harder.

File: cnd/src/btsieve/client.rs [hunk 2]
```diff
@@ -214,13 +222,51 @@ impl BtsieveHttpClient {
...
                 .map(|_| ())
                 .map_err(|e| {
                     Error::FailedRequest(format!("Failed to delete query because {:?}", e))
                 }),
         )
     }
+
+    pub fn health(&self) -> Box<dyn Future<Item = (), Error = ()> + Send> {
+        let mut url = self.endpoint.clone();
+        url.set_path("health");
```
D4nte: Considering that we are not sticking to the health RFC, shouldn't we call it something else than `/health` ?

api_tests/btsieve/bitcoin_only/test.ts
luckysori: I did try, but `.set()` can only be done after `.get()` :disappointed:

cnd/src/btsieve/client.rs
luckysori: No strong opinion on this, so happy to change it.

cnd/src/btsieve/client.rs
```diff
@@ -214,13 +222,51 @@ impl BtsieveHttpClient {
...
+                .headers(construct_headers())
+                .send()
+                .map(|response| {
+                    let endpoint = &response.url()[Position::BeforeScheme..Position::BeforePath];
+
+                    if response.status().is_success() {
+                        log::info!("Btsieve running at {}", endpoint)
+                    } else {
+                        log::error!(
+                            "Version of btsieve at {} does not match expected version",
```
tcharding: From this code it is not immediately apparent why this error message is what it is i.e., the code does not show _why_ this error is caused by a version mismatch.

luckysori: I agree. I'm not too happy with the logs. I should probably just log the HTTP API Problem returned by btsieve. The bot merged this before I could do it though, so it'll have to be in a follow-up PR.

PR #1292, opened 2019-08-27
Move blockchain_contracts to root

### LINKED ISSUE #1291: Move `blockchain_contracts` to root
It is not possible to do coblox/blockchain-contracts#4 as it is blocked (need to sort out `internal/bitcoin_witness`).

However, `internal/blockchain_contracts` crate is (almost) ready to go in its own repo, it is only depending on `internal/bitcoin_witness`.

Because of that, we will move the crate to the root folder as this crate is meant to be in its own repo.

Other `internal` crates are staying in `internal` until they are `sorted out` (whatever it means for each crate).

This issue is meant as a replacement of coblox/blockchain-contracts#4 within the sprint, coblox/blockchain-contracts#4 to be done once unblocked.

D4nte:
Resolves #1291.


Cargo.toml
```diff
@@ -1,2 +1,2 @@
 [workspace]
-members = ["cnd", "btsieve", "internal/*"]
```
luckysori: :question: Doesn't `cargo-tomlfmt` enforce alphabetical order?

D4nte: Nope! (I just tested)

luckysori: It looks like the toml specification [doesn't guarantee order for items in tables](https://github.com/toml-lang/toml#table).

PR #1293, opened 2019-08-27
Improve logging of negative btsieve health check

luckysori:
As @tcharding pointed out, the logs generated by cnd's health check on btsieve knew too much about how btsieve is implemented. This tries to fix that by leveraging the Problem Details JSON object returned with unsuccessful responses to btsieve.

Cool finding:

Functions that expect a closure as an argument also accept blocks that return a closure:

```rust
.map({
    let foo = self.foo.clone();
    move |_| println!("Foo: {}", foo)
})
```

This is particularly useful when working with futures.


cnd/src/btsieve/client.rs
```diff
@@ -234,26 +235,45 @@ impl BtsieveHttpClient {
...
+                    move |mut response| {
+                        if response.status().is_success() {
+                            log::info!("Btsieve running at {}", endpoint);
+                        } else {
+                            let _ = response
+                                .json()
+                                .map({
+                                    let endpoint = endpoint.clone();
+                                    move |problem: HttpApiProblem| match problem {
+                                        HttpApiProblem {
```
luckysori: hadouken-indentation

cnd/src/btsieve/client.rs
```diff
@@ -234,26 +235,48 @@ impl BtsieveHttpClient {
...
-                    if response.status().is_success() {
-                        log::info!("Btsieve running at {}", endpoint)
-                    } else {
-                        log::warn!(
-                            "Version of btsieve at {} does not match expected version",
-                            endpoint
-                        )
-                    }
+                .map_err(HealthCheckError::Reqwest)
+                .and_then(|response| HealthCheckResponse {
```
D4nte: I don't understand why you wrapped it in a Health Check Response

D4nte: Ah ok, it's because you wrap it in a Future.

tcharding:
> Cool finding:
> 
> Functions that expect a closure as an argument also accept blocks that return a closure:
> 
> ```rust
> .map({
>     let foo = self.foo.clone();
>     move |_| println!("Foo: {}", foo)
> })
> ```

Isn't that semantically the same as writing this?

```rust
.map(move |_| {
    let foo = self.foo.clone();
    println!("Foo: {}", foo)
})
```

thomaseizinger:
> > Cool finding:
> > Functions that expect a closure as an argument also accept blocks that return a closure:
> > ```rust
> > .map({
> >     let foo = self.foo.clone();
> >     move |_| println!("Foo: {}", foo)
> > })
> > ```
> 
> @luckysori isn't that semantically the same as writing this?
> 
> ```rust
> .map(move |_| {
>     let foo = self.foo.clone();
>     println!("Foo: {}", foo)
> })
> ```

Nope! Not quite :)
The block is executed immediately and the closure is returned!

thomaseizinger:
I tweeted about it here because I am so excited about this :D
https://twitter.com/oetzn/status/1166912187174023169

PR #1302, opened 2019-08-29
Refactor e2e tests to store config inside typescript files

### LINKED ISSUE #1295: Revise cnd & btsieve config in e2e tests
In order to achieve #1246, I propose to slightly revise the way our current e2e tests deal with config files.

At the moment, we check in the config file to be used by cnd and btsieve and read relevant stuff from it at execution time of the tests.

I propose to store the desired configuration inside typescript objects and generate the configuration files to passed to cnd and btsieve from that as temporary files.

This allows us to simply `import` the configuration into every test and guarantees us that this is the one that the binaries are actually going to use.
This will make the current code a lot simpler and will allow us to create a smoother migration path to the new e2e test suite.

thomaseizinger:
Sorry for the single patch PR guys but this was almost impossible to do in separate patches so I just always amended.

Basically, this PR does the following things:
- The main source of config is now in the typescript code
- We use testcontainers to boot up the nodes (instead of docker-compose)
- All config files within the tests now only declare what the want without specifying any details
- Configuration files for cnd and btsieve are written on the fly

This allows us to simply import the configuration inside the typescript files and work with it. For example, exposing the HTTP API port of cnd within the `Actor` is now a breeze :)

The way it works is that the individual runners will import the configuration objects from `lib/config.ts`, write it to a file and then start the node with the given configuration.
The `Actor` and `Btsieve` classes will simply _also_ import that configuration and by specifying the same actor name, they have access to the same configuration.

Fixes #1295.

The e2e tests should now also be slighly faster because I start/stop ledger and comit nodes in parallel.


api_tests/lib/ledger_runner.ts
```diff
@@ -1,93 +1,187 @@
...
+}
 
-    public getLedgersConfig(): any {
-        return this.ledgersConfig;
+/*
+ * For some weird reason, the log buffer contains weird prefixes
+ *
+ * This function removes those prefixes so that the logs can be printed to the file.
+ */
+function sanitizeBuffer(buffer: Buffer) {
```
bonomat: WTH... did you find out where this is coming from?

thomaseizinger: Nope, no idea :sweat_smile:

api_tests/harness.ts
```diff
@@ -59,43 +60,63 @@ async function runTests(testFiles: string[]) {
         logDir
     );
 
-    process.on("SIGINT", () => {
+    async function cleanupAll() {
+        try {
+            btsieveRunner.stopBtsieve();
+            nodeRunner.stopCnds();
```
D4nte:
```suggestion
            cndRunner.stopCnds();
```
🤡

api_tests/harness.ts
```diff
@@ -59,43 +60,63 @@ async function runTests(testFiles: string[]) {
...
+            await ledgerRunner.stopLedgers();
+        } catch (_) {
+            // we don't care much if the cleanup fails
```
D4nte: We can of do if we have some rogue processes/docker runnings. A `console.log` should be enough.

api_tests/lib/actor.ts
```diff
@@ -38,55 +26,36 @@ const MOVE_CURSOR_UP_ONE_LINE = "\x1b[1A";
...
+        return `http://127.0.0.1:${this.cndConfig.httpApiPort}`;
     }
 
     public cndNetworkListenAddress() {
-        const addr = multiaddr(this.cndConfig.network.listen[0]);
-        // Need to convert 0.0.0.0 to 127.0.0.1
-        return `/${addr.protoNames()[0]}/${this.host}/${addr.protoNames()[1]}/${
-            addr.nodeAddress().port
-        }`;
+        return `/ip4/0.0.0.0/tcp/${this.cndConfig.comitPort}`;
```
D4nte: This function looks useless. Is it used?

thomaseizinger: Haha, they are actually wrong :upside_down_face: 
Need to be `127.0.0.1`. Will send a follow up PR!

D4nte: Who were "they"?

Issue #1339, opened 2019-09-04
Mob programming: Future-based API for query matching in btsieve


thomaseizinger:
1. Create a function that returns results of a query: `fn(query) -> impl Stream<Item = Transaction>`
2. Implementation is based on a stream of blocks given from a specific timestamp and filters them (stream doesn't necessarily have to be ordered): `fn (timestamp, confirmations) -> impl Stream<Item = Block>`
3. Use this new interface when creating a query: get the stream of results, `and_then` store it in the database
4. Figure out how to implement the interface for (2) based on what we currently have: BlockchainDag, Blocksource etc.
a. Probably a manual implementation of `Stream`
b. has two dependencies: block by hash and latest block
c. Either fulfill these dependencies through our BlockchainDag or directly consult a blocksource (would need caching in the future then)

thomaseizinger:
Done!
Result is here: https://github.com/comit-network/comit-rs/tree/btsieve-mob-programming

Follow-up ticket created here: https://github.com/comit-network/comit-rs/issues/1346

PR #1344, opened 2019-09-05
Add save query with id method to query repository

luckysori:
Whilst working on #1339 we noticed that there is a bug with the `get_or_create` endpoint introduced with #1248. I wonder who caused it.............


api_tests/btsieve/bitcoin_only/test.ts
```diff
@@ -133,23 +122,31 @@ setTimeout(async function() {
...
-                        to_address: "mndvsV4weBdPFTarHQbg71rCXRy8z79SH5",
-                    };
+                const otherQueryId = "4BvFBixM4HmhV8AJe5RC8v8csxxhDBscMxwpiK5e";
+                const otherLocation =
+                    "/queries/bitcoin/regtest/transactions/" + otherQueryId;
+                const otherQuery = {
+                    to_address: "mndvsV4weBdPFTarHQbg71rCXRy8z79SH5",
+                };
 
+                it('btsieve should respond with 204 NO_CONTENT when "getting-or-creating" a new query', async function() {
```
D4nte: I started a refactor locally, I believe the `PUT` tests should be in a separate file as `PUT` and `POST` are not supposed to be used together.

By having a separate file you ensure there is no mix up nor (un)voluntary usage of side effects.

btsieve/src/in_memory_query_repository.rs
```diff
@@ -38,15 +38,23 @@ impl<T: Send + Sync + Clone + 'static> QueryRepository<T> for InMemoryQueryRepos
     fn save(&self, entity: T) -> Result<String, Error<T>> {
         let mut state = self.state.write().unwrap();
 
-        // this ID will eventually be generated by cnd, its current length is not
-        // relevant
+        // this ID will eventually be generated by cnd and the method save_with_id below
```
D4nte:
```suggestion
        // This ID will eventually be generated by cnd and the method save_with_id below
```

PR #1363, opened 2019-09-07
Simplify serialization of new bitcoin amount type

thomaseizinger:
This PR simplifies the serialization for the new bitcoin amount slightly.

We already have dedicated methods/traits in place that handle serialization/deserialization (`FromBamHeader`, `ToBamHeader`, `FromHttpAsset` and the generic `Http` struct).
This PR moves the serialization/deserialization logic of custom types into those trait implementations.


thomaseizinger:
> Thanks dude, looks better indeed :)

Unfortunately, there are still a few obstacles :(
I actually wanted to go even further but the serialization code in the HTTP layer is quite messy :sweat_smile: 
That was actually the main driver of why I created the technical debt issue here: https://github.com/comit-network/comit-rs/issues/1364

In general, I think we should actually just have dedicated types and simply convert between them for stuff like serialization. That would allow us to leverage `serde`'s powerful configuration capabilities instead of having generic stuff like `HttpAsset`.

PR #1368, opened 2019-09-09
Compile and use static lib openssl

D4nte:
Fixes #1366

Issue #1131, opened 2019-07-23
Replace `zmq` with `BlockSource` implementation for HTTP API of bitcoind


thomaseizinger:
Goal of the implementation:
We no longer use zmq but HTTP API to get blocks.

Consider doing https://github.com/comit-network/comit-rs/issues/1348 at the same time / before because it might make your life easier.

da-kami:
Get latest block:

* latest block hash: https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md#chaininfos - use `bestblockhash`
* block by hash: https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md#blocks - use `hex` format and directly deserialise into `Block`

PR #1370, opened 2019-09-10
Statically link latest openssl lib in release binary

### LINKED ISSUE #1366: Can't run cnd/btsieve release binaries on Ubuntu 19.04
When running cnd's binary from the [first release](https://github.com/comit-network/comit-rs/releases/tag/0.1.0)
```
./cnd
```
I get the following error:
```
./cnd: error while loading shared libraries: libssl.so.1.0.0: cannot open shared object file: No such file or directory
```

The same thing happens for btsieve.

D4nte:
Fixes #1366

Please note openssl is the only problematic library.
ZMQ is also an issue but will be removed with #1131.
All other dynamically linked libraries are system libraries:
```
# ldd ./target/release/btsieve
    linux-vdso.so.1 =>  (0x00007fff7bfef000)
    libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fe6cd8fd000)
    libzmq.so.5 => /usr/lib/x86_64-linux-gnu/libzmq.so.5 (0x00007fe6cd697000)
    libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fe6cd493000)
    librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fe6cd28b000)
    libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fe6cd06e000)
    libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fe6cce58000)
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fe6cca8e000)
    /lib64/ld-linux-x86-64.so.2 (0x00007fe6ce7de000)
    libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fe6cc785000)
    libsodium.so.18 => /usr/lib/x86_64-linux-gnu/libsodium.so.18 (0x00007fe6cc527000)
    libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fe6cc1a5000)
# ldd ./target/release/cnd
    linux-vdso.so.1 =>  (0x00007ffdedffb000)
    libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f63589db000)
    libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f63587d7000)
    librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f63585cf000)
    libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f63583b2000)
    libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f635819c000)
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f6357dd2000)
    /lib64/ld-linux-x86-64.so.2 (0x00007f635a96f000)
    libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f6357ac9000)
```

The resulting binaries have been tested on:
- Ubuntu 19.094
- Latest debian
- Latest fedora

No issue encountered (except for ZMQ).

Fixes #1366 
Supersedes #1368

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,6 @@ jobs:
           command: |
             set -v
             cd ~/ && curl -O https://www.openssl.org/source/openssl-1.1.1c.tar.gz
-            # TODO: checksum, see https://www.openssl.org/source/openssl-fips-ecp-2.0.16.tar.gz.sha256
             tar xaf openssl-1.1.1c.tar.gz
             mkdir ~/openssl
             cd openssl-1.1.1c && ./config --prefix=$HOME/openssl -static
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,17 +56,18 @@ jobs:
       - run:
           name: "Compile openssl static libs"
           command: |
+            set -v
             cd ~/ && curl -O https://www.openssl.org/source/openssl-1.1.1c.tar.gz
             # TODO: checksum, see https://www.openssl.org/source/openssl-fips-ecp-2.0.16.tar.gz.sha256
-            tar xavf openssl-1.1.1c.tar.gz
+            tar xaf openssl-1.1.1c.tar.gz
             mkdir ~/openssl
-            cd openssl-1.1.1c && ./config --prefix ~/openssl -static
+            cd openssl-1.1.1c && ./config --prefix=$HOME/openssl -static
             make
             make install
       - run:
           name: Release build and test using cargo make
           command: |
-            export OPENSSL_DIR=~/openssl/
+            export OPENSSL_DIR=$HOME/openssl/
             export OPENSSL_STATIC=1
             export CARGO_MAKE_CARGO_BUILD_TEST_FLAGS="--all --release"
             export BTSIEVE_BIN=~/comit/target/release/btsieve
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,8 +54,20 @@ jobs:
       - install_node_devlibs
       - print_current_versions
       - run:
+          name: "Compile openssl static libs"
+          command: |
+            cd ~/ && curl -O https://www.openssl.org/source/openssl-1.1.1c.tar.gz
+            # TODO: checksum, see https://www.openssl.org/source/openssl-fips-ecp-2.0.16.tar.gz.sha256
+            tar xavf openssl-1.1.1c.tar.gz
+            mkdir ~/openssl
+            cd openssl-1.1.1c && ./config --prefix ~/openssl -static
+            make
+            make install
+      - run:
           name: Release build and test using cargo make
           command: |
+            export OPENSSL_DIR=~/openssl/
+            export OPENSSL_STATIC=1
             export CARGO_MAKE_CARGO_BUILD_TEST_FLAGS="--all --release"
             export BTSIEVE_BIN=~/comit/target/release/btsieve
             export CND_BIN=~/comit/target/release/cnd


.circleci/config.yml
```diff
@@ -53,9 +53,21 @@ jobs:
       - install_rust
       - install_node_devlibs
       - print_current_versions
+      - run:
```
bonomat: 🔧 Instead of having this step in here, we probably should add it to `Install node and development libraries` step.
Because, if I understand it correctly, we will now use a different ssl version for the release-build and for our debug-test build.

D4nte: yes correct, it uses a different openssl version for release-build and debug-build.

But then, we all use different versions locally anyway (mac does not use openssl but libressl).

Note that we run all e2e with the binary release in any case so we do validate that before releasing.

If we were to static build for debug-test, we would need to compile openssl for each CI build, which will significantly increase the build time by 3min30s.
I think this build time increase is not worth it.

.circleci/config.yml
```diff
@@ -53,9 +53,21 @@ jobs:
...
+          name: "Compile openssl static libs"
+          command: |
+            set -v
+            cd ~/ && curl -O https://www.openssl.org/source/openssl-1.1.1c.tar.gz
```
thomaseizinger: Is there a way of how we can make this a more obious dependency?
Having to remember that we need to update this version string here regularly seems like a PITA :(
What are we even using openssl for in cnd/btsieve? Can we maybe feature flag that away so that we don't even depend on this?

D4nte: > Is there a way of how we can make this a more obvious dependency?
What do you mean?

> Having to remember that we need to update this version string here regularly seems like a PITA :(
True but I don't see how we can do better

> What are we even using openssl for in cnd/btsieve? Can we maybe feature flag that away so that we don't even depend on this?

There is actually a `vendored` feature flag on `rust-ssl` crate that would give the same result (as far as I understand) and make all this stuff easier.

I quickly check why we depend on ssl when doing this issue:
```
openssl < native-tls < reqwest*
                     < tokio-tls < websocket < web3*
                     < web3*
                     < websocket < web3*
                     < hyper-tls < reqwest*
                                 < web3*
```

So in short, it's brought it by `reqwest` and `web3` crate.

## reqwest
Reqwest has a [`default-tls-vendored` feature](https://github.com/seanmonstar/reqwest/blob/87a09322d6fef48e7a7c79a854acaf08bb16f606/Cargo.toml#L78) that could help.
Also, all tls crates are optional so maybe we could fully remove tls by setting features to `[]`.

## web3
No `vendored` option but the [TLS crates are optionals](https://github.com/tomusdrw/rust-web3/blob/db4443ee16765c0754a87d9359d8f6b9e0c04d3a/Cargo.toml#L45) so same than reqwest.

---

trying with `features = []` for `web3` and `reqwest` now.

D4nte: I am not it is possible to go down the features rabbit hole due to the dependency on testcontainers that also bring tls in.

D4nte: Please note that because we use ubuntu LTS in our CI, we don't even use the latest openssl library (1.1) as only 1.0 is available on ubuntu 16.04.

Hence, I'd be keen to have a follow-up ticket to (try) remove the openssl dependency.

In any case, we do not really on openssl functionally so I am not worried about upgrading openssl version.

What I am worried about is that if someone download our binary right now, it is likely to not work without them messing around symlink in their `/lib` (and installing libzmq), meaning our binary release is pretty much useless as it would actually be more straightforward to compile from source than fix the dynamic lib error they'd have.

D4nte: Ok, the features path lead nowhere. Most likely because test containers also bring openssl in.

This is the best I can do in a reasonable amount of time.

thomaseizinger: Nice research, thanks! :)
Agree on the timing problem! Let's fix it fast and have an issue that tracks how to fix it properly (by removing ssl dependency).
I am curious on why testcontainers brings in SSL, will investigate! :)

.circleci/config.yml
bonomat: Fair, thanks for the clarification.

.circleci/config.yml
D4nte: testcontainers bring in SSL because of web3 and also because of our bitcoin_rpc_client which bring in json_client that brings in tls and hence ssl.

PR #1373, opened 2019-09-11
Remove testcontainers dev dependency from cnd

luckysori:
`testcontainers` is an unused dev dependency in cnd.


cnd/Cargo.toml
```diff
@@ -70,5 +70,4 @@ spectral = "0.6"
 tc_bitcoincore_client = { path = "../internal/tc_bitcoincore_client" }
```
D4nte: 📣 Also removed the `tc_*` crates.

PR #1376, opened 2019-09-11
Remove openssl static build in release CI

### LINKED ISSUE #1372: Remove dependency on openssl
To ensure that our Linux binaries can be easily executed on most Linux distros, we statically link openssl libraries with #1370. However, openssl is actually not needed by comit-rs as we do not use SSL/TLS at all.

openssl is a coincidental dependency brought in by various crates include web3, testcontainers & reqwest. See [this comment thread](https://github.com/comit-network/comit-rs/pull/1370#discussion_r322575344) for more details.

In this ticket you should try to remove openssl dependency of cnd and btsieve

DoD:
- no `openssl` present in the Cargo.lock
- `ldd cnd` & `ldd btsieve` (Linux, try `otool -L cnd` on Mac) must not display a libssl or libcrypto entry.
- No more building of openssl as part of the release build.

Note:
Possible strategies:
- Only exclude openssl from web3 and reqwest. Try to ensure that testcontainers is not brought in for release mode
- Exclude openssl from testcontainers web3 and reqwest

You may want to wait #887 to be done as it will be easier to control openssl dependencies per crate.

D4nte:
It does not fully remove the dependency as `internal/comit_i` needs TLS to fetch the comit-i bundle.
For some reason, this makes `cnd` executable linked to openssl even though it's only needed in the build.
To resolve that, we use the `vendored-tls` feature of `reqwest` for `internal/comit_i` which statically link openssl.

For the rest, `web3` and other `reqwest` imports, we deactivate default features to not depend on openssl.

openssl is still needed to test of course, but the resulting release binaries do not link to openssl:

```
# ldd target/release/cnd
        linux-vdso.so.1 (0x00007ffe03ef2000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f441c73e000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f441c536000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f441c317000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f441c0ff000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f441bd0e000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f441e5b0000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f441b970000)

# ldd target/release/btsieve
        linux-vdso.so.1 (0x00007fff47f99000)
        libzmq.so.5 => /usr/lib/x86_64-linux-gnu/libzmq.so.5 (0x00007faa616c6000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007faa614c2000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007faa612ba000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007faa6109b000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007faa60e83000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007faa60a92000)
        /lib64/ld-linux-x86-64.so.2 (0x00007faa622ff000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007faa606f4000)
        libsodium.so.23 => /usr/lib/x86_64-linux-gnu/libsodium.so.23 (0x00007faa604a3000)
        libpgm-5.2.so.0 => /usr/lib/x86_64-linux-gnu/libpgm-5.2.so.0 (0x00007faa60257000)
        libnorm.so.1 => /usr/lib/x86_64-linux-gnu/libnorm.so.1 (0x00007faa5ff27000)
        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007faa5fb9e000)
```

Resolves #1372

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,20 +54,8 @@ jobs:
       - install_node_devlibs
       - print_current_versions
       - run:
-          name: "Compile openssl static libs"
-          command: |
-            set -v
-            cd ~/ && curl -O https://www.openssl.org/source/openssl-1.1.1c.tar.gz
-            tar xaf openssl-1.1.1c.tar.gz
-            mkdir ~/openssl
-            cd openssl-1.1.1c && ./config --prefix=$HOME/openssl -static
-            make
-            make install
-      - run:
           name: Release build and test using cargo make
           command: |
-            export OPENSSL_DIR=$HOME/openssl/
-            export OPENSSL_STATIC=1
             export CARGO_MAKE_CARGO_BUILD_TEST_FLAGS="--all --release"
             export BTSIEVE_BIN=~/comit/target/release/btsieve
             export CND_BIN=~/comit/target/release/cnd
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -302,11 +302,6 @@ dependencies = [
 
 [[package]]
 name = "bitflags"
-version = "0.9.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
-name = "bitflags"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
@@ -2470,6 +2465,14 @@ version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "openssl-src"
+version = "111.5.0+1.1.1c"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "cc 1.0.26 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "openssl-sys"
 version = "0.9.49"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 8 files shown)


D4nte:
> I think what you are facing is that cargo features are accumulated across all dependencies. i.e. as soon as one part of the dependency graph needs certain features, they are activated for all of them.
> 
> Maybe there are other ways to avoid this for runtime/build/devDependencies ..
> We could potentially solve that by using a different library for fetching the `comit_i` release. A different version of reqwest might also work :)
> While this is certainly an improvement, I think cnd and btsieve now just ship of openssl being statically embedded :(

Yes I must say I am surprised that a build dependency leads to the final binary linking to this lib.

Yes, you are absolutely right, this is due to the fact that a crate is imported with the same feature flag across all dependencies. It is not possible to use the same crate twice with different feature flags.

Yes, an alternative would be to use a different crate than reqwest for `internal/comit_i`.

PR #1371, opened 2019-09-10
bitcoin-d HTTP API

da-kami:
completely remove `zmq`. Use bitcoin-d HTTP API instead

PR #1378, opened 2019-09-11
Couple of improvements to the BitcoindHttpBlockSource

thomaseizinger:
While reviewing #1371, I noticed a couple of things that could be optimized.
Here is my attempt in doing so! :)

My main motivation was the implementation of the `remove_trailing_newline` function because it can actually we be replaced with `.trim()` since newlines are whitespaces characters. However, that led to it no longer being tested or at least the tests were no longer useful. Hence I got example responses from bitcoind and extracted the whole response decoding into a function again which makes to test (because the test actually fails if we don't include the `trim`).

I am opening this PR directly against #1371, however, feel free to merge #1371 first and then re-target this to master! Whatever you prefer :)

The PR contains three commits, each explaining why I am suggesting certain changes.


thomaseizinger:
@da-kami since this was not merged into the original pull request, I changed the base to master now and resolved the conflicts :)

Either merge or close if you disagree with the proposed changes!

PR #1379, opened 2019-09-12
Update installation instructions

thomaseizinger:
With PR #1376, we no longer need to have openssl installed for development since it is compiled automatically.

This PR also removes the installation instruction for zmq, which is actually only done in #1371. However, otherwise the joke would have not worked.


D4nte:
I'll approve once #1371 is merged.

PR #1380, opened 2019-09-12
Replace Reqwest with ureq

D4nte:
Removed dependency on openssl as ureq uses rustls.

On `ubuntu:latest` Docker image (19.04):

```
## Before:

▶ du -sh ./target/release/{cnd,btsieve}
33M     ./target/release/cnd
12M     ./target/release/btsieve
▶ du -s ./target/release/{cnd,btsieve}
33544   ./target/release/cnd
12104   ./target/release/btsieve


## After:

▶ du -sh ./target/release/{cnd,btsieve}
31M     ./target/release/cnd
12M     ./target/release/btsieve
▶ du -s ./target/release/{cnd,btsieve}
30728   ./target/release/cnd
12088   ./target/release/btsieve
```

2M saved on `cnd`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -92,6 +92,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "ascii"
+version = "0.9.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "asn1_der"
 version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


thomaseizinger:
OMG! So nice! Thank you @D4nte :)

I didn't release `rustls` was a pure Rust implementation, very nice :)
`reqwest` also has a `rustls` feature btw: https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml#L54

Could have maybe also used that :)

D4nte:
> OMG! So nice! Thank you @D4nte :)
> 
> I didn't release `rustls` was a pure Rust implementation, very nice :)
> `reqwest` also has a `rustls` feature btw: [seanmonstar/reqwest:Cargo.toml@`master`#L54](https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml#L54) [](/seanmonstar/reqwest/blob/HEAD@{2019-09-12T05:01:14Z}/Cargo.toml#L54)
> 
> Could have maybe also used that :)

`ureq` is good enough for what we do :)

thomaseizinger:
> ureq is good enough for what we do :)

That might be the case for all our uses of HTTP requests 😂

PR #1387, opened 2019-09-13
Pre deterministic ids refactoring

tcharding:
Initial attempt at doing some refactoring to aid the deterministic id PR.

The `matches` functions got a re-working.

Done in preparation for this PR: https://github.com/comit-network/comit-rs/pull/1389

### DIFF
--- a/btsieve/src/ethereum/queries/transaction.rs
+++ b/btsieve/src/ethereum/queries/transaction.rs
@@ -124,7 +124,7 @@ mod tests {
             let query = TransactionQuery {
                 from_address: Some(from_address),
                 to_address: None,
-                is_contract_creation: Some(true),
+                is_contract_creation: true,
                 transaction_data: None,
                 transaction_data_length: None,
             };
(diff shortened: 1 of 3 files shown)


btsieve/src/ethereum/queries/transaction.rs
```diff
@@ -18,46 +18,42 @@ use serde::{Deserialize, Serialize};
 pub struct TransactionQuery {
     from_address: Option<Address>,
     to_address: Option<Address>,
-    is_contract_creation: Option<bool>,
```
D4nte: Why?

cnd/src/btsieve/ethereum.rs
```diff
@@ -15,7 +15,7 @@ pub enum EthereumQuery {
     Transaction {
         from_address: Option<Address>,
         to_address: Option<Address>,
-        is_contract_creation: Option<bool>,
```
thomaseizinger: This change is semantically incorrect :)
We need three states: None, true and false.

With the proposed change, we cannot express a query where we "don't care" about whether it is a contract creation or not.

btsieve/src/ethereum/queries/transaction.rs
thomaseizinger: @tcharding 

This change is semantically incorrect :)
We need three states: None, true and false.

With the proposed change, we cannot express a query where we "don't care" about whether it is a contract creation or not.

PR #1400, opened 2019-09-13
Embed migrations through `diesel_migrations` crate

thomaseizinger:
I've noticed that there was probably a communication issue of what I meant with "embedding migrations".
Here is a PR that uses the `embed_migrations!` macro provided by `diesel_migrations` to embed the migrations folder of `cnd` into the resulting binary.
This gives us a module `embedded_migrations` that allows us to run the migrations against a database given a connection to it.

Developers are expected to use the `diesel_cli` to create these migrations through the `diesel migration generate XYZ` command. Make sure to do this from the `cnd` folder.


thomaseizinger:
cc @tcharding

PR #1401, opened 2019-09-13
Introduce `libp2p-comit` crate

### LINKED ISSUE #1161: Merge `bam` crate and `libp2p_bam` into `libp2p-comit` crate
Currently, we still use the term "BAM" to refer to the message structs that we are using inside our network layer.

With the proposed changes to RFC001 here https://github.com/comit-network/RFCs/issues/65, the term `BAM` will be gone. Once that is the case, we should create a dedicated `libp2p-comit` crate that encapsulates the whole implementation of the protocol over libp2p.

The name is inspired by the remaining libp2p crates which are all named `libp2p-X` where X is some component.

Result:
- `libp2p-comit` crate should be in the root
- `libp2p-comit` should not depend on `cnd`, hence, the types to represent COMIT messages and their elements, that are dedicated to be sent over the wire, will need to be moved in `libp2p-comit`. This is to be investigated/confirmed before proceeding.

Note: Beware of circular dependencies, hence the investigation needed for the types.

thomaseizinger:
This crate basically implements the `/comit/1.0.0` protocol defined in RFC001.

With this PR, the term "bam" is now officially gone from the code base! :tada: 

Resolves #1161.


README.md
```diff
@@ -10,14 +10,15 @@ This is a reference implementation for the COMIT protocol.
...
-The repository contains several main folders: `cnd`, `btsieve`, `blockchain_contracts` and `internal`.
+The repository contains several main folders: `cnd`, `btsieve`, `blockchain_contracts`, `libp2p-comit` and `internal`.
 
 ### Root crates
 
 Crates at the root of the repository hold primary binaries and libraries:
 - `cnd`: the COMIT Node Daemon binary.
 - `btsieve`: the binary that sifts through Blockchain Transactions and blocks from the network.
 - `blockchain_contracts`: the library that contains the HTLCs, to be migrated to its own repo once some dependencies are sorted.
+- `libp2p-comit`: implementation of the comit messaging protocol on top of libp2p
```
D4nte:
```suggestion
- `libp2p-comit`: implementation of the COMIT messaging protocol on top of libp2p
```

PR #1402, opened 2019-09-13
Rephrase the mainnet disclaimer

thomaseizinger:
The previous wording read a bit silly with the three exlamation marks at the end. I also figured that adding two sentences on _why_ it is not mainnet ready might be appreciated by users.


README.md
```diff
@@ -6,7 +6,10 @@
 COMIT is an open protocol facilitating trustless cross-blockchain applications.
 This is a reference implementation for the COMIT protocol. 
 
-## WARNING - We do not recommend running COMIT on mainnet for now!!!
+## DISCLAIMER: This is not mainnet ready!
+
+- Extensive testing on mainnet from our side still remains to be done
+- Several features for achieving production-ready robustness are still under works
```
D4nte:
```suggestion
- Several features for achieving production-ready robustness are still under works
- Once we consider comit-rs production ready, we will release version 1.0.0
```

thomaseizinger:
> When I read the description I thought you'd mention a few points we think need to be done first before we think this is mainnet ready...

Does it not? :)
Extensive testing and robustness.

bonomat:
> Does it not? :)
> Extensive testing and robustness.

Still a bit too vague imho but definitely better than before :)

thomaseizinger:
> > Does it not? :)
> > Extensive testing and robustness.
> 
> Still a bit too vague imho but definitely better than before :)

@bonomat It is vague, I agree but I didn't want to include too many details so that we constantly have to update it as we change/refine the direction :)

Issue #1364, opened 2019-09-07
Refactor how we deserialize the body of HTTP requests


thomaseizinger:
At the moment, we use serde's untagged enums to deserialize the request body of the `POST /swaps/rfc003/` endpoint.

This worked well initially because we didn't have to write a lot of code to get it done.

However, this approach has drawbacks:

1. We cannot return a useful error to the user if we fail to deserialize (https://github.com/comit-network/comit-rs/issues/754)
2. We cannot customize the deserialization easily. Switching to the new Amount type of rust-bitcoin highlighted this very well: https://github.com/comit-network/comit-rs/pull/1303

Instead of having a `SwapRequestBody` that is generic of assets and ledgers, we should just use a dedicated enum for those fields. Something like:

```rust
enum HttpAssetKind {
    Bitcoin(BitcoinAmount),
    Ether(EtherQuantity),
    ...
}
```

Note that we can still make use of the `Http<T>` type if we want to customize the deserialization of a foreign type here. For example:

```rust
enum HttpAssetKind {
    Bitcoin(Http<BitcoinAmount>),
    Ether(EtherQuantity),
    ...
}
```

Most importantly, this approach gives use control over the _structure_ that we want to serialize/deserialize while allowing us to compose the different types together the way we want.

Since `HttpAssetKind` is a type which is only to be used in the `http_api` module, it is welcome to derive `Serialize` and `Deserialize`. Serde's [Internally tagged](https://serde.rs/enum-representations.html#internally-tagged) feature is probably really useful here to produce the structure that we want in the API.

Issue #1327, opened 2019-09-03
COMIT as a library (on crates.io)


thomaseizinger:
This epic tracks all issues that are related to separating out the core parts of `cnd` into a `comit` library.

Requirements:
- A library released on crates.io cannot have path dependencies

Scope:
- Everything defined in the RFCs would be implemented as part of this library, every functionality _not_ specified in RFCs would need to go into the application

Other notes:
- Need extensive testing on the API of the library (probably replacing our current e2e tests)

bonomat:
The main effort was to remove path dependencies and that is done.

PR #1404, opened 2019-09-13
Extract the core of `cnd` into a `comit` crate

thomaseizinger:
This is a work-in-progress PR for achieving #1327. 

The main reason I started working on it was because I wanted to see how far we can actually go an what the blockers are for achieving this.

I'll open this as a long-running draft PR and continue to update it every now and then (hopefully that won't be too much of a pain :D).

Additionally, I'll block this on certain (technical-debt) issues that we already have where it is now obvious that we have to do them before tackling this:

- #1364: Our `Asset` and `Ledger` trait definitions depend on the `FromHttp{Asset,Ledger}` and `ToHttp{Asset,Ledger}` traits. Circular dependencies like this are impossible outside of a single crate.
- #1406: We cannot extract a `comit` library as long as the swap state machine is actually doing the communication. What we probably want is for the `comit` library to define the messages but not the actual communication.


thomaseizinger:
Closing this for now as there is no value in being open all the time.

PR #1415, opened 2019-09-18
Send SwapId as part of Swap Request

### LINKED ISSUE #1414: Send SwapId to other party as part of SWAP REQUEST
Send ID generated by Alice to Bob as part of swap request.

thomaseizinger:
Fixes #1414.


api_tests/dry/multiple_peers.ts
```diff
@@ -82,6 +82,21 @@ import "../lib/setup_chai";
             aliceSwapWithBobHref.should.be.a("string");
```
D4nte: Isn't there more e2e test simplifications? For the test where we look for the swap on Bob side after Alice sent it?

thomaseizinger:
> I believe that the bitcoin transient keys are generated by having swap id + a secret concatenated and hashed.

@D4nte The initial instance of `Seed` is in the config file. From there, we derive a "per-swap" `Seed` by incorporating the bytes of the SwapId.

Unless two people have the same root seed, using the same SwapId should not be an issue on both sides.

api_tests/dry/multiple_peers.ts
thomaseizinger: Probably, I'll create a follow-up ticket for that if you don't mind!

thomaseizinger: Actually, this is going away with the new e2e tests so I think we don't need to change it now :)

PR #1416, opened 2019-09-19
Add a Dockerfile for cnd and btsieve

### LINKED ISSUE #1341: Create docker images for `cnd` and `btsieve` for releases
Do this:

- where to build the container
  - Dockerhub (free)
- parent image
  - Ubuntu (works)
- how config files are handled
  - Hard coded default config file in a volume.
- default entrypoint
  - binaries (nothing else makes sense)
- one vs two Dockerfiles
  - two dockerfiles (simple)

thomaseizinger:
Resolves #1341.


.docker/btsieve/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM rust:1.37-buster as builder
+
+WORKDIR /src
+COPY ./btsieve ./btsieve
```
D4nte: Instead, use a `.dockerignore` file to exclude `api_tests`, `target`, etc.
This would allow you to do `COPY ./ ./`.
See https://docs.docker.com/engine/reference/builder/

.docker/btsieve/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY ./internal ./internal
+COPY ./cnd ./cnd
+COPY ./libp2p-comit ./libp2p-comit
+COPY ./blockchain_contracts ./blockchain_contracts
+COPY ./Cargo.toml ./Cargo.toml
+COPY ./Cargo.lock ./Cargo.lock
+
+RUN cargo check
```
D4nte: I understand you want to fail fast but considering how slow is docker and the fact that this would only be done for a release, I think it can be skipped

.dockerignore
```diff
@@ -0,0 +1,7 @@
+target/
```
D4nte: You already have the dockerignore file, why not doing `COPY ./ ./` then?

.docker/btsieve/Dockerfile
thomaseizinger: I did that first yes but that causes problems with the `btsieve.toml` config file.
I can give it another go :)

.docker/btsieve/Dockerfile
thomaseizinger: Can potentially be skipped yes but it also doesn't include _that_ much overhead since most time is spent downloading dependencies :(

.dockerignore
thomaseizinger: Will give it another try but it caused problems with poisoned caches :(

thomaseizinger: Actually, I'd categorize this as an optimisation and do it later. Fine for you?

D4nte: Yes it's just a risk because if we move the crates around we may notice that we forgot to update the dockerfile very late in the process (at release time)

D4nte:
@thomaseizinger you also mentioned something about pre-releases, what do you mean? Or is it only a docker hub setup?

.docker/cnd/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    libsqlite3-dev \
+ && rm -rf /var/lib/apt/lists/*
+
+RUN useradd --create-home --shell /bin/bash cnd
+USER cnd
+
+COPY --from=builder /src/target/release/cnd /usr/local/bin
+
+EXPOSE 9939
+EXPOSE 8000
```
da-kami: 🙃 	The Web-UI (Comit-i) on `8080` is not exposed by default although it is part of the default `toml` - Since we don't do much on Comit-i at the moment this is probably wanted.

thomaseizinger:
> Just an idea for the future if we don't want to build anymore:
> It seems like there is an event triggered if a new release was attached to a repository. We could build an GitHub action which then triggers the docker build on DockerHub whenever there is a new release.
> 
> [developer.github.com/v3/activity/events/types/#releaseevent](https://developer.github.com/v3/activity/events/types/#releaseevent)

@bonomat Great idea! Can you make a ticket for that? :)

thomaseizinger:
> @thomaseizinger you also mentioned something about pre-releases, what do you mean? Or is it only a docker hub setup?

@D4nte yes, it is part of the build setup on DockerHub.

D4nte:
Release candidates are usually `-rc` not `-RC`. (see https://semver.org/)

Issue #1421, opened 2019-09-19
Fix the error handling in the `db` module


tcharding:
I have started this work already, adding issue to fit my work flow in with the team process and also to help keep me in scope.  This is a pre-issue ticket for an issue already in this sprint, pulling this one straight into the sprint.  There will be more Sqlite issues to come.

https://github.com/comit-network/comit-rs/issues/1243

The database module got merged with messy error handling as a result of Tobin's inexperience.  This needs cleaning up

Tasks

- Add custom `Error` to db (Sqlite)
- Have a single error variant in metadata_store that holds the Sqlite Error
- Define custom error variants for all error cases in `db` and embed errors in them i.e., don't use `Io(String` use instead `Io(io::Error)`

PR #1426, opened 2019-09-20
Replace String associated data in Error variants

tcharding:
Currently database errors are define in the metadata_store module, they
should be in the database module.  We can then just have a single Error
enum variant in metadata_store with the Sqlite error as its associated
type.  Secondly, the error variants all use String as their associated
type, this is lazy, we can do better than this.

- Move all the Sqlite specific Error enum variants to the database
module.

- Create a single Error enum variant in metadata_store that has an
Sqlite error as its associated data.

- Remove the use of String types from associate data for the custom
Error used by the db module.  If there is another type of error
available use that as the associated type.

- Use `use crate::metadata_store as ms` to make the code less verbose.

Resolves: #1421


cnd/src/db/mod.rs
```diff
@@ -107,35 +141,37 @@ fn create_database_if_needed(db: &Path) -> Result<(), Error> {
...
-
-    SqliteConnection::establish(&database_url).map_err(|err| Error::Connect(err.to_string()))
+    let database_url = db.to_str().ok_or(Error::PathStr)?;
+    let conn = SqliteConnection::establish(&database_url)?;
+    Ok(conn)
 }
 
-impl TryFrom<Metadata> for metadata_store::Metadata {
-    type Error = metadata_store::Error;
+impl TryFrom<Metadata> for ms::Metadata {
```
luckysori: :upside_down_face: Could be clearer.
```suggestion
impl TryFrom<db::models::Metadata> for metadata_store::Metadata {
```
Not too sure about using an acronym, but it's a minor thing.

tcharding: Discussed offline; aliasing (acronym) for package makes the code clearer here because it makes the code less cluttered without introducing unnecessary extra cognitive load on the developer.  This is different to aliasing types which causes grief in typical IDE workflow.  Thanks.

cnd/src/db/mod.rs
```diff
@@ -107,35 +141,37 @@ fn create_database_if_needed(db: &Path) -> Result<(), Error> {
...
-        let swap_id = SwapId::from_str(md.swap_id.as_str())?;
-        let alpha_ledger = LedgerKind::from_str(md.alpha_ledger.as_str())?;
-        let beta_ledger = LedgerKind::from_str(md.beta_ledger.as_str())?;
-        let alpha_asset = AssetKind::from_str(md.alpha_asset.as_str())?;
-        let beta_asset = AssetKind::from_str(md.beta_asset.as_str())?;
-        let role = Role::from_str(md.role.as_str())?;
-        let counterparty = PeerId::from_str(md.counterparty.as_str()).unwrap();
-
-        Ok(metadata_store::Metadata {
+        // These map_err calls can be removed once Metadata uses types and
```
D4nte: Is there an issue to track this comment?

PR #1439, opened 2019-09-22
Remove Compare trait

tcharding:
Currently we have a custom trait `Compare` that is bound to the `Asset` trait.  This is unnecessary because all the types we use as assets implement the `std::cmp::Ord` trait, we can use it instead.

The unit tests are now probably redundant because they test the trait implementation of the asset libraries, I'm not totally sure on best practice in this regard - so I left them in.


D4nte:
> Currently we have a custom trait `Compare` that is bound to the `Asset` trait. This is unnecessary because all the types we use as assets implement the `std::cmp::Ord` trait, we can use it instead.
> 
> The unit tests are now probably redundant because they test the trait implementation of the asset libraries, I'm not totally sure on best practice in this regard - so I left them in.

Can you please delete the unit tests?

thomaseizinger:
> > Currently we have a custom trait `Compare` that is bound to the `Asset` trait. This is unnecessary because all the types we use as assets implement the `std::cmp::Ord` trait, we can use it instead.
> > The unit tests are now probably redundant because they test the trait implementation of the asset libraries, I'm not totally sure on best practice in this regard - so I left them in.
> 
> Can you please delete the unit tests?

I cshould change them to tests that assert that the `Ord` trait is implemented:

```rust
fn assert_ord<T: Ord>(_t: T) {}

#[test]
fn bitcoin_amount_is_ord() {
	assert_ord(Amount::from_btc(1.0).unwrap())
}
```

This will generate compile errors if we ever remove `Ord`. I don't mind either way :)

tcharding:
> I cshould change them to tests that assert that the Ord trait is implemented
> ...
> This will generate compile errors if we ever remove Ord

Can you explain what benefit this adds please?  The compiler will already complain if we take away `Ord` and still call `cmp()`.  If we do not call `cmp()` we do not need `Ord`?

tcharding:
And we have
```
pub trait Asset:
    Clone
    + Copy
    + Debug
    + Display
    + Send
    + Sync
    + 'static
    + PartialEq
    + Eq
    + Hash
    + FromHttpAsset
    + Into<AssetKind>
    + Ord
{
}

impl Asset for Amount {}

impl Asset for EtherQuantity {}

impl Asset for Erc20Token {}
```

tcharding:
Merging without your test @thomaseizinger, can do a follow up PR to add it - just say the word.

thomaseizinger:
> > I cshould change them to tests that assert that the Ord trait is implemented
> > ...
> > This will generate compile errors if we ever remove Ord
> 
> Can you explain what benefit this adds please? The compiler will already complain if we take away `Ord` and still call `cmp()`. If we do not call `cmp()` we do not need `Ord`?

Yeah fair enough, it is a question of whether we want to assert it explicitly should be fine :)

Issue #1106, opened 2019-07-17
Move to a PoW engine for all tests involving an Ethereum node


thomaseizinger:
Currently we are using the `instantSeal` engine of parity whenever we need an Ethereum node to test against.

However, that is quite brittle as we perceived while developing the `EthereumWallet` for bobtimus. In addition, this causes a divergence in how we structure our e2e tests, especially the `refund` tests where we need to account for time passing.

Moving to a PoW-based engine would allow us to continuously generate blocks which is hopefully more resilient.

Labelling this with `bug` because it may be involved in our build to be flacky at the moment.

Tasks:
- Use a PoW-engine for the Ethereum node in the e2e tests
- Convert the e2e tests to "wait until" the appropriate time is reached, same as we currently do for Bitcoin
- Use a PoW-engine for the Ethereum node in the `blockchain_contracts` tests

thomaseizinger:
To be re-discussed once we have more data based on: #1119.

bonomat:
Note: when doing this, also fix some ugly sleeps in api_tests/lib/ethereum.ts

bonomat:
Note: this ticket was labeled as a *bug* because btsieve was missing blocks due to the instant seal of parity's `dev` setting. Since btsieve can't go into the past, this was indeed an issue. 
However, with the sleep introduced some time ago, this is not so criticial anymore. 
I've removed the *bug* label for now. 

Here some of my findings: 

PoW is not suitable for our needs due its nature of being *proof of work*, i.e. it computes a cryptographic puzzle for each block. The only way to generate blocks every X seconds is by configuring the difficulty, however, this means that the CPU will be heavily used for these X seconds. If we want to make the puzzle simpler, we would just find blocks faster. 

--> PoA is the only alternative. 

I tried to setup a parity node with a PoA consensus which generated a block every 1 second. However, I had the negative side effect that some of our tests timed out due to the lack of `waits` for transactions to being mined. 

I'm about to fix this but before finishing this properly, we will need to release a docker container for a custom PoA network. 
Do we want to go this path or do we want to stick with parity's `dev` chain?

thomaseizinger:
Resolved with https://github.com/comit-network/comit-rs/pull/1441.

PR #1441, opened 2019-09-23
Use parity proof of authority

bonomat:
Note:
* Due to the removal of auto-seal we need to wait for blocks from time to time.
* btsieve tests can have multiple results due to re-matching queries on blocks btsieve saw already
* the new docker container can be found in: https://github.com/coblox/parity-poa


api_tests/lib/ethereum.ts
```diff
@@ -90,20 +94,21 @@ export class EthereumWallet {
...
         const parityPrivateKey =
             "0x4d5db4107d237df6a3d58ee5f70ae63d73d7658d4026f2eefd2f204c81682cb7";
 
         const weiAmount = ethers.utils.parseEther(ether);
-
         const tx = {
             to: this.address(),
             value: weiAmount.toHexString(),
             gasLimit: 21000,
+            chainId: 17,
```
thomaseizinger: Should we fetch the chainId through `ethers` to make sure it is correct?

api_tests/lib/ledger_runner.ts
```diff
@@ -171,15 +171,8 @@ async function startBitcoinContainer(): Promise<StartedTestContainer> {
...
-        .withCmd([
-            "--config=dev",
-            "--jsonrpc-apis=all",
-            "--unsafe-expose",
-            "--tracing=on",
-            "--logging=debug,ethcore-miner=trace,miner=trace,rpc=trace,tokio_core=warn,tokio_reactor=warn",
-            "--jsonrpc-cors='all'",
-        ])
+    return new GenericContainer("coblox/parity-poa", "v2.5.8-stable")
+        .withCmd(["--config=/parity/config/config.toml"])
```
thomaseizinger: Why do we need to pass this?
We do we no longer pass all the other configuration options?

bonomat: All the config is now in the `config.toml` which is included in the container and we need to tell parity which config file to load.

api_tests/lib/ethereum.ts
bonomat: 👍Done

api_tests/lib/ledger_runner.ts
thomaseizinger: Can we not default to that _inside_ the image?

da-kami: You mean you would put it hardcoded into the container upon container creation and then just point to it inside the container?

bonomat: Could probably, but then it's literally a single-use container. I'd assume, once specified, you can't overwrite it from the outside.

D4nte:
>     * btsieve tests can have multiple results due to re-matching queries on blocks btsieve saw already

Should we have a btsieve ticket to track that?

bonomat:
> > ```
> > * btsieve tests can have multiple results due to re-matching queries on blocks btsieve saw already
> > ```
> 
> Should we have a btsieve ticket to track that?

This will go away with the new update of btsieve where each query fetches the blocks.

D4nte:
>     * the new docker container can be found in: https://github.com/coblox/parity-poa

The link is correct yet the repo has a typo in the title: It is "docker-partiy-poa", should be "docker-parity-poa"

api_tests/lib/ledger_runner.ts
thomaseizinger: See this ticket here: https://github.com/coblox/docker-parity-poa/issues/3

api_tests/lib/ledger_runner.ts
```diff
@@ -171,15 +171,8 @@ async function startBitcoinContainer(): Promise<StartedTestContainer> {
 }
 
 async function startEthereumContainer(): Promise<StartedTestContainer> {
-    return new GenericContainer("parity/parity", "v2.5.0")
-        .withCmd([
-            "--config=dev",
```
thomaseizinger: Reposting my comment here since it got ignored in the other thread:

We do we no longer pass all the other configuration options?

api_tests/lib/ledger_runner.ts
```diff
@@ -171,15 +171,8 @@ async function startBitcoinContainer(): Promise<StartedTestContainer> {
...
-            "--jsonrpc-apis=all",
```
thomaseizinger: Reposting my comment here since it got ignored in the other thread:

We do we no longer pass all the other configuration options?

thomaseizinger: cc @bonomat

bonomat: Tried to answer this in the other thread :) : 
All the configs are set in the config file on default. So there is no need to pass in them here again.

thomaseizinger: Alright, I misunderstood, thanks!
I am not 100% if that is good but since any PoA network is probably used for testing, it should be fine.

thomaseizinger: It might be a bit surprising to users so we should probably mention it in the README of the Docker container. At the moment, it only says that it is standard parity + PoA but actually it is standard parity + PoA + certain logging + exposed APIs.

bonomat: Fair, updated the docker readme.

thomaseizinger:
Once this is merged, I think we should have a follow-up ticket that changes the wait logic for the Ethereum refund tests.

Thoughts? @bonomat @D4nte

bonomat:
> Once this is merged, I think we should have a follow-up ticket that changes the wait logic for the Ethereum refund tests.

what in particular would you change there?

thomaseizinger:
> > Once this is merged, I think we should have a follow-up ticket that changes the wait logic for the Ethereum refund tests.
> 
> what in particular would you change there?

Change this:

https://github.com/comit-network/comit-rs/blob/f0b3e2cf1c32b0469c05bee6135c995836290f82/api_tests/lib/actor.ts#L238-L259

to match the logic we do for Bitcoin:

https://github.com/comit-network/comit-rs/blob/f0b3e2cf1c32b0469c05bee6135c995836290f82/api_tests/lib/actor.ts#L189-L215

Basically, now that we continuously generate blocks on Ethereum, we no longer have to rely on systemtime to wait for refund but can actually rely on the blockchain time.

bonomat:
@thomaseizinger : created a ticket for this in #1447

api_tests/lib/ledger_runner.ts
D4nte: I am not sure I agree with the strategy.

**Inside** the docker you have a config file but you need to tell the docker from the **outside** that such file should be used.

To me a proper solution is either:
1. The docker load the file automatically
2. The docker does not contain the file but you mount a volume to pass a file you generate from a client side.

thomaseizinger: The ticket I created and linked above is about implementing solution 1 as a follow-up of this. We can also do that first and block this PR on that if we want to have it in one go. I don't mind either as long as the current situation is not the final result :)

thomaseizinger: Note that, as long as the file is in the default location, you can still override it from the outside with --config and a mounted volume!

D4nte: ok, @bonomat feel free to discard my "request changes".

bonomat: Let's do it properly and wait for https://github.com/coblox/docker-parity-poa/pull/5
Once the container is released I will update this PR so that it uses the default already.

api_tests/lib/ethereum.ts
```diff
@@ -188,16 +192,16 @@ export class EthereumWallet {
...
     }
 
     public async signAndSend(tx: TransactionRequest) {
-        await sleep(1001); // instant seal is a beautiful thing, if we fire transactions too fast, than btsieve can't see the block
         const wallet = new ethers.Wallet(this.keypair.privateKey, ethersClient);
-        const transactionResponse = await wallet.sendTransaction(tx);
-        return transactionResponse.wait(1);
+        tx.nonce = await wallet.getTransactionCount("latest");
+        const signedTx = await wallet.sign(tx);
+        return await ethersClient.sendTransaction(signedTx);
```
D4nte:
```suggestion
        return ethersClient.sendTransaction(signedTx);
```

luckysori:
Did this fix #1106? Because that one is still open.

thomaseizinger:
> Did this fix #1106? Because that one is still open.

Yes, can be closed now.

Issue #1423, opened 2019-09-19
Current database struct does not encode all required information for a swap


tcharding:
Problem: Currently we are loosing data when writing `Metadata` to the database.  For example we drop information from the `LedgerKind` (information encoded in the enum variants like network name).

Side note: I cannot remember why we did this but it seemed like the correct thing to do at the time, maybe @thomaseizinger remembers.  I believe it was because when we retrieved information from the store using the `with_swap_types` macro we were not using this information - so we removed it.

Goal: In order to make swaps fully persistent we need to write to the database _all_ information about a swap.  

Big picture goal: PR https://github.com/comit-network/comit-rs/pull/1446 renames the struct that is written to the database to `swap` (with table name `swaps`) and removes all fields except the `swap_id`.  This is the struct that needs to hold _all_ the data to describe a completed swap.

tcharding:
Closing, invalidated by #1446

PR #1454, opened 2019-09-25
Add relevant Metadata fields to Swap

tcharding:
Currently a database Swap object has only a swap_id.  Add all the fields from Metadata which are used when displaying swap information to a client i.e., everything used by `with_swap_types!`.  Heavily clean up the unit tests.

This PR is not directly linked to an issue (I closed the (un-groomed) issue https://github.com/comit-network/comit-rs/issues/1423) that was an attempt to describe this problem since the problem is a not easily defined since we still do not know _exactly_ what has to be written to disk.  This PR adds data that is used by the `with_swap_types` macro.

Part of the Sqlite database Epic.


cnd/src/db/models.rs
```diff
@@ -60,3 +70,101 @@ where
         Ok(SwapId(uuid))
     }
 }
+
+#[repr(i32)]
```
thomaseizinger: `repr` is only used for alignment, I think we should not abuse this here. I know that the diesel docs do it as well but I think it is wrong.

Better add an explicit method that returns an integer for your variant if you need that:

```rust
impl Role {
   fn index(&self) -> u32 {
      match self {
          Role::Alice => 0,
          Role::Bob => 1,
	  }
   }
}
```

thomaseizinger: However, as mentioned, I think they should be strings.

cnd/src/db/mod.rs
```diff
@@ -159,158 +159,227 @@ pub fn default_db_path() -> Option<PathBuf> {
...
+        pub alpha_asset: AssetKind,
+        pub beta_asset: AssetKind,
+        pub role: Role,
     }
 
-    fn insertable_swap() -> InsertableSwap {
-        let id = example_swap_id();
-        create_insertable_swap(id)
+    impl Swap {
+        fn instantiate_test_case_1() -> Swap {
```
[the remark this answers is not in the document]
tcharding: This stuff _is_ within `#[cfg(test)]`???

cnd/src/db/models.rs
```diff
@@ -60,3 +70,101 @@ where
...
+#[derive(Debug, Clone, Copy, PartialEq, FromSqlRow, AsExpression)]
+#[sql_type = "Integer"]
```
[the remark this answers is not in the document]
tcharding: Good points, will implement as suggested.

cnd/src/db/models.rs
tcharding: Cool, thanks for the review.

cnd/src/db/mod.rs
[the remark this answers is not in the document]
thomaseizinger: Ah, my mistake sorry!
I thought this was the production code.
What do we need all this for though? :see_no_evil: 

As far as I can see, you only use them to instantiate test data and then convert to the `InsertableSwap`? Wouldn't functions that give you `InsertableSwap` instances be sufficient?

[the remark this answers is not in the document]
tcharding: Nice observation, fixed as suggested.

cnd/src/db/models.rs
tcharding: Implemented as suggested.  Discussed use of macro to generate the `to_sql` and `from_sql` trait implementations for enums.  Was only able to do `to_sql` - `from_sql` should be possible with some extra thought.  Leaving for another time.

tcharding: If a way to do the macro jumps out at anyone feel free to say and I'll implement it.

cnd/src/db/models.rs
```diff
@@ -6,57 +6,214 @@ use diesel::{
     sql_types::Text,
     Insertable, Queryable, *,
 };
-use std::{io::Write, str::FromStr, string::ToString};
+use failure::_core::{
```
luckysori: :mega: `Debug`, `Display` and `Deref` should come from the standard library. `failure` is just re-exporting them here.

cnd/src/db/mod.rs
```diff
@@ -159,158 +159,188 @@ pub fn default_db_path() -> Option<PathBuf> {
...
     #[test]
-    fn can_count_empty_database() {
+    fn can_count_empty_db() {
         let db = new_temp_db();
-        assert_eq!(0, db.count().unwrap());
+        assert_eq!(0, db.count().expect("db count error"));
     }
 
     #[test]
     fn insert_return_vaule_is_as_expected() {
```
D4nte:
```suggestion
    fn given_insert_return_value_is_as_expected() {
```

cnd/src/db/mod.rs
```diff
@@ -159,158 +159,188 @@ pub fn default_db_path() -> Option<PathBuf> {
...
 
-        let swap = insertable_swap();
-        let rows_inserted = db.insert(swap).unwrap();
+        let swap = instantiate_test_case_1();
+        let rows_inserted = db.insert(swap).expect("db insert error");
         assert_eq!(rows_inserted, 1);
     }
 
     #[test]
     fn count_works_afer_insert() {
```
D4nte:
```suggestion
    fn count_works_after_insert() {
```

cnd/src/db/mod.rs
```diff
@@ -159,158 +159,188 @@ pub fn default_db_path() -> Option<PathBuf> {
...
-        let _rows_inserted = db.insert(swap).unwrap();
-        assert_eq!(1, db.count().unwrap());
+        let swap = instantiate_test_case_1();
+        let _rows_inserted = db.insert(swap).expect("db insert error");
+
+        assert_eq!(1, db.count().expect("db count error"));
     }
 
     #[test]
     fn can_not_add_same_record_twice() {
```
D4nte:
```suggestion
    fn cannot_add_same_record_twice() {
```

cnd/src/db/models.rs
D4nte: Also, they are always in scope, no need to import them at all.

PR #1472, opened 2019-09-29
Remove key_gen crate

tcharding:
Looks like we are not using this crate anymore.  At least `cargo make all` runs locally without it.  Am I missing something?


tcharding:
I've got this uneasy feeling that I tried to remove this when I first started working on COMIT :)

D4nte:
key_gen is a binary crate that helped us in the early days.

Happy to remove it.

PR #1478, opened 2019-09-30
Use Testcontainer exports

cristianrgreco:


cristianrgreco:
Should be good now @thomaseizinger 🙂

thomaseizinger:
Nice, thank you :)

PR #1479, opened 2019-10-01
Fix flaky quickcheck test in btsieve module

### LINKED ISSUE #1477: Quickcheck test for URL construction sometimes panics
## Problem & History:

We have a quickcheck test in our codebase that sometimes panics. One of example is here: https://app.circleci.com/jobs/github/comit-network/comit-rs/2062

The test was introduced because the function contains a possible panic that is called in our production code. This approach was chosen because the panic actually should never happen at runtime given the values that we expect to be passed to this function.
To assert this, a quickcheck test was introduced to make sure we can actually never reach this panic.

This is what the test fails with:
```
thread 'bitcoin::bitcoind_connector::tests::build_sub_url_should_never_fail' panicked at '[quickcheck] TEST FAILED (runtime error). Arguments: ("\\\\\u{0}")

Error: "building url should work: InvalidDomainCharacter"', /home/circleci/.cargo/registry/src/github.com-1ecc6299db9ec823/quickcheck-0.7.2/src/tester.rs:176:28

stack backtrace
```

## Goal:

Have tests that don't randomly panic.

## Recommendation:

As we can see in the output of the test, it fails because quickcheck triggered an `InvalidDomainCharacter` with the input it passed to the function.

This is actually a result of us using `String` as the argument to this function. If we were to use a more specific type, this error should no longer be possible.

The argument to this function is actually a transaction hash. If we encode it as a `Sha256dHash` (a type from `rust-bitcoin`), this should no longer happen.

A `Sha256dHash` is internally a 32byte array. However, the type doesn't implement `Arbitrary` (the trait required by quickcheck to being able to generate arbitrary instances for it). Hence, we will need to make use of the generic newtype pattern that is already used in [btsieve/src/ethereum/queries/quickcheck.rs](https://github.com/comit-network/comit-rs/blob/ac827bbdd415d49827e7d105cd603d2a23671912/btsieve/src/ethereum/queries/quickcheck.rs) in order to implement `Arbitrary` on it. Given that, we will be able to use `Sha256dHash` as the type for our `raw_block_by_hash_url` function and generate arbitrary instances for it using quickcheck to assert that it will never panic at runtime.

Since any 32 bytes are a valid hash and the string representation of those 32 bytes is always in hex, the particular panic that was triggered in this test should no longer occur.

thomaseizinger:
Fixes #1477.

### DIFF
--- a/btsieve/src/quickcheck.rs
+++ b/btsieve/src/quickcheck.rs
@@ -1,0 +1,14 @@
+use std::ops::Deref;
+
+/// Generic newtype that allows us to implement quickcheck::Arbitrary on foreign
+/// types
+#[derive(Clone, Debug)]
+pub struct Quickcheck<I>(pub I);
+
+impl<I> Deref for Quickcheck<I> {
+    type Target = I;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
--- a/btsieve/src/bitcoin/bitcoind_connector.rs
+++ b/btsieve/src/bitcoin/bitcoind_connector.rs
@@ -9,7 +9,7 @@ use serde::Deserialize;
 
 #[derive(Deserialize)]
 struct ChainInfo {
-    bestblockhash: String,
+    bestblockhash: bitcoin_support::BlockId,
 }
 
 #[derive(Clone)]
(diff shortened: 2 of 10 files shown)


btsieve/src/quickcheck.rs
```diff
@@ -0,0 +1,14 @@
+use std::ops::Deref;
+
+/// Generic newtype that allows us to implement quickcheck::Arbitrary on foreign
+/// types
+#[derive(Clone, Debug)]
+pub struct Quickcheck<I>(pub I);
+
+impl<I> Deref for Quickcheck<I> {
```
luckysori: :drooling_face:

PR #1485, opened 2019-10-04
Replace PubkeyHash with PublicKey

### LINKED ISSUE #1296: Change cnd to use public keys for the Bitcoin identity
Right now our we use the "pubkey hash" as the identity in [RFC005](https://github.com/comit-network/RFCs/blob/master/RFC-005-SWAP-Basic-Bitcoin.md). This is the wrong choice for two reasons:

1. This blocks the adoption of miniscript because the HTLC created from the miniscript policy uses public keys and not pubkey hashes. This makes sense because it is obfuscated through the P2WSH address anyway.
2. In the [taproot proposal](https://github.com/sipa/bips/blob/bip-schnorr/bip-taproot.mediawiki) outputs are locked to public keys. So sending to someone's public key will become the standard way of transferring Bitcoin (hopefully).

You would probably want to start by changing this line to a `PublicKey`: 

https://github.com/comit-network/comit-rs/blob/5702e0a207784745e5bb256014487cb6cce1b20f/cnd/src/swap_protocols/ledger/bitcoin.rs#L31

and simply follow the compile errors :)

bonomat:
Resolves #1296.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Changed
+- **Breaking:** Change the identity for the Bitcoin Ledger from a public key hash to a public key. This change impacts the HTTP and the COMIT API of cnd. 
+
 ## [0.3.0] - 2019-10-02
 ### Changed
 - Embed btsieve as a library inside cnd: From now on, you'll only need to run cnd to use COMIT.
--- a/cnd/src/bitcoin.rs
+++ b/cnd/src/bitcoin.rs
@@ -1,0 +1,65 @@
+//! This module is the home of bitcoin-specific types and functionality that is
+//! needed across several places in cnd.
+//!
+//! This involves:
+//!     - Creating wrapper types to allow for more ergonomic APIs of upstream
+//!       libraries
+//!     - Common functionality that is not (yet) available upstream
+
+use bitcoin_support::secp256k1_omni_context::secp256k1;
+use serde::{
+    de::{self, Visitor},
+    Deserialize, Deserializer, Serialize, Serializer,
+};
+use std::fmt;
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct PublicKey(bitcoin_support::PublicKey);
+
+impl PublicKey {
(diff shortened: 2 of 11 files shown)


cnd/src/bitcoin.rs
```diff
@@ -0,0 +1,26 @@
...
+use bitcoin_support::secp256k1;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+pub struct PublicKey(bitcoin_support::PublicKey);
+
+impl PublicKey {
+    pub fn new(key: secp256k1::PublicKey) -> Self {
+        Self(bitcoin_support::PublicKey {
+            compressed: true,
```
luckysori: :question: Why is it compressed? Should there be a comment here explaining?

cnd/src/http_api/mod.rs
```diff
@@ -38,10 +38,19 @@ use serde::{
     ser::SerializeStruct,
     Deserialize, Deserializer, Serialize, Serializer,
 };
+use std::ops::Deref;
 
 #[derive(Debug)]
 pub struct Http<I>(pub I);
 
+impl<I> Deref for Http<I> {
```
luckysori: :question: Where is this used? Please explain :smiley:

cnd/src/bitcoin.rs
thomaseizinger: That is there to indicate, how we want it to be serialized.

cnd/src/http_api/mod.rs
thomaseizinger: I think it is no longer used, can removed it :)

thomaseizinger:
> changelog!

Changelog entry added!

cnd/src/bitcoin.rs
D4nte: I think it's fine. The current Bitcoin standard is to use compressed keys. Non-compressed keys are old school and unexpected.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Changed
+- **Breaking:** Change the identity for the Bitcoin Ledger from a PubKey hash to a public key. This change impacts the HTTP and the COMIT API of cnd.
```
D4nte:
```suggestion
- **Breaking:** Change the identity for the Bitcoin Ledger from a public key hash to a public key. This change impacts the HTTP and the COMIT API of cnd. 
```
or use `pubkey` both times.

Issue #1255, opened 2019-08-21
Refactor `secp256k1-keypair` into a SECP-context management crate


thomaseizinger:
The `secp256k1_support` crate has been stripped down to only include the `Keypair` struct.

Instead of having a keypair, we want to have a crate that provides an abstraction on top of `secp256k1` which allow you to not care about the secp-context but just use all the features associated with a secret-key:

- Get the public key
- Sign something

API-wise, we want the user to initialize the SECP-context but then pass it during key generation. The key should take ownership (`Arc` to the context?) of that context so that it can generate the public key at any time and sign stuff.
Overall, the crate should be a way of conveniently managing the secp context.

The crate should be publishable (no path dependencies, no git-dependencies)!

PR #1505, opened 2019-10-08
Secp Keypair crate #1

D4nte:
This is a first PR, more work needs to be done for #1255.

Achievements:
- secp is carried in with the keypair
- new builder interface for keypair
- less use of static secp context in libraries
- rename the keypair crate

Yet to be done in a follow-up PR:
- remove all secp context static vars and only have one in cnd
- make the crate ready for publication (document)
- improve bitcoin_witness interface


internal/secp256k1_omni_context/src/keypair.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    pub fn rng<R: Rng>(self, rng: &mut R) -> Builder {
+        Builder {
+            secret_key: Some(SecretKey::new(rng)),
+            ..self
+        }
+    }
+
+    pub fn build(self) -> Result<KeyPair, Error> {
+        let mut secret_keys_count = 0;
```
luckysori: :wrench: I think you could represent all the possibilities with an enum or even just transforming the slice or hex into a `SecretKey` when the `secret_key_slice` or `secret_key_hex` method is called.

thomaseizinger: > even just transforming the slice or hex into a SecretKey when the secret_key_slice or secret_key_hex method is called.

I'd suggest doing that and simply returning a `Result<Builder, Error>` from each of these. It is still chainable if you `?` inbetween.

D4nte: hum, I see what you mean and I wondered the same.
However, I designed it this way because when using a builder, I expect to be able to chain initialisation parameters and then call `Builder::build()` and only expect to worry about a result when building.
Worrying about a result during initialisation of the builder seems weird to me.

Thoughts?

thomaseizinger: How many valid combinations of these build steps are there though?

It seems like they are just different constructors, where all of them should yield a `Result<KeyPair, Error>`.

D4nte: ok, I'll implement this comment in the next PR.

PR #1514, opened 2019-10-10
Fetch missing blocks for Bitcoin connectors

### LINKED ISSUE #1333: Match query against missed bitcoin blocks
When fetching blocks from a block source you can miss blocks, especially if the "poll interval" is relatively large because of limited requests per hour.

Tasks:
- Refactor the `MatchingTransactions` to return an actual `Stream` object of matching transactions
  - Create our own `Stream` implementation and implement `poll`
  - `poll` should then contain the logic what has already been seen (...) @thomaseizinger can elaborate
- Store blockhashes and their prev_blockhash.
- Based on the stored blockhashes, when grabbing a new block detect if we have missed a block and fetch it.

luckysori:
- Introduce `async/.await` into the codebase :tada:
- Keep track of all the blocks that have been checked against a query.
- Look for missing blocks if the parent of a block returned by `latest_block` is unknown.
- Add scripts to generate Bitcoin test data more easily.

Thanks, @thomaseizinger and @tcharding for helping me :smile: 

---

Fixes #1333.
Ethereum side of things tracked in https://github.com/comit-network/comit-rs/issues/1516.
Cleaning up `MatchingTransactions` tracked in #1519.


btsieve/tests/bitcoin_missing_blocks.rs
```diff
@@ -0,0 +1,324 @@
...
+    assert_eq!(transaction, expected_transaction);
+}
+
+#[test]
+fn find_transaction_in_missing_block_with_big_gap() {
+    let block1 = from_hex(
+        include_str!("./test_data/find_transaction_in_missing_block_with_big_gap/block1.hex")
+            .trim(),
+    );
+    let block2 = from_hex(
```
thomaseizinger: Variable name doesn't match file name.

btsieve/tests/test_data/find_transaction_if_blockchain_reorganisation_with_long_chain/generate_test_data.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+set -e
+
+source "../lib.sh"
+
+# This directory is created by docker as root
+sudo rm -rf /tmp/bitcoin
```
D4nte: Looks like this can be done in `docker_run`.
Also, `sudo` should not be needed.

btsieve/tests/test_data/lib.sh
```diff
@@ -0,0 +1,58 @@
...
+rm_file_if_exists() {
+    local file=$1
+
+    if [ -f $file ] ; then
+        rm $file
+    fi
+}
+
+docker_run() {
+    container=$(docker run -d -v /tmp/bitcoin:/root/.bitcoin coblox/bitcoin-core -regtest)
```
D4nte: do the `rm -rf /tmp/bitcoin/` here.

Also, better if you use `mktemp`

btsieve/tests/test_data/lib.sh
```diff
@@ -0,0 +1,58 @@
+container="" # Set by docker_run
+
+rm_file_if_exists() {
```
D4nte: Just replace with `rm -f`

btsieve/tests/test_data/lib.sh
```diff
@@ -0,0 +1,58 @@
...
+generate_101_blocks() {
+    docker exec $container bitcoin-cli -regtest generate 101 > /dev/null
+    echo "Generated 101 blocks"
+}
+
+generate_block() {
+    local path="$1"
+
+    docker exec $container bitcoin-cli -regtest generate 1 > /dev/null
+    blockhash=$(docker exec $container bitcoin-cli -regtest getbestblockhash)
```
D4nte: Save time by removing this line and instead use the result returned by the generate (it gives you the new block hash)

btsieve/tests/test_data/lib.sh
```diff
@@ -0,0 +1,58 @@
...
+    sleep 2
+}
+
+
+generate_101_blocks() {
+    docker exec $container bitcoin-cli -regtest generate 101 > /dev/null
+    echo "Generated 101 blocks"
+}
+
+generate_block() {
```
D4nte:
```suggestion
generate_block() {
    if [ $# -ne 1 ]; then
        echo "Incorrect number of parameters on generate_block"
        exit 2
    fi
```
That can be helpful to avoid issues

btsieve/tests/test_data/find_transaction_if_blockchain_reorganisation_with_long_chain/generate_test_data.sh
```diff
@@ -0,0 +1,50 @@
...
+generate_block "./block3.hex"
+generate_block "./block4.hex"
+
+docker_stop
+
+sudo cp -r /tmp/bitcoin /tmp/bitcoin-101
+
+docker_start
+
+create_transaction "./address" "./transaction.hex"
```
D4nte: `create_transaction` seems to write in `./address`. Who reads it?

btsieve/tests/test_data/find_transaction_if_blockchain_reorganisation/generate_test_data.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+set -e
+
+source "../lib.sh"
+
+# This directory is created by docker as root
```
D4nte: haaa, that's why there is sudo everywhere?
pro-tip: create the directory yourself with mktemp instead.

btsieve/tests/test_data/find_transaction_in_missing_block/transaction.hex
```diff
@@ -0,0 +1 @@
+020000000001024facc5f3eacd84a66b6e4b6a67a0083c2e7aa373a632e6c157dd9a8a19e17d4e0100000017160014b0d841f1f3f262601a40d7fd38483537cf0b2745feffffff1dad244cd6462c045c4eddf37552bb53739c24067174c4c8e82533ebd00cade4000000001716001464c6c98da700017cbb244120aa5acfd5756674e5feffffff0200cdf5050000000017a914655058043be28697efceabe405541dd346eac2a78700e1f5050000000017a914b31e69596d6838cbc30139bf9ca53a9482c51444870247304402203af9df942a7c23380eecbc4f229e6a3d8caeadc3d50e6c2ef70c5bbf6fade93a02200aa09e6c56b85020efd91cba16de9f01111d2308296ff5fb408f56f1bef84f42012102811419de9710ecc41207e1d03698c6c4b416c53602b2aea9e7279e0d583a1bdf0247304402205a9f1a5c6c7dd73e74fe12fd4e6af8c1512fd329ede8e3f06f5d1a66bfaf0e2a02200fe57021fcc606530bbba64c50f2239c122feb63653d43a4dcdaa0f2f293c4e50121028396603687ad920e9b43654a49d3c0116abc7138dd975a16de2bdd2109af3f5493010000
```
D4nte: So was this generated by "generate_test_data.sh`?

btsieve/src/bitcoin/bitcoind_connector.rs
```diff
@@ -41,7 +41,7 @@ impl LatestBlock for BitcoindConnector {
     type BlockHash = bitcoin_support::BlockId;
 
     fn latest_block(
-        &self,
+        &mut self,
```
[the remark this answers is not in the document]
luckysori: At the moment it is not needed but it was introduced for testing purposes (see implementation of `LatestBlock` for `BitcoinConnectorMock`). Since we can see future, smarter implementations of `LatestBlock` needing to modify state inside the blockchain connector we decided that it was okay.

btsieve/src/bitcoin/mod.rs
```diff
@@ -11,56 +11,123 @@ pub use self::{
...
+    loop {
+        // Delay so that we don't overload the connector,
+        // should probably be defined by the blockchain connector
+        Delay::new(std::time::Instant::now().add(std::time::Duration::from_secs(1)))
+            .compat()
+            .await
+            .unwrap_or_else(|e| log::warn!("Failed to wait for delay: {:?}", e));
+
+        let mut new_missing_block_futures = Vec::new();
+        for (block_future, blockhash) in missing_block_futures.into_iter() {
```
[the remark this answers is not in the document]
luckysori: On the first iteration, yes. `new_missing_block_futures` is not the same as `missing_block_futures`. As we iterate through the vector of missing_block_futures if we don't find a matching transaction inside the block and we don't know the parent of the block, we store a future to get the parent block and store it inside `new_missing_block_futures`.

After the loop we replace the contents of `missing_block_futures` with what's inside `new_missing_block_futures`.

btsieve/tests/test_data/lib.sh
luckysori: It gives you an array containing one blockhash though. I was told to avoid dealing with arrays in bash...

luckysori: I've been trying to use the array returned instead and it is not working lol

btsieve/tests/test_data/find_transaction_if_blockchain_reorganisation_with_long_chain/generate_test_data.sh
luckysori: The tests need to know the `to_address` to construct the `TransactionQuery`.

btsieve/tests/test_data/find_transaction_if_blockchain_reorganisation/generate_test_data.sh
luckysori: Gonna give it a go! Thanks for the suggestion :)

btsieve/tests/test_data/find_transaction_in_missing_block/transaction.hex
luckysori: The scripts do generate all the test data. We also check in all the test data, so you don't have to run the scripts every time you run the tests.

btsieve/src/bitcoin/mod.rs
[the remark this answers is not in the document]
D4nte: Ok, I see the mindfudge.

btsieve/tests/test_data/lib.sh
D4nte: use jq :)

D4nte:
@luckysori I thought I asked the question but I cannot find the comment... What are the shell scripts for? Can you please document? When are they suppose to be run?

btsieve/tests/test_data/find_transaction_in_missing_block/transaction.hex
D4nte: Please document that somewhere.

btsieve/tests/test_data/find_transaction_if_blockchain_reorganisation/generate_test_data.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Strategy followed to cause a blockchain reorganisation:
+#
+# 1. Start a bitcoin node inside a docker container with a volume.
+# 2. Generate some blocks.
+# 3. Stop the container and save the state by copying it to the file system.
+# 4. Restart the container and generate blocks.
+# 5. Stop the container again and overwrite its current state with the state
+#    saved in step 3.
+# 6. Restart the container and generate blocks. These will be different to
```
D4nte: hum.. ok? I trust you here.

thomaseizinger: Well, hopefully the blockchain is not deterministic :D

thomaseizinger: Well, actually I guess it is deterministic lol.
But since blocks contain timestamps and the time has advances, the blocks will be different.
Also we are including a TX in step 4 so that is also different.

D4nte: I should have used multiline comments :) the comment applies to all the steps.

btsieve/tests/test_data/lib.sh
thomaseizinger: Yeah could use jq. Or just leave it like that and have less dependencies :)

It is already working so ...

luckysori:
> Looks good, please document these shell scripts so we have an idea of what they are for in 6 months time.

Done!

PR #1527, opened 2019-10-12
Improve version printing for cnd

karen:
Fixes #1417.

PR #1528, opened 2019-10-14
Make http_api and network optional in config file

### LINKED ISSUE #1448: cnd's config file has sections that are required but don't need to
**Problem:**
The sections `http_api` and `network` are required to exist in the config file.
1. This makes changes to these sections more painful because it invalidates the old config file.
2. We only write default values to those sections upon the first startup anyway.

**Recommendation:**
Instead of writing default values for `http_api` and `network` to the configuration file, we should make them optional and default to the values we would have written otherwise.

Note:
This ticket, in the long run, helps with making the config file totally optional because:
- the secret seed will be removed in #948 
- the `btsieve` section went away with #1408

karen:
Fixes #1448.

Tested:
- With no config file, starting `cnd` produces a default file without `network` and `http_api`, but `Settings` takes the default
- With a config file specifying `network` and `http_api`, `Settings` uses the value from the config file

### DIFF
--- a/cnd/src/config/settings.rs
+++ b/cnd/src/config/settings.rs
@@ -2,6 +2,7 @@ use super::file::{Comit, Database, File, HttpSocket, Network};
 use crate::config::file::{Bitcoin, Ethereum};
 use log::LevelFilter;
 use reqwest::Url;
+use std::net::{IpAddr, Ipv4Addr};
 
 /// This structs represents the settings as they are used through out the code.
 ///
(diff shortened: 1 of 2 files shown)


thomaseizinger:
@karen, if you would like us to merge this, please sign the CLA :)

cnd/src/config/settings.rs
```diff
@@ -3,6 +3,8 @@ use crate::config::file::{Bitcoin, Ethereum};
 use log::LevelFilter;
 use reqwest::Url;
```
D4nte: Please remove this extraneous newline.

thomaseizinger:
Closed for now, see https://github.com/comit-network/comit-rs/pull/1527#issuecomment-552249777.

PR #1529, opened 2019-10-14
Set content-type for error responses as PROBLEM_JSON_MEDIA_TYPE

### LINKED ISSUE #1438: Error responses should correctly identify themselves as RFC7807: `application/problem+json`
Currently, we simply send back our instances of `HttpApiProblem` as JSON:

https://github.com/comit-network/comit-rs/blob/6616f56206b3a4d40feaaf15de40b9925a9dae6c/cnd/src/http_api/problem.rs#L149-L161

While not being wrong, it would be more correct to set the "content-type" header to `application/problem+json` to allow clients to actively identify them as "problems".

The http-api-problem library provides a constant for that: https://github.com/chridou/http-api-problem/blob/3e59dc7b1bdd6e35eb9ae749e93aef9579d57371/src/lib.rs#L133-L134

karen:
Fixes #1438.

### DIFF
--- a/cnd/src/http_api/problem.rs
+++ b/cnd/src/http_api/problem.rs
@@ -3,7 +3,7 @@ use crate::swap_protocols::{
     rfc003::{self, actions::ActionKind, state_store},
 };
 use http::StatusCode;
-use http_api_problem::HttpApiProblem;
+use http_api_problem::{HttpApiProblem, PROBLEM_JSON_MEDIA_TYPE};
 use serde::Serialize;
 use warp::{Rejection, Reply};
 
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Thanks @karen !
Could you also please add a test for that to the "dry-test" suite?
They are located in `api_tests/dry`.

The `sanity.ts` file would probably be the best place! It already contains tests that trigger error responses. All we would need is an assertion, that the `Content-Type` for these scenarios is indeed "application/problem+json".

thomaseizinger:
Closed for now, see https://github.com/comit-network/comit-rs/pull/1527#issuecomment-552249777.

PR #1533, opened 2019-10-14
Merge Ethereum transaction and event queries

### LINKED ISSUE #799: Merge transaction + event query into one query type
Conceptually, both types of queries target _transactions_ so the functionality should be merged in one query.

The fact that a bloom filter is used for the latter should only be an implementation detail IMO.

luckysori:
Fixes #799.


btsieve/src/ethereum/mod.rs
```diff
@@ -5,6 +5,107 @@ mod web3_connector;
...
+                log::warn!("Could not get latest block");
+                continue;
+            }
+            Err(e) => {
+                log::warn!("Could not get latest block: {:?}", e);
+                continue;
+            }
+        };
+
+        if !query.event_matches_block(&latest_block) {
```
thomaseizinger: Could be renamed to `can_skip_block` as discussed offline.

btsieve/src/ethereum/mod.rs
```diff
@@ -5,6 +5,107 @@ mod web3_connector;
...
+            continue;
+        }
+
+        for transaction in latest_block.transactions.iter() {
+            if !query.matches(transaction) {
```
thomaseizinger: This could move to the bottom and be combined with `||` with `event_matches_transaction_receipt` then the `continue` could change to a return at the end of the loop.

btsieve/src/ethereum/queries.rs
```diff
@@ -0,0 +1,655 @@
...
+};
+
+#[derive(Clone, Default, Debug, Eq, PartialEq)]
+pub struct TransactionQuery {
+    pub from_address: Option<Address>,
+    pub to_address: Option<Address>,
+    pub is_contract_creation: Option<bool>,
+    pub transaction_data: Option<Bytes>,
+    pub transaction_data_length: Option<usize>,
+    pub events: Vec<Event>, // Empty if we are not matching events.
```
thomaseizinger: Could make a this an `Option` to neglect the need for the comment.

luckysori: I think we considered this when the `EventQuery` was originally introduced, but I still don't like it because then we can have `Some(vec![])`, and what does that express? Isn't that just equivalent to `None`? Seems weird to me.

thomaseizinger: Hmm, `Some(vec![])` would mean you want a transaction that doesn't have any events whereas `None` would mean, I don't want to match against the events of the transaction. Not sure if we need it but it would be more consistent with the rest :)

tcharding: I disagree, I don't feel that `Some(vec![])` adds any meaning over an empty vector.  There is no time, AFAICT, that processing an empty vector is different from `None` - so to me it adds no extra meaning.  Any implied meaning would have to be added as a comment anyways because it is would be unclear as to what is the difference.

thomaseizinger:
bors retry

luckysori:
bors retry

PR #1535, opened 2019-10-15
Stale bot init

bonomat:
Feel free to propose alternative timeouts

resolves: coblox/admin#93


thomaseizinger:
Cool!
I think the config file by itself doesn't do anything though. We will also need a workflow that triggers the actual bot :)

bonomat:
> Cool!
> I think the config file by itself doesn't do anything though. We will also need a workflow that triggers the actual bot :)

true that :D 
I've installed the app already (you prob. should get a notification email)

PR #1536, opened 2019-10-15
Test transaction queries

tcharding:
For some time now I have been postulating that, when designing struct objects, the use of Option types with an optional type that has a nil value makes the code hard to reason about.  I have argued this point at length with other members of the team (@thomaseizinger and @luckysori) but have thus far been unconvincing.

In an effort to lay out in code why I believe this practice is a bad idea this patch set attempts to test the logic when matching transaction queries.  The struct member used as an example is `is_contract_creation: Option<bool>`.  The argument against my position is that `None` and `Some(false)` have a different semantic meaning, I am not disputing this, my claim is that this different meaning is _not_ apparent from reading the code and reasoning about the code that uses this struct member is difficult.  This patch adds tests that more thoroughly test the logic that we have.  All tests pass - I do not know if, logically, they should pass or fail.

If  you have the patience to humour me, please review the tests in the last commit of this PR and see if they are valid tests that we want.  If they are valid, all well and good - the code is hard to reason about (for me), nothing lost nothing gained.  If they are _not_ valid tests then the difficulty in reasoning about the code has opened us up to bugs either present or future.

Thanks for you patience.

One side note; I may lack the domain knowledge to understand what `is_contract_creation` is and how it would be used.  This domain knowledge may make the semantic difference more obvious.

### DIFF
--- a/btsieve/src/ethereum/queries/transaction.rs
+++ b/btsieve/src/ethereum/queries/transaction.rs
@@ -166,6 +166,8 @@ mod tests {
 
     #[test]
     fn given_query_transaction_data_transaction_matches() {
+        let to_address = "0bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb".parse().unwrap();
+
         let query_data = TransactionQuery {
             from_address: None,
             to_address: None,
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Thanks for this effort @tcharding !
However, I think that instead of adding more tests to the existing ones, we should replace the current ones with tests that have high-quality test data (i.e. mainnet blocks/transactions).

Issue #1506, opened 2019-10-09
Unstable property-based test


thomaseizinger:
## Problem

We have another unstable property based test: https://app.circleci.com/jobs/github/comit-network/comit-rs/2158

## Recommendation

Don't write any more tests. JK - fix it!

PR #1539, opened 2019-10-16
Use BlockId instead of String

tcharding:
When doing property based testing on a hash data field (of type
`String`) quickcheck generates input that causes a panic.  The generated
input is not possible during normal program flow.  To inhibit quickcheck
from generating input that is not useful to us we can use a specific
type instead of a string.  We already have code in tree doing the same
sort of thing, we can do the same here.

Use `bitcoin_support::BlockId` instead of `String`.

Resolves: #1506

### DIFF
--- a/btsieve/src/bitcoin/blockchain_info_connector.rs
+++ b/btsieve/src/bitcoin/blockchain_info_connector.rs
@@ -9,7 +9,7 @@ use tokio::prelude::Future;
 
 #[derive(Deserialize)]
 struct BlockchainInfoLatestBlock {
-    hash: String,
+    hash: bitcoin_support::BlockId,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


btsieve/src/bitcoin/blockchain_info_connector.rs
```diff
@@ -38,7 +38,8 @@ impl BlockchainInfoConnector {
         })
     }
 
-    fn block_by_hash_url(block_hash: &str) -> Url {
+    fn block_by_hash_url(block_hash: &bitcoin_support::BlockId) -> Url {
+        let block_hash = format!("{}", block_hash);
```
D4nte:
```suggestion
        let block_hash = block_hash.to_string();
```

tcharding:
> Judging from your PR description, I get the feeling you were 100% aware of what you are doing here, even though it was quite a small change and could have been blindly copied from the old PR! Well done :)

Thanks man, that is one of the reasons for writing descriptive PRs and commit messages IMO - to make sure one _realy_ understands what they are doing.  Unfortunately that implies that you can almost always read lack of information in these places from me as not understanding.

thomaseizinger:
@D4nte I hope you don't mind:

bors r+

D4nte:
> @D4nte I hope you don't mind:
> 
> bors r+

Shouln't the author of the PR be the one to `bors r+`?

thomaseizinger:
> > @D4nte I hope you don't mind:
> > bors r+
> 
> Shouln't the author of the PR be the one to `bors r+`?

Anyone can I guess, but since this one fixes a bug with for a flaky build, I wanted to get it in ASAP.

PR #1540, opened 2019-10-16
Make btsieve matching code clearer

D4nte:
Current matching logic in btsieve uses a mutable variable and logical and (`&&`) statements to mutate this variable.  This is unnecessary, also idiomatic Rust typically only uses mutable variables when needed.  We should change this code to not use mutable variables.

While we are cleaning up also do:
- Rename `TransactionQuery` to `Pattern`: since this struct holds data used to pattern match against transactions.
- Add documentation to the newly named `Pattern` struct and the pattern matching function.
- Rename the queries modules (bitcoin and ethereum) to `pattern_matching`.

### DIFF
--- a/btsieve/src/bitcoin/mod.rs
+++ b/btsieve/src/bitcoin/mod.rs
@@ -1,13 +1,13 @@
 mod bitcoind_connector;
 mod blockchain_info_connector;
-mod queries;
+mod transaction_pattern;
 
 #[cfg(test)]
 mod quickcheck_impls;
 
 pub use self::{
     bitcoind_connector::BitcoindConnector, blockchain_info_connector::BlockchainInfoConnector,
-    queries::TransactionQuery,
+    transaction_pattern::TransactionPattern,
 };
 
 use crate::{BlockByHash, LatestBlock, MatchingTransactions};
@@ -20,7 +20,7 @@ use tokio::{
     timer::Delay,
 };
 
-impl<C, E> MatchingTransactions<TransactionQuery> for C
+impl<C, E> MatchingTransactions<TransactionPattern> for C
 where
     C: LatestBlock<Block = bitcoin_support::Block, Error = E>
         + BlockByHash<Block = bitcoin_support::Block, BlockHash = bitcoin_support::BlockId, Error = E>
@@ -31,16 +31,16 @@ where
 
     fn matching_transactions(
         &self,
-        query: TransactionQuery,
+        pattern: TransactionPattern,
     ) -> Box<dyn Stream<Item = Self::Transaction, Error = ()> + Send + 'static> {
-        let matching_transaction = Box::pin(matching_transaction(self.clone(), query)).compat();
+        let matching_transaction = Box::pin(matching_transaction(self.clone(), pattern)).compat();
         Box::new(stream::futures_unordered(vec![matching_transaction]))
     }
 }
 
 async fn matching_transaction<C, E>(
     mut blockchain_connector: C,
-    query: TransactionQuery,
+    pattern: TransactionPattern,
 ) -> Result<bitcoin_support::Transaction, ()>
 where
     C: LatestBlock<Block = bitcoin_support::Block, Error = E>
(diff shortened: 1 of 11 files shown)


btsieve/src/bitcoin/mod.rs
```diff
@@ -31,7 +31,7 @@ where
 
     fn matching_transactions(
         &self,
-        query: TransactionQuery,
+        query: Pattern,
```
D4nte:
```suggestion
        pattern: Pattern,
```

btsieve/src/ethereum/pattern_matching.rs
```diff
@@ -25,34 +30,45 @@ impl TransactionQuery {
...
-                    result = result && (transaction.to == Some(*to_address));
+                    if transaction.to != Some(*to_address) {
+                        return false;
+                    }
                 }
 
                 if let Some(is_contract_creation) = is_contract_creation {
-                    // to_address is None for contract creations
-                    result = result && (*is_contract_creation == transaction.to.is_none());
+                    // transaction.to_address is None iff the transaction creates a contract
```
thomaseizinger:
```suggestion
                    // transaction.to_address is None if the transaction creates a contract
```

btsieve/src/bitcoin/pattern_matching.rs
```diff
@@ -3,39 +3,51 @@ use bitcoin_support::{
 };
 
 #[derive(Clone, Default, Debug, Eq, PartialEq)]
-pub struct TransactionQuery {
+// If the field is set to Some(foo) then only transactions matching foo are
+// returned Otherwise, when the field is set to None, no pattern matching is
+// done for this field
+pub struct Pattern {
```
thomaseizinger: I like the new name but I think it is too generic. What about `TransactionPattern`?

btsieve/src/bitcoin/mod.rs
```diff
@@ -1,13 +1,13 @@
 mod bitcoind_connector;
 mod blockchain_info_connector;
-mod queries;
+mod pattern_matching;
```
thomaseizinger: IMO module names + the structs they define should match up, esp. if the modules are private and we re-export the type from the inside.

btsieve/src/ethereum/pattern_matching.rs
```diff
@@ -5,7 +5,10 @@ use ethereum_support::{
 };
 
 #[derive(Clone, Default, Debug, Eq, PartialEq)]
-pub struct TransactionQuery {
+// If the field is set to Some(foo) then only transactions matching foo are
```
thomaseizinger: Should be a rust-doc comment (triple slash) :)

btsieve/src/ethereum/pattern_matching.rs
D4nte: not a typo :) https://www.merriam-webster.com/dictionary/iff

btsieve/src/bitcoin/mod.rs
tcharding: Cool, done.  This raises an issue that this PR only touches btsieve - all through cnd we still use the term 'query'.  I intend on opening an issue for that.

btsieve/src/ethereum/pattern_matching.rs
tcharding: Discussed offline - agreed to use full English.

btsieve/src/bitcoin/pattern_matching.rs
tcharding: Done, thanks.

btsieve/src/bitcoin/mod.rs
tcharding: Implemented as suggested.

btsieve/src/ethereum/pattern_matching.rs
tcharding: Sweet - done.

tcharding:
This PR commit history has gotten a bit wild - note to self: squash before merging.

PR #1541, opened 2019-10-16
Improve comment explaining the delay

tcharding:
This comment has caused some confusion in the team; attempt to improve it.

- 'the machine' is ambiguous - the machine we are running on, the machine the
blockchain connector is connecting to?  Use 'the CPU' instead since
non-ambiguously refers to the CPU running the current process.

- 'given the assumption' - this is not an assumption since it does not always
  hold true, better to say 'in the event that'.


tcharding:
bors  r+

Issue #1459, opened 2019-09-25
ethereum::TransactionPattern is not very well tested


thomaseizinger:
**Context**:

The implementation of `ethereum::TransactionPattern` is mostly tested with handcrafted data (blocks and transactions). Additionally, with #1408 a lot of tests were deleted.

With PR #1543, one `ethereum::TransactionPattern` test was modified to use a block and transaction receipt from mainnet:

https://github.com/comit-network/comit-rs/blob/491c285b79283c58f79acc76381581c265925935/btsieve/tests/ethereum_pattern.rs#L6-L28

**Problem**:

Lacklustre test data used to test `ethereum::TransactionPattern`.

**Goal**:

Very good tests.

**Recommendation**:

Keep going with what was started in #1543 and remember to include ERC20 transfer.

PR #1543, opened 2019-10-16
1459 improve event query tests

tcharding:
Paired with @luckysori on this one.

Makes a starts on #1459.  The issue was not completed within the sprint and deemed not high enough to continue.

These changes should help the next dev who comes along to pick up this ticket when it gets included in the sprint.

PR adds an Etherem block from mainnet (with all but one transaction removed) in a file in the repo.  Reads this file and uses the block for a single unit test.  Adds some trickery so we only read the file once (ready for the rest of the unit tests to use the block).  I wrote the trickery coming from other languages - if there is a more Rusty way to do it please say.

Thanks


tcharding:
> Every include_str statement is going to embed the payload from the file into the source-code so all this trickery only optimizes how often it is parsed and not read :)

Oh, very nice - I did not know that was the behaviour of `include_str`.  Trickery is un-needed.

PR #1563, opened 2019-10-23
Simplify the data-flow when accepting & declining swap requests

### LINKED ISSUE #1406: Decouple communication from swap state machine
Currently, the swap state machine handles the communication (swap request & response) as the first step. This creates a dependency from the the swap state machine to the communication layer (libp2p). At the very least, this makes it impossible for us to separate out the core part of the application into a library. There are very likely other problems with this as well (The awkward constructions of `AliceToBob` and `BobToAlice` are another example).

To simplify this, we should separate out the communication step from the swap state machine. What we want is to have all the data ready then we start the swap instead of the data accumulating as the state machine executes.

This will also make it easier to move the state machine away from a poll-based model. (Probably even a pre-condition to that.)

Something to remember is that the transition within the state machine calls back to the state store.
If the communication is moved outside, storing this data will have to be done explicitely.

### LINKED ISSUE #1507: Change our network-layer API to allow sending a response at any time
## Problem

Currently, upon receiving a swap request, the network layer emits an event that contains a channel which you have to use to send the response back to the other party:

https://github.com/comit-network/comit-rs/blob/0e9bacbc3e5a2dddcfae3ebc38edf593870fbad9/cnd/src/network.rs#L132-L149

This is problematic because it means we have a source-code dependency between the part of the code that handles the receiving of a message and the part that computes the response. Concretely, the code path from the above linked function leads to spawning the state machine for Bob where we perform some channel trickery to receive the response on this end once the user submits it via the HTTP API.

## Goal

Ideally, all we want to do when receiving a message from the other party is to store it in a persistent place. Other parts of the codebase, like the HTTP API can then perform appropriate actions as they inspect the current state.

## Recommendation

The use of channels inside the network layer should be an implementation detail an not leak to the application code.
Instead, the network layer should offer an API to send a response at any time. This will require one of two things:

a. If we stick to the request-response based message model, the API needs to give an identifier to the caller upon receiving an event that can be used to send the response back
b. ~The messaging model moves to something that is not request-response. That means the network layer can close the substream after receiving a message and sending a response back is entirely application logic.~

### LINKED ISSUE #1508: Remove the use of channels for accepting/declining a swap
## Problem

Currently, we are using channels inside the `Accept` and `Decline` action to send the whatever the user gave us via the HTTP API through to the state machine and eventually to the other party via the network layer.

## Goal

Removing these channels allows us to:

a) make the dataflow clearer in our codebase
b) simply store all the data structures that are involved because they no longer contain channels

## Recommendation

Given the data from the user:

- spawn the state machine now that we have all the data we need (this assumes https://github.com/comit-network/comit-rs/issues/1406)
- send the message to the other party via the network layer (this assumes https://github.com/comit-network/comit-rs/issues/1507)

tcharding:
This PR is a combination of work done by @thomaseizinger and @tcharding.  It removes the communication from the state machine and changes the way the bob spawning is done.  With this when a swap request arrives from the peer node we save the channel in a hashmap and store the swap request in the state store but do **not** spawn bob until later when the action handler gets triggered by the client withe accept/decline.  We also use async/await when spawning bob.

Resolves #1507.
Resolves #1406.
Resolves #1508.

### DIFF
--- a/cnd/src/http_api/mod.rs
+++ b/cnd/src/http_api/mod.rs
@@ -24,20 +24,33 @@ use crate::{
         asset::{FromHttpAsset, HttpAsset},
         ledger::{FromHttpLedger, HttpLedger},
     },
-    network::DialInformation,
+    network::{DialInformation, Network},
     swap_protocols::{
+        asset::Asset,
         ledger::{Bitcoin, Ethereum},
-        SwapId, SwapProtocol,
+        metadata_store,
+        rfc003::{
+            alice::{self, AliceSpawner},
+            bob::BobSpawner,
+            messages::{AcceptResponseBody, DeclineResponseBody, Request, ToRequest},
+            state_machine::SwapStates,
+            state_store::{self, StateStore},
+            ActorState, CreateLedgerEvents, Ledger,
+        },
+        LedgerEventDependencies, Metadata, MetadataStore, SwapId, SwapProtocol,
     },
 };
 use bitcoin_support::{amount::Denomination, Amount as BitcoinAmount};
 use ethereum_support::{Erc20Token, EtherQuantity};
+use futures::sync::oneshot::Sender;
 use libp2p::PeerId;
+use libp2p_comit::frame::Response;
 use serde::{
     de::{self, MapAccess},
     ser::SerializeStruct,
     Deserialize, Deserializer, Serialize, Serializer,
 };
+use std::sync::Arc;
 
 #[derive(Debug)]
 pub struct Http<I>(pub I);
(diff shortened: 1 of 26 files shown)


cnd/src/http_api/routes/rfc003/mod.rs
```diff
@@ -51,15 +57,18 @@ pub fn get_swap<T: MetadataStore, S: StateStore>(
         .map_err(into_rejection)
 }
 
+// TODO: Remove two arguments here (they are in the dependencies struct)
```
thomaseizinger: Uff, I left this here :sweat_smile: 

Do you mind removing it, otherwise it is gonna open an issue ...

tcharding: We have a million clippy issues and I only have 10 minuets until I'm off to get Milla.

cnd/src/network.rs
```diff
@@ -87,18 +92,195 @@ impl<TSubstream, B> ComitNode<TSubstream, B> {
...
+                    }
+                }
+            }
+
+            // This case is just catered for, because of rust. It can only happen
+            // if there is a typo in the request_type within the program. The request
+            // type is checked on the messaging layer and will be handled there if
+            // an unknown request_type is passed in.
+            request_type => {
+                log::warn!("request type '{}' is unknown", request_type);
```
D4nte: A `warn` is a bit light if this is not supposed to happen at all.

cnd/src/swap_protocols/rfc003/alice/spawner.rs
```diff
@@ -80,32 +83,93 @@ impl<T: MetadataStore, S: StateStore, C: Client> AliceSpawner
...
-            alice.new_state_machine(
-                ledger_events.create_ledger_events(),
-                ledger_events.create_ledger_events(),
-                self.client.clone(),
-                bob_dial_info,
-                Arc::new(sender),
-            )
+            let state_store = Arc::clone(&self.state_store);
+
+            // For legacy reasons, we also store the communication in the state store
```
D4nte: Can we have a "Will be remove with issue #xx" comment?

cnd/src/swap_protocols/rfc003/alice/mod.rs
```diff
@@ -52,56 +45,50 @@ pub enum SwapCommunication<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> {
...
+            error: None,
+        }
     }
-
-    pub fn request(&self) -> Request<AL, BL, AA, BA> {
-        match &self.swap_communication {
-            SwapCommunication::Accepted { request, .. }
-            | SwapCommunication::Proposed { request }
-            | SwapCommunication::Declined { request, .. } => request.clone(),
+    pub fn accepted(
```
luckysori: :wrench: 
```suggestion
    
    pub fn accepted(
```

cnd/src/swap_protocols/rfc003/state_machine.rs
```diff
@@ -116,10 +98,10 @@ impl<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> OngoingSwap<AL, BL, AA, BA> {
...
 #[derive(Debug, PartialEq, Clone)]
 pub enum SwapOutcome<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> {
-    Declined {
-        start: Start<AL, BL, AA, BA>,
-        reason: DeclineResponseBody,
-    },
+    //    Declined {
+    //        request: Request<AL, BL, AA, BA>,
+    //        reason: DeclineResponseBody,
+    //    },
```
D4nte:
```suggestion
```

cnd/src/swap_protocols/rfc003/alice/mod.rs
```diff
@@ -52,56 +45,50 @@ pub enum SwapCommunication<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> {
...
+        response: DeclineResponseBody,
+        secret_source: impl SecretSource,
+    ) -> Self {
         Self {
-            swap_communication: SwapCommunication::Proposed { request },
+            swap_communication: SwapCommunication::Declined { request, response },
             alpha_ledger_state: LedgerState::NotDeployed,
             beta_ledger_state: LedgerState::NotDeployed,
-            secret_source,
+            secret_source: Arc::new(secret_source),
```
luckysori: It's definitely strange that you need the secret source when instantiating the `State` as declined.

cnd/src/swap_protocols/rfc003/state_machine.rs
```diff
@@ -116,10 +98,10 @@ impl<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> OngoingSwap<AL, BL, AA, BA> {
 
 #[derive(Debug, PartialEq, Clone)]
 pub enum SwapOutcome<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> {
-    Declined {
-        start: Start<AL, BL, AA, BA>,
-        reason: DeclineResponseBody,
-    },
+    //    Declined {
```
luckysori: :mega: Woops

cnd/src/http_api/mod.rs
```diff
@@ -232,6 +245,139 @@ impl<'de> Deserialize<'de> for DialInformation {
...
+
+impl<M, S, A, B, N: Network> Network for Dependencies<M, S, A, B, N>
+where
+    M: Send + Sync + 'static,
+    S: Send + Sync + 'static,
+    A: Send + Sync + 'static,
+    B: Send + Sync + 'static,
+    N: Send + Sync + 'static,
+{
+    fn comit_peers(
```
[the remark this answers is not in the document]
thomaseizinger: This used to be called like that before, keen on not touching it in this PR :)

cnd/src/network.rs
thomaseizinger: This used to be there before, we tried to minimize the impact of this PR :)
Happy to accept a follow-up PR on this!

cnd/src/swap_protocols/rfc003/alice/mod.rs
thomaseizinger: Yeah, future refactoring I'd say :D

cnd/src/http_api/routes/rfc003/handlers/action.rs
```diff
@@ -8,65 +8,90 @@ use crate::{
...
+                                        .ok_or_else(problem::missing_channel)?;
+
+                                    let decline_body =
+                                        action.decline(to_swap_decline_reason(body.reason));
+
+                                    let response = rfc003_decline_response(decline_body.clone());
+                                    channel.send(response).map_err(problem::send_over_channel)?;
+
+                                    let request = state.request();
+                                    BobSpawner::spawn(&dependencies, request, Err(decline_body));
```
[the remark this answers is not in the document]
thomaseizinger: We actually don't spawn anything because the spawner inside checks for the communication to be in state accepted. Needs refactoring later :)

cnd/src/swap_protocols/rfc003/alice/spawner.rs
thomaseizinger: I removed the comment because it is more confusing than it actually helps and I don't yet know when we are going to "fix" it. Also because I don't yet know how :)

PR #1564, opened 2019-10-23
Remove `bitcoin_support`

thomaseizinger:
This PR removes the `bitcoin_support` crate.

~There are no observable differences in how we use `cnd` at the moment. The only potentially noticeable difference is the serialization of the `bitcoin::Network` enum.~

~Our enum used to serialize it as `Mainnet` whereas the `rust-bitcoin` one serializes to `Bitcoin`. I don't see this as a big issue because~

~a) we currently don't have any API tests validating that, hence I consider it unspecified behaviour~
~b) we can always adapt this behaviour with a serde-module or a generic newtype once we actually need it (when are we going mainnet?)~

I added tests and restored the old behaviour.

If anyone objects with that, please raise your voice now or remain silent forever.


D4nte:
Can we merge #1565 first so I better understand the COMIT API impact?

cnd/src/http_api/action.rs
```diff
@@ -34,13 +37,13 @@ pub enum ActionExecutionParameters {
...
-        to: bitcoin_support::Address,
+        to: bitcoin::Address,
         amount: String,
-        network: bitcoin_support::Network,
+        network: bitcoin::Network,
     },
     BitcoinBroadcastSignedTransaction {
         hex: String,
-        network: bitcoin_support::Network,
+        network: bitcoin::Network,
```
D4nte: How does the HTTP API output looks like know? Can you add a test so that it's obvious?

tcharding:
> I think we should have more serialisation tests on both API (a bit what I did with #1565). Of course, I do not want to test serialization brought in by other libs, however I think this kind of tests are important:
> - for documentation purposes
> - for non-regression purposes (after all, we do not want to break our API by upgrading rust-bitcoin)

I thought about this myself and agree, I think we should have serialization tests for upstream libraries.

thomaseizinger:
@tcharding @D4nte please check the updated PR :)

cnd/src/config/file.rs
```diff
@@ -94,7 +94,7 @@ pub struct Database {
 
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct Bitcoin {
-    pub network: bitcoin_support::Network,
+    pub network: bitcoin::Network,
```
D4nte: Sorry, I thought I originally made a comment on that but apparently I did not...

So what is the expected configuration now for mainnet? (ie, same topic that for COMIT API and HTTP API...)

Issue #948, opened 2019-05-09
Remove secret seed from configuration


da-kami:
The secret seed comes from the user and should not be in the configuration file of the `comit_node` (a default secret seed would not make much sense...). Instead, a pem file should be used.

Take a look at @LLFourn thoughts: https://github.com/comit-network/spikes/blob/master/0007-secret-seed-storage-and-key-derivation.md

As per the spike, no need to encrypt for now.

Related to: https://github.com/comit-network/comit-rs/issues/919

thomaseizinger:
If we want to encrypt the secret seed, we may want to use something like this: https://docs.rs/pwbox/0.2.1/pwbox/

D4nte:
So, should it be encrypted or not? #refine

PR #1569, opened 2019-10-24
Remove secret key

tcharding:
Remove secret key from the config file and put it in a pem file in the same default data directory used for the config file.  Add a config file option to specify the location and a command line option also.

WIP because e2e tests are failing.  Can someone please review the config file option choices made in this patch and the patch to `api_tests/lib/config.ts`.  I'm uncertain if I've updated the e2e test lib incorrectly or if he PR has bugs.

Thanks

Fixes: #948

### DIFF
--- a/cnd/src/cli.rs
+++ b/cnd/src/cli.rs
@@ -8,6 +8,6 @@ pub struct Options {
     pub config_file: Option<PathBuf>,
 
     /// Path to secret seed file.
-    #[structopt(long = "secret-seed-file", parse(from_os_str))]
+    #[structopt(short = "s", long = "secret-seed-file", parse(from_os_str))]
     pub secret_seed_file: Option<PathBuf>,
 }
--- a/cnd/src/cli.rs
+++ b/cnd/src/cli.rs
@@ -3,7 +3,11 @@ use std::path::PathBuf;
 #[derive(structopt::StructOpt, Debug)]
 #[structopt(name = "COMIT network daemon")]
 pub struct Options {
-    /// Path to configuration file
+    /// Path to configuration file.
     #[structopt(short = "c", long = "config", parse(from_os_str))]
     pub config_file: Option<PathBuf>,
+
+    /// Path to secret seed file.
+    #[structopt(long = "secret-seed-file", parse(from_os_str))]
+    pub secret_seed_file: Option<PathBuf>,
 }
--- a/cnd/src/cli.rs
+++ b/cnd/src/cli.rs
@@ -8,6 +8,6 @@ pub struct Options {
     pub config_file: Option<PathBuf>,
 
     /// Path to secret seed file.
-    #[structopt(short = "s", long = "secret-seed-file", parse(from_os_str))]
-    pub secret_seed_file: Option<PathBuf>,
+    #[structopt(short = "s", long = "seed-file", parse(from_os_str))]
+    pub seed_file: Option<PathBuf>,
 }
--- a/cnd/src/config/file.rs
+++ b/cnd/src/config/file.rs
@@ -112,7 +112,7 @@ impl File {
                 "No configuration file found, creating default at {}",
                 PrintablePath(&path)
             );
-            Self::default(rand).write_to(&path)
+            Self::default().write_to(&path)
         }
     }
 
@@ -190,6 +190,7 @@ impl File {
 mod tests {
     use super::*;
     use log::LevelFilter;
+    use reqwest::Url;
     use spectral::prelude::*;
     use tempfile::NamedTempFile;
 
--- a/cnd/src/config/file.rs
+++ b/cnd/src/config/file.rs
@@ -1,8 +1,7 @@
-use crate::{seed::Seed, std_ext::path::PrintablePath};
+use crate::std_ext::path::PrintablePath;
 use config as config_rs;
 use libp2p::Multiaddr;
 use log::LevelFilter;
-use rand::Rng;
 use serde::{Deserialize, Serialize};
 use std::{
     ffi::OsStr,
@@ -20,7 +19,7 @@ use std::{
 /// for filling in default values for absent configuration options.
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct File {
-    pub comit: Comit,
+    pub comit: Option<Comit>,
     pub network: Network,
     pub http_api: HttpSocket,
     pub database: Option<Database>,
@@ -31,14 +30,13 @@ pub struct File {
 }
 
 impl File {
-    pub fn default<R: Rng>(rand: R) -> Self {
+    pub fn default() -> Self {
         let comit_listen = "/ip4/0.0.0.0/tcp/9939"
             .parse()
             .expect("cnd listen address could not be parsed");
-        let seed = Seed::new_random(rand).expect("Could not generate random seed");
 
         File {
-            comit: Comit { secret_seed: seed },
+            comit: None,
             network: Network {
                 listen: vec![comit_listen],
             },
@@ -66,7 +64,7 @@ pub struct Logging {
 
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct Comit {
-    pub secret_seed: Seed,
+    pub secret_seed_file: PathBuf,
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
@@ -107,7 +105,7 @@ pub struct Ethereum {
 }
 
 impl File {
-    pub fn read_or_create_default<R: Rng>(rand: R) -> Result<Self, config_rs::ConfigError> {
+    pub fn read_or_create_default() -> Result<Self, config_rs::ConfigError> {
         let path = Self::default_config_path()?;
 
         if path.exists() {
@@ -198,18 +196,13 @@ impl File {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use rand::rngs::mock::StepRng;
-    use reqwest::Url;
+    use log::LevelFilter;
     use spectral::prelude::*;
     use tempfile::NamedTempFile;
 
     #[derive(serde::Deserialize, PartialEq, Debug)]
     struct LoggingOnlyConfig {
         logging: Logging,
-    }
-
-    fn rng() -> StepRng {
-        StepRng::new(0, 0)
     }
 
     #[test]
@@ -278,7 +271,7 @@ mod tests {
     fn complete_logging_section_is_optional() {
         let config_without_logging_section = File {
             logging: None,
-            ..File::default(rng())
+            ..File::default()
         };
         let temp_file = temp_toml_file();
         let temp_file_path = temp_file.into_temp_path().to_path_buf();
@@ -301,7 +294,7 @@ mod tests {
 
     #[test]
     fn read_and_write_config_work() {
-        let config = File::default(rng());
+        let config = File::default();
         let temp_file = temp_toml_file();
         let path = temp_file.into_temp_path().to_path_buf();
 
--- a/cnd/src/config/file.rs
+++ b/cnd/src/config/file.rs
@@ -19,7 +19,6 @@ use std::{
 /// for filling in default values for absent configuration options.
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct File {
-    pub comit: Option<Comit>,
     pub network: Network,
     pub http_api: HttpSocket,
     pub database: Option<Database>,
@@ -36,7 +35,6 @@ impl File {
             .expect("cnd listen address could not be parsed");
 
         File {
-            comit: None,
             network: Network {
                 listen: vec![comit_listen],
             },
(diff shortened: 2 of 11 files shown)


cnd/src/cli.rs
```diff
@@ -3,7 +3,11 @@ use std::path::PathBuf;
 #[derive(structopt::StructOpt, Debug)]
 #[structopt(name = "COMIT network daemon")]
 pub struct Options {
-    /// Path to configuration file
+    /// Path to configuration file.
     #[structopt(short = "c", long = "config", parse(from_os_str))]
     pub config_file: Option<PathBuf>,
+
+    /// Path to secret seed file.
+    #[structopt(long = "secret-seed-file", parse(from_os_str))]
```
thomaseizinger: Have a short option here? Maybe `s`?

cnd/src/config/file.rs
```diff
@@ -66,7 +64,7 @@ pub struct Logging {
 
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct Comit {
-    pub secret_seed: Seed,
+    pub secret_seed_file: PathBuf,
```
thomaseizinger: I think you can make your life easier if we don't have this path.
A default path + the override from the commandline should be enough for now.
If anyone actually requests such a feature, we can always implement it. Doesn't seem to be worth the complexity for now :)

cnd/src/seed.rs
```diff
@@ -40,6 +48,168 @@ impl Seed {
...
+
+    fn write_to(&self, seed_file: PathBuf) -> Result<(), Error> {
+        ensure_directory_exists(seed_file.clone())?;
+        self.write_to_file(seed_file)?;
+        Ok(())
+    }
+
+    fn write_to_file(&self, path: PathBuf) -> Result<(), Error> {
+        let out = format!(
+            "-----BEGIN SECRET SEED-----\n{}\n-----END SECRET SEED-----\n",
```
thomaseizinger: If you are doing all of this manually, what is the `pem` library good for? :thinking:

tcharding: Good question, I asked this myself - that library is practically useless.

tcharding: On a serious note, it does the validation of the pem format i.e., opening tag, data, cloning tag.

thomaseizinger: According to the docs, you should be able to parse and encode the file contents with it?

- https://docs.rs/pem/0.6.1/pem/fn.parse.html
- https://docs.rs/pem/0.6.1/pem/struct.Pem.html#structfield.contents
- https://docs.rs/pem/0.6.1/pem/fn.encode.html

cnd/src/config/file.rs
tcharding: Implemented as suggested.

cnd/src/cli.rs
tcharding: Implemented as suggested.

tcharding: I had another go at the typescript stuff but have not managed to get it working.  If anyone wants to pick this up in order to get it in before the end of sprint they are welcome.  Otherwise I'll try and typescript on Monday.

tcharding: @luckysori @D4nte @thomaseizinger - shameless bump to bring attention to why this is still WIP.

thomaseizinger: I'll take a look :)

thomaseizinger: Implemented serializing a pem file for you in the `cnd_runner` file. I didn't run all the tests but the first couple ones definitely pass :)

cnd/src/seed.rs
thomaseizinger: @tcharding Please address this before merging, I think a fair bit of the logic here can be offloaded to the library :)

cnd/src/cli.rs
D4nte: Let's work together Monday?

cnd/src/seed.rs
tcharding: ha ha - I sat down to write a pem library tonight after the kids went to bed because of the terrible code I wrote in this PR.  Before starting I thought I had better take one more look at the library - well, as you posted here, turns out I can't read a Rust library.  Adding patches to fix this and use the library correctly.  Thanks for looking into this @thomaseizinger and for doing the Typescript stuff, I owe you a beer.

tcharding:
> Maybe I am seeing the diff wrong on my phone but it doesn't seem like you are using the pem library to its fullest yet? :)
> The writing still seems to be manual.

You are correct, my bad.  Will fix.

> Also, you don't seem to validate tag of the pem entry, but I am not sure how critical that actually is!

Yeah I thought to not bother validating the tag, its just one more thing to document and since the file only supports a single pem entry I don't see it adding any value for the added cost to the user and to us maintaining documentation.

> Both things are optimizations though, so I am happy for this one to go in!

Cheers

cnd/src/seed.rs
```diff
@@ -51,8 +191,14 @@ impl From<[u8; 32]> for Seed {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use pem;
     use rand::rngs::OsRng;
 
+    #[test]
+    fn seed_byte_string_must_be_32_bytes_long() {
+        let _seed = Seed::from(*b"this string is exactly 32 bytes!");
```
D4nte: Are you testing the failure?

cnd/src/seed.rs
```diff
@@ -40,6 +48,138 @@ impl Seed {
         rand.try_fill(&mut arr[..])?;
         Ok(Seed(arr))
     }
+
+    /// Construct a seed from base64 data read in from pem file.
+    pub fn from_file<D: AsRef<OsStr>>(seed_file: D) -> Result<Seed, Error> {
+        let file = Path::new(&seed_file);
+        log::info!(
+            "Found secret seed file, reading from {}",
```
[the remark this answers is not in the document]
tcharding: Good points, thanks for verifying the output.  Will fix

cnd/src/seed.rs
```diff
@@ -40,6 +48,138 @@ impl Seed {
...
+            return Err(Error::IncorrectLength(pem.contents.len()));
+        }
+
+        let mut array = [0; SEED_LENGTH];
+        for (i, b) in pem.contents.iter().enumerate() {
+            array[i] = *b;
+        }
+
+        Ok(Seed::from(array))
+    }
```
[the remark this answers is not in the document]
tcharding: This was intentional - is it not done like this in Rust (this is how its done in kernel C and in Golang)?

cnd/src/seed.rs
```diff
@@ -40,6 +48,138 @@ impl Seed {
...
+        log::info!(
+            "No seed file found, creating default at {}",
+            PrintablePath(&path)
+        );
+
+        Ok(())
+    }
+}
+
+fn ensure_directory_exists(file: PathBuf) -> Result<(), Error> {
```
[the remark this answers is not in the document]
tcharding: Yes but we have a logger here and not in config.

cnd/src/seed.rs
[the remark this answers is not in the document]
D4nte: This is not a blocking change and happy to merge it like that.

A `return` breaks the flow. I understand why it is intentional but to me it creates an `exception` to the logic that you have to keep in mind to move 
forward (mindload).

I guess it's a personal preference.

cnd/src/seed.rs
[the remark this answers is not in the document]
D4nte: Good point.

cnd/src/seed.rs
tcharding: Not like this but further down as part of pem string testing the failure case is tested.

cnd/src/seed.rs
```diff
@@ -40,6 +48,138 @@ impl Seed {
...
+        }
+    }
+    Ok(())
+}
+
+fn default_seed_path() -> Result<PathBuf, Error> {
+    crate::data_dir()
+        .map(|dir| Path::join(&dir, "secret_seed.pem"))
+        .ok_or_else(|| {
+            panic!("TODO: error handling");
```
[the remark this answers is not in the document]
tcharding: woops, fixed. Thanks.

cnd/src/seed.rs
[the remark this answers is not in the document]
tcharding: Discussed offline.  Yes _is_ personal preference and also case by case.

cnd/src/cli.rs
```diff
@@ -3,7 +3,11 @@ use std::path::PathBuf;
 #[derive(structopt::StructOpt, Debug)]
 #[structopt(name = "COMIT network daemon")]
 pub struct Options {
-    /// Path to configuration file
+    /// Path to configuration file.
     #[structopt(short = "c", long = "config", parse(from_os_str))]
     pub config_file: Option<PathBuf>,
+
+    /// Path to secret seed file.
+    #[structopt(short = "s", long = "secret-seed-file", parse(from_os_str))]
```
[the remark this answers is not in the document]
tcharding: Done!

cnd/src/seed.rs
```diff
@@ -40,6 +47,141 @@ impl Seed {
...
+        let pem_string = encode(&pem);
+
+        let mut file = File::create(path.clone())?;
+        file.write_all(pem_string.as_bytes())?;
+
+        Ok(())
+    }
+}
+
+fn ensure_directory_exists(file: PathBuf) -> Result<(), Error> {
```
luckysori: :question: Does this need to take an owned `PathBuf`?

cnd/src/seed.rs
```diff
@@ -85,4 +233,60 @@ mod tests {
...
     }
+
+    // 32 bytes base64 encoded.
+    const PEM: &str = "-----BEGIN SECRET SEED-----
+syl9wSYaruvgxg9P5Q1qkZaq5YkM6GvXkxe+VYrL/XM=
+-----END SECRET SEED-----
+";
+
+    #[test]
+    fn pem_library_works_as_expected() {
```
luckysori: :question: Is this test necessary, since you are only testing that a hopefully already tested library works as advertised?

cnd/src/seed.rs
```diff
@@ -85,4 +233,60 @@ mod tests {
...
+        let pem = pem::parse(PEM).unwrap();
+        assert_eq!(pem.tag, "SECRET SEED");
+    }
+
+    #[test]
+    fn seed_from_pem_works() {
+        let want = base64::decode("syl9wSYaruvgxg9P5Q1qkZaq5YkM6GvXkxe+VYrL/XM=").unwrap();
```
luckysori: :question: Isn't it a little weird to have the secret seed hardcoded twice? Maybe we should have a constant for the secret seed, embed it in the PEM string and then compare the outcome of deriving the seed from the PEM string to the original seed? Not sure if this is better, just wondering.

cnd/src/seed.rs
```diff
@@ -85,4 +233,60 @@ mod tests {
...
+        let pem = pem::parse(PEM).unwrap();
+        let got = Seed::from_pem(pem).unwrap();
+        assert_eq!(got.0, *want);
+    }
+
+    #[test]
+    #[should_panic]
```
luckysori: :question: Would it be better to match against the error instead of panicking? Again, I don't know what's best. Maybe this is fine!

cnd/src/seed.rs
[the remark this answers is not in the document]
tcharding: Spent a few hours on this - it turns out it is not a trivial problem.  Will leave it like this for now and create an issue.

cnd/src/seed.rs
```diff
@@ -40,6 +47,141 @@ impl Seed {
...
+            PrintablePath(&file.to_path_buf())
+        );
+
+        let contents = fs::read_to_string(file)?;
+        let pem = pem::parse(contents)?;
+
+        Seed::from_pem(pem)
+    }
+
+    pub fn from_pem(pem: pem::Pem) -> Result<Seed, Error> {
```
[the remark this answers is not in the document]
tcharding: Made private.

cnd/src/seed.rs
tcharding: Removed, thanks.

cnd/src/seed.rs
tcharding: Yes, IIUC `PathBuf` is like `String` and parameters typically always consume it.  Discussed offline, posting here in case I'm wrong.

cnd/src/seed.rs
tcharding: With the above mentioned test now removed these consts can be moved into the unit test that uses them.

cnd/src/seed.rs
tcharding: Nice idea, code is a bit more verbose but more accurate.

cnd/src/main.rs
```diff
@@ -123,8 +128,8 @@ fn main() -> Result<(), failure::Error> {
     Ok(())
 }
 
-fn derive_key_pair(secret_seed: &Seed) -> identity::Keypair {
-    let bytes = secret_seed.sha256_with_seed(&[b"NODE_ID"]);
+fn derive_key_pair(seed: &Seed) -> identity::Keypair {
+    let bytes = seed.sha256_with_seed(&[b"NODE_ID"]);
```
D4nte: ❓  What is that about?

tcharding: rename _all_ instances of 'secret seed' to 'seed'.  In line with the suggestion you made on shortening the command line option from `--secret-seed-file` in `--seed-file`.

D4nte: I wondered about the `"NODE_ID"` but I see it is actually not new.

tcharding: Yes, the logic behind `sha256_with_seed()` is a mystery to me also.

Issue #1455, opened 2019-09-25
The way we configure ledgers is redundant and we don't validate on whether it is correct


thomaseizinger:
Currently, the configuration file contains, among other things, the following:

```toml
[bitcoin]
network = "regtest"
node_url = "http://localhost:18443"

[ethereum]
network = "regtest"
node_url = "http://localhost:8545"
```

**Problem:** This is redundant + we don't validate whether it is actually true (i.e. whether the node at `localhost:8545` is actually a regtest node).

**Goal:**
1. Provide a non-redunant and easy to configure way of which ledgers the software should connect to.
2. Validate, which ledgers we are connecting to.

**Recommendation:**

1. The sections should be optional (I think this is already the case)
2. When specified, the minimum should be to specify a network. Given a network, the software should:
- for Bitcoin, always use the bitcoind connector
- for Ethereum, always use the web3 connector

Further customization is purposely deferred to the future to not over-complicate the configuration. Once the p2p connectors (https://github.com/comit-network/comit-rs/issues/1250, https://github.com/comit-network/comit-rs/issues/1409) are implemented, we will have to think of a useful, yet simple way to configure these.

bonomat:
Closed in favor of: #1844

PR #1573, opened 2019-10-25
Use ChainId for Ethereum

### LINKED ISSUE #1440: Use chainId to identify Ethereum networks
Currently, we use our own `Network` enum to differentiate between different Ethereum chains:

https://github.com/comit-network/comit-rs/blob/bdaeaa879133de67d73b4692944475d96659a921/internal/ethereum_support/src/network.rs#L16-L25

**Problem:**
This actually caused a lot of pain already because the rest of the Ethereum ecosystems works with chain and network IDs and not with names.

**Recommendation:**
- [ ] Change this enum to `Chain` and encode the chainId with each variant.
- [ ] Use the chainId when encoding the Ledger inside COMIT messages
- [ ] Add the chainId to the payload when returning actions to clients (i.e. don't make a breaking change right away, otherwise it is going to be a pain to upgrade clients)
- [ ] Create follow-up tickets to remove current mapping logic from clients (for example: https://github.com/coblox/bobtimus/blob/98ed937d5b87f664ee4d6e1f931ad2d1eb5a82de/src/ledgerExecutor.ts#L231-L235)

D4nte:
Resolves #1440.

- Internally use chain id on the Ethereum swap protocol ledger.
- Replace `network` with `chain_id` on the COMIT API.
- Add `chain_id` to the HTTP API: return it along side `network`, accept it on the request. Still accept `network` on the request for backward compatibility.
- Move Ethereum ledger `network` code to http_api module (from ethereum_support).


api_tests/lib/ethereum.ts
```diff
@@ -179,20 +178,20 @@ export class EthereumWallet {
...
         data: string = "0x0",
         value: BigNumber | number | string = "0",
+        chainId: number = 17,
         gasLimit = "0x3D0900"
     ) {
         const nonce = await ethersClient.getTransactionCount(this.address());
 
         value = ethers.utils.bigNumberify(value);
 
-        const network = await ethersClient.getNetwork();
```
thomaseizinger: Careful here, I _think_ this should stay. In the end, you are connecting to this network! :)

cnd/src/config/file.rs
```diff
@@ -102,7 +102,6 @@ pub struct Bitcoin {
 
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct Ethereum {
-    pub network: ethereum_support::Network,
```
thomaseizinger: I think last time we touched this, we ought to keep it because it is the primary way we want to specify things, even though we are not using yet :/

thomaseizinger: See here https://github.com/comit-network/comit-rs/issues/1455.

D4nte: If it is not use that not only it's useless but also dangerous are we luring the user in thinking it's used and possibly that some checks are done.

To be re-introduced with #1455 then.

api_tests/lib/ethereum.ts
D4nte: Let's look at the difference of behaviour:

**Before:**
The network returned by cnd for the blockchain action (deploy contract) is ignored (not even checked). The network used to do the transaction is the one that is taken from the node we are connected to.

**After:**
The chain id returned by cnd for the blockchain action (deploy contract) is used to do the transaction.

I think the **after** behaviour makes more sense.

api_tests/lib/comit.ts
```diff
@@ -41,7 +41,8 @@ export interface Asset {
 
 export interface Ledger {
     name: string;
-    network: string;
+    network?: string;
```
da-kami: ❓ should maybe be refactored?

After quick discussion with @D4nte the network has to stay here because the `Ledger` is used for Bitcoin and Ethereum and for Bitcoin we still have a network name. 

I feel it is a bit weird to have the network name just in use for Bitcoin and the chainId just in use for Ethereum. 

❓ For Bitcoin it will always be by name?

cnd/src/http_api/action.rs
```diff
@@ -32,6 +32,7 @@ pub enum ActionExecutionParameters {
     None {},
 }
 
+/// `network` field here for backward compatibility, to be removed with #TODO
```
da-kami: 📣 please create ticket and remove TODO

cnd/src/http_api/ethereum_network.rs
```diff
@@ -35,6 +37,25 @@ impl Network {
     }
 }
 
+impl From<ChainId> for Network {
```
da-kami: ❓ Is this for backwards compatibility? Will it be removed in the future?

cnd/src/swap_protocols/ledger/ethereum.rs
```diff
@@ -1,21 +1,24 @@
 use crate::swap_protocols::ledger::{Ledger, LedgerKind};
-use ethereum_support::{Address, Network, Transaction};
+use ethereum_support::{Address, Transaction};
+use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+/// `network` is only kept for backward compatibility with client
+/// and must be removed with issue #TODO
```
da-kami: 📣 related to same comment about TODO above.

cnd/src/http_api/mod.rs
```diff
@@ -90,8 +91,37 @@ impl Serialize for Http<bitcoin::Transaction> {
...
+
+        let mut state = serializer.serialize_struct("", 3)?;
+        state.serialize_field("name", "ethereum")?;
+        state.serialize_field("chain_id", &self.0.chain_id)?;
+        state.serialize_field("network", &network)?;
+        state.end()
+    }
+}
+
+// Can re-use macro once `network` is removed with #TODO
```
da-kami: 📣 related to same comment about TODO

api_tests/lib/comit.ts
D4nte: 1. Happy to use dedicated separate interface for Bitcoin and Ethereum.
2. The requirement to change to chainId comes from the fact for each of our clients, we end-up doing a mapping to ensure that we are connected to an Ethereum node with the right network. However, for Bitcoin, we do not even do such checks.
The conclusion from the discussion with @thomaseizinger was that on Bitcoin side, we need to learn more about the usage. See how the bitcoin network is used by cnd clients. Only then, if we realise there is duplication to map the value return by cnd to something else (`testnet3`? Magic block number?) then we can do the change in cnd. If we were to change bitcoin now, it would just be a wild gues.

cnd/src/http_api/ethereum_network.rs
D4nte: Yes, it is used by the HTTP API, as per the issue's description.

api_tests/lib/comit.ts
da-kami: Thanks for explaining @D4nte. Refactoring if needed can be in a follow-up PR.

D4nte: Already done :)

Issue #1570, opened 2019-10-24
Cannot re-spawn state machine for Alice through AliceSpawner


tcharding:
## Problem

In order to achieve https://github.com/comit-network/comit-rs/issues/1244, the interface of `AliceSpawner` needs to only accept the _data_ it needs to start the state machine. At the moment, this is not possible.

## Goal 

Be able to re-spawn state machine for Alice.

## Recommendation

Several steps are required/possible to achieve this.

1. The interface of `AliceSpawner` needs to be aligned with `BobSpawner`
2. (1) makes the interfaces functionally equivalent and hence the duplication can be removed in favor of a single `Spawner` trait. (this also unifies `alice::CommunicationState` and `bob::CommunicationState`)
3. (1) also moves a lot of code to the caller (the HTTP controller). A lot of code can likely be simplified by getting rid of `alice::ProtocolDependencies` and merging it into `http_api::Dependencies`.
4. The `comit_client::Client` trait can likely be merged into the `Network` trait

EDIT: (2) and (3) are actually, AFAICT, not possible without re-designing/re-thinking the `Depenedencies` facade.  The facade requires the traits `BobSpawner` and `AliceSpawner` to exist in order to be able have these traits as trait bounds.

The purpose of this issue is to move the database implementation forwards.  Points 2-4 are nice to have but not required to achieve this.  After spending about 4 hours working on these points I (tcharding) decided to open new issues for points 2-4.

Opened issue: https://github.com/comit-network/comit-rs/issues/1596 to see to (2) and (3).

~Opened~ Closed issue: https://github.com/comit-network/comit-rs/issues/1597 to see to (4).  Please see issue for reasoning behind close.`

tcharding:
Closed by: https://github.com/comit-network/comit-rs/pull/1604

PR #1598, opened 2019-10-30
Refactor AliceSpawner

tcharding:
We would like to be able to re-spawn Alice for swaps that are in progress if cnd restarts.  Currently this is not possible because of the way that `AliceSpawner` is implemented.

Refactor `AliceSpawner` so that Alice can be re-spawned.  This is in line with how `BobSpawner` currently is.  This means
- Store the swap in the metadata store and the state store
- Send the request to peer node (Bob)
- Spawn Alice.

Resolves: #1570


cnd/src/comit_client.rs
```diff
@@ -37,3 +48,117 @@ pub enum RequestError {
...
+    >(
+        &self,
+        dial_information: DialInformation,
+        request: rfc003::Request<AL, BL, AA, BA>,
+    ) -> Box<dyn Future<Item = rfc003::Response<AL, BL>, Error = RequestError> + Send>
+    where
+        LedgerEventDependencies: CreateLedgerEvents<AL, AA> + CreateLedgerEvents<BL, BA>,
+    {
+        let request = build_swap_request(request)
+            .expect("constructing a frame::OutoingRequest should never fail!");
```
D4nte:
```suggestion
            .expect("constructing a frame::OutgoingRequest should never fail!");
```

cnd/src/http_api/mod.rs
```diff
@@ -306,33 +309,31 @@ impl<'de> Deserialize<'de> for DialInformation {
 /// to another implementation. This allows us to keep the number of arguments to
 /// HTTP API controllers small and still access all the functionality we need.
 #[derive(Debug)]
-pub struct Dependencies<M, S, A, B, N> {
-    pub metadata_store: Arc<M>,
-    pub state_store: Arc<S>,
+pub struct Dependencies<A, B, S> {
+    pub metadata_store: Arc<InMemoryMetadataStore>,
```
D4nte:
```suggestion
    pub metadata_store: Arc<MetadataStore>,
```
I understand that a MedatadaStore will always be in memory. As all Rust struct are :)

tcharding: Commented in another thread on this.  Perhaps we can chat offline about the motivation - basically I am not a big fan of generics when there is only, and only likely ever going to be, a single struct that implements the trait.  It seems to add needless complexity, of course you have much more Rust experience so I am open to your opinions on this.

D4nte: I don't understand your comment.

What I am saying is: Why do we call this `InMemoryMedataStore` when, as per your previous comments, will ever be only one type of `MetadataStore` due to the fact that we always load db items in this `InMemoryMedataStore`.

I am just proposing a rename here. Not sure you why you are going about generics and all.

I don't like the fact that we prefix the name with `InMemory`. If there is not going to be a `OnFilesystem` version of this then let's not prefix it with `InMemory`. Let's rename to `MetadataStore`.

tcharding: Cool, I agree with you totally.  I have it on my list of clean ups to remove the generics and just have a single MetatadataStore.  And do the same for StateStore.

tcharding:
If its ok with the team I will merge https://github.com/comit-network/comit-rs/pull/1604 after review as it is a superset of this one (and has been rebased on top of current master already).  Then I will close this one, leaving open for now however until the other merges.

thomaseizinger:
https://github.com/comit-network/comit-rs/pull/1604 seems to be a strict superset and we decided to merge that. Will close this then!

PR #1601, opened 2019-10-31
Use concrete types

tcharding:
Use of generics for traits that have only a single implementation adds complexity.

Remove usage of `K` and use `SwapId`.

Remove usage of the `MetadataStore` and `StateStore` traits and use the structs that satisfy these traits, `InMemoryMetadataStore` and `InMemoryStateStore` respectively.

### DIFF
--- a/cnd/src/http_api/mod.rs
+++ b/cnd/src/http_api/mod.rs
@@ -29,7 +29,7 @@ use crate::{
     swap_protocols::{
         asset::Asset,
         ledger::{ethereum, Bitcoin, Ethereum},
-        metadata_store,
+        metadata_store::{self, InMemoryMetadataStore, MetadataStore},
         rfc003::{
             alice::{self, AliceSpawner},
             bob::BobSpawner,
@@ -38,7 +38,7 @@ use crate::{
             state_store::{self, StateStore},
             ActorState, CreateLedgerEvents, Ledger,
         },
-        LedgerEventDependencies, Metadata, MetadataStore, SwapId, SwapProtocol,
+        LedgerEventDependencies, Metadata, SwapId, SwapProtocol,
     },
 };
 use bitcoin::{util::amount::Denomination, Amount as BitcoinAmount};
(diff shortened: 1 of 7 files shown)


cnd/src/http_api/mod.rs
```diff
@@ -306,15 +306,15 @@ impl<'de> Deserialize<'de> for DialInformation {
 /// to another implementation. This allows us to keep the number of arguments to
 /// HTTP API controllers small and still access all the functionality we need.
 #[derive(Debug)]
-pub struct Dependencies<M, S, A, B, N> {
-    pub metadata_store: Arc<M>,
-    pub state_store: Arc<S>,
+pub struct Dependencies<A, B, N> {
+    pub metadata_store: Arc<InMemoryMetadataStore>,
```
D4nte: 📣 This `M` trait bound was here so that once the database is implemented, we could choose between initialising in the `main` our structs with either `InMemoryMetadataStore` or a new implementation `SqliteMetadataStore` (I made up the name).

If now you are forcing the use of `InMemory` down in the library (`http_api` module), then how will you replace with sqlite?

tcharding: You are correct, this was the initial idea behind `MetadataStore`.  During previous work on the database it came to light that this was not going to work and that we would keep the metatadata store in memory always and load it with data from the database.  Same applies to the `StateStore`.

tcharding:
I opened this PR because I thought https://github.com/comit-network/comit-rs/pull/1598 was stuck (these patches are the first patches of that PR).  That work has now been done to completion.  Closing this in favour of the other PR.

cnd/src/http_api/mod.rs
D4nte: If the metadata store would always be in memory then can we just rename `InMemoryMetadataStore` to `MetadataStore`? To me, the `InMemory` qualification means that there is an alternative (because after all, everything is in memory....)

D4nte:
delete branch?

PR #1603, opened 2019-11-01
Remove task executor from ComitNode

tcharding:
Recent work saved the channel (future) for an incoming swap request instead of
spawning the future.  At that time we still had a future for the error path.  It
was not noticed that this error path future was only there because of the now
removed spawning.  We can remove this error path future also, this means the
task executor is no longer needed in the ComitNode.


D4nte:
Why is bors not merging that??

Issue #1616, opened 2019-11-05
Ethereum: Allow queries to be run from a timestamp


luckysori:
Fetch all blocks up until we have one that is before the timestamp of a query and match the query against the blocks.

Remove the sleep added in #1288 as part of this ticket.

luckysori:
Blocked it on #1516, because that also will touch the implementation of `MatchingTransactions`.

PR #1617, opened 2019-11-05
Look for matching transactions in all Bitcoin blocks since provided timestamp

### LINKED ISSUE #1289: Bitcoin: Allow queries to be run from a timestamp
Fetch all blocks up until we have one that is _before_ the timestamp of a query and match the query against the blocks.

luckysori:
In order to keep this PR small I decided to only work on Bitcoin. Doing this for Ethereum is tracked in #1616.

---

Fixes #1289.


btsieve/src/bitcoin/mod.rs
```diff
@@ -37,22 +37,29 @@ where
...
     pattern: TransactionPattern,
+    timestamp: Option<u32>,
 ) -> Result<bitcoin::Transaction, ()>
 where
     C: LatestBlock<Block = bitcoin::Block, Error = E>
         + BlockByHash<Block = bitcoin::Block, BlockHash = sha256d::Hash, Error = E>
         + Clone,
     E: Debug + Send + 'static,
 {
+    let reference_timestamp = timestamp.unwrap_or(u32::max_value());
```
D4nte:
```suggestion
    let reference_timestamp = timestamp.unwrap_or(u32::max_value());
```
I lol'd.

btsieve/src/bitcoin/mod.rs
```diff
@@ -92,13 +99,36 @@ where
         }
         missing_block_futures = new_missing_block_futures;
 
+        if let Some(block) = oldest_block.clone() {
+            if block.header.time >= reference_timestamp {
```
D4nte: 🔧 I find this piece of code quite hacky. Can't we have 2 different logic depending on whether `reference_timestamp` is `Some(..)` or `None`?
Now I have to think "ok, if there is no timestamp provided then this condition is always false..." too much mental load.

```suggestion
		if let Some(reference_timestamp) = reference_timestamp {
            if block.header.time >= reference_timestamp {
```

Issue #1610, opened 2019-11-04
Spawners contain duplicate code


tcharding:
## Problem

The spawner functions for Alice and Bob are functionally identical except for a single line of code.  The two files `alice/spawn_alice.rs` and `bob/spawn_bob.rs` are each 118 lines of code; all lines, excluding function/trait names and use statements, are identical except for:

```
            state_store.update::<alice::State<AL, BL, AA, BA>>(&id, update);
```

Further moer, one of the reasons we currently have _two_ spawners is because knowledge of the role is currently within the swap_protocols logic.  If we move all role knowledge over to the http_api then a lot of the swap_protocols code will/may become more simple.  There is more to this argument (concerning type parameters being passed down the function call chain `AL, BL, AA, BA`) explained to me by @thomaseizinger but I do not recall it fully now.

## Goal

Reduce the code duplication between `alice::SpawnAlice` and `bob::SpawnAlice`.

PR #1620, opened 2019-11-06
Unify Alice and Bob spawners into a single Spawn trait

tcharding:
Latest work on refactoring in order to get us closer to persistent storage.

- Patch 1 heavily refactors `Connector` and all the Alice/Bob traits that are passed to the http_api layer.  Moves all (see patch 4 below) knowledge of the role to the http_api layer.
- Patch 2 unifies the spawners.
- Patch 3 does refactoring within `post_swap.rs` (handling accept/decline for alice).
- Patch 4 fixes bugs introduced in patch 1 - not squashing in case anyone wants to review patches 1 and 3 separately. 

Fixes: #1610


cnd/src/http_api/connector.rs
```diff
@@ -0,0 +1,163 @@
...
+use std::sync::Arc;
+
+/// Collect all the connector trait bounds together under one trait.
+pub trait Connect:
+    Clone + MetadataStore + StateStore + Network + SendRequest + SpawnAlice + SpawnBob + SwapSeed
+{
+}
+
+#[allow(missing_debug_implementations)]
+pub struct Connector<S> {
```
thomaseizinger: What is the difference between this and `http_api::Dependencies`?

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
```diff
@@ -61,6 +110,87 @@ pub fn handle_post_swap<C: Connect>(
...
+            let id = swap_request.id;
+            let alice_state = match response.clone() {
+                Ok(accept) => State::accepted(swap_request.clone(), accept, seed),
+                Err(decline) => State::declined(swap_request.clone(), decline, seed),
+            };
+            StateStore::insert(&dependencies, swap_request.id, alice_state.clone());
+
+            let receiver = Spawn::spawn(&dependencies, swap_request, response);
+
+            tokio::spawn(receiver.for_each(move |update| {
```
thomaseizinger: To harmonize this with the other usage, we could do this also outside of this function :)

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
```diff
@@ -1,40 +1,89 @@
...
+pub fn handle_post_swap<D: Clone + StateStore + MetadataStore + SendRequest + Spawn + SwapSeed>(
+    dependencies: D,
     request_body_kind: SwapRequestBodyKind,
 ) -> Result<SwapCreated, HttpApiProblem> {
     let id = SwapId::default();
 
     match request_body_kind {
         SwapRequestBodyKind::BitcoinEthereumBitcoinErc20Token(body) => {
-            con.initiate_request(id, body.peer.clone(), Box::new(body.clone()))?;
+            let seed = dependencies.swap_seed(id);
```
thomaseizinger: I feel like all of this code could go into a single helper-function `handle_supported_swap` (because it handles all combinations of supported swap requests). The only two parameters should be `dependencies` and `body`.

cnd/src/dependencies.rs
```diff
@@ -0,0 +1,25 @@
...
+        LedgerConnectors, SwapId,
+    },
+};
+use std::sync::Arc;
+
+/// Core cnd dependencies required by both the libp2p network layer and the HTTP
+/// API layer.
+#[allow(missing_debug_implementations)]
+#[derive(Clone)]
+pub struct Dependencies {
```
thomaseizinger: Personal opinion ahead: I think you are following DRY a bit to rigorous here :see_no_evil: 
Just because two structs have the same fields, doesn't mean it is (bad) duplication. This kind of de-duplication often leads to dependencies between modules that can actually bite you fairly badly. Plus, it is a layer of indirection that readers of the code will need to follow.

I you feel strongly about it, I am happy to keep it and see how it goes :)

cnd/src/swap_protocols/rfc003/spawn.rs
```diff
@@ -1,62 +1,58 @@
...
-use futures::{sync::mpsc, Stream};
+use futures::sync::mpsc;
 use futures_core::{
     compat::Future01CompatExt,
     future::{FutureExt, TryFutureExt},
 };
 use std::sync::Arc;
 
-pub trait SpawnAlice: Send + Sync + 'static {
+pub trait Spawn: Send + Sync + 'static {
```
thomaseizinger: We've started to talk about an "execution layer" in COMIT. This trait is basically representing that. What do you think about renaming it to `ExecuteSwap` or `SwapExecutor` (analogous to `TaskExecutor`).

cnd/src/http_api/routes/rfc003/handlers/action.rs
```diff
@@ -81,7 +93,17 @@ pub fn handle_action<C: Connect>(
                                     channel.send(response).map_err(problem::send_over_channel)?;
 
                                     let request = state.request();
-                                    con.spawn_bob(request, Err(decline_body));
+                                    let receiver =
```
thomaseizinger: The tests are likely failing because you need to insert the initial state here before you call `spawn`, i.e. we never insert the declined state here.

Actually, for declined, we no longer need to call spawn here because nothing is happening after `decline`. That is one of the benefits of splitting up the work `spawn` is/was doing!

cnd/src/http_api/connector.rs
tcharding: My bad, that file was supposed to be deleted.  This was re-named to `http_api::Dependencies`.

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
tcharding: Good idea, will attempt.

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
tcharding: Yes, I had made zero attempt to refactor this code - well spotted.

cnd/src/dependencies.rs
tcharding: Interesting point, your view is somewhat expressed by the way I declared the second instance of `Dependencies` in main explicitly instead of cloning the first instance passed into the `ComitNode`
```
    let http_api_dependencies = http_api::Dependencies {
        dependencies: Arc::new(Dependencies {
            ledger_events: ledger_events.clone(),
            metadata_store: Arc::clone(&metadata_store),
            state_store: Arc::clone(&state_store),
            seed,
        }),
        swarm: Arc::clone(&swarm),
    };
```

I actually tried putting the `Dependencies` fields into a struct that was used by the http_api layer but this did not work out as nice - open to being corrected but it seems we need _some_ object to implement `Spawn` on.  `spawn()` needs access to the seed and the `ledger_events` so AFAICT we need an object that holds both of these at least.  Am I missing someting?  Actually the seed could be passed into spawn then spawn could be implemented on `LedgerConnectors` but the next developer is going to struggle to see why that is the case?  I can have a play with this and see if I come up with something.

cnd/src/swap_protocols/rfc003/spawn.rs
tcharding: I may have missed something in that discussion but I am under the impression that the request/response COMIT messages are still grouped under the execution phase (probably need renaming if we do add a separate negotiation protocol.  Happy to rename but may need further thought.

cnd/src/http_api/routes/rfc003/handlers/action.rs
tcharding: Thanks for the suggestion, I do not think this is the case.  We insert the proposed state in `initiate_request()` then insert the accept/decline state in `send_request_and_spawn_alice()`.  You are right there used to be a state insert during spawn but this was already removed in the previous patch (that passes e2e tests).

thomaseizinger: Fair point, will do some more investigation to see why the tests fail then :)

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
tcharding: Discussed offline.

cnd/src/swap_protocols/rfc003/spawn.rs
tcharding: No need for comment, discussed offline.

tcharding:
> Awesome work man!
> Loving how this refactoring is going :)

Word.  Feels good huh.

tcharding:
@thomaseizinger I just pushed another commit onto this branch that removes the `Dependencies` struct.  This is as you suggested during our call and was an unforeseen benefit of the spawner unification bug fixing I did earlier - I did not realise at the time `seed` is not needed in `spawn()` now so BOOM away goes the need for `Dependencies`.

cnd/src/http_api/routes/rfc003/handlers/action.rs
```diff
@@ -53,16 +57,34 @@ pub fn handle_action<C: Connect>(
...
+                                        swap_request.clone(),
+                                        accept_body.clone(),
+                                        seed,
+                                    );
+                                    StateStore::insert(&dependencies, id, state);
+
+                                    let receiver =
+                                        Spawn::spawn(&dependencies, swap_request, accept_body);
+
+                                    tokio::spawn(receiver.for_each(move |update| {
```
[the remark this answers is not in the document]
luckysori: I found the answer: [`UnboundedReceiver`](https://docs.rs/futures/0.1.29/futures/sync/mpsc/struct.UnboundedReceiver.html) implements `Stream`.

Issue #1553, opened 2019-10-21
No CORS problems when using cnd during the workshop/hackathon


da-kami:
## Problem

A browser-based application will currently run into CORS limitations as it tries to connect to cnd because the HTTP API is running on a different port than the webserver of the application.

## Goal

Allow browser-based applications to interact with cnd.

## Recommendation

Add an optional configuration parameter `cors` to the config file in the `http_api` section.

Possible values should be :

- `all`: allow requests from all hosts
- `none` (default value): don't include cors headers
- `host:port`: only allow requests from the given host and port

Create-comit-app should then set this value to `all` for the instances it launches during developlment.

This recommendation will require to implement our own `cors` module for warp because the existing one cannot handle the `all` case. Fortunately, the cors headers support wildcards (`*`), so we just have to write a little module that sets all the headers to `*` if `all` is specified.

PR #1626, opened 2019-11-07
Remove preflight

thomaseizinger:
We are no longer embedding comit-i, hence there is no need (for now), to specify these cors headers for the arbitrary origin of localhost:3000.


thomaseizinger:
> So we don't need any of this code for #1553?

Parts maybe, but since we don't know the host & port for #1553, implementing #1553 is going to be more difficult compared to what we have now.
For now it is dead code, so there is no reason to keep it :)

Whoever implements #1553 can always look at the history if they want to get inspired, but as already mentioned in the ticket:

```
This recommendation will require to implement our own cors module for warp because the existing one cannot handle the all case. Fortunately, the cors headers support wildcards (*), so we just have to write a little module that sets all the headers to * if all is specified.
```

What warp provides will likely not be sufficient.

PR #1627, opened 2019-11-07
Allow for an eager `bors r+`

thomaseizinger:
Currently, one has to wait for a successful build before the PR can be handled to bors. However, sometimes we now upfront a build will succeed (textfile only changes for example). Hence, we can immediately delegate to bors without waiting for the build.


thomaseizinger:
bors retry

thomaseizinger:
bors retry

PR #1630, opened 2019-11-08
Add table declaration statements for request & response messages

thomaseizinger:
This patch adds 6 table declarations to our database module. To keep the writing and reading code simple, we add a table per combination of AL/BL/AA/BA that is currently supported by cnd.

This work was separated to allow for parallel work on the reading and writing code because the macros generate the glue code for accessing the database.

### DIFF
--- a/cnd/src/db/schema.rs
+++ b/cnd/src/db/schema.rs
@@ -9,3 +9,99 @@ table! {
         role -> Text,
     }
 }
+
+table! {
+   rfc003_bitcoin_ethereum_bitcoin_ether_request_messages {
+       id -> Integer,
+       swap_id -> Text,
+       bitcoin_network -> Text,
+       ethereum_chain_id -> Integer,
+       bitcoin_amount -> Text,
+       ether_amount -> Text,
+       hash_function -> Text,
+       bitcoin_refund_identity -> Text,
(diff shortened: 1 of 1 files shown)


cnd/src/db/schema.rs
```diff
@@ -9,3 +9,99 @@ table! {
...
 }
+
+table! {
+   rfc003_bitcoin_ethereum_bitcoin_ether_request_messages {
+       id -> Integer,
+       swap_id -> Text,
+       bitcoin_network -> Text,
+       ethereum_chain_id -> Integer,
+       bitcoin_amount -> Text,
+       ether_amount -> Text,
```
luckysori: :question: Could/should amounts be `Integer`s?

thomaseizinger: We are storing wei and satoshis, storing those as integers could be dangerous in regards to overflows.

thomaseizinger:
@tcharding, we might actually want to store certain things as bytes (identities, the secret hash, etc) and not as hex-encoded strings. What do you think?

cnd/src/db/schema.rs
luckysori: True that.

PR #1632, opened 2019-11-08
Store request, accept and decline messages in the database

thomaseizinger:


cnd/src/db/save_message.rs
```diff
@@ -0,0 +1,353 @@
...
+    + SaveMessage<Request<Bitcoin, Ethereum, bitcoin::Amount, ethereum_support::Erc20Token>>
+    + SaveMessage<Request<Ethereum, Bitcoin, ethereum_support::EtherQuantity, bitcoin::Amount>>
+    + SaveMessage<Request<Ethereum, Bitcoin, ethereum_support::Erc20Token, bitcoin::Amount>>
+    + SaveMessage<Accept<Bitcoin, Ethereum>>
+    + SaveMessage<Accept<Ethereum, Bitcoin>>
+    + SaveMessage<Decline>
+{
+}
+
+macro_rules! impl_save_message {
```
D4nte: 📣 From the `Http<I>` experience I'd be keen to not use this macro.
It really does not give much (only function signatures and `let $conn = self.connect().unwrap();`) but decrease readability.

The next person who implement a new `save_message` will have to understand how to use the macro to then use it. I think it increases your mental load for little gain.

cnd/src/db/save_message.rs
```diff
@@ -0,0 +1,353 @@
...
+    ethereum_redeem_identity: Text<EthereumAddress>,
+    bitcoin_expiry: U32,
+    ethereum_expiry: U32,
+    secret_hash: Text<SecretHash>,
+}
+
+impl_save_message! {
+    fn save_message(connection: SqliteConnection, message: Request<Bitcoin, Ethereum, bitcoin::Amount, ethereum_support::EtherQuantity>) -> Result<(), Error> {
+        let Request {
+            id,
```
D4nte: The use of `id` here not to represent the table `id` but to represent the table `swap_id` only reinforce my conviction that we should remove `id` from the table (and only have `swap_id`).

cnd/src/db/save_message.rs
```diff
@@ -0,0 +1,353 @@
...
+            ethereum_chain_id: U32(beta_ledger.chain_id.into()),
+            bitcoin_amount: Text(Satoshis(alpha_asset.as_sat())),
+            erc20_amount: Text(DecimalU256(beta_asset.quantity.0)),
+            erc20_token_contract: Text(EthereumAddress(beta_asset.token_contract)),
+            hash_function: Text(hash_function),
+            bitcoin_refund_identity: Text(alpha_ledger_refund_identity.into_inner()),
+            ethereum_redeem_identity: Text(EthereumAddress(beta_ledger_redeem_identity)),
+            bitcoin_expiry: U32(alpha_expiry.into()),
+            ethereum_expiry: U32(beta_expiry.into()),
+                    secret_hash: Text(secret_hash)
```
D4nte: Format issue?

cnd/src/db/save_message.rs
```diff
@@ -0,0 +1,353 @@
...
+#[table_name = "rfc003_decline_messages"]
+struct InsertableDeclineMessage {
+    swap_id: Text<SwapId>,
+    reason: Option<String>,
+}
+
+impl_save_message! {
+    fn save_message(connection: SqliteConnection, message: Decline) -> Result<(), Error> {
+        let Decline {
+            id, reason: _reason // we don't map reason to a DB type because will be gone soon (hopefully)
```
D4nte: Can you open an issue so we can track it properly?

cnd/migrations/2019-11-11-034058_create-message-tables/up.sql
```diff
@@ -0,0 +1,91 @@
...
+    bitcoin_redeem_identity  NOT NULL,
+    ethereum_refund_identity NOT NULL,
+    bitcoin_expiry           NOT NULL,
+    ethereum_expiry          NOT NULL,
+    secret_hash              NOT NULL
+);
+
+CREATE TABLE rfc003_bitcoin_ethereum_bitcoin_erc20_request_messages
+(
+    id INTEGER               NOT NULL PRIMARY KEY,
```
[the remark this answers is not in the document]
bonomat: Good idea. Since `swap_id` is `unique` it can also just serve as the `id`.

cnd/src/http_api/routes/rfc003/handlers/action.rs
```diff
@@ -53,29 +56,32 @@ pub fn handle_action<D: MetadataStore + StateStore + Network + Spawn + SwapSeed>
                 .select_action(action_kind, method)
                 .and_then({
                     |action| match action {
-                        Action::Accept(action) => serde_json::from_value::<AcceptBody>(body)
+                        Action::Accept(_) => serde_json::from_value::<AcceptBody>(body)
```
D4nte: If we do not use the inner value then should(can?) we remove it from the enum?

cnd/src/main.rs
```diff
@@ -68,12 +69,15 @@ fn main() -> Result<(), failure::Error> {
     let local_peer_id = PeerId::from(local_key_pair.clone().public());
     log::info!("Starting with peer_id: {}", local_peer_id);
 
+    let database = Sqlite::new(settings.database.as_ref().map(|d| d.sqlite.clone())).unwrap();
```
D4nte:
```suggestion
    let database = Sqlite::new(settings.database.as_ref().map(|d| d.sqlite.clone())).expect("Could not initialize database handler");
```

cnd/migrations/2019-11-11-034058_create-message-tables/up.sql
```diff
@@ -0,0 +1,91 @@
...
+CREATE TABLE rfc003_bitcoin_ethereum_bitcoin_ether_request_messages
+(
+    id INTEGER               NOT NULL PRIMARY KEY,
+    swap_id UNIQUE           NOT NULL,
+    bitcoin_network          NOT NULL,
+    ethereum_chain_id        NOT NULL,
+    bitcoin_amount           NOT NULL,
+    ether_amount             NOT NULL,
+    hash_function            NOT NULL,
+    bitcoin_refund_identity  NOT NULL,
```
bonomat: ❓I'm a bit rusty in SQL, so please help me out here: 
What's the type of these fields? Does it default to the type `text`? 
Should we mention it nevertheless to avoid ambiguity?

thomaseizinger: In SQlite, the types of a field is tracked together with the field, not the table: https://www.sqlite.org/datatype3.html

cnd/migrations/2019-11-11-034058_create-message-tables/up.sql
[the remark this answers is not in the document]
thomaseizinger: This is actually very much on purpose. Having a primary key separate from any domain concepts makes it easier to establish foreign key relations.

Since that one is naturally ascending, it also helps the database in creating indicies and layout on disk.

I am very reluctant to changing this as I would like to keep the primary key an implementation detail of the database and not mix it with our data.

cnd/src/db/custom_sql_types.rs
```diff
@@ -0,0 +1,78 @@
...
+{
+    fn from_sql(bytes: Option<&DB::RawValue>) -> deserialize::Result<Self> {
+        let s = String::from_sql(bytes)?;
+        let parsed = T::from_str(s.as_ref())?;
+
+        Ok(Text(parsed))
+    }
+}
+
+// Custom diesel new type for enforcing storage of a u32
```
luckysori: :upside_down_face: 
```suggestion
// Custom diesel newtype for enforcing storage of a u32
```

cnd/src/db/save_message.rs
thomaseizinger: Not having this macro makes you write all the types two times though :(
I would not consider this macro to be the long-term way of dealing with this duplication.

I see it as the starting point for iterating on a different way of dealing with this complexity in our codebase.

Thanks for pointing out the unwrap, that needs changing.

cnd/src/db/custom_sql_types.rs
```diff
@@ -0,0 +1,78 @@
+use diesel::{
+    backend::Backend,
+    deserialize::{self, FromSql},
+    serialize::{self, Output, ToSql},
+    sql_types,
+};
+use std::{convert::TryFrom, fmt, ops::Deref, str::FromStr};
+
+/// Custom diesel new-type that works as long as T implements `Display` and
```
luckysori: :upside_down_face: 
```suggestion
/// Custom diesel newtype that works as long as T implements `Display` and
```

cnd/src/http_api/routes/rfc003/handlers/action.rs
thomaseizinger: Yes, that is a possible future refactoring. I'd hesitate to add more complexity to this PR though :)

cnd/src/http_api/routes/rfc003/handlers/action.rs
```diff
@@ -97,29 +101,32 @@ pub fn handle_action<
...
+                        Action::Decline(_) => serde_json::from_value::<DeclineBody>(body)
                             .map_err(problem::deserialize)
                             .and_then({
                                 |body| {
                                     let channel = Network::pending_request_for(&dependencies, id)
                                         .ok_or_else(problem::missing_channel)?;
 
-                                    let decline_body =
-                                        action.decline(to_swap_decline_reason(body.reason));
+                                    let decline_message = rfc003::Decline {
```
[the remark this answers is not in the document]
thomaseizinger: Too trivial I'd say, we are only constructing a single struct and we don't need different behaviour for different ledger combinations like we do for accept.

cnd/src/db/custom_sql_types.rs
```diff
@@ -0,0 +1,78 @@
...
+        let parsed = T::from_str(s.as_ref())?;
+
+        Ok(Text(parsed))
+    }
+}
+
+// Custom diesel new type for enforcing storage of a u32
+#[derive(Debug, Clone, Copy, PartialEq, FromSqlRow, AsExpression)]
+#[sql_type = "sql_types::BigInt"]
+pub struct U32(pub u32);
```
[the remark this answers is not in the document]
thomaseizinger: > Does u32 make sense for absolutely all integer types we have?

No :)
We can always add a different newtype as soon as we run into the need for one.

cnd/src/db/mod.rs
```diff
@@ -159,188 +102,104 @@ pub fn default_db_path() -> Option<PathBuf> {
...
+        let number_of_variants = E::iter().count();
 
-        let res = db.get(swap.swap_id).expect("db get error");
-        let got = res.unwrap();
-
-        assert_eq!(got, swap);
+        assert_eq!(
+            number_of_variants,
+            expected_number_of_variants,
+            "the number of variants for this enum seem to have changed, please add a serialization format test for the new variant and update the expected variant count"
```
luckysori: :wrench: 
```suggestion
            "the number of variants for this enum seems to have changed, please add a serialization format test for the new variant and update the expected variant count"
```

cnd/src/db/new_types.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+impl fmt::Display for DecimalU256 {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct EthereumAddress(pub ethereum_support::Address);
```
luckysori: :question: I get what you're doing for `U256`, but why do you need this for addresses?

cnd/src/db/new_types.rs
```diff
@@ -0,0 +1,44 @@
+use ethereum_support::{FromDecimalStr, U256};
```
luckysori: :upside_down_face: You could rename the file to `newtype.rs` since I think it's [the correct term](https://wiki.haskell.org/Newtype) :stuck_out_tongue:

cnd/src/db/new_types.rs
```diff
@@ -0,0 +1,44 @@
...
+use std::{fmt, str::FromStr};
+
+#[derive(Debug, Clone, Copy, PartialEq, derive_more::FromStr, derive_more::Display)]
+pub struct Satoshis(pub u64);
```
luckysori: :wrench: I think the correct plural is
```suggestion
pub struct Satoshi(pub u64);
```

cnd/src/db/new_types.rs
luckysori: :question: Also, would it make sense for this one to be something like `ledger_newtypes.rs` and the other one to go from `custom_sql_types.rs` to `sql_newtypes.rs`? It may make the separation clearer.

cnd/src/http_api/problem.rs
```diff
@@ -15,6 +15,11 @@ pub struct MissingQueryParameter {
     pub description: &'static str,
 }
 
+pub fn database(e: crate::db::Error) -> HttpApiProblem {
+    log::error!("database failure {:?}", e);
```
luckysori: :wrench: 
```suggestion
    log::error!("Database failure: {:?}", e);
```

cnd/src/db/new_types.rs
```diff
@@ -0,0 +1,44 @@
...
+
+/// The `FromStr` implementation of U256 expects hex but we want to store
+/// decimal numbers in the database to aid human-readability.
+///
+/// This type wraps U256 to provide `FromStr` and `Display` implementations that
+/// use decimal numbers.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct DecimalU256(pub U256);
```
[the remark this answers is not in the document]
thomaseizinger: It is tested somewhere else!

I've made this more explicit by moving them to a separate module :)

cnd/src/db/save_message.rs
thomaseizinger: Meh, really? :(

This stupid decline reason is in the way EVERYWHERE and we are going to remove it very soon hopefully. Not sure if it adds more noise to our backlog and we just waste time looking at it or if it has value to have it documented.

cnd/src/http_api/routes/rfc003/handlers/action.rs
```diff
@@ -150,7 +164,7 @@ trait SelectAction<Accept, Decline, Deploy, Fund, Redeem, Refund>:
 }
 
 fn rfc003_accept_response<AL: rfc003::Ledger, BL: rfc003::Ledger>(
```
luckysori: Being in `rfc003/handlers/action.rs` couldn't this method just be like this?
```suggestion
fn accept_response<AL: Ledger, BL: Ledger>(
```
Same goes for any other fully qualified imports here that reference RFC003.

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
```diff
@@ -119,8 +131,10 @@ where
 
             match response {
                 Ok(accept) => {
-                    let state = State::accepted(swap_request.clone(), accept.clone(), seed);
+                    let state = State::accepted(swap_request.clone(), accept, seed);
                     StateStore::insert(&dependencies, id, state.clone());
+                    SaveMessage::save_message(&dependencies, accept)
+                        .expect("failed to save message to db");
```
luckysori: :question: Is it intentional to panic if this fails?

cnd/migrations/2019-11-11-034058_create-message-tables/up.sql
thomaseizinger: Sorry for such a short initial reply. The reasoning of not specifying the column types (which is something that one can do in standard SQL) was that SQLite actually ignores them because it automatically determines the type and stores them accordingly with the field.

I didn't know this neither before writing this code but it actually became handy because we don't have to specify all those types but only make sure the format we store is correct :)

cnd/src/db/custom_sql_types.rs
thomaseizinger: Please open a follow-up PR for these kind of changes! I don't want to wait for another build to land it :see_no_evil:

cnd/src/db/mod.rs
thomaseizinger: Me bad at grammar.

cnd/src/db/new_types.rs
thomaseizinger: Because the default `Display` for addresses only prints the short form with `...` in the middle :roll_eyes: - very annoying.

cnd/src/db/new_types.rs
thomaseizinger: I like the idea!

Would you mind sending a follow-up PR to get it in? :)

cnd/src/db/new_types.rs
thomaseizinger: Highly likely to be true! Can I blatantly also ask for a follow-up PR here (all can go in one ofc!) :grimacing:

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
thomaseizinger: It was a trade-off we had to make:

We wanted to know if it fails but since this is inside a future, we can't quite handle the error properly :(

thomaseizinger:
I am merging this in so that @tcharding and me can continue building on this!

Thank you for the useful comments @D4nte and @luckysori. I will happily accept follow-up PRs with improvements :)

Issue #1551, opened 2019-10-21
Fetched transaction receipts can end up unused


luckysori:
**Problem**:

We eagerly fetch transaction receipts even though for `TransactionPattern`s that don't care about logs we don't need them until we find the matching transaction and want to return the `TransactionAndReceipt`. This is wasteful.

https://github.com/comit-network/comit-rs/blob/49271bce29a2dad60fca77b830000d101a316101/btsieve/src/ethereum/mod.rs#L76-L104

Also function `matches(transaction: &Transaction) -> bool` defined on `&bitcoin::TransactionPattern` and  function `matches(transaction: &Transaction, receipt: &TransactionReceipt) -> bool` defined on `&ethereum::TransactionQuery` are still a bit awkward to read because "pattern matches transaction" is not really correct since it should be the other way around: "transaction matches pattern".

**Goals**:

- Clear code.
- Efficient code.

**Recommendation**:

- Create trait `FindTransactionMatching` that defines a function `find_transaction_matching(self, pattern: TransactionPattern) -> Option<Transaction>`.

Implement on `bitcoin::Block`, `ethereum::Block<Transaction>` and interestingly also on `ethereum::Block<TransactionAndReceipt>`, since an Ethereum block takes a generic type parameter which is normally either a transaction or a transaction ID, but which can hold our custom `struct` instead.

- Create trait `HasBloomFilterMatching` that defines a function `has_bloom_filter_matching(self, pattern: TransactionPattern) -> Outcome`, where

```rust
enum Outcome {
  Match,
  NoMatch,
  Inconclusive,
}
```

which can encode all the possible paths:

1. `Outcome::Match`: the block's bloom logs match the pattern; look at the transactions (and their receipts) in this block.
2. `Outcome::NoMatch`: the blocks' bloom logs do not match the pattern; skip this block.
3. `Outcome::Inconclusive: the pattern cannot be used to decide based on bloom logs; look at the transactions in this block.

Implement on `ethereum::Block<T>` and use to decide what to do with each block.

---

The names defined here are not definitive (nor should be the proposal, really), in particular `HasBloomFilterMatching` and `Outcome`.

thomaseizinger:
This should be taken care of by https://github.com/comit-network/comit-rs/issues/1781.

PR #1633, opened 2019-11-08
Match query against missed and old Ethereum blocks

### LINKED ISSUE #1516: Match query against missed ethereum blocks
**Problem**:
As outlined for Bitcoin in #1333, blocks can be missed when polling `latest_block` for Ethereum. This means that, currently, if a block containing a matching transaction is missed, the corresponding query will never be resolved.

**Goal**:
Make the btsieve library as resilient as possible.

**Recommendation**:
Do something similar to what was done in #1514 for Bitcoin.

### LINKED ISSUE #1616: Ethereum: Allow queries to be run from a timestamp

luckysori:
New design using channels; also includes matching a pattern against transactions given a timestamp.

Fixes #1516.
Fixes #1616.


btsieve/src/ethereum/mod.rs
```diff
@@ -58,8 +75,71 @@ where
...
+                            }
+                        };
+
+                        if pattern.matches(transaction, &receipt) {
+                            return Ok(TransactionAndReceipt {
+                                transaction: transaction.clone(),
+                                receipt,
+                            });
+                        };
+                    }
```
bonomat: 🔧 Duplicated code. Same as below starting from line 169-197.

btsieve/src/ethereum/mod.rs
```diff
@@ -58,8 +75,71 @@ where
...
+                                log::warn!(
+                                    "Could not retrieve transaction receipt for {}: {:?}",
+                                    transaction.hash,
+                                    e
+                                );
+                                continue;
+                            }
+                        };
+
+                        if pattern.matches(transaction, &receipt) {
```
bonomat: ❓📣 
Just wondering: I think to remember that I was the one implementing it :D but please correct me if I'm wrong: 

When we introduced the bloomfilter query, the idea was that we don't have to fetch every `transaction_receipt` for every `transaction` in every `block`. Right now this is not the case, do you know what happened?

luckysori: Explained offline. Issue #1551 tracks some missing optimisations.

PR #1650, opened 2019-11-14
Configure CORS

### LINKED ISSUE #1553: No CORS problems when using cnd during the workshop/hackathon

luckysori:
Fixes #1553.

### DIFF
--- a/cnd/src/config/file.rs
+++ b/cnd/src/config/file.rs
@@ -42,7 +42,7 @@ impl File {
                     address: IpAddr::V4(Ipv4Addr::UNSPECIFIED),
                     port: 8000,
                 },
-                cors: Some(Cors {
+                cors: Option::Some(Cors {
                     allowed_origins: AllowedOrigins::None(None::None),
                 }),
             },
@@ -176,7 +176,7 @@ impl File {
     fn ensure_directory_exists(config_file: &Path) -> Result<(), config_rs::ConfigError> {
         match config_file.parent() {
             Option::None => Ok(()),
-            Some(path) => {
+            Option::Some(path) => {
                 if !path.exists() {
                     println!(
                         "Config path does not exist, creating directories recursively: {:?}",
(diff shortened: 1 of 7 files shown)


D4nte:
If it's ugly (and it is) then implement a custom deserializer.

cnd/src/http_api/route_factory.rs
```diff
@@ -33,14 +34,29 @@ pub fn create<
         + SaveRfc003Messages,
 >(
     peer_id: PeerId,
+    allowed_foreign_origins: AllowedForeignOrigins,
     dependencies: D,
+    peer_id: PeerId,
```
tcharding: We don't need the second `peer_id` here :).

cnd/src/main.rs
```diff
@@ -138,9 +138,16 @@ fn spawn_warp_instance<
     runtime: &mut tokio::runtime::Runtime,
     dependencies: D,
 ) {
-    let routes = route_factory::create(peer_id, dependencies);
+    let routes = route_factory::create(
+        dependencies,
+        settings.http_api.cors.allowed_foreign_origins.clone(),
+        peer_id,
```
tcharding:
```
    let routes = route_factory::create(
        peer_id,
        dependencies,
        settings.http_api.cors.allowed_foreign_origins.clone(),
```

cnd/src/config/file.rs
```diff
@@ -252,14 +286,52 @@ mod tests {
...
+    fn cors_deserializes_correctly() {
+        let file_contents = vec![
+            r#"
+            allowed_origins = "all"
+            "#,
+            r#"
+             allowed_origins = "none"
+            "#,
+            r#"
+             allowed_origins = ["http://localhost:8000", "https://192.168.1.55:3000"]
```
D4nte: 👍

Issue #1194, opened 2019-08-14
How to define/deal with custom errors?


tcharding:
## Problem

Error handling in Rust is non-trivial and can lead to a lot of boiler plate code.

1. We currently define a new error type per module for functions within the module to share them.

2. If function needs to return an error by bubbling up from 2 different Error enums then such function needs to have its own Error enum to wrap those and need conversions From those.

## Goal

To reach project consensus on a nice way to do error handling and start to move towards doing so in a consistent manner in our codebase.

## Ideas

List of libraries (to be extended):

- https://crates.io/crates/custom_error
- https://crates.io/crates/error-chain
- https://crates.io/crates/failure
- https://crates.io/crates/snafu
- https://jeltef.github.io/derive_more/derive_more/from.html#enums
- https://github.com/dtolnay/thiserror

Conclusion: those interested can do it as a side task and use a new way in a PR (ie, PoC). If it's good, we could make it the CoBloX standard for handling errors.

thomaseizinger:
I am not really a fan of this kind of macro-based programming. Either hide all of the generated code from my completely (like custom derives) or don't generate code I am supposed to use :D

Autocompletion is usually also bad for these things (although CLion got drastically better at this already).

There have many several attempts at making this better in the Rust ecosystem:
- error_chain
- failure
- custom_error

Personally, I would like to see a PoC that clearly shows the added benefit of using such a library without compromising the flexibility that we have now.

Regarding the number of errors types: Did you grep across the whole repository? Then the number is quite misleading because it completely ignores the fact that we have certain module boundaries that exist on purpose and hence need to define their own error type.

thomaseizinger:
Nevertheless, we can/should do something about the situation.
I'd propose creating a spike that investigates the state of the art error handling.
We can collect some libraries here that may help:
- https://github.com/shepmaster/snafu

thomaseizinger:
There is a crate called `derive_more` that allows you to derive the `From` trait which could be useful to not having to write that impl manually for our custom error types:

https://jeltef.github.io/derive_more/derive_more/from.html#enums

thomaseizinger:
@bonomat why did this get closed? I don't see this issue being resolved and it actually came up recently here: https://github.com/comit-network/comit-rs/issues/1582

tcharding:
The `anyhow` crate is likely the answer to this issue.  We have started using it already (see `cnd/src/db/mod.rs`).

[...]

PR #1652, opened 2019-11-14
Use `thiserror` to derive std::error::Error

tcharding:
Move the whole code base to use `thiserror` any where it is appropriate.  While some of the call sites appear to add extra code we are actually getting a lot of extras for free.  All errors now implement `std::error::Error` and have useful debugging output.

This is related to #1194, once I've hacked about with `anyhow` I will do brief write up for the team on error handling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,12 +75,7 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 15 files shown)


thomaseizinger:
@tcharding rebased to fix the merge conflict :)

PR #1664, opened 2019-11-18
Make the remaining sections in the config file optional and don't generate one on startup

### LINKED ISSUE #1448: cnd's config file has sections that are required but don't need to

thomaseizinger:
Resolves #1448.

The first patch makes those sections optional and removes the code for generating the config file as this is no longer needed now.

The second patch moves the conditional logic for when we try to read which config file to `main.rs` and prints more specific messages.


thomaseizinger:
bors retry

thomaseizinger:
> I think we should tackle at some point the documentation of this config file. Any plans @thomaseizinger ?

No concrete plans yet. Our current config is solely based on assumptions and our own needs. Following the trend of backing up decisions with data/surveys, I wouldn't doing nothing until someone complains and telld us about their usecase :)

Issue #1244, opened 2019-08-20
Resume SWAP after restart of cnd


tcharding:
## Problem

Currently, we don't continue with swaps that are still in progress after a restart of a node.

## Goal

Allow users to restart the node while having an active swap.

## Recommendation

Building on top of the work of #1509, we can define a read-only interface to the database that returns us all swaps that have been fully negotiated. For those, we simply need to then re-spawn the state machine and it should continue looking for transactions. If some of the transactions already happened (either while the node was not running or before that), we should simply be able to find those again by scanning the blockchain from a specific point in the time (the timestamp when the negotiation completed.)

PR #1666, opened 2019-11-19
Store and load swaps from SQLite + respawn the state machine on startup

tcharding:
This PR is the final missing piece of the database implementation.

- Moves the metadata store to the database.
- Loads swaps on startup from the database.
- Respawns the state machines for accepted swaps.

# Highly recommend reviewing this one patch by patch!

Resolves: #1243 (the persistent storage epic - Epic!)

PR #1669, opened 2019-11-20
Load swaps on startup

tcharding:
Load swaps from the database on cnd startup.  Needs testing.

This should resolve: #1244 and also show swap history.


api_tests/lib_sdk/actors/actor.ts
```diff
@@ -218,8 +218,9 @@ export class Actor {
     }
 
     public async restart(configFile: CndConfigFile) {
+        const doNotClobberLog = true;
```
D4nte: I don't see the point of that, passing a bool is fine.

api_tests/lib_sdk/cnd_instance.ts
```diff
@@ -57,10 +61,8 @@ export class CndInstance {
                         this.logDir + "/cnd-" + this.actorConfig.name + ".log",
                         "w"
                     ), // stdout
-                    await openAsync(
-                        this.logDir + "/cnd-" + this.actorConfig.name + ".log",
-                        "w"
-                    ), // stderr
+                    await output,
```
D4nte: should be `output`

api_tests/lib_sdk/cnd_instance.ts
```diff
@@ -46,6 +49,7 @@ export class CndInstance {
         const pemObject = new PEMObject("SEED", this.actorConfig.seed);
         const seedFile = await tempWrite(pemObject.encoded, "seed.pem");
 
+        const output = this.output(doNotClobberLog);
```
D4nte: put the `await` here

thomaseizinger:
Included in #1666.

PR #1670, opened 2019-11-20
Dump config

### LINKED ISSUE #1657: Document configuration
## Problem

With our current auto-generated files that only generate mandatory values and the fact that the rest of the configuration values are defaulted, a user simply does not know what are the available options.
Hence, if such user were to try to use cnd on mainnet or testnet, they would need to look at Rust code to know what needs to be configured.

## Goal

Make it clear for the user what can be configured on cnd.

## Recommendation

A. A markdown file that explains the available options (like [bitcoind does](https://github.com/bitcoin/bitcoin/blob/master/share/examples/bitcoin.conf)).
B. The auto-generated files contains comments that explains all options

tcharding:
`cnd --dump-config` starts cnd, dumps the running config as toml, and the exits.

This work was inspired by:
- Recent discussion within the team about the lack of documentation for configuring cnd.
- While attempting to manually test the database work I was struggling to work out how to configure cnd.  
- While hacking on my trading bot I also was struggling to configure cnd.  
- `geth` subcommand `dumpconfig`

Example usage looks like:
```
$ cnd --dump-config
[network]
listen = ["/ip4/0.0.0.0/tcp/9939"]
[http_api.socket]
address = "0.0.0.0"
port = 8000

[http_api.cors]
allowed_origins = "none"

[database]
sqlite = "/home/tobin/.local/share/comit/cnd.sqlite"

[logging]
level = "DEBUG"
structured = false

[bitcoin]
network = "regtest"
node_url = "http://localhost:18443/"

[ethereum]
```

This was supposed to be a quick bit of fun in the evening, ended up taking me 4 hours thanks to all the `serda` stuff.  It is not perfect yet, see the missing blank line after `[Network]` section. ~, my guess is this is because of the square bracket in `listen` array being mistaken for a section heading.  Also I wrote a round trip test but it fails, and from visually parsing I cannot see why, it would be nice to have this fixed.~  Now includes passing serialization unit test (cannot work out why `Network` doesn't get an empty trailing newline. 

There is a bunch of code that I cargo cult programmed by reading `serde_url` and do not fully understand.  Its for the custom `serialize` funcitons in `config/serde_*.rs`.

Resolves #1657


cnd/src/config/settings.rs
```diff
@@ -14,7 +15,7 @@ use std::{
 /// meaning in cnd. Contrary to that, many configuration values are optional in
 /// the config file but may be replaced by default values when the `Settings`
 /// are created from a given `Config`.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
```
D4nte: Why do you need `Deserialize`?

cnd/src/config/settings.rs
```diff
@@ -267,4 +269,37 @@ mod tests {
                 listen: vec!["/ip4/0.0.0.0/tcp/9939".parse().unwrap()],
             })
     }
+
+    // Note toml library serializes Network without a trailing new
```
D4nte: Open an issue on toml's library?

cnd/src/cli.rs
```diff
@@ -10,4 +10,8 @@ pub struct Options {
     /// Path to secret seed file.
     #[structopt(short = "s", long = "seed-file", parse(from_os_str))]
     pub seed_file: Option<PathBuf>,
+
+    /// Dump the current configuration and exit
+    #[structopt(short = "d", long = "dump-config")]
```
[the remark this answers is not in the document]
tcharding: lol, neither would I.  I tried to pre-empt @thomaseizinger who previously asked for a short option for `--seed-file`.  Will remove.

cnd/src/config/file.rs
```diff
@@ -252,15 +284,13 @@ structured = false
 
 [bitcoin]
 network = "mainnet"
-node_url = "http://example.com"
+node_url = "http://example.com/"
```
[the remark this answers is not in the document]
tcharding: Trailing slash is not required when parsing the toml.  It is added when serializing though.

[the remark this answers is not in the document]
tcharding: Yes, none of the tests have a port number.  I will add.

cnd/src/config/serde_bitcoin_network.rs
```diff
@@ -28,3 +28,50 @@ where
 
     deserializer.deserialize_str(Visitor)
 }
+
+// reference: serde_url crate.
```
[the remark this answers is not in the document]
tcharding: I cut and pasted all the code below this comment, doesn't feel right to do that with out some comment saying I stole it.  Previous patch review has nack'ed comments like 'shamelessly stolen from ...' so this time I wrote 'ref: ...'.

cnd/src/config/settings.rs
tcharding: Will check, I think I added this to satisfy the compiler.

cnd/src/config/settings.rs
tcharding: It seems to be something to with the `Multiaddr` serialization.  When I tried to make a small repro toml crate handled array of ints and and also array of custom type correctly.

cnd/src/config/serde_bitcoin_network.rs
```diff
@@ -28,3 +28,50 @@ where
...
+
+/// Serialises `value` with a given serializer.
```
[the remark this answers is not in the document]
tcharding: Cheer, will remove.

tcharding:
Thanks for the review @D4nte

cnd/src/config/settings.rs
D4nte: strange as I would expect the new line **after** the end of the multiaddr array...

cnd/src/config/settings.rs
D4nte: In this case it'd be good to functionally understand whether it's actually needed.

cnd/src/config/serde_bitcoin_network.rs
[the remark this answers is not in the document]
D4nte: You are doing a standard custom implementation of serde serialisation. It's fine IMHO to copy paste, it's how they would expect you to do it.

cnd/src/config/file.rs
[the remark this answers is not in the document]
D4nte: actually the test that test a full config does have a port so keep it like that.

cnd/src/main.rs
```diff
@@ -37,6 +38,11 @@ fn main() -> anyhow::Result<()> {
 
     let settings = read_config(&options).and_then(Settings::from_config_file_and_defaults)?;
 
+    if options.dump_config {
```
thomaseizinger: What is a bit annoying about the rest of the code is that it also prints to `stdout`. This means redirecting the output with this option will _not_ yield a valid config file. Should we just change all the remaining print statements to `stderr`, including the logger?

thomaseizinger: This can actually be done as a follow-up PR. However, if you do it now, we can write a nice test that simply does:

```bash
cargo run -p cnd -- --dump-config > ./my_config.toml
cargo run -p cnd -- --config ./my_config.toml
```

That verifies that all of this works :)

tcharding: I don't think it is an issue to have this going to stdout, in my opinion the use case for this is exploratory not to output a valid config file that can be re-parsed.  Either way there is no output other than valid config file if the `--config` option is not present.

In regard to the second post, I like a unit test that has the hard coded config file as as string better than one that writes to the filesystem, although I would not NACK a test that did this.

tcharding:
> I like the idea!
> 
> Regarding the serialization, you could take a look at the PR where I removed it and simply bring that back: #1664
> I feel like some of the stuff is a bit overly complex :)

Awesome, I used the implementation of `serialize` as it was removed.  Thanks

tcharding:
> Are you sure the changelog does not need updating?

Done

D4nte:
@tcharding excellent demonstration of initiative. I marked it as solving #1657 for now. Well done & cheers.

tcharding:
> Nice, I like the idea!
> 
> @tcharding would be cool if you could drop a line in the README as well, so people know that the `--dump-config` flag exists.

I'll add a comment about using `cnd --help` instead if you don't mind - saves the README going stale.

cnd/src/config/settings.rs
```diff
@@ -267,4 +268,37 @@ mod tests {
...
+listen = ["/ip4/0.0.0.0/tcp/9939"]
+[http_api.socket]
+address = "0.0.0.0"
+port = 8000
+
+[http_api.cors]
+allowed_origins = "none"
+
+[database]
+sqlite = "/home/tobin/.local/share/comit/cnd.sqlite"
```
thomaseizinger: This is not going to work on any machine other than yours :sweat_smile:

tcharding:
> So actually, with this PR, we would have the `File` struct to deserialize the config and the `Settings` struct to serialize is.
> 
> That doesn't make much sense from my POV, we should only use the `File` struct to read from and write to the config file.
> 
> I am gonna push a commit that moves the serialization to the `File` struct :)

Cool, happy with what you did.  Some time in person I'd like to hear why you don't like the separation of `Settings` serialising and `File` de-serialising.  Its not important just interesting.  I'm going to merge this.

cnd/src/config/settings.rs
tcharding: Yeah I laughed at myself when I first saw the build failure.

PR #1674, opened 2019-11-22
Merge save traits

tcharding:
This is just a single commit (on top of #1666).  Needs manual merging with #1666 and #1669.  To be discussed with @thomaseizinger.


thomaseizinger:
Last patch needs to be re-applied onto master after #1666 is merged.

Issue #1678, opened 2019-11-25
Unify `Save` and `SaveMessage` traits


thomaseizinger:
## Problem

During design and implementation of the database layer we have ended up with two traits that are identical except for the type they act on i.e., we have duplication in the functionality provided by the traits.  This issue is further worsened by the fact that the `SaveMessage` trait is already generic, it makes no sense to have a generic trait and a specific trait that do the same thing.

## Goal

Unify the `SaveMessage` and `Save` traits.

## Solution

(DONE: #1688 - waiting for DB work to merge).

Merge `SaveMessage<M>` and `Save` into `Save<T>`.

PR #1688, opened 2019-11-26
Merge `Save` and `SaveMessage` traits

tcharding:
Now that we have all the database saving implemented it is clear that the `Save` trait (used for saving what was swap metadata) is a a specific case of the generic `SaveMessage<M>` if 'message' was the more general 'anything'.

Merge the two traits together and remove reference to 'messages'.
 
Fixes: #1678


thomaseizinger:
@tcharding I hope you don't mind by I rebased your PR onto latest master :)

Issue #1481, opened 2019-10-02
Store a timestamp to the start of a swap


luckysori:
**Goal**:
To be able to (eventually) persist a timestamp of the start of a swap which could then be used to create btsieve queries that [look back in time](https://github.com/comit-network/comit-rs/issues/1289).

**Recommendation**:
Take the timestamp of when the swap was accepted and store it in the state store.

---

Related to #1289.
Related to https://github.com/comit-network/comit-rs/pull/1453.

luckysori:
Not now because we are still coming up with the data model to be persisted.

tcharding:
We did this already but closed it because without using it in btsieve the code introduced is dead.

https://github.com/comit-network/comit-rs/pull/1453

PR #1689, opened 2019-11-26
1481 store timestamp

tcharding:
Add a timestamp column to the two accepted swap database tables.  Use SQLite auto-generation of timestamp to populate this column when an accepted swap is stored in the database.  Add a database function to retrieve the timestamp for an [accepted] swap id.  Add an integration test that saves and retrieves a swap in order to test the validity of timestamp.

Resolves: #1481


D4nte:
Please mark PRs that build on top of other non-merged PRs as draft.

cnd/src/db/integration_tests.rs
```diff
@@ -54,17 +54,17 @@ macro_rules! db_roundtrip_test {
                         db.save(saved_accept.clone()).await?;
 
                         let loaded_swap = Retrieve::get(&db, &swap_id).await?;
-                        let (loaded_request, loaded_accept) = db.load_accepted_swap(swap_id).await?;
+                        let (loaded_request, loaded_accept, _at) = db.load_accepted_swap(&swap_id).await?;
```
D4nte: 📣  Why not testing the timestamp too?

tcharding: I could not think of any meaningful way to do so, that's why I added `accepted_swap_has_valid_timestamp()` in `cnd/tests`.  Open to suggestions.

tcharding:
> Did I miss a PR? I don't see where the timestamp is actually taken.

What do you mean by taken?  Do you mean where the timestamp is generated?  If so its done by in this PR by the database insert.

D4nte:
> > Did I miss a PR? I don't see where the timestamp is actually taken.
> 
> What do you mean by taken? Do you mean where the timestamp is generated? If so its done by in this PR by the database insert. If you mean where it is used, its not used yet - we need to plug this PR into the ledger timestamp work @luckysori is doing/has done. To do that the timestamp most likely needs to be taken from the database _after_ accept message is inserted and put into the state_store (most likely in the swap communication).

I mean when is it generated/set. I'd expect a `now()` somewhere.
Can you comment in the code where you do that? Are you saying the DB framework does that for you?

tcharding:
> > > Did I miss a PR? I don't see where the timestamp is actually taken.
> > 
> > 
> > What do you mean by taken? Do you mean where the timestamp is generated? If so its done by in this PR by the database insert. If you mean where it is used, its not used yet - we need to plug this PR into the ledger timestamp work @luckysori is doing/has done. To do that the timestamp most likely needs to be taken from the database _after_ accept message is inserted and put into the state_store (most likely in the swap communication).
> 
> I mean when is it generated/set. I'd expect a `now()` somewhere.
> Can you comment in the code where you do that? Are you saying the DB framework does that for you?

Yes, the timestamp is generated using
```
CREATE TABLE rfc003_ethereum_bitcoin_accept_messages
(
    id INTEGER               NOT NULL PRIMARY KEY,
    swap_id UNIQUE           NOT NULL,
    bitcoin_refund_identity  NOT NULL,
    ethereum_redeem_identity NOT NULL,
    at DATETIME DEFAULT CURRENT_TIMESTAMP
);
```

D4nte:
I don't know why I am not too fond of it. Maybe because I would not have done it this way? Which is not justification to ask it to be done differently. Approved. Thanks.

tcharding:
> I don't know why I am not too fond of it. Maybe because I would not have done it this way? Which is not justification to ask it to be done differently. Approved. Thanks.

On a previous iteration I generated the timestamp using `now()` as you suggest but then when doing the insert we passed the timestamp into the database.  @thomaseizinger rightly pointed out that its strange to pass a timestamp into a database for what is functionally a timestamp on the column creation.  So we moved on to using the database to create the timestamp.  This does have the unfortunate side effect that now in order to insert into the state store we have to do the database insert _and_ a database read to get back the timestamp - conceptually I feel that the timestamp should be on when cnd first _sees_ the accept not when cnd writes it to storage.  Open to suggestions, or we leave it like this for now and re-visit the design when all the timestamp stuff is tied together.  I don't think its super important how its done, its trivial to change.

PR #1695, opened 2019-11-27
Update CHANGELOG for recent database work

tcharding:
No update to the changelog was done during any of the database PRs.  Now that the database implementation is merged we should mention it in the changelog.


tcharding:
> I believe these changes are unreleased so they should go under that heading

Good catch, thank you.

PR #1698, opened 2019-11-27
Replace cargo-make with regular Makefile

### LINKED ISSUE #1046: Investigate how to get rid of cargo-make

thomaseizinger:
Closes #1046.


Makefile
```diff
@@ -0,0 +1,93 @@
...
+TOOLCHAIN = $(shell cat rust-toolchain)
+CARGO = $(RUSTUP) run --install $(TOOLCHAIN) cargo --color always
+
+NIGHTLY_TOOLCHAIN = "nightly-2019-07-31"
+CARGO_NIGHTLY = $(RUSTUP) run --install $(NIGHTLY_TOOLCHAIN) cargo --color always
+
+GIT_HOOKS_PATH = ".githooks"
+GIT_HOOKS = $(wildcard $(GIT_HOOKS_PATH)/*)
+
+BUILD_ARGS = ""
```
D4nte: I am not a makefile expert, can you test that indeed `-e` would override this?

Makefile
```diff
@@ -0,0 +1,93 @@
...
+
+install_rust_nightly:
+	$(RUSTUP) install $(NIGHTLY_TOOLCHAIN)
+
+## Dev environment
+
+install_clippy: install_rust
+	$(RUSTUP) component list --installed --toolchain $(TOOLCHAIN) | grep -q clippy || $(RUSTUP) component add clippy --toolchain $(TOOLCHAIN)
+
+# need nightly toolchain to get access to `merge_imports`
```
D4nte: I think we use more than `merge_imports` on fmt nightly.. no? I'd just remove this comment.

Makefile
```diff
@@ -0,0 +1,93 @@
...
+install_rustfmt: install_rust_nightly
+	$(RUSTUP) component list --installed --toolchain $(NIGHTLY_TOOLCHAIN) | grep -q rustfmt || $(RUSTUP) component add rustfmt --toolchain $(NIGHTLY_TOOLCHAIN)
+
+install_tomlfmt: install_rust
+	$(CARGO) --list | grep -q tomlfmt || $(CARGO) install cargo-tomlfmt
+
+install_yarn:
```
D4nte:
```suggestion
yarn_install:
```
Name is confusing as all `install_foo` install `foo` on the machine.
Here, we are running `yarn install`.

Makefile
```diff
@@ -0,0 +1,93 @@
...
+## Development tasks
+
+all: format build clippy test doc e2e_scripts
+
+format: install_rustfmt install_tomlfmt install_yarn
+	$(CARGO_NIGHTLY) fmt
+	$(CARGO) tomlfmt -p Cargo.toml
+	(cd ./api_tests; yarn run fix)
+
+ci: check_format clippy build test e2e
```
D4nte: no `doc`?

Makefile
thomaseizinger: I tested it :)
`make` actually has a very useful `-n` option to dry-run a task!

Makefile
thomaseizinger: True!

D4nte:
Also, maybe add BUILD_ARGS to test, clippy, doc etc tasks like I propose in my create-comit-app Release 0.5.0 PR

thomaseizinger:
> Also, maybe add BUILD_ARGS to test, clippy, doc etc tasks like I propose in my create-comit-app Release 0.5.0 PR

Well those are not `build` commands. I can add `TEST_ARGS` etc or generalize it to `CARGO_ARGS` but I feel like that doesn't make any sense because you don't want to run the unit tests against `--release` I believe (is that even possible?)

D4nte:
> > Also, maybe add BUILD_ARGS to test, clippy, doc etc tasks like I propose in my create-comit-app Release 0.5.0 PR
> 
> Well those are not `build` commands. I can add `TEST_ARGS` etc or generalize it to `CARGO_ARGS` but I feel like that doesn't make any sense because you don't want to run the unit tests against `--release` I believe (is that even possible?)

Leave it like this. I don't care about `--release` but more for other argument (features?) that one might want to pass down.

thomaseizinger:
> > > Also, maybe add BUILD_ARGS to test, clippy, doc etc tasks like I propose in my create-comit-app Release 0.5.0 PR
> > 
> > 
> > Well those are not `build` commands. I can add `TEST_ARGS` etc or generalize it to `CARGO_ARGS` but I feel like that doesn't make any sense because you don't want to run the unit tests against `--release` I believe (is that even possible?)
> 
> Leave it like this. I don't care about `--release` but more for other argument (features?) that one might want to pass down.

Yeah good point. I was thinking about that but ended up not doing it because there is currently no need for it. I'd suggest we do very specific arguments for now (`BUILD_ARGS`, `INSTALL_ARGS`) depending on what we need. Once that gets annoying, we can always introduce something better :)

PR #1699, opened 2019-11-27
Build docker image from existing release binary upon release

### LINKED ISSUE #1425: Reuse release binaries in docker container
**Problem**:
Right now we build cnd (and btsieve) inside the docker container. However, it would be more efficient if we just download the release from GitHub and use it inside the container. 
This can be achieved with Github actions, i.e. it can be triggered based on `release` events (see [1]). 

**Goal**:
* No more build of cnd in the docker container
* Docker images use the release from Github
* The docker container build is triggered through Github actions
  * It seems like the container can be build *in* Github or *in* Dockerhub and it is up to the developer to chose how to solve this. 

* [1] information about the release event: https://developer.github.com/v3/activity/events/types/#releaseevent
* [2] related to this PR: https://github.com/comit-network/comit-rs/pull/1416

thomaseizinger:
Resolves #1425.

I am not using GitHub actions as mentioned in the ticket but instead do it as part of the release step in the CircleCI build. This way we don't need to figure out how to wait for the proper event when the binary gets attached.

Need to remove the dockerhub build once this is merged.


.circleci/config.yml
```diff
@@ -108,6 +109,13 @@ jobs:
             ls ~/package/
             ghr -t ${GITHUB_TOKEN_FOR_RELEASES} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -c ${CIRCLE_SHA1} -delete ${VERSION} ~/package/
 
+            echo $DOCKER_HUB_TOKEN | docker login -u thomaseizinger --password-stdin
+
+            docker build -f .docker/Dockerfile . -t comitnetwork/cnd:${VERSION}
```
bonomat: This should work according the documentation: 
https://docs.docker.com/engine/reference/commandline/build/#tag-an-image--t

```suggestion
            docker build -f .docker/Dockerfile . -t comitnetwork/cnd:${VERSION}  -t comitnetwork/cnd:latest
```

.circleci/config.yml
```diff
@@ -108,6 +109,13 @@ jobs:
...
+            docker build -f .docker/Dockerfile . -t comitnetwork/cnd:latest
```
bonomat:
```suggestion
```

.circleci/config.yml
```diff
@@ -108,6 +109,13 @@ jobs:
             ls ~/package/
             ghr -t ${GITHUB_TOKEN_FOR_RELEASES} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -c ${CIRCLE_SHA1} -delete ${VERSION} ~/package/
 
+            echo $DOCKER_HUB_TOKEN | docker login -u thomaseizinger --password-stdin
```
D4nte: Seriously? You can't have a generic token for a repo not associated to a specific user?? :(

.dockerignore
```diff
@@ -1,4 +1,5 @@
 target/
```
D4nte: Trash this fine now that we only copy one binary

.circleci/config.yml
thomaseizinger: Don't think so. Up until September you couldn't even do tokens :sweat_smile:

.dockerignore
thomaseizinger: Without this, copying over the context takes long because it has to copy the whole target folder which is usually 3GB or something.

.circleci/config.yml
thomaseizinger: Works!

.dockerignore
D4nte: why does it copy over the context??

thomaseizinger:
@D4nte @bonomat I actually found an even better way of handling things:

I moved the Dockerfile to the root because we only have one anyway. In addition, we can ignore everything else apart from the release binary to only rebuild if that one changes :)

thomaseizinger:
@D4nte. FYI Just removes the DockerHub build configuration

.dockerignore
D4nte: @thomaseizinger please clarify this point.

Issue #1594, opened 2019-10-30
Rewrite e2e tests for `inverted_refund` and `refund`


luckysori:
## Problem

Only one e2e test (btc-eth happy path) was rewritten in PR #1584 following the new design outlined in [spike 11](https://github.com/comit-network/spikes/blob/master/0011-revise-e2e-test-framework.adoc#the-ideal-test).

## Goal

Use the new design for all our e2e tests.

## Proposal

Do for albtc-eth tests: `inverted_refund.ts`, `refund.ts`.

bonomat:
The PR in #1709 converts `refund` and `inverted_refund` to the new format. However, for over/under-funding our SDK is too correct, i.e. it does not let you misuse the `fund` function (which is good).

In order to test this nevertheless and not have to write the action interpretation again just for the e2e tests, I would propose a change in the sdk and make the functions public: 

https://github.com/comit-network/comit-js-sdk/blob/563cbff7df763d71f5496d00548fc7be6b9ab0ff/src/swap.ts#L49-L57
and 
https://github.com/comit-network/comit-js-sdk/blob/563cbff7df763d71f5496d00548fc7be6b9ab0ff/src/swap.ts#L59

@comit-network/rust-devs any thoughts on that?

thomaseizinger:
Hmm, interesting insight. I'd hesitate a bit in extending the public API surface too much. Can we maybe achieve the same functionality by adding more parameters to the `fund` etc actions? Just in case users want to override something?

> it does not let you misuse the fund function (which is good).

Maybe these tests should just use a lower level of the SDK and do things "manually"?

D4nte:
Can you bypass the fact that they are not exported by any chance? (I am guessing it's a JS thing and not a TS thing you could bypass but better try).

thomaseizinger:
> Can you bypass the fact that they are not exported by any chance? (I am guessing it's a JS thing and not a TS thing you could bypass but better try).

I don't think you can because I believe that is actually enforced on the module level, even in JavaScript.

bonomat:
> Can we maybe achieve the same functionality by adding more parameters to the fund etc actions? Just in case users want to override something?

I would rather not allow passing something into the `fund` method. Because at the end we want the sdk to be foolproof.

Providing some low level APIs which enable the user to have more fine grained control over the actions sounds reasonable. 

But let's not go this path for now and rather rethink how/what we want to achieve with the e2e tests. 

[...]

PR #1709, opened 2019-11-29
New e2e tests

bonomat:
This PR converts the e2e tests `refund` and `inverted_refund` to the new format as described in #1594


api_tests/e2e/rfc003/btc_eth/happy.ts
```diff
@@ -11,11 +11,23 @@ setTimeout(function() {
 
             await alice.sendRequest(AssetKind.Bitcoin, AssetKind.Ether);
             await bob.accept();
+
             await alice.fund();
+            await alice.assertAlphaFunded();
```
thomaseizinger: Do we really want to have these on the outside here? Makes the tests fairly verbose (again). WDYT?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Updated e2e tests (`refund` and `inverted_refund`) to new e2e test format.
```
thomaseizinger: Do our users care that much that we need a changelog entry? :grimacing:

api_tests/e2e/rfc003/btc_eth/happy.ts
bonomat: Good question, I'm not strongly opinionated about where this should be. 
Tobin and me had a discussion on the flight around these asserts in general and agreed that we should add them. 

Would you rather see these asserts inside `fund()` and so on?

D4nte: This is back to the initial debate. One of the major criticism of the last e2e refactoring was the fact that asserts were done in an hidden manner.

It looks like some team members are keen to only have one assert at the end.

I am more keen to have regular asserts to help with debugging.

bonomat: > I am more keen to have regular asserts to help with debugging.

This was actually a driving factor and we tried to have a failed test with the inline asserts and without. Having these asserts inline improved the readability big times. 

Actually, since we want to assert that both cnd's did indeed transition to `alphaFunded` we can't just move it inside `fund()` as inside that function we only have access to either alice's or bob's cnd (depending on who was funding).

I opt for keeping it as proposed in this PR.

CHANGELOG.md
bonomat: hehe... mergify complained so I was a good boy and updated it :D

thomaseizinger: The "complaint" is more a reminder that you should think about whether it needs updating and not that you should update it, regardless of what the change is.

I don't think we need to bother users with internal changes like this.

api_tests/e2e/rfc003/btc_eth/happy.ts
thomaseizinger: I am totally pro useful output but I'd also like to have these tests less verbose.

> One of the major criticism of the last e2e refactoring was the fact that asserts were done in an hidden manner.

Yes, in a way. The major criticism was that logic for talking to cnd was intertwined with assertions. This prevented that code from being used in other occasions. All that logic is now in the SDK so this is no longer a concern IMO.

What we are talking about here is code that is specifically written for testing (like the `Actor::fund` method). I cannot possibly see any other way the `Actor::fund` method could be used that would make the assertion wrong if it were in the fund method.

> Actually, since we want to assert that both cnd's did indeed transition to alphaFunded we can't just move it inside fund() as inside that function we only have access to either alice's or bob's cnd (depending on who was funding).

Every actor has access to all other actors through the `actors` field.

Ideally, we would have better output even without these explicit assertions. Ideally, a test should follow the arrange/act/assert pattern. If the "act" as in our case is doing a lot, then we just need to make the errors better (which is also something the users of our SDK would benefit).

The proposed way seems to be suboptimal to me for a number of reasons:

- It is verbose
- It solves a problem (bad output in failure cases) in a way that only helps in our scenario. Other SDK users suffer from the same problems and don't benefit.

It seems like a proper fix would require more work that we are currently willing to invest into this PR. I'd propose to make a decision between:

- Remove the assertions completely, leaving us with the pain of bad error messages to incentivize us to fix the problem properly in the SDK.
- Move the assertions into the respective methods as a short term gain for better output in the e2e test.

For me, either way is fine :)

thomaseizinger:
We should probably also fix https://github.com/comit-network/comit-rs/issues/1719 first before we make more tests depend on the new framework.

api_tests/e2e/rfc003/btc_eth/happy.ts
bonomat: > Ideally, we would have better output even without these explicit assertions. Ideally, a test should follow the arrange/act/assert pattern. If the "act" as in our case is doing a lot, then we just need to make the errors better (which is also something the users of our SDK would benefit).

I think in this case the arrange/act/assert pattern is hard to apply because we are testing multiple things at once. 

> The proposed way seems to be suboptimal to me for a number of reasons:

> * It is verbose
> * It solves a problem (bad output in failure cases) in a way that only helps in our scenario. Other SDK users suffer from the same problems and don't benefit.

I don't think this is verbose but you are right about the second part: It solves the problem of bad output. I think this is fine though as an application using the sdk should do the same thing, i.e. check with the blockchain if a transaction was included, if you received the money, etc. 


> Move the assertions into the respective methods as a short term gain for better output in the e2e test.

This does not work unfortunately. 
Assuming we want to consolidate:
```javascript
            await alice.fund();
            await alice.assertAlphaFunded();
            await bob.assertAlphaFunded();
```

currently, inside the `fund` method we do not know whether we are about to fund alpha or beta or if the actor is alice or bob. One could build this into these functions but that would clutter the `fund` function.

Hence, I still opt for sticking to the proposed approach as it gives a way to  see where an end2end test failed.

bonomat:
closing this in favor of: #1771 to keep the history clean.

api_tests/e2e/rfc003/btc_eth/happy.ts
bonomat: Maybe I was wrong, have a look at the new branch please: #1771

PR #1710, opened 2019-11-29
Run generate task only before testing instead of on every install

thomaseizinger:
We run `yarn install` quite a lot with the tasks in the Makefile.
Not running this tasks shaves of some time of that.


thomaseizinger:
Doesn't work because we need the generated files for running tsc.

Issue #1583, opened 2019-10-28
Having two command line options that mention files in the same directory is cumbersome


tcharding:
## Problem

Running `cnd` with a custom config file _and_ a custom seed file currently requires two config options.  If the assumption that these files will typically be located on the host filesystem in the same directory then this is overly cumbersome.

## Goal

Make `cnd` a sexy piece of software with well thought out command line options.

## Suggestion

Use a single command line option `config-dir` and use default file names within this directory.  Bonus points: this gives us profiles for free (https://github.com/comit-network/comit-rs/issues/1501)

PR #1711, opened 2019-11-29
Add data directory configuration option

tcharding:
Currently we have `--seed-file` for the seed file and a config option for the database.  Both of these files default to the systems data directory with `comit/` appended.  It makes more sense to configure these two items in the same fashion.

Add configuration option
```
[Data]
dir = path/to/data/dir
```

Use this path with `cnd.sqlite` and `seed.pem` appended for the database and seed file respectively.

Remove `--seed-file` command line option and the database configuration option.

Resolves: #1583


cnd/src/config/settings.rs
```diff
@@ -143,14 +140,14 @@ impl Settings {
...
                 .unwrap_or_default(),
-            database: {
-                let default_database_path = crate::data_dir()
-                    .map(|dir| Path::join(&dir, "cnd.sqlite"))
-                    .context("unable to determine default database path")?;
-                database.unwrap_or_else(|| Database {
-                    sqlite: default_database_path,
+            data: {
+                let default_data_dir =
+                    crate::data_dir().context("unable to determine default database path")?;
```
thomaseizinger:
```suggestion
                    crate::data_dir().context("unable to determine default data path")?;
```

cnd/src/seed.rs
```diff
@@ -49,17 +49,41 @@ impl Seed {
...
+
+        Ok(random_seed)
+    }
+
+    fn from_file<D: AsRef<OsStr>>(seed_file: D) -> Result<Seed, Error> {
+        let file = Path::new(&seed_file);
         let contents = fs::read_to_string(file)?;
         let pem = pem::parse(contents)?;
 
+        log::info!("Read in seed from file: {}", &file.to_path_buf().display());
```
thomaseizinger: `Path` has `.display` as-well :)

https://doc.rust-lang.org/std/path/struct.Path.html#method.display

cnd/src/config/settings.rs
tcharding: Cheers man, fixed.

cnd/src/seed.rs
tcharding: thanks

api_tests/lib_sdk/cnd_instance.ts
```diff
@@ -39,27 +38,20 @@ export class CndInstance {
             "config.toml"
         );
 
-        const pemObject = new PEMObject("SEED", this.actorConfig.seed);
```
thomaseizinger: Now that this is removed, we can no longer control the seed of the created nodes. It seems be fine because the tests pass but it means we should probably remove the seed from the E2EActorConfig class :)

D4nte:
Sure the readme did need update @tcharding ?

tcharding:
> Sure the readme did need update @tcharding ?

After a quick read of the readme I don't see anything that needs adding.  Did you have something specific in mind please @D4nte or @thomaseizinger?

thomaseizinger:
> > Sure the readme did need update @tcharding ?
> 
> After a quick read of the readme I don't see anything that needs adding. Did you have something specific in mind please @D4nte or @thomaseizinger?

I think @D4nte was referring to the changelog. That one should probably have an entry about the new behaviour.

D4nte:
> > > Sure the readme did need update @tcharding ?
> > 
> > 
> > After a quick read of the readme I don't see anything that needs adding. Did you have something specific in mind please @D4nte or @thomaseizinger?
> 
> I think @D4nte was referring to the changelog. That one should probably have an entry about the new behaviour.

I was talking about the readme, wasn't sure whether we were documenting the `--seed` there :)

Issue #1547, opened 2019-10-17
libp2p Transport creation is not production ready


tcharding:
We build the libp2p `Transport` using `libp2p::build_development_transport`.  The [docs](https://docs.rs/libp2p/0.12.0/libp2p/fn.build_development_transport.html) explicitly state that this function should not be used in production.

> Note: This Transport is not suitable for production usage, as its implementation reserves the right to support additional protocols or remove deprecated protocols.

## Goal

Do not rely on code that is explicitly stated to be non-production code.

## Suggestion

Cut and paste the code out of the libp2p library (excluding all the feature flagged stuff we don't nee) and create our own `Trasport` in cnd.

thomaseizinger:
@tcharding please refine this further before we go through it in grooming, i.e. work out concrete action items (kind of what the issue says yes but otherwise it is just one more ticket that is sitting around).

tcharding:
No worries, done.

PR #1713, opened 2019-11-29
Implement a libp2p transport

tcharding:
Cut n paste the code out of libp2p stripping unnecessary complexity.  In other words implement a transport that is functionally identical to the one we are currently using (`libp2p::build_development_transport`).

Initial patch updates libp2p from 0.12 -> 0.13 and libp2p-swarm from 0.2 -> 0.3

Resolves: #1547


tcharding:
> There some internal optimizations I could think of but most importantly, I'd say the transport implementation should be inside `cnd` and not `libp2p_comit` because the latter is just the messaging protocol implementation which is independent of the transport that is in use.

Cool, will move.  Thanks

thomaseizinger:
bors retry

PR #1722, opened 2019-12-03
Make sure we build cnd before executing the e2e tests

thomaseizinger:


thomaseizinger:
I believe from the outside (using the Makefile), this is functionally equivalent.

If this is still breaking someone's workflow, please send a follow-up PR - I can't think of anything right now!

bors r+

api_tests/package.json
```diff
@@ -6,6 +6,7 @@
     "scripts": {
         "check": "tsc && prettier --check '**/*.{ts,json,yml}' && tslint --project .",
         "postinstall": "mkdir -p ./gen && json2ts -i ./siren.schema.json -o ./gen/siren.d.ts",
+        "pretest": "cargo build --bin cnd",
```
D4nte: I would recommend `make build` so that if we change the way we build by default, it is automatically reflected here.

PR #1723, opened 2019-12-03
Refactor deserialization and serialization of ledger and asset values

### LINKED ISSUE #1364: Refactor how we deserialize the body of HTTP requests

thomaseizinger:
We make heavy use of serde's "from" and "into" features to explicitly model the format we expect on the HTTP API and the format we send out.

This allows us to shift most of the error handling to serde, which in turn gives us better error messages. They are not perfect but about as good as it gets if we stick to deserializing JSON with serde.

We could potentially have better error messages by:

- either taking apart the JSON manually but that would be a huge
coding effort and cumbersome to maintain because we could not make
use of many features of serde.
- simplifying the JSON model to reduce the complexity in the
deserialization process.

Fixes #1364.

As a side-benefit, we also remove a circular dependency between the `Ledger` & `Asset` traits and the `http_api` module. This brings us one step closer to the extracting a library.


api_tests/dry/sanity.ts
```diff
@@ -63,7 +63,7 @@ setTimeout(async function() {
                 "content-type",
                 "application/problem+json"
             );
-            expect(res.body.title).to.equal("Swap not supported.");
+            expect(res.body.title).to.equal("Invalid body.");
```
luckysori: :upside_down_face: Test names are slightly misaligned with what the tests do now.

cnd/src/http_api/ethereum_network.rs
```diff
@@ -22,43 +22,46 @@ pub enum Network {
...
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error("chain with id {0} is unknown")]
+pub struct UnknownChainId(String);
+
 impl Network {
-    pub fn from_network_id(s: String) -> Self {
-        match s.as_str() {
+    pub fn from_network_id(s: &str) -> Result<Self, UnknownChainId> {
```
luckysori: :question: Should it be called `from_chain_id`?

cnd/src/http_api/mod.rs
```diff
@@ -21,40 +17,34 @@ pub use self::{
...
+        ledger::{self, ethereum::ChainId},
         SwapId, SwapProtocol,
     },
 };
 use bitcoin::util::amount::Denomination;
 use libp2p::PeerId;
+use libp2p_core::Multiaddr;
 use serde::{
-    de::{self, MapAccess},
+    de::{self, Error as _, MapAccess},
```
luckysori: :question: What is this hackery?

cnd/src/swap_protocols/ledger/mod.rs
```diff
@@ -3,27 +3,13 @@ pub mod ethereum;
...
-    + Send
-    + Sync
-    + 'static
-    + Default
-    + PartialEq
-    + Eq
-    + Hash
-    + FromHttpLedger
-    + Into<LedgerKind>
+    Clone + Copy + Debug + Send + Sync + 'static + Default + PartialEq + Eq + Hash + Into<LedgerKind>
```
luckysori: Best change :sparkles:

PR #1727, opened 2019-12-04
Massively improve Makefile performance

thomaseizinger:
Instead of running all the commands all the time, we conditionally define rules based on the dev's machine and the files that need to be committed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,9 @@ INSTALLED_COMPONENTS = $(shell $(RUSTUP) component list --installed --toolchain 
 INSTALLED_NIGHTLY_COMPONENTS = $(shell $(RUSTUP) component list --installed --toolchain $(NIGHTLY_TOOLCHAIN))
 AVAILABLE_CARGO_COMMANDS = $(shell $(CARGO) --list)
 
+# All our targets go into .PHONY because none of them actually create files
+.PHONY: init_git_hooks default install_rust install_rust_nightly install_clippy install_rustfmt install_tomlfmt install clean all ci build clippy test doc e2e check_format format check_rust_format check_toml_format check_ts_format
+
 default: init_git_hooks build format
 
 init_git_hooks:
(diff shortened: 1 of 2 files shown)


thomaseizinger:
Needs more work.

Makefile
```diff
@@ -77,16 +85,42 @@ test:
 doc:
 	$(CARGO) doc
 
+e2e: build
+	(cd ./api_tests; yarn install; yarn test)
+
 check_format: check_rust_format check_toml_format check_ts_format
 
+STAGED_FILES = $(shell git diff --staged --name-only)
```
D4nte: 📣 you have changed the behaviour and considering how I use git/CLion I disagree with the usage of `--staged` and would prefer to keep the original behaviour `git status --untracked-files=no --short` simply because I/CLion don't stage changes before a commit meaning I am going to have a lot of CI fails due to formatting.

Makefile
```diff
@@ -77,16 +85,42 @@ test:
...
+STAGED_RUST_FILES = $(filter %.rs,$(STAGED_FILES))
```
D4nte: Is that only executed if the invoked task is below?

Makefile
thomaseizinger: I don't quite understand what you mean. You cannot commit files without staging them and AFAIK CLion actually does the opposite of what you are saying - it usually stages files automatically.

I don't see how the new behaviour is bad. On the contrary, it is actually more correct because it only runs the `check_XY_format` tasks for the files that are about to be committed. If you have changes in TS files but you are not committing them, the Makefile does no longer care.

Makefile
thomaseizinger: No, I think that is always executed. Why?

Makefile
thomaseizinger: ping @D4nte

D4nte: ok. let me test.

Makefile
D4nte: I was hoping for it to be optimized.

Makefile
D4nte: ok, looks like my memory is playing with me.

Makefile
thomaseizinger: It is pretty freakin' fast :)

PR #1751, opened 2019-12-10
Merge the `SendRequest` trait into the `Network` trait

### LINKED ISSUE #1681: `Network` and `SendRequest` traits are un-clean
Currently the `Network` trait has three un-related functions
```    
fn comit_peers(&self) -> Box<dyn Iterator<Item = (PeerId, Vec<Multiaddr>)> + Send + 'static>;
fn listen_addresses(&self) -> Vec<Multiaddr>;
fn pending_request_for(&self, swap: SwapId) -> Option<oneshot::Sender<Response>>;
```

We also have a trait with a single function `SendRequest`.  These traits make up the core functionality of the libp2p network layer.  

## Problem

There is no obvious reason _why_ there is a trait with three methods and a stand alone trait/method.  This code is non-uniform and not clear/clean.  We should strive for having code with clear explicit design choices in order to reduce the cost of maintenance.

## Solution

Either separate all the methods into separate traits i.e., as a Rust purest would do _or_ combine them all together as a single trait for the `ComitNode`.

thomaseizinger:
Fixes #1681.


cnd/src/network/mod.rs
```diff
@@ -340,10 +354,22 @@ where
     Ok(())
 }
 
+/// Defines all the operations the Comit Node can perform in regards to the
+/// Comit network.
+///
+/// Ideally, this trait would not be necessary and we would instead have one
+/// trait per function. Unfortunately, an instance of `Swarm` is very hard to
+/// "name" (see the complex traits bound below). To avoid this kind of code 4
```
tcharding:
```suggestion
/// name (see the complex traits bound below). To avoid this kind of code 4
```

PR #1754, opened 2019-12-10
Remove unused dependencies

thomaseizinger:
Thanks cargo-udeps!

### DIFF
--- a/cnd/Cargo.toml
+++ b/cnd/Cargo.toml
@@ -10,20 +10,17 @@ anyhow = "1"
 async-std = { version = "1", features = ["unstable"] }
 async-trait = "0.1"
 bigdecimal = "0.1.0"
-binary_macros = "0.6"
 bitcoin = { version = "0.19.1", features = ["use-serde"] }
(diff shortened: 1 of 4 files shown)


cnd/Cargo.toml
```diff
@@ -34,11 +31,10 @@ http-api-problem = "0.13"
 hyper = "0.12"
 lazy_static = "1"
 libp2p = { version = "0.13" }
-libp2p-core = { version = "0.13" }
```
D4nte: How come it's here? don't we have tomlfmt running in the CI?

thomaseizinger: Should actually, not know why it slipped through :man_shrugging:

tcharding:
`cargo-udeps` cool, good thing I didn't try to hard to write one.

cnd/Cargo.toml
D4nte: I think you fixed it, for tomlfmt you have to explicitly pass each cargo toml file.

Issue #1075, opened 2019-07-10
cnd refund availability


da-kami:
## Problem

cnd makes refund available as action even before the expiry timestamp is reached. 

## Goal

Applications on top of COMIT don't have to deal with the time calculation when refund is available, but cnd does that for the application.

## Proposal 

The cnd REST API _hides_ the `refund` action if it is not available yet due to time constraints. 
The state machine does still emit that redeem and refund is available, but the HTTP interface does not emit the refund action until the expiry is reached. It should still be possible to retrieve the refund action through the API before the expiry time is reached, but it would not be returned by default.

The solution is to retrieve the current network time through btsieve which gets it from block stored in the internal DAG.

Note: due to the missing confirmation feature on btsieve, this solution is not perfect. If btsieve is aware of a block before the node to which comit-i/bobtimus is connected to then comit-i/bobtimus refund transaction may be rejected.

thomaseizinger:
Needs refinement after btsieve refactoring.

tcharding:
Seems one cannot connect to a draft PR.  [This PR](https://github.com/comit-network/comit-rs/pull/1755) potentially resolves this (needs testing).

thomaseizinger:
Removing `no-activity` because I think this is still something quite important to do but just not yet.

PR #1755, opened 2019-12-11
ON ICE: Only show refund after expiry time has passed

tcharding:
Do not show the refund action if expiry time has not passed.  This PR _totally_ removes the refund action, there is no way to get it if the expiry time has not passed.

Draft, and on ice, because still needs testing (breaks current e2e tests also).  There is also some doubt whether we should even do this ticket - leaving to discuss with the team after Christmas. 

Resolves: #1075

### DIFF
--- a/cnd/src/swap_protocols/rfc003/mod.rs
+++ b/cnd/src/swap_protocols/rfc003/mod.rs
@@ -28,6 +28,9 @@ pub use self::{
 };
 
 pub use self::messages::{Accept, Decline, Request};
+
+use crate::{swap_protocols::asset::Asset, timestamp::Timestamp};
+
 /// Swap request response as received from peer node acting as Bob.
 pub type Response<AL, BL> = Result<Accept<AL, BL>, Decline>;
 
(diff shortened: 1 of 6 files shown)


cnd/src/swap_protocols/rfc003/mod.rs
```diff
@@ -42,3 +45,15 @@ pub enum Error {
     #[error("internal error: {0}")]
     Internal(String),
 }
+
+pub fn alpha_expiry_has_passed<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset>(
+    request: &Request<AL, BL, AA, BA>,
+) -> bool {
+    request.alpha_expiry < Timestamp::now()
```
thomaseizinger: :mega: This compares the expiry time to the current system time and not to the actual blockchain time of this ledger (which is used to determine whether or not the generated TX can go in). Not sure if we want to implement this in this fashion.

tcharding: Ok, @bonomat has schooled me on blockchain times and I agree that we need to hit the chain to get the time - I am also of the opinion, along with @D4nte, that this is an optimisation so we can do it later _after_ this PR actually works.  ATM this doesn't work because it breaks current e2e tests and has no added e2e test.  Also, I'm not convinced that we should even be doing this issue.  I'm putting this PR on ice until it can be discussed in person in the office (probably after Christmas)

thomaseizinger: Ok, I will close it for now to remove clutter from the PR page :)

Issue #1330, opened 2019-09-03
make code more stable by removing `unreachable!`s in `metadata_store.rs` 


tcharding:
## Problem

`metadata_store` has two `unreachable!` statements in the conversion code for ledger to metadata `LedgerKind` and `AssetKind`.

## Goal

Make the code more stable. 

## Proposal

In order to remove these we would have to remove the `Unknown` variant by handling during deserialization.  See `cnd/src/bam_api/mod.rs:from_bam_header()`.

PR #1756, opened 2019-12-11
Remove Unknown variants

tcharding:
Currently we have `Unknown` variants in `AssetKind`, `LedgerKind`, and
`SwapProtocol`.  These are unnecessary if we handle the errors during
de/serialization.

Resolves: #1330

### DIFF
--- a/cnd/src/comit_api/mod.rs
+++ b/cnd/src/comit_api/mod.rs
@@ -13,10 +13,6 @@ use libp2p_comit::frame::Header;
 use serde::de::Error;
 use std::fmt;
 
-fn fail_serialize_unknown<D: fmt::Debug>(unknown: D) -> serde_json::Error {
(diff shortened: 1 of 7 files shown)


tcharding:
Thanks, I was unsure exactly _where_ the new error would show up but from the code around where these changes are there are already other de/serialization errors there so I don't feel that this makes the code any worse.  And I agree, I don't think its worth investing time into improving the error handling within the messaging protocol right now :)

thomaseizinger:
> Thanks, I was unsure exactly _where_ the new error would show up but from the code around where these changes are there are already other de/serialization errors there so I don't feel that this makes the code any worse. And I agree, I don't think its worth investing time into improving the error handling within the messaging protocol right now :)

All good. Not sure when any of the others are available for review, feel free to hand it over to bors yourself :)

Issue #1349, opened 2019-09-05
Create cache proxies for blockchain connectors


thomaseizinger:
## Problem

With the changes to what a `BlockchainConnector` is in https://github.com/comit-network/comit-rs/issues/1347 and https://github.com/comit-network/comit-rs/issues/1346, we will fire a request towards the `LatestBlock` and `BlockByHash` traits. The same requests are sent several times from different invocations of `MatchingTransactions`.

## Goal

Use the network more efficiently. Ideally, we want to request each block only once.

## Recommendation

Make use of the proxy pattern to write caching implementations of for each connector: `BitcoindConnector` and `Web3Connector`.
The cache proxy can be specific to each chain but should be generic over the respective traits (`BlockByHash`, `LatestBlock` and `ReceiptByHash`) to allow any connector for this particular chain to be plugged in.

We'd recommend a struct with a generic inner (connector, implements `BlockByHash`, etc traits) field. The struct would store the result and then forward it to the inner. 

Ethereum needs to cache receipts and blocks. Recommendation would have a `BlockHash` and `ReceiptHash` structs. Or you could be generic over what you are caching.

Note that the facade needs to have all 3 traits implemented (`BlockByHash`, `LatestBlock` & `ReceiptByHash`) so the proxy struct will have to implement them too.

tcharding:
I feel that this is two separate issues.

a) `block_by_hash` should cache blocks and return local cached block if a lookup is made for a block we have already fetched from the connector.
b) Optimise `LatestBlock` so we don't fetch data when the latest block has already been fetched.

PR #1762, opened 2019-12-12
Add a block cache to the blockchain connectors

tcharding:
Add a `LruCache` for caching blocks fetched from the blockchain connectors (Just `BitcoindConnector` and `Web3Connector`.  During calls to `block_by_hash()` check the cache before going out to the connector.

Adds caching logic in each connector.

Resolves: #1349


tcharding:
> I'd like this functionality to be solved more generally :)
> With the current approach, the cache only works for specific connectors and we would need to add more caching logic to each new connector.
> 
> With the existing `BlockByHash` abstraction and the [proxy pattern](https://en.wikipedia.org/wiki/Proxy_pattern), we should be able to create something that can be used for any connector as long as it implements `BlockByHash`.

Interesting, I'll give it a go.  Thanks

tcharding:
> Interesting, I'll give it a go. Thanks

I've added a proxy but I cannot please the compiler.  Can you take a look please @thomaseizinger and see if this is what you meant and if you think it is possible to get this code to build.  I've spent a bunch of time on this (on a happily full stomach) and I'm not getting any further - so asking for help :)

tcharding:
The code that declares the type `FooBlockCache`, the impl block for it, and the impl block for `BlockCache`, is all the same.  We could use a macro but I couldn't get the macro expansion to work with `#[async_trait]`?

I tried this: 
```
#[macro_export]
macro_rules! impl_BlockCache {
    ($name:ty) => {
        #[async_trait]
        impl BlockCache for $name {
            type Block = Block;
            type BlockHash = Hash;

            async fn get(&self, block_hash: &Hash) -> anyhow::Result<Option<Block>> {
                let mut cache = self.map.lock().await;
                Ok(cache.get_mut(block_hash).map(|block| block.clone()))
            }

            async fn insert(
                &mut self,
                block_hash: Hash,
                block: Block,
            ) -> anyhow::Result<Option<Block>> {
                let mut cache = self.map.lock().await;
                Ok(cache.insert(block_hash, block))
            }
        }
    };
}
```
And got this build error:
```
error[E0424]: expected value, found module `self`
  --> cnd/src/btsieve/block_cache.rs:4:9
   |
4  |         #[async_trait]
   |         ^^^^^^^^^^^^^^ `self` value is a keyword only available in methods with `self` parameter
   | 
  ::: cnd/src/btsieve/bitcoin/bitcoind_connector.rs:52:1
   |
52 | impl_BlockCache!(BitcoindBlockCache);
   | ------------------------------------- in this macro invocation
```

tcharding:
Follow up issue: If/when this merges we should probably be caching the transaction receipts (thanks @bonomat).

tcharding:
> I'd like this functionality to be solved more generally :)

Now with added macros to declare the cache struct and add the imlp block (i.e., `BitcoindBlockCache::new`).  Does not add the `impl BlockCache` because doing so upsets the compiler, can no longer infer types.  It would be nice to have this imlp block in the macro too ...

cnd/src/btsieve/bitcoin/transaction_pattern.rs
```diff
@@ -75,7 +76,7 @@ mod tests {
 
     fn create_outpoint(tx: &str, vout: u32) -> OutPoint {
         OutPoint {
-            txid: sha256d::Hash::from_hex(tx).unwrap(),
```
thomaseizinger: Up until https://github.com/rust-bitcoin/rust-bitcoin/pull/349 is merged and we can upgrade the bitcoin library, I actually liked importing with the namespace to clearly identify this `Hash` as a transaction hash.

cnd/src/btsieve/block_cache.rs
```diff
@@ -0,0 +1,30 @@
+macro_rules! block_cache {
+    ($type:ident) => {
+        pub struct $type {
+            map: Arc<Mutex<LruCache<Hash, Block>>>,
```
thomaseizinger: To make the macro more contained, you could use full-qualified names here.

cnd/src/btsieve/block_cache.rs
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        impl Clone for $type {
```
thomaseizinger: This can be replaced with a `derive`.

cnd/src/btsieve/block_cache.rs
```diff
@@ -0,0 +1,30 @@
...
+            fn clone(&self) -> Self {
+                Self {
+                    map: Arc::clone(&self.map),
+                }
+            }
+        }
+
+        impl std::fmt::Debug for $type {
```
thomaseizinger: Same here.

cnd/src/btsieve/ethereum/web3_connector.rs
```diff
@@ -9,32 +9,60 @@ use crate::{
...
+block_cache!(Web3BlockCache);
+
+#[async_trait]
+impl BlockCache for Web3BlockCache {
+    type Block = Block;
+    type BlockHash = Hash;
+
+    async fn get(&self, block_hash: &Hash) -> anyhow::Result<Option<Block>> {
+        let mut cache = self.map.lock().await;
+        Ok(cache.get_mut(block_hash).cloned())
```
thomaseizinger: What do you need `get_mut` for?

cnd/src/btsieve/ethereum/web3_connector.rs
```diff
@@ -9,32 +9,60 @@ use crate::{
...
+    }
+
+    async fn insert(&mut self, block_hash: Hash, block: Block) -> anyhow::Result<Option<Block>> {
```
thomaseizinger: The `Map` is wrapped in a `Mutex`, you don't need `&mut self`. I am also not sure whether interior mutability is good here. We could also just have an `Arc<Mutex<BlockCache>>`. That would make it more explicit that the cache is shared.

cnd/src/btsieve/mod.rs
```diff
@@ -37,6 +41,19 @@ pub trait BlockByHash: Send + Sync + 'static {
     ) -> Box<dyn Future<Item = Self::Block, Error = Self::Error> + Send + 'static>;
 }
 
+#[async_trait]
+pub trait BlockCache: Send + Sync + 'static {
```
thomaseizinger: There is no point in requiring the caller to be sync if you have a method with `&mut`. How does this even compile? :D

tcharding:
Thanks for the review @thomaseizinger, I won't respond since I'm on holiday and all and I think it easiest to chat about this in person next week since there is a bunch of things involved.

thomaseizinger:
> Thanks for the review @thomaseizinger, I won't respond since I'm on holiday and all and I think it easiest to chat about this in person next week since there is a bunch of things involved.

Of course! I didn't expect you to respond at all actually :)
Enjoy your holidays!

D4nte:
@tcharding if it's "to-refine" please remove review requests :)

tcharding:
>  if it's "to-refine" please remove review requests :)

Done thanks

tcharding:
Putting this on ice.

Issue #1417, opened 2019-09-19
Improve version printing


D4nte:
## Problem

As a comit-rs dev, I sometimes install a compiled version of comit-rs with `cargo install -p .` to have quick easy access to `cnd`.

However, once it is installed, it is not possible to differentiate two non-tagged versions of `cnd`: all binaries show the version from the `Cargo.toml` file of master.

## Goal

Make `cnd` return its git commit hash in addition of the Cargo.toml version

## Recommendation

Include the git hash as part of the version:
- `--version` prints out version + commit hash and exit
- AND print version + commit hash as first line of logs when starting

Inspire yourself from:
- https://stackoverflow.com/questions/43753491/include-git-commit-hash-as-string-into-rust-program#44407625
- https://docs.rs/vergen/3.0.4/vergen/
- https://github.com/TeXitoi/structopt/issues/136#issuecomment-429119703

PR #1764, opened 2019-12-12
Add a custom version string

tcharding:
Add a custom version string to the cnd executable.  With this applied version option form is unchanged.  Output includes first 12 characters of the latest commit hash.
```
$ cnd --version
COMIT network daemon 0.5.0 (8e5181c4bab3)
```

Resolves: #1417

### DIFF
--- a/cnd/build.rs
+++ b/cnd/build.rs
@@ -1,0 +1,11 @@
+use std::process::Command;
+fn main() {
+    let output = Command::new("git")
+        .args(&["rev-parse", "HEAD"])
+        .output()
+        .expect("failed to get commit hash for tip of branch");
+
+    let git_hash = String::from_utf8(output.stdout).expect("failed to convert hash");
+
+    println!("cargo:rustc-env=GIT_HASH={}", git_hash);
+}
--- a/cnd/build.rs
+++ b/cnd/build.rs
@@ -1,11 +1,13 @@
 use std::process::Command;
 fn main() {
-    let output = Command::new("git")
-        .args(&["rev-parse", "HEAD"])
-        .output()
-        .expect("failed to get commit hash for tip of branch");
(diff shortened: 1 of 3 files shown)


thomaseizinger:
I stumbled over `git describe` the other day:  https://git-scm.com/docs/git-describe

What do you think of using that instead? It doesn't output the hash of the commit if the tag matches exactly (which it would for releases). Not sure if that is good or bad :)

tcharding:
> I stumbled over git describe the other day: git-scm.com/docs/git-describe

Cool, I've not seen this command before.

> What do you think of using that instead?

I don't really think it makes a lot of difference honestly, the difference is shown below , if any one really feels strongly I'm happy to implement it but IMO it does the job just fine as is.

1. `COMIT network daemon 0.5.0 (712b880de561)`
2. `COMIT network daemon 0.5.0-9-g712b880d`

'1' is this PR, '2' is using `git describe`, the 'middle' digit (9) is number of commits since the last tag.

cnd/build.rs
```diff
@@ -0,0 +1,11 @@
+use std::process::Command;
+fn main() {
+    let output = Command::new("git")
+        .args(&["rev-parse", "HEAD"])
+        .output()
+        .expect("failed to get commit hash for tip of branch");
```
bonomat: As mentioned during the call today: maybe it's nice to just print a warning instead of panicing. 

This is not a blocker, so feel free to just move forward with this.

thomaseizinger: In a build script, I'd say it is actually better to panic otherwise the application will panic at runtime (env variable not being set).

Alternatively to `expect`, you could use `?` and have the `main` return an `anyhow::Result` but for a simple build-script like this, it doesn't really make a difference I believe :)

Issue #1586, opened 2019-10-28
Errors propagated to main() do not use logger


tcharding:
## Problem

When an error gets propagated up to main() then printed via the `?` operator the error gets printed to stderr.  This is a problem because throughout cnd we use a logger to print error messages which prints to stdout not stderr.  It is unusual to have error messages printed to two different places.  This will almost certainly bite us at some stage if logs are redirected (or stderr) then later checked during a debugging session.  Also output from containers can easily be redirected differently for stdout and stderr.

## Recommendation

Use a single data stream to write all error messages, stderr.

tcharding:
@D4nte this is the issue I distilled out of your findings from the seed file parse errors.

bonomat:
I think the problem is that we initialize the logger only after reading the config file. That's why everything before can't be logged using the logger.

thomaseizinger:
> I think the problem is that we initialize the logger only after reading the config file.

What you are describing is actually completely unrelated.
The issue @tcharding is describing could be solved be adapting, where we print our log messages here: https://github.com/comit-network/comit-rs/blob/bb8577454c52dec5c56d9db29da2e102e08141bb/cnd/src/logging/initialize.rs#L12

However, I am not sure how we can cleanly separate out error messages from the remaining logs. Should we just print all logs to `stderr`?

tcharding:
> Should we just print all logs to stderr

In my opinion we should research the industry best practice (including considering containers etc.)

tcharding:
We should print all errors and diagnostics to stderr i.e., all logged output.

Reference: thread on [stack overflow](https://unix.stackexchange.com/questions/331611/do-progress-reports-logging-information-belong-on-stderr-or-stdout)

PR #1765, opened 2019-12-12
1586 stderr

tcharding:
Print all diagnostics and log messages to stderr instead of stdout.  With this applied the only output that goes to stdout is the output from `cnd --dump-config`.

Verify with: `target/debug/cnd 2> /tmp/cnd.stderr`.

Resolves: #1586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
```
bonomat:
```suggestion
```

PR #1771, opened 2019-12-16
New end 2 end tests

bonomat:
This PR converts the e2e tests refund and inverted_refund to the new format as described in #1594

The test log files are now somewhat useful :D 

```log
[2019-12-16T15:59:21.425] [DEBUG] alice - Funding as part of swap @ /swaps/rfc003/e7c5b9f2-9a4a-490f-8d4e-81b53ff71e65
[2019-12-16T15:59:26.114] [DEBUG] alice - Waiting for cnd to see alpha_ledger in state FUNDED for swap @ /swaps/rfc003/e7c5b9f2-9a4a-490f-8d4e-81b53ff71e65
[2019-12-16T15:59:26.358] [DEBUG] alice - cnd saw alpha_ledger in state FUNDED for swap @ /swaps/rfc003/e7c5b9f2-9a4a-490f-8d4e-81b53ff71e65
[2019-12-16T15:59:26.358] [DEBUG] bob - Waiting for cnd to see alpha_ledger in state FUNDED for swap @ /swaps/rfc003/e7c5b9f2-9a4a-490f-8d4e-81b53ff71e65
[2019-12-16T15:59:26.603] [DEBUG] bob - cnd saw alpha_ledger in state FUNDED for swap @ /swaps/rfc003/e7c5b9f2-9a4a-490f-8d4e-81b53ff71e65
[2019-12-16T15:59:26.603] [DEBUG] bob - Funding as part of swap @ /swaps/rfc003/e7c5b9f2-9a4a-490f-8d4e-81b53ff71e65
[2019-12-16T15:59:27.721] [DEBUG] alice - Waiting for cnd to see beta_ledger in state FUNDED for swap @ /swaps/rfc003/e7c5b9f2-9a4a-490f-8d4e-81b53ff71e65
[2019-12-16T15:59:29.417] [DEBUG] alice - cnd saw beta_ledger in state FUNDED for swap @ /swaps/rfc003/e7c5b9f2-9a4a-490f-8d4e-81b53ff71e65
[2019-12-16T15:59:29.418] [DEBUG] bob - Waiting for cnd to see beta_ledger in state FUNDED for swap @ /swaps/rfc003/e7c5b9f2-9a4a-490f-8d4e-81b53ff71e65
[2019-12-16T15:59:29.696] [DEBUG] bob - cnd saw beta_ledger in state FUNDED for swap @ /swaps/rfc003/e7c5b9f2-9a4a-490f-8d4e-81b53ff71e65
[...]

PR #1774, opened 2019-12-17
WIP: Rewrite e2e tests for BTC <-> ETH

tcharding:
Builds on top of #1771 (which cannot merge until #1783 merges). 

Works towards resolving #1594.  With this applied all the btc/eth and eth/btc tests that can be done using the current sdk are done i.e., does not do over/under funding tests.


tcharding:
Another case of test passing when run individually but not as a part of the whole suite.

tcharding:
Added these patches onto #1771, @bonomat hope you don't mind me pushing to your branch.

PR #1783, opened 2019-12-18
Run Blockchain nodes bare-metal 

bonomat:
Docker is dead, long live binaries!

This PR is an attempt to fix our random test failures by removing docker-based blockchain nodes. 

So far it was not clear what causes our end2end tests to be flaky and this PR provides a rather radical fix for it.

Reasoning:
Docker has known performance issues with i/o intensive applications. Specifically blockchain nodes are highly i/o intensive. My assumption was that our random timeouts are because the blockchain nodes are too slow to respond to the requests from cnd in time.


bonomat:
Merge if we have one green build!!!! gogogogo!!! Yolo!!!

api_tests/lib/bitcoind_instance.ts
```diff
@@ -0,0 +1,77 @@
...
+                    await openAsync(this.logDir + "/bitcoind.log", "w"), // stderr
+                ],
+            }
+        );
+
+        this.process.on("exit", (code: number, signal: number) => {
+            console.log(`bitcoind exited with ${code || "signal " + signal}`);
+        });
+
+        await sleep(10000); // allow the nodes to start up
```
bonomat: Todo: we should check the log file same as we did for the docker container to verify the node is up.

api_tests/lib/parity_instance.ts
```diff
@@ -0,0 +1,54 @@
...
+                    await openAsync(this.logDir + "/parity.log", "w"), // stdout
+                    await openAsync(this.logDir + "/parity.log", "w"), // stderr
+                ],
+            }
+        );
+
+        this.process.on("exit", (code: number, signal: number) => {
+            console.log(`parity exited with ${code || "signal " + signal}`);
+        });
+        await sleep(5000); // allow the nodes to start up
```
bonomat: Todo: we should check the log file same as we did for the docker container to verify the node is up.

bonomat:
aah... @tcharding ... what did you do? 😅

bonomat:
> What is the motivation of removing the docker containers?
> Reduce complexity?
Does the PR description not tell you the motivation for this? 

> Do the tests really run more stable without docker?

Yes, it seems like it, the end2end test even worked on my machine :D 

Docker has known performance issues with i/o intensive applications. Specifically blockchain nodes are highly i/o intensive. My assumption was that our random timeouts are because the blockchain nodes are too slow to respond to the requests from cnd in time. 
You can check the logs for hints. By reducing the log level the requests were answered faster but still I received random test failures so I decided to just give it a try. :) 

We had good results today also against some of the failing PR's until @tcharding broke this branch :D Let me do a rollback.

tcharding:
@bonomat, I've cleaned up this branch and pushed the new version to https://github.com/comit-network/comit-rs/pull/1785

- squashed this branch into a single commit
- update the makefile clean target to remove blockhcain artifacts
- gitignored the blockchain artifacts
- changed `local_ledger_runner` to `host_ledger_runner`
- removed printing of `ensure_parity.sh` and `ensure_bitcoin.sh` during `make`
- copied code from `ensure_bitcoin.sh` into `ensure_parity.sh` for conditionally calling `mkdir`
- deleted `docker_ledger_runner`

thomaseizinger:
Interesting approach.
To be honest, I had the same question as Daniel: It is the first time I hear about **docker** being a problem in our test setup.
Yes, blockchain nodes are IO intenstive and docker has some problems with IO intensive apps.
However, it does give us a lot of benefits as-well which we would lose with these changes.

Do we still have problems after https://github.com/comit-network/comit-rs/pull/1753 being merged?
Can we please collect more evidence that this is actually better and is worth the added complexity before merging it?

tcharding:
> Do we still have problems after #1753 being merged?

PRs were timing out still with the above PR merged.

tcharding:
> Can we please collect more evidence that this is actually better and is worth the added complexity before merging it?

I'm suspicious that this PR does not _totally_ fix the problem.  #1244 is timing out still (and still displaying the anomaly that the e2e test passes when run alone but fails when run as part of the whole suite).

bonomat:
> I'm suspicious that this PR does not totally fix the problem. #1244 is timing out still (and still displaying the anomaly that the e2e test passes when run alone but fails when run as part of the whole suite).

Did you investigate where and why it times out? 
You can try with the docker setup and just set all timeouts to 15 minutes, if the test still fails then it's a different issue and this PR won't help 

> Do we still have problems after #1753 being merged?

Yes, unfortunately. My findings was that the blockchains were not able to handle the load .

I was of the impression that with this PR we do not get random timeouts anymore.
Meaning, if this PR still does not solve the timeouts than I'm not sure we should merge it.

Thomas is right, docker has some advantages and I think the binaries only make sense if we get the random timeouts fixed.

bonomat:
> Can we please collect more evidence that this is actually better and is worth the added complexity before merging it?

Actually I think this is a really simple solution and easy to maintain. No Docker, no cookie file extracting from volumes or piping logoutputs ... Just starting the official binaries.

thomaseizinger:
> Actually I think this is a really simple solution and easy to maintain. No Docker, no cookie file extracting from volumes or piping logoutputs ... Just starting the official binaries.

Depends on how you look at simplicity.
I primarily look at the maintenance aspect and ANY additional script / file that we have in our repository makes this harder.

Anyway, the primary reason I am very reluctant to merge this is because:

> Yes, unfortunately. My findings was that the blockchains were not able to handle the load .

If we have a load problem, we should fix the load problem and not work around it!

D4nte:
I am surprise that docker would be the cause of the issue as we are not doing something that intensive in tests and I would expect to be able to deploy bitcoind in Docker for mainnet/production usage (as some do, e.g. noodl box).

thomaseizinger:
Some more thoughts:

I don't think we should do this to fix a potential load problem because it is not a proper fix but just applying band aid. There is no inherent need why our application should put a lot of load on down stream services, so any behavior that causes it to do so is bad and should be fixed (we are working on this with the caches around blockchain connectors). Docker has some problems with IO but I would be very surprised if the load that we _should_ cause is a problem. We shouldn't be doing more than a couple of requests per second.

However, we do have to do some weird stuff to make the log files work (which is something I still don't understand). I am happy for us to investigate, running the nodes directly to avoid this weird trickery (which improves the maintainability of the test suite).

In summary: I am on board with doing this if it improves the maintainability of our e2e test suite (ideally without adding complexity through config files / more scripts), every other reason stated here seems to be a problem that should be fixed in a different way.

Process-wise I would have appreciated discussing this first before all of this work is done so that I don't have to feel bad now for you having wasted it :(

api_tests/lib/bitcoind_instance.ts
```diff
@@ -0,0 +1,77 @@
...
+                    await openAsync(this.logDir + "/bitcoind.log", "w"), // stderr
+                ],
+            }
+        );
+
+        this.process.on("exit", (code: number, signal: number) => {
+            console.log(`bitcoind exited with ${code || "signal " + signal}`);
+        });
+
+        await sleep(3000); // allow the nodes to start up
```
thomaseizinger: :mega: To make the tests more deterministic, this should check the logs instead.

api_tests/lib/bitcoind_instance.ts
```diff
@@ -0,0 +1,77 @@
...
+        this.dbDir = tmp.dirSync();
+
+        this.process = spawn(
+            bin,
+            [
+                `-datadir=${this.dbDir.name}`,
+                "-regtest",
+                "-server",
+                "-printtoconsole",
+                "-bind=0.0.0.0:18444",
```
thomaseizinger: :mega: I'd be in favour of maintaining our current functionality and choose a free port here instead of assuming that 18444 will be available.

api_tests/lib/parity_instance.ts
```diff
@@ -0,0 +1,54 @@
...
+                    await openAsync(this.logDir + "/parity.log", "w"), // stdout
+                    await openAsync(this.logDir + "/parity.log", "w"), // stderr
+                ],
+            }
+        );
+
+        this.process.on("exit", (code: number, signal: number) => {
+            console.log(`parity exited with ${code || "signal " + signal}`);
+        });
+        await sleep(3000); // allow the nodes to start up
```
thomaseizinger: :mega: Please check the log file instead of just waiting 3 seconds.

blockchain_nodes/ensure_bitcoind.sh
```diff
@@ -0,0 +1,27 @@
...
+TARGET_FOLDER=./blockchain_nodes/bitcoin
+
+if [ ! -d "${TARGET_FOLDER}" ]
+then
+    mkdir -p "${TARGET_FOLDER}"
+fi
+
+TARGET_FILE="${TARGET_FOLDER}/bitcoin-0.17.0"
+if [ -d "$TARGET_FILE" ]; then
+  echo "${TARGET_FILE} exists; all good mate!"
```
thomaseizinger: I'd be in favor of not adding spam like this that we are gonna see each time we run the tests :)

blockchain_nodes/parity/home/parity/.local/share/io.parity.ethereum/config.toml
```diff
@@ -0,0 +1,41 @@
...
+disable = false
+port = 8545
+interface = "all"
+cors = ["all"]
+hosts = ["all"]
+apis = ["web3", "eth", "net", "parity", "traces", "rpc", "personal", "parity_accounts", "signer", "parity_set"]
+
+[network]
+bootnodes = [
+  "enode://147573f46fe9f5cc38fbe070089a31390baec5dd2827c8f2ef168833e4d0254fbee3969a02c5b9910ea5d5b23d86a6ed5eabcda17cc12007b7d9178b6c697aa5@172.16.0.10:30303",
```
thomaseizinger: What was `172.16.0.10` originally?

Does this have to change now that we are no longer using this inside docker?

api_tests/lib/bitcoind_instance.ts
bonomat: Good idea. How would we do that though? :)

api_tests/lib/bitcoind_instance.ts
bonomat: True, that was a todo which I removed again 😬

blockchain_nodes/ensure_bitcoind.sh
bonomat: I always smile when reading this message. It's like a good friend who pats you on the back and says: `All good mate!`

blockchain_nodes/parity/home/parity/.local/share/io.parity.ethereum/config.toml
```diff
@@ -0,0 +1,41 @@
+[parity]
+#chain = "./home/parity/.local/share/io.parity.ethereum/chain.json"
+#base_path = "./home/parity/.local/share/io.parity.ethereum"
+#db_path = "./home/parity/.local/share/io.parity.ethereum/chains"
+#keys_path = "./home/parity/.local/share/io.parity.ethereum/keys"
+
```
bonomat: 📣 Remove this as it's configured from the outside.

blockchain_nodes/parity/home/parity/.local/share/io.parity.ethereum/config.toml
bonomat: Mhmm... maybe I can even remove this then... I'll give it a try

blockchain_nodes/parity/home/parity/.local/share/io.parity.ethereum/config.toml
```diff
@@ -0,0 +1,41 @@
...
+hosts = ["all"]
+apis = ["web3", "eth", "net", "parity", "traces", "rpc", "personal", "parity_accounts", "signer", "parity_set"]
+
+[network]
+bootnodes = [
+  "enode://147573f46fe9f5cc38fbe070089a31390baec5dd2827c8f2ef168833e4d0254fbee3969a02c5b9910ea5d5b23d86a6ed5eabcda17cc12007b7d9178b6c697aa5@172.16.0.10:30303",
+]
+
+#[account]
+#password = ["./home/parity/authorities/authority.pwd"]
```
bonomat: 📣Remove this as it's configured from the outside

api_tests/lib/bitcoind_instance.ts
thomaseizinger: I would use something like this: https://www.npmjs.com/package/get-port

D4nte:
So we are rolling with docker removal?

bonomat:
@D4nte 
> So we are rolling with docker removal?

Yes, I'd say so

@thomaseizinger : 
Integrated your comments.

PR #1785, opened 2019-12-19
Run blockchain nodes on bare metal

tcharding:
`docker` has some know limitations in relation to IO.  We currently have e2e test failures thanks to timeout that may be caused by the blockchain nodes.  Lets run the blockchain nodes on bare metal to see if the timeout issues is resolved.

Most of this is @bonomat's good work.  Tobin just updated the Makefile, fixed the git ignore stuff, and squashed all the commits ready for general consumption.

### Note

With this applied one must run `make e2e` before running individual tests (with `yarn test ...`) so that the blockchain nodes are downloaded.  The error when one forgets to do so is not obvious
```
Error: spawn /home/tobin/SPOT/build/github.com/comit-network/comit-rs/blockchain_nodes/parity/parity ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:240:19)
    at onErrorNT (internal/child_process.js:415:16)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```


tcharding:
For reasons of devs being all over the world right now this branch was created as the 'mergable' version of #1783, yes this is kinda sloppy - please keep discussion in a single thread over on that PR.  thanks.

D4nte:
I think this brings quite a bit of complexity.
Maybe worth running the CI 10+ times to indeed confirm it does solve the issue.
I would not want to bring that in and then realize it does not fix the issue.

bonomat:
@tcharding: can we close this PR in favor of #1783 which I kept up2date with latest master?

PR #1784, opened 2019-12-19
Make e2e test quieter

tcharding:
In an ongoing effort to improve the dev experience when end to end testing:

- Quieten down the e2e tests, no more:
```
            Stopping cnds: alice, bob, charlie
            Starting cnd for alice, bob, charlie
            cnd charlie exited with signal SIGINT
            cnd bob exited with signal SIGINT
            cnd alice exited with signal SIGINT
```
- Add a command line flag `yarn test --verbose` to maintain the current behavior

PR #1787, opened 2019-12-20
Shorten test sleep time

tcharding:
Shorten sleep times to 200 milliseconds instead of 1 second.  200 milliseconds is an eternity for a modern processor.  We do not need to wait for a second for the CPU to process things.  As for the CI infrastructure ...

### DIFF
--- a/api_tests/lib/actor.ts
+++ b/api_tests/lib/actor.ts
@@ -74,7 +74,7 @@ export class Actor {
         if (predicate(response.body)) {
             return response.body;
         } else {
-            await sleep(500);
+            await sleep(200);
 
             return this.pollCndUntil(location, predicate);
         }
@@ -195,7 +195,7 @@ export class Actor {
                         let escapeSequence = "";
 
                         while (diff > 0) {
-                            await sleep(1000);
+                            await sleep(200);
 
                             currentMedianBlockTime = await fetchMedianTime();
                             diff =
@@ -252,7 +252,7 @@ export class Actor {
                             min_block_timestamp
                         );
                     }
-                    await sleep(delay * 1000);
+                    await sleep(delay * 200);
                 }
 
                 return this.wallet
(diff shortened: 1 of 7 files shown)


tcharding:
These patches were previously included in PR #1784.

api_tests/lib_sdk/cnd_instance.ts
```diff
@@ -54,13 +57,15 @@ export class CndInstance {
...
-            );
+            if (global.verbose) {
+                console.log(
+                    `cnd ${this.actorConfig.name} exited with ${code ||
+                        "signal " + signal}`
+                );
+            }
         });
 
-        await sleep(1000); // allow the nodes to start up
```
bonomat: 📣Especially on my machine this won't be enough.

api_tests/lib_sdk/wallets/index.ts
```diff
@@ -65,7 +65,7 @@ export async function pollUntilMinted(
     if (new BigNumber(currentBalance).gte(minimumBalance)) {
         return;
     } else {
-        await sleep(500);
```
bonomat: 📣We have a new block every 1 second. Which means requesting the balance every 200ms does not make much difference besides putting more load on your machine.

api_tests/lib/actor.ts
```diff
@@ -182,24 +185,28 @@ export class Actor {
...
+                            console.log(
+                                `Waiting for median time to pass %d`,
+                                min_median_block_time
+                            );
+                        }
 
                         let escapeSequence = "";
 
                         while (diff > 0) {
-                            await sleep(1000);
```
bonomat: 📣Our HTLC timeouts are set to +3/+8 seconds, so asking the chain every 200ms for the latest time won't make much difference.

tcharding:
I put the verbose flag patch into a separate PR (merged already) and forgot to remove it from this one.  Closing this since after the comments, thanks for the review @bonomat

Issue #1682, opened 2019-11-25
Bob can derive a (incorrect) secret - prevent this at compile-time


thomaseizinger:
## Problem

The current design of Bob's `State` holds an implementation of `SecretSource`. `SecretSource` allows you to derive a secret and keys from the underlying seed.
Bob doesn't need to have access to this functionality in the rfc003 protocol because Alice is the one to generate the secret.

## Goal

Alter the design of the `SecretSource` and `State` structs in a way that models the real world better, making it impossible for Bob the derive a secret.

## Recommendation

1. Split up the `SecretSource` into three traits:
a. `DeriveSecret`
b. `DeriveRedeemIdentity`
c. `DeriveRefundIdentity`

> `DeriveRedeemIdentity` and `DeriveRefundIdentity` might be also be combined into `DeriveIdentities`.

2. Replace all the usages of `SecretSource` with the actual trait bounds that are needed.

3. Change the instance of `SecretSource` inside the `State` struct to only hold the minimal interface that is actually needed for the code to work. (Bob should only need `DeriveIdentities` here).

bonomat:
I'm curious about this one. Do you have some more details please?

PR #1795, opened 2019-12-23
Encode the capability of deriving a Secret into the type system so that only Alice can do that

tcharding:
Currently Bob can derive the secret from seed found in the state store.  This is logically incorrect.  With this applied Bob's functionality when using the `secret_source` is reduced to only deriving identities.

Resolves: #1682

### DIFF
--- a/cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
+++ b/cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
@@ -10,7 +10,7 @@ use crate::{
         ledger,
         rfc003::{
             self, alice::State, events::HtlcEvents, state_store::StateStore, Accept, Decline,
-            Ledger, Request, SecretHash, SecretSource,
+            DeriveIdentities, DeriveSecret, Ledger, Request, SecretHash,
         },
         Facade, HashFunction, Role, SwapId,
     },
@@ -27,7 +27,7 @@ pub async fn handle_post_swap<S: Network>(
 ) -> anyhow::Result<SwapCreated> {
     let id = SwapId::default();
     let seed = dependencies.derive_swap_seed(id);
-    let secret_hash = seed.secret().hash();
+    let secret_hash = seed.derive_secret().hash();
 
     let body = serde_json::from_value(body)?;
 
@@ -289,7 +289,7 @@ struct Identities<AL: Ledger, BL: Ledger> {
 trait IntoIdentities<AL: Ledger, BL: Ledger> {
     fn into_identities(
         self,
-        secret_source: &dyn SecretSource,
+        secret_source: &dyn DeriveIdentities,
     ) -> anyhow::Result<Identities<AL, BL>>;
 }
 
@@ -322,7 +322,7 @@ pub enum IdentityKind {
 impl IntoIdentities<ledger::Bitcoin, ledger::Ethereum> for HttpIdentities {
     fn into_identities(
         self,
-        secret_source: &dyn SecretSource,
+        secret_source: &dyn DeriveIdentities,
     ) -> anyhow::Result<Identities<ledger::Bitcoin, ledger::Ethereum>> {
         let HttpIdentities {
             alpha_ledger_refund_identity,
@@ -346,7 +346,7 @@ impl IntoIdentities<ledger::Bitcoin, ledger::Ethereum> for HttpIdentities {
 
         let alpha_ledger_refund_identity = crate::bitcoin::PublicKey::from_secret_key(
             &*crate::SECP,
-            &secret_source.secp256k1_refund(),
+            &secret_source.derive_refund_identity(),
         );
 
         Ok(Identities {
@@ -359,7 +359,7 @@ impl IntoIdentities<ledger::Bitcoin, ledger::Ethereum> for HttpIdentities {
 impl IntoIdentities<ledger::Ethereum, ledger::Bitcoin> for HttpIdentities {
     fn into_identities(
         self,
-        secret_source: &dyn SecretSource,
+        secret_source: &dyn DeriveIdentities,
     ) -> anyhow::Result<Identities<ledger::Ethereum, ledger::Bitcoin>> {
         let HttpIdentities {
             alpha_ledger_refund_identity,
@@ -383,7 +383,7 @@ impl IntoIdentities<ledger::Ethereum, ledger::Bitcoin> for HttpIdentities {
 
         let beta_ledger_redeem_identity = crate::bitcoin::PublicKey::from_secret_key(
             &*crate::SECP,
-            &secret_source.secp256k1_redeem(),
+            &secret_source.derive_redeem_identity(),
         );
 
         Ok(Identities {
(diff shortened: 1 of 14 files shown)


cnd/src/swap_protocols/rfc003/mod.rs
```diff
@@ -42,3 +44,46 @@ pub enum Error {
...
+    fn derive_refund_identity(&self) -> SecretKey {
+        SecretKey::from_slice(self.sha256_with_seed(&[b"REFUND"]).as_ref())
+            .expect("The probability of this happening is < 1 in 2^120")
+    }
+}
+
+/// SecretHasher generates the secret from a hash of the current seed for
+/// inclusion in HTLC.  This 'secret' is what is revealed to Bob when Alice
+/// redeems the HTLC on beta ledger.
+pub trait SecretHasher {
```
thomaseizinger: Do we really need this trait?

I'd prefer if we find a model that doesn't rely on a concept like this because it doesn't exist in our domain.

cnd/src/swap_protocols/rfc003/bob/mod.rs
```diff
@@ -21,9 +24,9 @@ pub struct State<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> {
     pub swap_communication: SwapCommunication<AL, BL, AA, BA>,
     pub alpha_ledger_state: LedgerState<AL>,
     pub beta_ledger_state: LedgerState<BL>,
-    #[derivative(Debug = "ignore")]
-    pub secret_source: Arc<dyn SecretSource>,
-    pub secret: Option<Secret>,
+    #[derivative(Debug = "ignore", PartialEq = "ignore")]
+    pub secret_source: Seed, // Used to derive identities for refund/redeem.
```
thomaseizinger: If you want to add comments, please use rustdoc. (triple slash comment in the line above the field.

cnd/src/swap_protocols/rfc003/alice/actions/erc20.rs
```diff
@@ -64,8 +64,8 @@ where
             actions.push(Action::Redeem(<(BL, BA)>::redeem_action(
                 HtlcParams::new_beta_params(request, response),
                 htlc_location.clone(),
-                &*self.secret_source,
-                self.secret_source.secret(),
+                self.secret_source, // Alice uses this swaps seed to derive the secret.
```
thomaseizinger: Why are we passing the secret and the source if we only need the source for deriving the secret (assuming your comment is correct)?

cnd/src/swap_protocols/rfc003/bob/mod.rs
thomaseizinger: :mega: Bob can still derive a secret here. The main idea of the ticket is to make this impossible. To do that, this would need to be something like `dyn DeriveIdentities`.

cnd/src/swap_protocols/rfc003/alice/actions/erc20.rs
tcharding: That is the exact question that led me to write this comment, I can't remember the answer right now but clearly the addition of this comment has not made the code clear enough :)

cnd/src/swap_protocols/rfc003/bob/mod.rs
tcharding: I realise that this PR does in no way inhibit Bob from deriving the secret.  I played with using dyn and could not get it to work because there are times when we must pass a parameter that is for use by Alice _and_ Bob, hence needs `DeriveSecret`.  I could come up with no way in code to separate Bob and Alice call chains since in many other places in the code base there is no explicit separation.  Happy to be proved wrong.

tcharding: This is not, in my opinion, a comment for consumers of this module so it doesn't get a rustdoc.  Its a comment for internal usage of this module because the variable name/type needs further explanation.  If my understanding of use of rustdoc vs `//` is wrong I'm happy to be schooled.  Also, I think that if it does require a rustdoc comment then all of the fields here should have one - if this is the case though, we probably should have a lot more documentation.

tcharding:
Thanks for the review @thomaseizinger

cnd/src/swap_protocols/rfc003/bob/mod.rs
thomaseizinger: > we probably should have a lot more documentation.

Yes!

>  Also, I think that if it does require a rustdoc comment then all of the fields here should have one

Yes!

>  If my understanding of use of rustdoc vs // is wrong I'm happy to be schooled.

I wouldn't make a distinction here. rustdoc is Rust's way of adding documentation to the source code. Because it uses a well-defined format, it can then automatically generate documentation. IMO, there is no downside to using it, so I would just opt for rustdoc :)

thomaseizinger: It is unfortunate if it doesn't work (yet).

I think, making it work would probably result in a better model because it resembles the real world better (Bob never derives a secret). If you are keen, we can pair-program on this next week and see if we can make progress!

tcharding: > I wouldn't make a distinction here. rustdoc is Rust's way of adding documentation to the source code. Because it uses a well-defined format, it can then automatically generate documentation. IMO, there is no downside to using it, so I would just opt for rustdoc :)

I have to disagree with you here @thomaseizinger, rustdoc is for comments that should be included in the rendered documentation _not_ for random code comments (that arguably should not need to exist in the first place).  I definitely feel there is a usecase for `//` comments.  Whether this case in question is one of those is open to debate.

thomaseizinger: >  I definitely feel there is a usecase for // comments. Whether this case in question is one of those is open to debate.

I think having such comments can be worthwhile if they are inline with the source-code.
You are adding documentation to a struct field here, I'd argue that is what rustdoc is for (among other things).

tcharding:
Squashed and force pushed - sorry :)

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
```diff
@@ -346,7 +344,7 @@ pub enum IdentityKind {
 impl IntoIdentities<ledger::Bitcoin, ledger::Ethereum> for HttpIdentities {
     fn into_identities(
         self,
-        secret_source: &dyn SecretSource,
+        secret_source: &Seed,
```
thomaseizinger: Does this really need `Seed`? Should be fine with `DeriveIdentities` no?

cnd/src/swap_protocols/rfc003/alice/actions/erc20.rs
```diff
@@ -64,8 +64,8 @@ where
             actions.push(Action::Redeem(<(BL, BA)>::redeem_action(
                 HtlcParams::new_beta_params(request, response),
                 htlc_location.clone(),
-                &*self.secret_source,
-                self.secret_source.secret(),
+                &self.secret_source, // Derive identities with this.
+                self.secret_source.derive_secret(), // The secret used by Alice.
```
thomaseizinger: To make this clearer, we could derive the secret for Alice upon instantiation of the state and do `self.secret` here? What do you think?

tcharding: I gave it a shot, it just moves the confusion to the place where we derive the secret (since at that place we now hove both the secret_source and the secret).

thomaseizinger:
@tcharding I fixed up the merge conflicts for you so that we can land this. I'll go ahead and merge this!

bors r+

PR #1796, opened 2019-12-24
Ethereum: Refactor `matching_transactions()`

tcharding:
The function `matching_transactions` is a candidate for refactoring because it is super long and high in complexity.  We can simplify the code, making it easier to read, understand, and maintain, by factoring parts out into separate functions.

There are very clear boundaries in the function where we spawn closures.  We can factor each closure out into a separate function.

**Please note**:  Patch 3 is a logic change _not_ a refactoring.  It is an optimisation in line with other functions in the newly refactored code.

The bitcoin connector refactor of `matching_transacitions()` can be found [here](https://github.com/comit-network/comit-rs/pull/1797).


thomaseizinger:
Will review once the build is passing.

tcharding:
> Will review once the build is passing.

Build is failing because of a `FIXME` - I left this in on purpose so we can discuss it.

thomaseizinger:
> > Will review once the build is passing.
> 
> Build is failing because of a `FIXME` - I left this in on purpose so we can discuss it.

Was this behaviour already the case before or is it introduced with the refactoring?
If you can pull off this refactoring without a logic change, I am happy to blindly approve :)

tcharding:
> Was this behaviour already the case before or is it introduced with the refactoring?

Was there before, @luckysori is fixing it - he has picked up this PR.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -243,9 +243,9 @@ async fn process_next_find_parent<C, E>(
     }
 }
 
-/// Process hashes from the `next_look_in_the_past` receiver.  Gets the block
+/// Process hashes from the `next_look_in_the_past` receiver. Gets the block
```
tcharding: No need to change but I always use two spaces after a full stop.  Researching this for this post shows that I'm wrong - will now switch to using one space.  FTR my emacs setup does not do 're column width' i.e., update region to use the correct column width, correctly when there is only one space (that's why I started using two).

D4nte:
I have not touched btsieve since teh refactoring so I had a look and do not see any major issue but prefer to hold my approval due to lack of confidence.

tcharding:
> I have not touched btsieve since the refactoring so I had a look and do not see any major issue but prefer to hold my approval due to lack of confidence.

No worries, @thomaseizinger that leaves you :)

PR #1806, opened 2020-01-02
Migrate to tokio 0.2

thomaseizinger:
Tokio-compat is compatibility shim for tokio 0.1 and tokio 0.2. It allows us to poll futures of tokio 0.1 and 0.2 at the same time.


thomaseizinger:
This needs to be done at the same time as https://github.com/comit-network/comit-rs/issues/1804.

thomaseizinger:
This needs to have https://github.com/comit-network/comit-rs/issues/1804 otherwise it doesn't work.

thomaseizinger:
I have incorporated https://github.com/comit-network/comit-rs/pull/1804 into this because we need both of them at the same time!

PR #1810, opened 2020-01-03
Upgrade comit-sdk to 0.9.0

D4nte:


api_tests/lib/setup_chai.ts
```diff
@@ -1,21 +1,21 @@
 // A helper file to configure chai with all the plugins we need
 // This is to reduce noise in our actual test files
 
-import chaiBn = require("bn-chai");
 import { tv4, use } from "chai";
 import chaiAsPromised from "chai-as-promised";
+// @ts-ignore
```
thomaseizinger: :mega: Is there a way how we can _not_ introduce this? I'd like to have more verification at compile-time than runtime failures.

api_tests/types/bn-chai/index.d.ts
```diff
@@ -1,11 +1,11 @@
...
-
-    export = bnChaiFn;
-}
-
-declare namespace Chai {
-    interface NumberComparer {
-        BN(value: any): void;
-    }
-}
+// declare module "bn-chai" {
```
thomaseizinger: :mega: If we no longer need it, remove it please.

thomaseizinger: Although I'd prefer to remove the `ts-ignore` and provide correct typings.

luckysori: :question: What's going on here?

D4nte: no choice, because we moved from `bn.js` based `ethers.utils.BigNumber` type to `bignumber.js`.

D4nte: Yes need to be removed.
I am trying to make this work before wasted my time on typings (and I did try to add typings for 10min then gave up).

api_tests/lib/setup_chai.ts
D4nte: I don't know if there is a way to not introduce this which does not involve a 100 lines typing module.

thomaseizinger:
@D4nte please have a look and remove the label if keen for merging!

I added typings for `chai-bignumber` and squashed it all down into one commit to not have any intermediaries that don't build.

PR #1812, opened 2020-01-03
Make `HtlcEvents` an async trait

thomaseizinger:


thomaseizinger:
This PR is at a stage where I think we need to rewrite the state machine before we can continue :(

cnd/src/swap_protocols/facade.rs
```diff
@@ -169,67 +167,73 @@ where
...
+        self.bitcoin_connector.htlc_deployed(htlc_params).await
     }
 
-    fn htlc_funded(
+    async fn htlc_funded(
         &self,
         htlc_params: HtlcParams<Bitcoin, Amount>,
         htlc_deployment: &Deployed<Bitcoin>,
-    ) -> Box<FundedFuture<Bitcoin, Amount>> {
+    ) -> Result<Funded<Bitcoin, Amount>, rfc003::Error> {
```
tcharding: This line highlights some non-uniformity we currently have.  We could be returning an `anyhow::Result` here.  `rcf003::Error` is implemented using `thiserror` now so implements `std::error::Error`.  I had a quick go but its not a two minute job.  Want me to investigate further and write up an issue?

PR #1818, opened 2020-01-06
Return Siren links for /swaps and /swaps/rfc003 in index

### LINKED ISSUE #1557: Reduce the coupling of cnd's clients to its API
> This is a follow-up ticket from: https://github.com/comit-network/comit-rs/issues/1504

## Problem

Currently, clients of cnd have to hardcode the paths `/swaps` and `/swaps/rfc003` for retrieving all swaps and creating a new one. This creates coupling and makes it a breaking change for us to change those paths (as for example suggested in https://github.com/comit-network/comit-rs/issues/1504. This coupling restricts the solution-space for tackling issues like https://github.com/comit-network/comit-rs/issues/1504. Making a breaking change might just not be a viable solution (depending on the adoption of cnd and the comit-js-sdk), given the pain we'd have in upgrading things like the sdk and having to match the version of cnd used inside create-comit-app.

## Goal

Not to constrain the future coblox team in the solution space regarding changes to cnd's API.

## Recommendation

We already make good use of siren in cnd's API to reduce the coupling when it comes to executing actions. We can reduce the coupling further by returning a siren document on `GET /`that has two links:

```json
{
  "class": [],
  "properties": {},
  "entities": [],
  "actions": [],
  "links": [
    { "rel": [ "collection" ], "class": ["swaps"], "href": "/swaps" },
    { "rel": [ "collection", "edit" ], "class": ["swaps", "rfc003"], "href": "/swaps/rfc003" },
  ]
}
```

### Rationale

We try to use as many standard link relations as possible. By combining them with siren's `class` attribute, we can easily differentiate between all our usecases.

1. `collection` is a link-relation that describes a link that points to a collection. Hence, we can describe the link that points to a collection of swaps by adding the class `swaps`.

2. `edit` describes a link that can be used to change a link's context. Together with the `collection` relation and the two classes `swaps` and `rfc003`, we can describe the link that allows a client to **edit** the **rfc003** **collection** of **swaps**.

luckysori:
Fixes #1557.

### DIFF
--- a/api_tests/dry/sanity.ts
+++ b/api_tests/dry/sanity.ts
@@ -89,15 +89,26 @@ setTimeout(async function() {
             expect(res.body.peers).to.have.length(0);
         });
 
-        it("[Alice] Response for GET / is a valid siren document", async () => {
+        it("[Alice] Returns its peer ID and the addresses it listens on when you GET /", async () => {
+            const res = await request(alice.cndHttpApiUrl()).get("/");
+
+            expect(res.body.id).to.be.a("string");
+            expect(res.body.listen_addresses).to.be.an("array");
+            // At least 2 ipv4 addresses, lookup and external interface
+            expect(res.body.listen_addresses.length).to.be.greaterThan(1);
+        });
+
+        it("[Alice] Response for GET / with accept header set as application/vnd.siren+json is a valid siren document", async () => {
             const res = await request(alice.cndHttpApiUrl()).get("/");
 
             expect(res).to.have.status(200);
             expect(res.body).to.be.jsonSchema(sirenJsonSchema);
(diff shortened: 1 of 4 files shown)


api_tests/dry/sanity.ts
```diff
@@ -88,20 +89,42 @@ setTimeout(async function() {
             expect(res.body.peers).to.have.length(0);
         });
 
-        it("[Alice] Returns its peer ID when you GET /", async () => {
+        it("[Alice] Response for GET / is a valid siren document", async () => {
```
D4nte: The PR title and the commit does not mention changes on `/`.
Please update README and fix description.

PR #1820, opened 2020-01-06
Add two types `RootSeed` and `SwapSeed`

tcharding:
Currently we have a `Seed` type that is used as the root seed to derive
per swap seeds which are then used to derive identities (refund/redeem).
Using the same type for all of these seeds means the compiler cannot
give us type safety.  Since these two things are separate we can use
separate types to allow the compiler to differentiate the types and
provide safety.

Add `RootSeed` for the seed read in from the PEM file and derive a new
`SwapSeed` to use for identity derivation.

Note: I'm not totally happy with this one, open to any suggested improvements, thanks.

Done in preparation for #1682

### DIFF
--- a/cnd/src/seed.rs
+++ b/cnd/src/seed.rs
@@ -12,9 +12,6 @@ use std::{
 };
 use thiserror;
 
-const SEED_LENGTH: usize = 32;
-pub type Seed = [u8; SEED_LENGTH];
-
 /// We create a `RootSeed` either randomly or by reading in the PEM file from
 /// disk.  This `RootSeed` is used to generate a per swap `SwapSeed` which is
 /// then use as the secret source for deriving redeem/refund identities.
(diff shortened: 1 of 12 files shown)


cnd/src/seed.rs
```diff
@@ -12,82 +12,109 @@ use std::{
 };
 use thiserror;
 
-pub const SEED_LENGTH: usize = 32;
+const SEED_LENGTH: usize = 32;
+pub type Seed = [u8; SEED_LENGTH];
```
thomaseizinger: I am not sure if this only being a type-alias helps much?

I'd either go for a real newtype (so you can attach functionality to it), or not having the type alias.

cnd/src/seed.rs
```diff
@@ -12,82 +12,109 @@ use std::{
...
+    fn derive_swap_seed(&self, id: SwapId) -> SwapSeed {
+        SwapSeed(self.sha256_with_seed(&[b"SWAP", id.0.as_bytes()]))
+    }
+}
+
 #[derive(Clone, Copy, Serialize, Deserialize, PartialEq)]
-pub struct Seed(#[serde(with = "hex_serde")] [u8; SEED_LENGTH]);
+pub struct RootSeed(#[serde(with = "hex_serde")] Seed);
+
+impl fmt::Debug for RootSeed {
```
thomaseizinger: If you make `Seed` an actual newtype, you could implement `Debug` and `Display` once for that type and you would not need to repeat it here.

I would then forward the `Debug` and `Display` impls to the inner one.

cnd/src/seed.rs
```diff
@@ -12,82 +12,109 @@ use std::{
...
+/// `RootSeed` and `SwapSeed` are the same underlying type (`Seed`), they exist
+/// solely to allow the compiler to provide us with type safety.
+
+pub trait DeriveSwapSeed {
+    fn derive_swap_seed(&self, id: SwapId) -> SwapSeed;
+}
+
+impl DeriveSwapSeed for RootSeed {
+    fn derive_swap_seed(&self, id: SwapId) -> SwapSeed {
+        SwapSeed(self.sha256_with_seed(&[b"SWAP", id.0.as_bytes()]))
```
thomaseizinger: If you make `Seed` an actual newtype, the `sha256_with_seed` function can live on there so you don't have to abuse `SwapSeed` here :)

cnd/src/seed.rs
tcharding: Awesome, done as suggested.

cnd/src/seed.rs
tcharding: Done.

cnd/src/seed.rs
tcharding: Whats the abuse?  Totally honest question, other suggestions implemented but this function is now 
```
impl DeriveSwapSeed for RootSeed {
    fn derive_swap_seed(&self, id: SwapId) -> SwapSeed {
        SwapSeed(Seed(self.sha256_with_seed(&[b"SWAP", id.0.as_bytes()])))
    }
}
```

thomaseizinger: Don't mind me, I thought you are wrapping `self` inside `SwapSeed` to access a function. I can't read apparently.

tcharding: All g!

tcharding:
Rebased on top of master and force pushed - no other changes.

cnd/src/seed.rs
```diff
@@ -12,9 +12,27 @@ use std::{
...
+        let data = self.sha256_with_seed(&[b"SWAP", id.0.as_bytes()]);
+        SwapSeed(Seed(data))
+    }
+}
+
+const SEED_LENGTH: usize = 32;
+
 #[derive(Clone, Copy, Serialize, Deserialize, PartialEq)]
-pub struct Seed(#[serde(with = "hex_serde")] [u8; SEED_LENGTH]);
+struct Seed(#[serde(with = "hex_serde")] [u8; SEED_LENGTH]);
```
thomaseizinger: :+1:

PR #1824, opened 2020-01-07
Implement Copy on as many types as possible

thomaseizinger:
If types can be copy, they should according to the docs:
https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy.

Implementing Copy on as many types as possible allows us to remove `.clone()` in some circumstances.

### DIFF
--- a/cnd/src/http_api/swap_resource.rs
+++ b/cnd/src/http_api/swap_resource.rs
@@ -33,7 +33,7 @@ pub struct SwapResource<S> {
     pub state: Option<S>,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Debug, Clone, Copy, Serialize)]
 pub struct SwapParameters {
     alpha_ledger: HttpLedger,
     beta_ledger: HttpLedger,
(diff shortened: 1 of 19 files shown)


cnd/src/http_api/swap_resource.rs
```diff
@@ -33,15 +33,15 @@ pub struct SwapResource<S> {
...
+#[derive(Debug, Clone, Copy, Serialize)]
 pub struct SwapParameters {
     alpha_ledger: HttpLedger,
     beta_ledger: HttpLedger,
     alpha_asset: HttpAsset,
     beta_asset: HttpAsset,
 }
 
-#[derive(Debug, Serialize, PartialEq)]
+#[derive(Debug, Clone, Copy, Serialize, PartialEq)]
```
D4nte: Why do you still keep Clone?

cnd/src/lib.rs
```diff
@@ -1,4 +1,9 @@
-#![warn(unused_extern_crates, missing_debug_implementations, rust_2018_idioms)]
+#![warn(
```
D4nte: what about the `libp2p-comit` crate?

thomaseizinger: Fair enough, can add there as-well!

cnd/src/http_api/swap_resource.rs
thomaseizinger: Can't derive Copy without Clone.

PR #1825, opened 2020-01-07
Migrate state machine to async/await using stable generators

thomaseizinger:
This patch migrates the "state machine" to async/await.

It makes use of the `genawaiter` crate to use generators on stable Rust. As part of that, what used to be the state machine is now a simple stream of events, emitting what is happening on each of the ledgers, without performing logic or computation on that result.

Each of the events is fed into the state store where we take the event and transition the appropriate ledger state of the actor to the new one.

In the case of an invalid funding, we transition accordingly.

This patch also removes `Option<Secret>` from `bob::State`. I am not 100% sure that it is a good idea though. It was made possible because I now record the `Secret` as-well as the asset that was used for funding within the `LedgerState` variant.
Hence, I can just access the `Secret` that was used to redeem in the `impl Actions` implementation.

cc @LLFourn :)


cnd/src/swap_protocols/rfc003/bob/actions/erc20.rs
```diff
@@ -50,19 +50,19 @@ where
 
         use self::LedgerState::*;
 
-        let mut actions = match (alpha_state, beta_state, self.secret) {
-            (Funded { htlc_location, .. }, _, Some(secret)) => {
+        let mut actions = match (alpha_state, beta_state) {
+            (Funded { htlc_location, .. }, Redeemed { secret, .. }) => {
```
thomaseizinger: Bob can now access the funding by inspecting the `beta_ledger_state`. In case that one is `Redeemed`, the secret used by Alice is captured in there.

cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -0,0 +1,279 @@
...
+    sync::{Co, Gen},
+    GeneratorState,
+};
+
+/// Returns a future that tracks the swap negotiated from the given request and
+/// accept response on both ledgers.
+///
+/// The current implementation is naive in the sense that it does not take into
+/// account situations where it is clear that no more events will happen even
+/// though in theory, there could. For example:
```
D4nte:
```suggestion
/// though in theory, it could. For example:
```
?

cnd/src/swap_protocols/rfc003/state_store.rs
```diff
@@ -63,192 +51,51 @@ impl StateStore for InMemoryStateStore {
...
+
+                match expected_asset.cmp(&funded.asset) {
+                    Ordering::Equal => actor_state.alpha_ledger_mut().transition_to_funded(funded),
+                    _ => actor_state
+                        .alpha_ledger_mut()
+                        .transition_to_incorrectly_funded(funded),
                 }
             }
+            SwapEvent::AlphaRedeemed(redeemed) => {
+                // what if redeemed.secret.hash() != secret_hash in request ??
```
D4nte: If I understand correctly, it is assumed impossible because only the right secret can redeem.
So maybe worth making a note here.

cnd/src/swap_protocols/rfc003/ledger_state.rs
```diff
@@ -18,26 +25,136 @@ pub enum LedgerState<L: Ledger> {
...
+                }
+            }
+            other => panic!("expected state Deployed, got {}", HtlcState::from(other)),
+        }
+    }
+
+    pub fn transition_to_incorrectly_funded(&mut self, funded: Funded<L, A>) {
+        let Funded { transaction, asset } = funded;
+
+        match std::mem::replace(self, LedgerState::NotDeployed) {
```
luckysori: :question: Why are the `mem::replace`s needed and why are they always replaced with `LedgerState::NotDeployed`?

cnd/src/swap_protocols/rfc003/ledger_state.rs
```diff
@@ -18,26 +25,136 @@ pub enum LedgerState<L: Ledger> {
...
+        } = deployed;
+
+        match std::mem::replace(self, LedgerState::NotDeployed) {
+            LedgerState::NotDeployed => {
+                *self = LedgerState::Deployed {
+                    deploy_transaction: transaction,
+                    htlc_location: location,
+                }
+            }
+            other => panic!("expected state NotDeployed, got {}", HtlcState::from(other)),
```
luckysori: :thinking: I wish we could leverage the type system to avoid these panics. Hmmm.

cnd/src/swap_protocols/init_swap.rs
```diff
@@ -32,46 +29,29 @@ where
...
-                Role::Alice => {
-                    StateStore::update::<alice::State<AL, BL, AA, BA>>(&dependencies, &id, update)
-                }
-                Role::Bob => {
-                    StateStore::update::<bob::State<AL, BL, AA, BA>>(&dependencies, &id, update)
-                }
-            }
-            Ok(())
+            dependencies
+                .clone()
```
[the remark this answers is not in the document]
thomaseizinger: Unfortunately I do because spawning requires a mutable reference. We will probably rework this in the future though.

cnd/src/swap_protocols/init_swap.rs
```diff
@@ -3,16 +3,13 @@ use crate::{
...
         },
-        Role, SwapId,
+        Role,
     },
 };
-use futures::{Future, Stream};
+use futures_core::{FutureExt, TryFutureExt};
 use tokio::executor::Executor;
 
 #[allow(clippy::cognitive_complexity)]
```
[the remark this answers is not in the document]
thomaseizinger: Haven't checked actually :D

cnd/src/swap_protocols/rfc003/ledger_state.rs
thomaseizinger: Don't really know if we can, these panics are bugs IMO so it should be fine.

cnd/src/swap_protocols/rfc003/ledger_state.rs
D4nte: I wondered too but it then made sense ;)

PR #1832, opened 2020-01-08
http_api: Use Facade as arg instead of trait bounds

tcharding:
Currently we pass the dependencies throughout the http_api module using trait bounds.  This is super verbose.  Also since we now this is only ever going to be an actual instance of the `Facade` and we are not trying to limit the callers usage of the facade it adds no real benefit using trait bounds.


tcharding:
@thomaseizinger I did this now because while working on the libp2p stuff it seemed like it was a good time to do it - also my brain was kinda fried and this was low intensity work.  We talked about this before but I could not find an issue, is there an issue tracking this?

Issue #1776, opened 2019-12-17
Make it easier to handle an instance of `libp2p::Swarm`


thomaseizinger:
## Problem

At the moment, dealing with our instance of a `libp2p::Swarm` is annoying because it is hard to name due to the complexity of the type parameters.
This manifests in two ways:

- Our `Facade` type has a type parameter `S` and we need to repeat `S: Send + Sync + 'static` for all the trait implementations
- The network initialization code is all inside `main` because we cannot name the `Swarm` instance (and hence return it from a method or take it as a parameter somewhere.

## Goal

1. Consistency in our `Facade` struct
2. Simple `main` function that only deals with one level of detail per component of the application (config, logging, http-api, comit-api, etc). Currently, this is violated for the comit-api.

## Solution

Hard to say. The "standard" way would be to create a struct that has a trait object (`Box`) inside. This avoids bubbling up of type parameters but I am not sure if this approach works.
Probably worth asking on the `rust-libp2p` GitHub repo for help.

thomaseizinger:
Important work has been done here: https://github.com/comit-network/comit-rs/tree/better-libp2p-handling

PR #1839, opened 2020-01-09
Clean up Network code

tcharding:
Currently `main.rs` contains a bunch of networking code.  The root cause of much of this is due to libp2p types.  If we wrap the libp2p swarm in a struct then we can do a lot of cleaning up.

With this applied there is no more libp2p code in `main.rs` :)

Resolves: #1776

### DIFF
--- a/api_tests/lib_sdk/cnd_instance.ts
+++ b/api_tests/lib_sdk/cnd_instance.ts
@@ -32,6 +32,10 @@ export class CndInstance {
             ? process.env.CND_BIN
             : this.projectRoot + "/target/debug/cnd";
 
+        if (global.verbose) {
+            console.log(`[${this.actorConfig.name}] using binary ${bin}`);
+        }
+
         this.configFile = this.actorConfig.generateCndConfigFile(
             this.ledgerConfig
         );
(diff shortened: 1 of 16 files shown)


api_tests/lib_sdk/cnd_instance.ts
```diff
@@ -56,6 +60,12 @@ export class CndInstance {
             ],
         });
 
+        if (global.verbose) {
```
D4nte: Should we have a function helper that does the conditional check and log passed message?

cnd/src/http_api/routes/index/mod.rs
```diff
@@ -19,26 +19,31 @@ pub struct InfoResource {
...
+    let listen_addresses: Vec<Multiaddr> = facade.listen_addresses().await.to_vec();
 
     Ok(warp::reply::json(&InfoResource {
         id: Http(id),
         listen_addresses,
     }))
 }
 
-pub fn get_info_siren<S: Network>(
+pub fn get_info_siren(
```
D4nte: I am not sure to see the added value of going async but still having a wrapping compat layer here for such simple functions.

D4nte:
no-merge label in case you want to review my comments first.

api_tests/lib_sdk/cnd_instance.ts
thomaseizinger: Happy to have that but not in this PR. I just pushed it because the work was already done as part of debugging :)

cnd/src/http_api/routes/index/mod.rs
thomaseizinger: These functions will eventually be async once we upgrade warp.

PR #1841, opened 2020-01-09
Migrate init_accepted_swap to tokio 0.2

thomaseizinger:
Tokio 0.2 no longer provides an `Executor` trait. Instead, we use the `tokio::task::spawn` method to spawn an async task in the background.
This is not as explicit as calling `.spawn()` on dependencies, but as long as we only do this from application code (read cnd) and not from potential library code, this should be fine.

### DIFF
--- a/cnd/src/http_api/routes/rfc003/handlers/action.rs
+++ b/cnd/src/http_api/routes/rfc003/handlers/action.rs
@@ -8,11 +8,11 @@ use crate::{
         route_factory::new_action_link,
         routes::rfc003::decline::{to_swap_decline_reason, DeclineBody},
     },
+    init_swap::init_accepted_swap,
     libp2p_comit_ext::ToHeader,
     network::Network,
     seed::DeriveSwapSeed,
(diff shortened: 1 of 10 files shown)


cnd/src/http_api/routes/rfc003/handlers/action.rs
```diff
@@ -8,11 +8,11 @@ use crate::{
         route_factory::new_action_link,
         routes::rfc003::decline::{to_swap_decline_reason, DeclineBody},
     },
+    init_swap::init_accepted_swap,
```
tcharding: Hey Thomas I've see you bring functions into scope a couple of times just recently, is this a result of that lint we discussed?  I was under the impression that idiomatic Rust only brings types and traits into scope but keeps one layer of module on function calls.  At least that is what the Rust book says (source link omitted due to laziness).

tcharding:
> as long as we only do this from application code (read cnd) and not from potential library code, this should be fine

How does this mesh with cnd as a library?

PR #1858, opened 2020-01-12
Run stale bot on push to master

bonomat:
We want this bot to run, however, cron jobs does not seem to work at the moment. 
I've reached out to github support already, but in the meantime, let's run it on every push to master.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,4 +1,6 @@
 on:
+  schedule:
+    - cron: '0 * * * *'
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


D4nte:
Just keep both instead, no?

.github/workflows/stale.yml
```diff
@@ -1,6 +1,7 @@
 on:
-  schedule:
-    - cron: '0 * * * *'
+  push:
```
bonomat:
```suggestion
  schedule:
    - cron: '0 * * * *'
  push:
```

bonomat:
> Just keep both instead, no?

Really good idea :D

PR #1860, opened 2020-01-13
Yield blocks for Ethereum matching_transaction using a generator

### LINKED ISSUE #1780: Remove dependency on `TaskExecutor` from `btsieve::ethereum::MatchingTransactions`
## Problem

The implementation of `MatchingTransactions` depends on the `tokio::executor::TaskExecutor` trait to spawn background tasks. This is sub-optimal for a number of reasons:

- It couples us to the tokio-executor
- It makes it harder to clean-up resources of finished swaps

## Goal

To have code that is not more complicated than actually necessary.

## Recommendation

The ideal API of the `MatchingTransactions` trait (or with #1519, the function) is: `async (connector, pattern) -> transaction`. For this, we should not need to spawn background tasks, everything that needs to be implemented can be done as part of the future that is returned.

I believe it should be somehow possible to use the current implementation design (based on message passing through channels) without spawning them away but either composing through combinators or polling them manually.

luckysori:
- Stop spawning tasks using `tokio`'s executor in `matching_transaction`.
- Simplify the code and make the yielding of blocks more specific to our needs.

There is no need to fetch new and old blocks concurrently. Given that fetching old blocks by hash is fast (since they're all available) that work can always be done on demand, halting the fetching of new blocks in the meantime.

---
Fixes #1780.

Added @tcharding to reviewers so that he can eventually tell me whether he thinks this is a step forward in terms of clarity.

### DIFF
--- a/cnd/src/btsieve/ethereum/mod.rs
+++ b/cnd/src/btsieve/ethereum/mod.rs
@@ -9,9 +9,13 @@ use crate::{
     btsieve::{BlockByHash, LatestBlock, ReceiptByHash},
     ethereum::{Transaction, TransactionAndReceipt, TransactionReceipt, H256, U256},
 };
-use futures_core::{compat::Future01CompatExt, future::join};
+use anyhow;
+use futures_core::compat::Future01CompatExt;
+use genawaiter::{
+    sync::{Co, Gen},
+    GeneratorState,
+};
 use std::{collections::HashSet, fmt::Debug};
-use tokio::sync::mpsc::{self, Receiver, Sender};
 
 type Hash = H256;
 type Block = crate::ethereum::Block<Transaction>;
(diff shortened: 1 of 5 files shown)


cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -26,321 +30,198 @@ where
...
     C: LatestBlock<Block = Option<Block>, Error = E>
         + BlockByHash<Block = Option<Block>, BlockHash = Hash, Error = E>
         + ReceiptByHash<Receipt = Option<TransactionReceipt>, TransactionHash = Hash, Error = E>
         + Clone,
-    E: Debug + Send + 'static,
+    E: std::error::Error + Debug + Send + Sync + 'static,
 {
-    let (mut fetch_block_by_hash_queue, mut next_hash) = next_block_channel;
-    let mut sent_blockhashes: HashSet<H256> = HashSet::new();
+    let mut seen_blockhashes: HashSet<Hash> = HashSet::new();
```
tcharding: We currently have `blockhash` and `block_hash` in the code base.  Please consider adding a patch to the end of this PR converting the whole code base to use either `block_hash` or `blockhash`.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -26,321 +30,198 @@ where
...
-                        let _ = fetch_block_by_hash_queue.send(blockhash).await;
-                    }
-                };
+        let blockhash = block
+            .hash
+            .ok_or_else(|| anyhow::anyhow!("Connector returned latest block with nullable hash"))?;
+        seen_blockhashes.insert(blockhash);
+
+        if let Some(timestamp) = reference_timestamp {
+            if seen_blockhashes.len() == 1 && block.timestamp > U256::from(timestamp) {
```
tcharding: Can we have a helper function please e.g., `block_is_younger_than_time(block, time)` (note parameter order matches order in function name).  Then future developers do not need to check to think if the `<` and `>` are correct.

luckysori:
> This PR is well hot, good work.

Thanks for the review! I'll address your suggestions in a separate PR.

Issue #1844, opened 2020-01-09
Minimal viable ledger configuration strategy


bonomat:
## Problem

At the moment, users of cnd can configure the connection to a `bitcoind` node and an ethereum node implementing the web3 endpoints.
In addition, the user can specify a `network` but it is not used for anything.
This is confusing and should be resolved.

## Goal

Define a minimally viable ledger configuration strategy for cnd.
The inner workings of cnd are designed to work against anything that implements `BlockByHash` and `LatestBlock`.
Choosing a connector is non-trivial though and depends a lot on the usecase of the user.
To not make cnd overly complex, we should reduce the configuration surface to a bare minimum at only keep the flexibility on the inside (and for a future library).
The goal of this spike is to figure out, what the best way of configuring cnd with two fullnodes is. Take into account:

- how/whether to specify, which blockchain network we are connected to
- the configuration surface should be minimal

We can assume that cnd will always run against full-nodes for now.

Note: Think about how to derive blockchain network from connection details (or the opposite), what should be the mandatory information (socket or blockchain network), what should be the default if something/anything is not specified.

D4nte:
What is the difference with https://github.com/comit-network/comit-rs/issues/1503 ?

thomaseizinger:
> What is the difference with #1503 ?

I've updated the issue description.

thomaseizinger:
@D4nte Seems like you had a typo in the issue number :)

D4nte:
Yep, sorry

bonomat:
@thomaseizinger : Can we remove the blocker (or turn it around) of https://github.com/comit-network/comit-rs/issues/1952? 
I think that we can do either independently or use the learnings from https://github.com/comit-network/comit-rs/issues/1952 in this spike.

thomaseizinger:
> @thomaseizinger : Can we remove the blocker (or turn it around) of #1952?
> I think that we can do either independently or use the learnings from #1952 in this spike.

See https://github.com/comit-network/comit-rs/issues/1952#issuecomment-579522866.

How do you see technical-debt being cleaned up if not at the time when we need to touch stuff that needs changing?

thomaseizinger:
I guess this could be downgraded from a spike to an actual ticket. I know it is not estimated now but it was created before we had the talk about "what should spikes be".

Thoughts?

bonomat:
## Assumptions

* It should be easy for a user to define what networks (blockchain networks) he wants to connect to
[...]

PR #1865, opened 2020-01-13
Remove BitcoinInfoConnector

thomaseizinger:
This is dead code, the implementation is trivial and can easily be recreated if we need it.


thomaseizinger:
> For me the bigger question is, if we want to feature the HTTP blockchain connectors or not. If not: remove it. If yes: Leave it and embrace it more... It is only "dead" because we don't use it...
> 
> @bonomat thoughts?

The thing is, we don't have a way of configuring it. There is https://github.com/comit-network/comit-rs/issues/1844 to address our current configuration problems. The easiest option for now is: don't offer HTTP connectors but instead, eventually ship a library and then people can pick and choose and assemble their own cnd.

bonomat:
>  It is only "dead" because we don't use it...

I'm pro deleting _dead/unused_ code. We can always revert it and get it back once it's used. The upside of heaving an unused file in our codebase is rather low imho.

> The thing is, we don't have a way of configuring it. There is #1844 to address our current configuration problems. The easiest option for now is: don't offer HTTP connectors but instead, eventually ship a library and then people can pick and choose and assemble their own cnd.

I agree with @thomaseizinger that we need a way to configure it and then can think of it again. 
In the long run I see it more as a configurable feature in the config file though than _pick and chose_ programmatically (kinda that's what we had right now :) )

PR #1866, opened 2020-01-14
Replace web3 with reqwest

### LINKED ISSUE #1718: Replace web3 with reqwest
## Problem

The Rust web3 library uses an outdated version of tokio and doesn't have a good API design (EventLoopHandle ...).
We only need a handful of RPC requests so we actually don't need to depend on it for cnd.

## Goal

Less nasty dependencies -> fewer headaches -> happier developers.

## Recommendation

Replace our usages of the `Web3` client with `reqwest` and simple structs that define the JSON rpc requests we need.

thomaseizinger:
Fixes #1718.

This is the MVP of #1718. It includes _some_ duplication. Can be removed in future iterations.

### DIFF
--- a/cnd/src/btsieve/ethereum/web3_connector.rs
+++ b/cnd/src/btsieve/ethereum/web3_connector.rs
@@ -1,32 +1,26 @@
 use crate::{
     btsieve::{BlockByHash, LatestBlock, ReceiptByHash},
-    ethereum::{
-        web3::{
-            self,
-            transports::{EventLoopHandle, Http},
-            Web3,
-        },
-        BlockId, BlockNumber,
-    },
+    ethereum::{BlockId, BlockNumber},
 };
+use anyhow::Context;
 use futures::Future;
-use reqwest::Url;
+use futures_core::{FutureExt, TryFutureExt};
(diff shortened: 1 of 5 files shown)


cnd/src/btsieve/ethereum/web3_connector.rs
```diff
@@ -37,15 +31,75 @@ impl LatestBlock for Web3Connector {
...
+                        code,
+                        message
+                    );
+                    return Ok(None);
+                }
+            };
+
+            log::trace!(
+                "Fetched block from web3: {}",
+                block.hash.expect("blocks to have a hash")
```
bonomat: ❓ What happens if there was not hash?

thomaseizinger: We panic and the tokio executor thread dies.

I am unaware of a scenario where blocks don't have a hash. In case this is actually legit, it is a serious bug and we should immediately stop IMO. Hence the decision to panic.

thomaseizinger: I expect this to get simpler once we completely remove web3 and have control over those data structures.

PR #1880, opened 2020-01-17
bitcoin asset

da-kami:
wrap `bitcoin::Amount` in `asset::Bitcoin`

### DIFF
--- a/cnd/src/asset/bitcoin.rs
+++ b/cnd/src/asset/bitcoin.rs
@@ -1,7 +1,4 @@
-use bitcoin::{
-    util::amount::{Denomination, ParseAmountError},
-    Amount,
-};
+use bitcoin::{util::amount::Denomination, Amount};
 use std::fmt;
 
 #[derive(Debug, Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Hash)]
@@ -14,17 +11,6 @@ impl Bitcoin {
 
     pub fn as_sat(self) -> u64 {
         Amount::as_sat(self.0)
-    }
-
-    pub fn from_btc(btc: f64) -> Result<Bitcoin, ParseAmountError> {
-        Amount::from_btc(btc).map(Bitcoin)
-    }
-
-    pub fn from_str_in(
-        quantity: &str,
-        denomination: Denomination,
-    ) -> Result<Bitcoin, ParseAmountError> {
-        Amount::from_str_in(quantity, denomination).map(Bitcoin)
     }
 }
 
@@ -48,7 +34,7 @@ mod tests {
     #[test]
     fn display_bitcoin() {
         assert_eq!(
-            (asset::Bitcoin::from_btc(9000.0)).unwrap().to_string(),
+            asset::Bitcoin::from_sat(900_000_000_000).to_string(),
             "9000.00000000 BTC"
         );
     }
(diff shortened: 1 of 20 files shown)


cnd/src/db/load_swaps.rs
```diff
@@ -51,11 +51,11 @@ struct BitcoinEthereumBitcoinEtherAcceptedSwap {
 }
 
 #[async_trait]
-impl LoadAcceptedSwap<Bitcoin, Ethereum, bitcoin::Amount, asset::Ether> for Sqlite {
+impl LoadAcceptedSwap<Bitcoin, Ethereum, asset::Bitcoin, asset::Ether> for Sqlite {
```
thomaseizinger: Things start to be beautiful :heart_eyes:

cnd/src/asset/bitcoin.rs
```diff
@@ -0,0 +1,55 @@
...
+impl Bitcoin {
+    pub fn from_sat(sat: u64) -> Bitcoin {
+        Bitcoin(Amount::from_sat(sat))
+    }
+
+    pub fn as_sat(self) -> u64 {
+        Amount::as_sat(self.0)
+    }
+
+    pub fn from_btc(btc: f64) -> Result<Bitcoin, ParseAmountError> {
```
thomaseizinger: We actually only ever use this method in tests, so you could just remove it and change the tests.

Using floats for money is dangerous. If we don't have to provide this method, that would be very good :D

cnd/src/asset/bitcoin.rs
```diff
@@ -0,0 +1,55 @@
...
+        Amount::from_btc(btc).map(Bitcoin)
+    }
+
+    pub fn from_str_in(
```
thomaseizinger: I see only two usages of this, both of them use `Satoshi` as the `Denomonation`.

I think we could get rid of this if we require the caller to parse their string as `u64` and then just use the `from_sat` method.

cnd/src/comit_api/mod.rs
```diff
@@ -95,8 +95,9 @@ impl FromHeader for AssetKind {
         Ok(match header.value::<String>()?.as_str() {
             "bitcoin" => {
                 let quantity = header.take_parameter::<String>("quantity")?;
-                let amount = bitcoin::Amount::from_str_in(quantity.as_str(), Denomination::Satoshi)
-                    .map_err(|e| serde_json::Error::custom(e.to_string()))?;
+                let quantity =
+                    u64::from_str(quantity.as_str()).map_err(serde_json::Error::custom)?;
+                let amount = asset::Bitcoin::from_sat(quantity);
```
tcharding: This would read cleaner, IMO, if it was just 
```
let quantity = Bitcoin::from_sat(quantity);
```

cnd/src/db/integration_tests.rs
```diff
@@ -1,5 +1,5 @@
 use crate::{
-    asset::{Erc20, Ether},
+    asset::{Bitcoin as BitcoinAsset, Erc20, Ether},
```
tcharding: Prefer `asset::Bitcoin` here instead of the alias.

cnd/src/comit_api/mod.rs
tcharding: I see this is uniform with the rest of the PR though :)

cnd/src/db/integration_tests.rs
da-kami: sorry, just saw that now, PR already merged. I remember I had some problems with the macro when using `asset::Bitcoin` at this particular position. If it matters please create a follow up PR :)

PR #1888, opened 2020-01-20
POC: Add tracing

tcharding:
This is a proof of concept to see if we can add tracing to cnd.  With
this applied (and logging disabled) the happy path btc_eth e2e test
produces the following output

```
	$ cat api_tests/log/cnd-alice.log
	Jan 20 13:01:30.749  INFO cnd::trace: Initialized tracing
	Jan 20 13:01:30.749  INFO cnd: Started Comit Network Daemon
	Jan 20 13:01:30.788  INFO cnd::db: SQLite database file: /tmp/tmp-299589vquB9TPrwWz/cnd.sqlite
	Jan 20 13:01:34.238 TRACE swap{id=5b744af1-d3aa-4b3b-a54a-f5e45799ae40}: cnd::http_api::routes::rfc003::handlers::post_swap: Initiating new request
	Jan 20 13:01:35.924 TRACE swap: Initiating accepted swap
	Jan 20 13:01:35.924 TRACE cnd::init_swap: Initialising accepted swap
	Jan 20 13:01:49.144  INFO swap{id=5b744af1-d3aa-4b3b-a54a-f5e45799ae40}: cnd::swap_protocols::rfc003::create_swap: Swap finished
```

It can be seen that the swap_id context is not always present.  This is
because of the lifetime of spans, specifically, a span declared within a
future exists only as long as the future exists (I'm not exactly sure
_when_ a future ceases to exist though).  This can be countered by
declaring a new span each time we create a new future - whether there is
a better way to do this I'm not sure.  At the very minimum getting
context everywhere we want it is going to require careful thought about
how a swap flows through the codebase.

We can, using this POC, switch over to tracing simply by doing
`s/log::/tracing::/g`.  Making use of context can then be done
progressively.

Converting the configuration option for controlling level is not touched
by this work, also interacting with library log output is also not
considered.  Neither of these should present much of a challenge though
from my understanding.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -464,6 +464,10 @@ dependencies = [
  "tokio 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-compat 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "toml 0.5.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tracing 0.1.12 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tracing-core 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tracing-futures 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tracing-subscriber 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "url 2.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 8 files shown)


thomaseizinger:
Awesome, nice work @tcharding !

> I'm not exactly sure when a future ceases to exist though

A future stops existing after it resolves (the executor is going to drop it after that, hence all the memory associated with it is freed).

This looks good to me!

The `Instrument` trait looks also quite interesting: https://tracing.rs/tracing_futures/trait.instrument

D4nte:
Proper implementation is tracked somewhere.

PR #1894, opened 2020-01-20
Use custom Infallible

tcharding:
`std::convert::Infallible` is meant to be used as error type (according to the docs), seeing it in the `Ok` variant of `Result` is confusing. We are using `Infallible` to make explicit that no `Ok` result is ever returned by this function.  Using the std infallible violates the principle of least surprise.  We can better achieve this behaviour by adding a custom `Infallible` type and using it.

### DIFF
--- a/cnd/src/btsieve/ethereum/mod.rs
+++ b/cnd/src/btsieve/ethereum/mod.rs
@@ -53,6 +53,10 @@ where
     }
 }
 
+#[allow(missing_debug_implementations)]
+#[derive(Copy, Clone)]
+pub enum Infallible {}
+
 /// This function uses the `connector` to find blocks relevant to a swap.
 ///
 /// It yields those blocks as part of the process.
@@ -60,7 +64,7 @@ async fn find_relevant_blocks<C>(
     mut connector: C,
     co: &Co<Block>,
     start_of_swap: Option<u32>,
-) -> anyhow::Result<std::convert::Infallible>
+) -> anyhow::Result<Infallible>
 where
     C: LatestBlock<Block = Option<Block>>
         + BlockByHash<Block = Option<Block>, BlockHash = Hash>


D4nte:
> `std::convert::Infallible` is meant to be used as error type (according to the docs), seeing it in the `Ok` variant of `Result` is confusing. We are using `Infallible` to make explicit that no `Ok` result is ever returned by this function. Using the std infallible violates the principle of least surprise. We can better achieve this behaviour by adding a custom `Infallible` type and using it.

🤯 why no `Ok` result is ever returned?

tcharding:
> > `std::convert::Infallible` is meant to be used as error type (according to the docs), seeing it in the `Ok` variant of `Result` is confusing. We are using `Infallible` to make explicit that no `Ok` result is ever returned by this function. Using the std infallible violates the principle of least surprise. We can better achieve this behaviour by adding a custom `Infallible` type and using it.
> 
> 🤯 why no `Ok` result is ever returned?

Its a loop that either finds and yeilds the block or fails with an error.

tcharding:
> `Infallible` is going to be a an alias of the never-type `!` as soon as it is stabilized. Our usage of it is perfectly fine.
> 
> Source: [doc.rust-lang.org/std/convert/enum.Infallible.html#future-compatibility](https://doc.rust-lang.org/std/convert/enum.Infallible.html#future-compatibility)

From the top of the link provided (above)

'The error type for errors that can never happen.'

This is _not_ an error, therefore it violates the principle of least surprise because when one reads the code one thinks 'that's funny, why is there an error type being returned in the `Ok` variant of this `Result`.  I actually would prefer to log this (or panic) since the current code can hit this silently.

thomaseizinger:
> > `Infallible` is going to be a an alias of the never-type `!` as soon as it is stabilized. Our usage of it is perfectly fine.
> > Source: [doc.rust-lang.org/std/convert/enum.Infallible.html#future-compatibility](https://doc.rust-lang.org/std/convert/enum.Infallible.html#future-compatibility)
> 
> From the top of the link provided (above)
> 
> 'The error type for errors that can never happen.'
> 
> This is _not_ an error, therefore it violates the principle of least surprise because when one reads the code one thinks 'that's funny, why is there an error type being returned in the `Ok` variant of this `Result`. I actually would prefer to log this (or panic) since the current code can hit this silently.

Yes it is an error type but I'd rather avoid defining our own replacement for a never type if we have a perfectly fine one in stdlib.

> I actually would prefer to log this (or panic) since the current code can hit this silently.

Hitting this code path would mean you were able to construct an instance of `Infallible`. This is not possible without unsafe code, hence the code compiling gives us the guarantee that this codepath cannot be hit. That is point of the never type but it is not yet stable unfortunately.

thomaseizinger:
> This is not an error, therefore it violates the principle of least surprise because when one reads the code 

To me, not using Infallible violates the principle of least surprise because I am thinking:

"That is funny, why do they still have their own never type even though std already has such a concept."

Not sure where you want to go with this discussion. To me, us having less code to write is preferable. If you don't like the name we can define an alias somewhere in the root of our crate.

thomaseizinger:
Does this make you more happy?

https://github.com/comit-network/comit-rs/pull/1916

tcharding:
> Does this make you more happy? #1916

That type alias made me smile :)

tcharding:
@thomaseizinger fixed this :)

PR #1895, opened 2020-01-20
Use correct word 'inexistent' -> 'non-existent'

tcharding:
Minor grammar fix in e2e tests.

### DIFF
--- a/api_tests/lib_sdk/actors/actor.ts
+++ b/api_tests/lib_sdk/actors/actor.ts
@@ -178,7 +178,7 @@ export class Actor {
 
     public async accept() {
         if (!this.swap) {
-            throw new Error("Cannot accept inexistent swap");
+            throw new Error("Cannot accept non-existent swap");
         }
 
         await this.swap.accept(Actor.defaultActionConfig);
@@ -186,7 +186,7 @@ export class Actor {
 
     public async fund() {
         if (!this.swap) {
-            throw new Error("Cannot fund inexistent swap");
+            throw new Error("Cannot fund non-existent swap");
         }
 
         const txid = await this.swap.fund(Actor.defaultActionConfig);
@@ -207,7 +207,7 @@ export class Actor {
 
     public async refund() {
         if (!this.swap) {
-            throw new Error("Cannot refund inexistent swap");
+            throw new Error("Cannot refund non-existent swap");
         }
 
         const entity = await this.swap.fetchDetails();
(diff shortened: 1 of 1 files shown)


bonomat:
@D4nte : you might like this one as you commented exactly that on my PR 😅

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -178,15 +178,15 @@ export class Actor {
 
     public async accept() {
         if (!this.swap) {
-            throw new Error("Cannot accept inexistent swap");
+            throw new Error("Cannot accept non-existent swap");
```
D4nte:
```suggestion
            throw new Error("Cannot accept nonexistent swap");
```

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -178,15 +178,15 @@ export class Actor {
...
         }
 
         await this.swap.accept(Actor.defaultActionConfig);
     }
 
     public async fund() {
         if (!this.swap) {
-            throw new Error("Cannot fund inexistent swap");
+            throw new Error("Cannot fund non-existent swap");
```
D4nte:
```suggestion
            throw new Error("Cannot fund nonexistent swap");
```

tcharding:
> For pedantry's sake, it looks like inexistent is a [perfectly valid word](https://www.merriam-webster.com/dictionary/inexistent)

Its a word but it has the wrong meaning here.

api_tests/lib_sdk/actors/actor.ts
tcharding: Cool I learned about hyphens, the web seems to think if it is a 'real' word (which nonexistent is) it does _not_ get hyphens and if it is made up (which I also learned is totally acceptable) it gets hyphens :)

tcharding:
LOL! I misspelled in the commit log the word that this was all about - no `git commit --amend` on github eh?

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -207,7 +207,7 @@ export class Actor {
 
     public async refund() {
         if (!this.swap) {
-            throw new Error("Cannot refund inexistent swap");
+            throw new Error("Cannot refund non-existent swap");
```
D4nte: @tcharding maybe stick to one spelling?

PR #1904, opened 2020-01-21
Better logs

thomaseizinger:
Another attempt of making e2e test failures easier to debug.

### DIFF
--- a/api_tests/harness.ts
+++ b/api_tests/harness.ts
@@ -19,6 +19,8 @@ const rlp = readline.createInterface({
     terminal: true,
 });
 
+const IS_CI = process.env.CI ?? Boolean(JSON.parse(process.env.CI));
+
 commander
     .option("--dump-logs", "Dump logs to stdout on failure")
     .option("--verbose", "Verbose output")
@@ -28,6 +30,10 @@ commander
         false
     )
     .parse(process.argv);
+
+if (IS_CI) {
+    commander.verbose = true;
+}
 
 // ************************ //
 // Setting global variables //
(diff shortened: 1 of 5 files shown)


api_tests/harness.ts
```diff
@@ -19,6 +19,8 @@ const rlp = readline.createInterface({
     terminal: true,
 });
 
+const IS_CI = process.env.CI ?? Boolean(JSON.parse(process.env.CI));
```
tcharding: It would be nice to be able to trigger this output locally also.

cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -56,16 +56,17 @@ pub async fn create_swap<D, A: ActorState>(
...
             // every event that is yielded is passed on
             GeneratorState::Yielded(event) => {
+                log::info!("swap {} yielded event {}", id, event);
                 dependencies.update::<A>(&id, event);
             }
             // the generator stopped executing, this means there are no more events that can be
             // watched.
             GeneratorState::Complete(Ok(_)) => {
-                log::info!("Swap {} finished", id);
+                log::info!("swap {} finished", id);
```
tcharding: Oh, I've been converting everything I touch to upercase first character, @luckysori and I discussed and didn't reach consensus.  We should get team consensus to save churn.  I don' care either way, as long as we do it just one way.

api_tests/harness.ts
thomaseizinger: You can, use `--verbose`.

cnd/src/swap_protocols/rfc003/create_swap.rs
thomaseizinger: I prefer lowercase but I think that is a topic we shouldn't spend too much time on :sweat_smile:

PR #1909, opened 2020-01-21
Clarify direction of time

tcharding:
Time is a funny thing; when we talk about time it is not always precisely clear
what we mean, special care should be taken so as not to be ambiguous.

The term 'past' is ambiguous, it could mean the time has 'passed' or that a
certain time is past another (worse still, this can be in either direction).

We can use the term 'predates' when discussing two times, this term is
unambiguous.  If time A predates time B it means exactly that event that
occurred at time A occurred chronologically before an event that occurred at
time B - there is no other interpretation.

Use 'predates' in the code and documentation instead of 'past'.


D4nte:
Is the `predates` stuff tested at all?

tcharding:
> Is the predates stuff tested at all?

We have integration tests that test this.  No unit tests.

Issue #1879, opened 2020-01-17
Make `LatestBlock`, `BlockByHash` and `ReceiptByHash` async-traits


thomaseizinger:
## Problem

To fully migrate cnd to async/await, we cannot use old futures anywhere. Those functions are returning old futures and thus need to be migrated.

## Goal

Use async/await as much as possible.

## Recommendation

- Use the `async_trait` proc-macro to make the functions of this trait `async`.
- Remove any kind of compatibility layer in their implementations and on the call-sites.

thomaseizinger:
This is not necessarily blocked by https://github.com/comit-network/comit-rs/issues/1878 but doing https://github.com/comit-network/comit-rs/issues/1878 first makes things easier in the migration process.

PR #1912, opened 2020-01-22
add cache for Bitcoin and Ethereum connectors

da-kami:
Add cacheing layer to btsieve
- Add block cache for Bitcoin
- Add block and transaction-receipt cache for Ethereum

Note on the implementation:
We decided to create a separate cache for Bitcoin and Ethereum. The first version was fully generic (including the Key+Value from the cache), but that made the code difficult to read/change and actually added problems for `LatestBlock` as the hash extraction of a block differs on Ethereum and Bitcoin.

### DIFF
--- a/cnd/src/btsieve/bitcoin/bitcoin_cache.rs
+++ b/cnd/src/btsieve/bitcoin/bitcoin_cache.rs
@@ -1,0 +1,94 @@
+use crate::btsieve::{BlockByHash, LatestBlock};
+use bitcoin::{hashes::sha256d, util::hash::BitcoinHash};
+use fmt::Debug;
+use futures::Future;
+use futures_core::{
+    compat::Future01CompatExt,
+    future::{FutureExt, TryFutureExt},
+};
+use std::{fmt, fmt::Formatter, sync::Arc};
+use tokio::sync::Mutex;
+
+pub struct BitcoinCache<T> {
+    pub inner: T,
+    pub cache: Arc<Mutex<lru::LruCache<sha256d::Hash, bitcoin::Block>>>,
+}
+
+impl<T> Clone for BitcoinCache<T>
+where
+    T: Clone,
+{
+    fn clone(&self) -> Self {
+        BitcoinCache {
+            inner: self.inner.clone(),
+            cache: self.cache.clone(),
+        }
+    }
+}
+
+impl<T> Debug for BitcoinCache<T>
+where
+    T: Debug,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), fmt::Error> {
+        self.inner.fmt(f)
+    }
+}
+
+impl<T> LatestBlock for BitcoinCache<T>
+where
+    T: LatestBlock<Block = bitcoin::Block, BlockHash = sha256d::Hash> + Clone,
+{
+    type Block = bitcoin::Block;
+    type BlockHash = sha256d::Hash;
+
+    fn latest_block(
+        &mut self,
+    ) -> Box<dyn Future<Item = Self::Block, Error = anyhow::Error> + Send + 'static> {
+        let cache = self.cache.clone();
+        let mut inner = self.inner.clone();
+
+        let future = async move {
+            let block = inner.latest_block().compat().await?;
+            let mut guard = cache.lock().await;
+            guard.put(block.bitcoin_hash(), block.clone());
+            Ok(block)
+        }
+        .boxed()
+        .compat();
+
+        Box::new(future)
+    }
+}
+
+impl<T> BlockByHash for BitcoinCache<T>
+where
+    T: BlockByHash<Block = bitcoin::Block, BlockHash = sha256d::Hash> + Clone,
+{
+    type Block = bitcoin::Block;
+    type BlockHash = sha256d::Hash;
+
+    fn block_by_hash(
+        &self,
+        block_hash: Self::BlockHash,
+    ) -> Box<dyn Future<Item = Self::Block, Error = anyhow::Error> + Send + 'static> {
+        let cache = self.cache.clone();
+        let inner = self.inner.clone();
+
+        let future = async move {
+            match cache.lock().await.get(&block_hash) {
+                Some(block) => Ok(block.clone()),
+                None => {
+                    let block = inner.block_by_hash(block_hash.clone()).compat().await?;
+                    let mut guard = cache.lock().await;
+                    guard.put(block_hash, block.clone());
+                    Ok(block)
+                }
+            }
+        }
+        .boxed()
+        .compat();
+
+        Box::new(future)
+    }
+}
--- a/cnd/src/btsieve/bitcoin/bitcoin_cache.rs
+++ b/cnd/src/btsieve/bitcoin/bitcoin_cache.rs
@@ -1,16 +1,19 @@
 use crate::btsieve::{BlockByHash, LatestBlock};
 use bitcoin::{hashes::sha256d, util::hash::BitcoinHash};
-use fmt::Debug;
+use derivative::Derivative;
 use futures::Future;
 use futures_core::{
     compat::Future01CompatExt,
     future::{FutureExt, TryFutureExt},
 };
-use std::{fmt, fmt::Formatter, sync::Arc};
+use std::sync::Arc;
 use tokio::sync::Mutex;
 
+#[derive(Derivative)]
+#[derivative(Debug)]
 pub struct BitcoinCache<T> {
     pub inner: T,
+    #[derivative(Debug = "ignore")]
     pub cache: Arc<Mutex<lru::LruCache<sha256d::Hash, bitcoin::Block>>>,
 }
 
@@ -23,15 +26,6 @@ where
             inner: self.inner.clone(),
             cache: self.cache.clone(),
         }
-    }
-}
-
-impl<T> Debug for BitcoinCache<T>
-where
-    T: Debug,
-{
-    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), fmt::Error> {
-        self.inner.fmt(f)
     }
 }
 
--- a/cnd/src/btsieve/bitcoin/cache.rs
+++ b/cnd/src/btsieve/bitcoin/cache.rs
@@ -24,10 +24,8 @@ pub struct Cache<C> {
 
 impl<T> Cache<T> {
     pub fn new(connector: T, capacity: usize) -> Cache<T> {
-        Cache {
-            connector,
-            cache: Arc::new(Mutex::new(LruCache::new(capacity))),
-        }
+        let cache = Arc::new(Mutex::new(LruCache::new(capacity)));
+        Cache { connector, cache }
     }
 }
 
--- a/cnd/src/btsieve/bitcoin/cache.rs
+++ b/cnd/src/btsieve/bitcoin/cache.rs
@@ -15,13 +15,16 @@ use tokio::sync::Mutex;
 pub struct Cache<C> {
     pub connector: C,
     #[derivative(Debug = "ignore")]
-    pub cache: Arc<Mutex<LruCache<Hash, Block>>>,
+    pub block_cache: Arc<Mutex<LruCache<Hash, Block>>>,
 }
 
 impl<C> Cache<C> {
     pub fn new(connector: C, capacity: usize) -> Cache<C> {
-        let cache = Arc::new(Mutex::new(LruCache::new(capacity)));
-        Cache { connector, cache }
+        let block_cache = Arc::new(Mutex::new(LruCache::new(capacity)));
+        Cache {
+            connector,
+            block_cache,
+        }
     }
 }
 
@@ -35,7 +38,7 @@ where
     fn latest_block(
         &mut self,
     ) -> Box<dyn Future<Item = Self::Block, Error = anyhow::Error> + Send + 'static> {
-        let cache = Arc::clone(&self.cache);
+        let cache = Arc::clone(&self.block_cache);
         let mut connector = self.connector.clone();
 
         let future = async move {
@@ -56,43 +59,4 @@ where
     }
 }
 
-impl<C> BlockByHash for Cache<C>
-where
-    C: BlockByHash<Block = Block, BlockHash = Hash> + Clone,
-{
-    type Block = Block;
-    type BlockHash = Hash;
-
-    fn block_by_hash(
-        &self,
-        block_hash: Self::BlockHash,
-    ) -> Box<dyn Future<Item = Self::Block, Error = anyhow::Error> + Send + 'static> {
-        let connector = self.connector.clone();
-        let cache = Arc::clone(&self.cache);
-        Box::new(Box::pin(block_by_hash(connector, cache, block_hash)).compat())
-    }
-}
-
-async fn block_by_hash<C>(
-    connector: C,
-    cache: Arc<Mutex<LruCache<Hash, Block>>>,
-    block_hash: Hash,
-) -> anyhow::Result<Block>
-where
-    C: BlockByHash<Block = Block, BlockHash = Hash> + Clone,
-{
-    if let Some(block) = cache.lock().await.get(&block_hash) {
-        log::trace!("Found block in cache: {:x}", block_hash);
-        return Ok(block.clone());
-    }
-
-    let block = connector.block_by_hash(block_hash.clone()).compat().await?;
-    log::trace!("Fetched block from connector: {:x}", block_hash);
-
-    // We dropped the lock so at this stage the block may have been inserted by
-    // another thread, no worries, inserting the same block twice does not hurt.
-    let mut guard = cache.lock().await;
-    guard.put(block_hash, block.clone());
-
-    Ok(block)
-}
+impl_block_by_hash!();
--- a/cnd/src/btsieve/bitcoin/cache.rs
+++ b/cnd/src/btsieve/bitcoin/cache.rs
@@ -1,51 +1,70 @@
 use crate::btsieve::{BlockByHash, LatestBlock};
-use bitcoin::{hashes::sha256d, util::hash::BitcoinHash};
+use bitcoin::{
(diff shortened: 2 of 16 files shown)


cnd/src/main.rs
```diff
@@ -45,10 +49,18 @@ fn main() -> anyhow::Result<()> {
 
     let bitcoin_connector = {
         let config::Bitcoin { node_url, network } = settings.clone().bitcoin;
-        BitcoindConnector::new(node_url, network)?
+
+        BitcoinCache {
+            inner: BitcoindConnector::new(node_url, network)?,
+            cache: Arc::new(Mutex::new(lru::LruCache::new(100))),
```
thomaseizinger: You could provide a constructor `BitcoinCache::new(inner, capacity)`. That would make this more ergonomic because it is always the same anyway :)

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
```diff
@@ -0,0 +1,94 @@
...
+};
+use std::{fmt, fmt::Formatter, sync::Arc};
+use tokio::sync::Mutex;
+
+pub struct BitcoinCache<T> {
+    pub inner: T,
+    pub cache: Arc<Mutex<lru::LruCache<sha256d::Hash, bitcoin::Block>>>,
+}
+
+impl<T> Clone for BitcoinCache<T>
```
tcharding: I'm pretty sure we can just derive Clone for this.  @thomaseizinger any reason not too?  One reason may be to explicitly use
```
cache: Arc::clone(&self.cache),
```
But I do not see any real value in doing so.

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
```diff
@@ -0,0 +1,94 @@
...
+where
+    T: LatestBlock<Block = bitcoin::Block, BlockHash = sha256d::Hash> + Clone,
+{
+    type Block = bitcoin::Block;
+    type BlockHash = sha256d::Hash;
+
+    fn latest_block(
+        &mut self,
+    ) -> Box<dyn Future<Item = Self::Block, Error = anyhow::Error> + Send + 'static> {
+        let cache = self.cache.clone();
```
tcharding: Please consider using, as per Rust convention
```
let cache = Arc::clone(&self.cache)
```
This makes it explicit that we are increasing a refcount and not doing an expensive Clone operation.

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
```diff
@@ -0,0 +1,94 @@
...
+        let inner = self.inner.clone();
+
+        let future = async move {
+            match cache.lock().await.get(&block_hash) {
+                Some(block) => Ok(block.clone()),
+                None => {
+                    let block = inner.block_by_hash(block_hash.clone()).compat().await?;
+                    let mut guard = cache.lock().await;
+                    guard.put(block_hash, block.clone());
```
tcharding: We take the lock twice here, perhaps it could be written like:
```
diff --git a/cnd/src/btsieve/bitcoin/bitcoin_cache.rs b/cnd/src/btsieve/bitcoin/bitcoin_cache.rs
index 3f82bda4..ea9d6f7e 100644
--- a/cnd/src/btsieve/bitcoin/bitcoin_cache.rs
+++ b/cnd/src/btsieve/bitcoin/bitcoin_cache.rs
@@ -70,12 +70,12 @@ where
         let inner = self.inner.clone();
 
         let future = async move {
-            match cache.lock().await.get(&block_hash) {
+            let mut cache = cache.lock().await;
+            match cache.get(&block_hash) {
                 Some(block) => Ok(block.clone()),
                 None => {
                     let block = inner.block_by_hash(block_hash.clone()).compat().await?;
-                    let mut guard = cache.lock().await;
-                    guard.put(block_hash, block.clone());
+                    cache.put(block_hash, block.clone());
                     Ok(block)
                 }
             }
```

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
```diff
@@ -0,0 +1,94 @@
...
+            Ok(block)
+        }
+        .boxed()
+        .compat();
+
+        Box::new(future)
+    }
+}
+
+impl<T> BlockByHash for BitcoinCache<T>
```
tcharding: We may want to convert these traits to new futures as a first step to this, it would save us converting all the new code.  @thomaseizinger is there any reason these have _not_ been done before (before @da-kami goes on a wild goose chase)?

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn latest_block(
+        &mut self,
+    ) -> Box<dyn Future<Item = Self::Block, Error = anyhow::Error> + Send + 'static> {
+        let cache = self.cache.clone();
+        let mut inner = self.inner.clone();
+
+        let future = async move {
+            let block = inner.latest_block().compat().await?;
+            let mut guard = cache.lock().await;
+            guard.put(block.bitcoin_hash(), block.clone());
```
tcharding: Likely we should check for existence of the key before calling `put` here otherwise we clone the block and update the cache every call.  From docs for `LruCache`:
```
    /// Puts a key-value pair into cache. If the key already exists in the cache, then it updates
    /// the key's value and returns the old value. Otherwise, `None` is returned.
```

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
tcharding: In non-diff form
```
        let future = async move {
            let mut cache = cache.lock().await;
            match cache.get(&block_hash) {
                Some(block) => Ok(block.clone()),
                None => {
                    let block = inner.block_by_hash(block_hash.clone()).compat().await?;
                    cache.put(block_hash, block.clone());
                    Ok(block)
                }
            }
        }
```

cnd/src/main.rs
tcharding: Can we have a const as well please instead of a magic number.  Eventually we will likely want to expose this value as a config option.

cnd/src/btsieve/ethereum/ethereum_cache.rs
```diff
@@ -0,0 +1,142 @@
...
+        let future = async move {
+            let block = inner.latest_block().compat().await?;
+
+            match block.clone() {
+                Some(block) => {
+                    let mut guard = cache.lock().await;
+                    guard.put(block.hash.expect("no blocks without hash"), Some(block));
+                }
+                None => (),
+            }
```
tcharding: Clippy suggests
```
            if let Some(block) = block.clone() {
                let mut guard = cache.lock().await;
                guard.put(block.hash.expect("no blocks without hash"), Some(block));
            };
```

tcharding: Also, should guard the `put` with a check for the blocks existence (same is suggested for bitcoin cache).

cnd/src/btsieve/ethereum/ethereum_cache.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    fn block_by_hash(
+        &self,
+        block_hash: Self::BlockHash,
+    ) -> Box<dyn Future<Item = Self::Block, Error = anyhow::Error> + Send + 'static> {
+        let cache = self.block_cache.clone();
+        let inner = self.inner.clone();
+
+        let future = async move {
+            match cache.lock().await.get(&block_hash) {
```
tcharding: Same locking suggestions as for bitcoin.

cnd/src/btsieve/ethereum/ethereum_cache.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    fn receipt_by_hash(
+        &self,
+        transaction_hash: Self::TransactionHash,
+    ) -> Box<dyn Future<Item = Self::Receipt, Error = anyhow::Error> + Send + 'static> {
+        let cache = self.receipt_cache.clone();
+        let inner = self.inner.clone();
+
+        let future = async move {
+            match cache.lock().await.get(&transaction_hash) {
```
tcharding: And here.

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
```diff
@@ -0,0 +1,94 @@
...
+use fmt::Debug;
+use futures::Future;
+use futures_core::{
+    compat::Future01CompatExt,
+    future::{FutureExt, TryFutureExt},
+};
+use std::{fmt, fmt::Formatter, sync::Arc};
+use tokio::sync::Mutex;
+
+pub struct BitcoinCache<T> {
```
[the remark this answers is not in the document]
da-kami: Yes and no. Yes, in the definition of the struct it is the connector that is concerned; but we actually implement it for the specific traits identifying a connector, i.e. `LastBlock`, `BlockByHash` and `ReceiptByHash` - `C` does not really make sense there, but is still the type passed into the cache. I feel the `T` is fine there as `Type`.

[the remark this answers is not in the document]
da-kami: I have a different feeling about this, but I tend to name things more specific. I don't like it when files with the same name pop up in a search because you always have to imply context then by looking at other things. If you insist I can change it, but I feel it's not too important.

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub inner: T,
```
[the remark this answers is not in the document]
da-kami: I don't mind too much - I would have been more abstract for the same reason as the `C` above, but changed it to your suggestion for now.

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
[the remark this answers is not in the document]
thomaseizinger: > I have a different feeling about this, but I tend to name things more specific. I don't like it when files with the same name pop up in a search because you always have to imply context then by looking at other things. If you insist I can change it, but I feel it's not too important.

I do agree with Tobin here, it is more common in Rust to add clarity to an identifier by using it through a module like `bitcoin::Cache<BitcoindConnector>`.

The search thing is slightly annoying but at least CLion does a good job of adding the context, so personally I feel like it is not that bad.

Having more concise names makes the code overall more readable because at any point in the source-code you can choose, how much context you want to add. If the name already includes the context, you'd have to do aliasing which is strictly worse IMO.

cnd/src/swap_protocols/facade.rs
```diff
@@ -183,14 +186,46 @@ impl HtlcEvents<Bitcoin, asset::Bitcoin> for Facade {
 }
 
 #[async_trait::async_trait]
-impl<A> HtlcEvents<Ethereum, A> for Facade
-where
-    A: Asset + Send + Sync + 'static,
-    Web3Connector: HtlcEvents<Ethereum, A>,
-{
+impl HtlcEvents<Ethereum, asset::Ether> for Facade {
```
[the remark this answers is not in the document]
thomaseizinger: That was from a time when we tried to do it more generically. It is actually no longer needed now.

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
thomaseizinger: Now that we are back to one type parameter, it can be derived yes.

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
thomaseizinger: If we only take the lock once, we have it open for too long, including one async operation which might timeout, blocking all other requests.

It was on purpose to only lock for as short as possible :)

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
thomaseizinger: See https://github.com/comit-network/comit-rs/issues/1879 plus dependency chain.

da-kami:
I worked in the changes, rebased on master and ran the e2e tests again.

The e2e tests still fail (we seem to not process the blocks properly). This seems to be a problem of how the cache and the connector work together. I don't really have time to take a look at the moment because I am outlining the new homepage. @tcharding would you maybe have time to take a look tomorrow?

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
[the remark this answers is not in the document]
tcharding: Thanks for taking my suggestion, very decent of you.  Implementing it for specific traits and using `C` is nice IMO, they are 'connector' traits after all.

cnd/src/btsieve/bitcoin/bitcoin_cache.rs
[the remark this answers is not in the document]
tcharding: Yeah, the lipp2p codebase uses inner all the time, I find it un-descriptive unless the outer thing is just a wrapper.  This is not really a wrapper because it has two fields with two distinct purposes.  Thanks for changing it.

tcharding:
> I worked in the changes, rebased on master and ran the e2e tests again.
> 
> The e2e tests still fail (we seem to not process the blocks properly). This seems to be a problem of how the cache and the connector work together. I don't really have time to take a look at the moment because I am outlining the new homepage. @tcharding would you maybe have time to take a look tomorrow?

Hey mate, looks like you didn't push your changes.  I'm keen to debug this.

tcharding:
Nice macro `block_cache`!

tcharding:
@da-kami please review if you want to (cannot add you because you are assigned).

cnd/src/btsieve/bitcoin/cache.rs
```diff
@@ -0,0 +1,62 @@
...
+        Cache {
+            connector,
+            block_cache,
+        }
+    }
+}
+
+impl<C> LatestBlock for Cache<C>
+where
+    C: LatestBlock<Block = Block, BlockHash = Hash> + Clone,
```
thomaseizinger: Personally, I don't like these imports :/

It feels like pointless information to me but we can't remove it. I'd opt for `bitcoin::Block` here so that it at least adds some value (you know that this is implementation is bitcoin without having to check the imports or the file name)

Yes we are in the _context_ of bitcoin here but I find that mental jump annoying.

Thoughts?

cnd/src/btsieve/block_by_hash.rs
```diff
@@ -0,0 +1,44 @@
+macro_rules! impl_block_by_hash {
+    () => {
+        impl<C> BlockByHash for Cache<C>
```
thomaseizinger: Macros should never rely on their context to import certain things.

I would suggest to make:

- `Block` type
- `BlockHash` type
- `Cache` type

arguments of the macro and refer to all other traits through full qualified names.

cnd/src/btsieve/block_by_hash.rs
```diff
@@ -0,0 +1,44 @@
...
+            ) -> Box<dyn Future<Item = Self::Block, Error = anyhow::Error> + Send + 'static> {
+                let connector = self.connector.clone();
+                let cache = Arc::clone(&self.block_cache);
+                Box::new(Box::pin(block_by_hash(connector, cache, block_hash)).compat())
+            }
+        }
+
+        async fn block_by_hash<C>(
+            connector: C,
+            cache: Arc<Mutex<LruCache<Hash, Block>>>,
```
thomaseizinger: These types should be fully qualified here to not have ambiguous imports that lead to weird compile errors.

cnd/src/main.rs
```diff
@@ -43,12 +43,22 @@ fn main() -> anyhow::Result<()> {
 
     let mut runtime = Runtime::new()?;
 
+    const BITCOIN_BLOCK_CACHE_CAPACITY: usize = 144;
```
thomaseizinger: Since these are consts anyway, should we just inline them inside the constructor?

cnd/src/btsieve/ethereum/cache.rs
```diff
@@ -0,0 +1,126 @@
...
+use futures::Future;
+use futures_core::{
+    compat::Future01CompatExt,
+    future::{FutureExt, TryFutureExt},
+};
+use lru::LruCache;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+// This makes it a bit obscure that we have an option, the compile will point it
```
thomaseizinger: With my other suggestions, this comment goes away :)

tcharding:
> Would have been nice to see that without all the other refactoring being in place.

I just refactored it and it worked.  My guess is it was the locking but that's just a guess.

cnd/src/btsieve/bitcoin/cache.rs
tcharding: Can't use the macro if we don't shorted the types.  I agree the mental jump is not nice, I feel all this type info is just crap to keep the compiler happy anyways so it doesn't bother me to shorten it.  Main reason is the macro though to reduce code duplication.

cnd/src/btsieve/block_by_hash.rs
tcharding: oh yeah?  At first I was going to add Block and Hash then I didn't.  Perhaps I was treating Rust macros too much like C macros?

cnd/src/btsieve/block_by_hash.rs
tcharding: I'm getting the feeling you don't like how terse I made everything :) I added no-mergify back in.  Lets chat about it in person, easier to explain the nuances then.

cnd/src/main.rs
tcharding: I personally am religious about removing magic numbers (you might have seen me request them on every single review I've done ever :)  Its not a big enough deal for me to ACK stuff but I always ask for them.

PR #1910, opened 2020-01-22
Remove web3 from Erc20 definition + remove U256 from db module

D4nte:
Resolves #1844

PR #1913, opened 2020-01-22
Remove U256 from db types

D4nte:
Remove the use of `web3::U256` as inner for db types.


D4nte:
Everything is in #1910 now

PR #1914, opened 2020-01-22
Prefer usage of standard traits

D4nte:
Not sure I made it better.
For sure we need the tests in case rust-bitcoin change their serialisation behaviour.

### DIFF
--- a/cnd/src/bitcoin.rs
+++ b/cnd/src/bitcoin.rs
@@ -11,32 +11,46 @@ use serde::{
     de::{self, Visitor},
     Deserialize, Deserializer, Serialize, Serializer,
 };
-use std::fmt;
+use std::{fmt, str::FromStr};
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct PublicKey(bitcoin::PublicKey);
 
 impl PublicKey {
-    pub fn new(key: secp256k1::PublicKey) -> Self {
+    pub fn from_secret_key<C: secp256k1::Signing>(
+        secp: &secp256k1::Secp256k1<C>,
+        secret_key: &secp256k1::SecretKey,
+    ) -> Self {
+        secp256k1::PublicKey::from_secret_key(secp, secret_key).into()
+    }
+}
+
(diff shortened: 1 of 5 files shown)


cnd/src/bitcoin.rs
```diff
@@ -67,10 +81,45 @@ impl<'de> Deserialize<'de> for PublicKey {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::str::FromStr;
+
+    #[test]
+    fn serialize_bitcoin_identity() {
+        let secp_pubkey = secp256k1::PublicKey::from_str(
+            "02c2a8efce029526d364c2cf39d89e3cdda05e5df7b2cbfc098b4e3d02b70b5275",
```
tcharding: Perhaps a const instead of these identical strings.

cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -45,14 +45,12 @@ pub async fn create_swap<D, A: ActorState>(
...
         let dependencies = dependencies.clone();
-        |co| {
-            async move {
-                future::try_join(
-                    watch_alpha_ledger(&dependencies, &co, swap.alpha_htlc_params(), at),
-                    watch_beta_ledger(&dependencies, &co, swap.beta_htlc_params(), at),
-                )
-                .await
-            }
+        |co| async move {
```
tcharding: LOL, here it is again (poke @thomaseizinger).  That's three PRs this morning that have included this (and one I noticed of my own the other day).  For reference, @D4nte, this keeps being changed back and forwards from 
```
 |co| { async move {
```
to
```
|co| async move {
```

D4nte: I don't think it keeps being changed back and forth. all PRs I see brings in the new format `|co| async move {` because it was not yet in master.
Now that I rebased this PR you can see that my PR does not bring it back anymore.

cnd/src/bitcoin.rs
D4nte: Well some are in JSON format and some are not. Which means it's just duplicated once. Let's refactor on the 2rd occurrence :)

PR #1917, opened 2020-01-22
Use NaiveDateTime for Predates trait

thomaseizinger:
This allows for a more ergonomic usage because it avoids conversion at the callsite.


cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -45,14 +45,12 @@ pub async fn create_swap<D, A: ActorState>(
...
         let dependencies = dependencies.clone();
-        |co| {
-            async move {
-                future::try_join(
-                    watch_alpha_ledger(&dependencies, &co, swap.alpha_htlc_params(), at),
-                    watch_beta_ledger(&dependencies, &co, swap.beta_htlc_params(), at),
-                )
-                .await
-            }
+        |co| async move {
```
tcharding: This block keeps getting re-formatted in various PRs.  Seems some tools are arguing about whether the `async move` needs to be inside another block or not.

PR #1918, opened 2020-01-22
Remove public cndIsRunning method

thomaseizinger:
We only use this functionality internally, hence we can inline the method at the usage site.


tcharding:
Added @luckysori to reviewers because this PR overrides you suggestion on using a GET request to check if cnd is up.  I'm not fussed either way so I ack'ed this one since its done already :)  Big ups @thomaseizinger for doing my work.

tcharding:
Looks like @luckysori was right and the resume swap PR got through CI by chance.

PR #1919, opened 2020-01-22
Use the process field instead of the config file check if cnd is running

thomaseizinger:
Using the process seems more natural because the process is dead anyway after it has been killed and hence, the reference to the object is worthless.

### DIFF
--- a/api_tests/lib_sdk/cnd_instance.ts
+++ b/api_tests/lib_sdk/cnd_instance.ts
@@ -80,10 +80,10 @@ export class CndInstance {
 
     public stop() {
         this.process.kill("SIGINT");
-        this.configFile = null; // Please see function is_running() below before changing this.
+        this.process = null;
     }
 
     public isRunning() {
-        return !this.configFile == null;
+        return this.process != null;
     }
 }


api_tests/lib_sdk/cnd_instance.ts
```diff
@@ -80,10 +80,10 @@ export class CndInstance {
 
     public stop() {
         this.process.kill("SIGINT");
-        this.configFile = null; // Please see function is_running() below before changing this.
+        this.process = null;
     }
 
     public isRunning() {
-        return !this.configFile == null;
+        return !this.process == null;
```
tcharding: Is there some Typescript reason why this is not
```
return this.process != null;
```
??

tcharding: On looking up the order of precedence for Javascript this does not do what you want it to do I believe?  `!` has higher precedence that `==` so we need either
```
!(this.process == null)
```
or (preferably)
```
this.process != null
```

thomaseizinger:
Nah you are totally right! I wasn't thinking when I made the changes and just changed the field.

It should definitely be `!=` !

tcharding:
Added @luckysori to reviewers because this PR overrides you suggestion on using a GET request to check if cnd is up. I'm not fussed either way so I ack'ed this one since its done already :) Big ups @thomaseizinger for doing my work.

(moved from incorrect posting on other PR.)

thomaseizinger:
@tcharding Update the PR with the better way of comparing the process field :)

PR #1921, opened 2020-01-22
Resume swap follow up

tcharding:
Follow up patches on top of the work that merged yesterday for resuming ongoing swaps.

Patch 1 increases timeout in 'resume' tests from 1 minute to 2.  This may not be necessary but does not hurt, AFAICT.

The other three patches are from review suggestions by @luckysori, docs and integration test fixes.  Thanks Lucky.


api_tests/e2e/rfc003/btc_eth/resume.ts
```diff
@@ -109,4 +109,4 @@ setTimeout(function() {
     );
 
     run();
-}, 60000);
+}, 120_000);
```
thomaseizinger: NO!

This has to be 0 :see_no_evil: 

This has nothing to do with the timeout of the test but is a _delay_ up until the tests will actually run. We only need this because we run mocha with a special flag where we need to call `run` at the end. This allows us to do async work before running the tests.

Sorry about this, you cannot possible know this :sweat_smile: 

This can be removed once we migrated all of our tests.

tcharding: All good, glad you caught it :)

tcharding:
It would be nice if we had a way of turning 'request changes' into an approval if the person did, as you have here, 'ack with xyz'.  Once xyz is done its a nuisance for the reviewer to have to hit approved.

thomaseizinger:
> It would be nice if we had a way of turning 'request changes' into an approval if the person did, as you have here, 'ack with xyz'. Once xyz is done its a nuisance for the reviewer to have to hit approved.

Just dismiss my review and delgate to bors ;)

image

PR #1922, opened 2020-01-23
Forward all non-generic traits in Facade using the ambassador crate

thomaseizinger:
Found this crate the other day. Support for generic traits is coming soon, so hopefully we will be able to remove all of this boilerplate eventually :)

What do you guys think of this?
Cool or too much magic?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,6 +45,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ambassador"
+version = "0.2.1"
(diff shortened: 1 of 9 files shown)


D4nte:
Looks ok to me.

thomaseizinger:
> Looks ok to me.
> should we cargo geiger everytime we bring a new dep?

Would make it part of the build process then to cover updates to crates as-well.

tcharding:
The docs are light on for the `ambassador` crate.  The code gave me a nice little nudge, there is so much more Rust to learn.

PR #1924, opened 2020-01-23
Macro in load swaps

D4nte:
Trying to slightly alleviate the work for when I am going to multiple the number of variants by 3 (from Bitcoin to bitcoin::Mainnet, bitcoin::Regtest, bitcoin::Test.

### DIFF
--- a/cnd/src/db/load_swaps.rs
+++ b/cnd/src/db/load_swaps.rs
@@ -29,6 +29,48 @@ pub trait LoadAcceptedSwap<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> {
     ) -> anyhow::Result<AcceptedSwap<AL, BL, AA, BA>>;
 }
 
+macro_rules! impl_load_accepted_swap {
+    ($al:tt, $bl:tt, $aa:tt, $ba:tt, $accept_messages:tt, $request_messages:tt,
+    $record:tt, $select:tt) => {
+        paste::item! {
+            // Parenthesis are needed because we use `tt`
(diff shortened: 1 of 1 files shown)


D4nte:
It is quite interesting that I only saved 10 lines at this stage :)

D4nte:
> Cool!
> 
> There is gotta be a way of making this less verbose :D

The new `From` Implementations could use some more `From` implementations for the attributes of `Reqwest` I think.

cnd/src/db/load_swaps.rs
```diff
@@ -29,6 +29,48 @@ pub trait LoadAcceptedSwap<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> {
     ) -> anyhow::Result<AcceptedSwap<AL, BL, AA, BA>>;
 }
 
+macro_rules! impl_load_accepted_swap {
+    ($al:tt, $bl:tt, $aa:tt, $ba:tt, $accept_messages:tt, $request_messages:tt,
+    $record:tt, $select:tt) => {
+        paste::item! {
```
tcharding: We don't need this.  Its only for concatinating identifiers.  https://docs.rs/paste/0.1.6/paste/

cnd/src/db/load_swaps.rs
```diff
@@ -50,50 +92,11 @@ struct BitcoinEthereumBitcoinEtherAcceptedSwap {
     at: NaiveDateTime,
```
tcharding: This should have been renamed to `start_of_swap` during the resume swap PR.  I added it to my TODO list :)

Issue #1945, opened 2020-01-28
Cnd let clients continue with swap despite error


bonomat:
_follow up ticket from #1940_

# Problem

In #1940 matching an ethereum transaction failed with an error. However, cnd still returns the actions to continue with the swap. 

# Insights

c&p a comment from Thomas: https://github.com/comit-network/comit-rs/issues/1940#issuecomment-579024286

use an invalid url for the ethereum endpoint to reproduce.

> [...] I guess we don't update the state correctly somewhere so the API keeps emitting actions. 

# Recommendation

Outcome:
- Return 500 on the swap endpoint
- No longer return actions for swap that failed
- Still return the swap in the list of swaps (with no actions)

thomaseizinger:
It is not mentioned in the issue but it would be good to not just return a 500 but also a body with a problem ( application/problem+json).

PR #1965, opened 2020-01-31
Handle swap failure

tcharding:
Currently there is no indication on the `/swap/<foo>` endpoint if there
has been an error for a given swap.  This is not useful for a user.

This problem exists for `/swaps` endpoint also but needs to be handled
differently.

Do:
- For the `/swap/<foo>` endpoint return a 500 error if the swap has erred.
- For the `/swaps` endpoint return an erred swap but do not add any
  actions.

**untested**

Fixes: #1945

### DIFF
--- a/cnd/src/http_api/swap_resource.rs
+++ b/cnd/src/http_api/swap_resource.rs
@@ -108,7 +108,7 @@ pub enum IncludeState {
     No,
 }
 
-/// Build siren entity.  For failed swaps do not return any actions.
+/// Build siren entity. For failed swaps do not return any actions.
 pub fn build_rfc003_siren_entity_no_err<S: StateStore>(
     state_store: &S,
     swap: Swap,
@@ -118,7 +118,7 @@ pub fn build_rfc003_siren_entity_no_err<S: StateStore>(
     build_rfc003_siren_entity(state_store, swap, types, include_state, false)
 }
 
-/// Build siren entity.  For failed swaps return 500 error.
+/// Build siren entity. For failed swaps return 500 error.
 pub fn build_rfc003_siren_entity_err<S: StateStore>(
     state_store: &S,
     swap: Swap,
--- a/cnd/src/http_api/swap_resource.rs
+++ b/cnd/src/http_api/swap_resource.rs
@@ -131,7 +131,7 @@ pub fn build_rfc003_siren_entity<S: StateStore>(
             .get::<ROLE>(&id)?
             .ok_or_else(|| anyhow!("state store did not contain an entry for {}", id))?;
 
-        if on_fail == OnFail::Error && state.swap_failed() {
+        if state.swap_failed() && on_fail == OnFail::Error {
             return Err(anyhow!(HttpApiProblem::with_title_and_type_from_status(
                 StatusCode::INTERNAL_SERVER_ERROR,
             )));
@@ -183,13 +183,14 @@ pub fn build_rfc003_siren_entity<S: StateStore>(
                 .with_class_member("protocol-spec"),
             );
 
-        if on_fail == OnFail::NoAction {
-            let entity = actions.into_iter().fold(entity, |acc, action| {
-                let action = action.to_siren_action(&id);
-                acc.with_action(action)
-            });
+        if state.swap_failed() && on_fail == OnFail::NoAction {
             return Ok(entity);
         }
+
+        let entity = actions.into_iter().fold(entity, |acc, action| {
+            let action = action.to_siren_action(&id);
+            acc.with_action(action)
+        });
 
         Ok(entity)
     })
--- a/cnd/src/http_api/swap_resource.rs
+++ b/cnd/src/http_api/swap_resource.rs
@@ -108,35 +108,21 @@ pub enum IncludeState {
     No,
 }
 
-/// Build siren entity. For failed swaps do not return any actions.
-pub fn build_rfc003_siren_entity_no_err<S: StateStore>(
-    state_store: &S,
-    swap: Swap,
-    types: SwapTypes,
-    include_state: IncludeState,
-) -> anyhow::Result<siren::Entity> {
-    build_rfc003_siren_entity(state_store, swap, types, include_state, false)
-}
-
-/// Build siren entity. For failed swaps return 500 error.
-pub fn build_rfc003_siren_entity_err<S: StateStore>(
-    state_store: &S,
-    swap: Swap,
-    types: SwapTypes,
-    include_state: IncludeState,
-) -> anyhow::Result<siren::Entity> {
-    build_rfc003_siren_entity(state_store, swap, types, include_state, true)
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum OnFail {
+    Error,
+    NoAction,
 }
 
 // This is due to the introduction of a trust per Bitcoin network in the
 // `with_swap_types!` macro and can be iteratively improved
 #[allow(clippy::cognitive_complexity)]
-fn build_rfc003_siren_entity<S: StateStore>(
+pub fn build_rfc003_siren_entity<S: StateStore>(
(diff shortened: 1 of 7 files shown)


D4nte:
Can we test it? Do you aim to add a test as part of this PR?

cnd/src/http_api/swap_resource.rs
```diff
@@ -107,21 +108,48 @@ pub enum IncludeState {
     No,
 }
 
+/// Build siren entity.  For failed swaps do not return any actions.
```
D4nte:
```suggestion
/// Build siren entity. For failed swaps do not return any actions.
```

cnd/src/http_api/swap_resource.rs
```diff
@@ -107,21 +108,48 @@ pub enum IncludeState {
...
+pub fn build_rfc003_siren_entity_no_err<S: StateStore>(
+    state_store: &S,
+    swap: Swap,
+    types: SwapTypes,
+    include_state: IncludeState,
+) -> anyhow::Result<siren::Entity> {
+    build_rfc003_siren_entity(state_store, swap, types, include_state, false)
+}
+
+/// Build siren entity.  For failed swaps return 500 error.
```
D4nte:
```suggestion
/// Build siren entity. For failed swaps return 500 error.
```

cnd/src/http_api/swap_resource.rs
```diff
@@ -107,21 +108,48 @@ pub enum IncludeState {
...
+pub fn build_rfc003_siren_entity_err<S: StateStore>(
```
D4nte: I am trying to think of a better name...
```suggestion
pub fn build_rfc003_siren_entity_err_on_fail<S: StateStore>(
```

cnd/src/http_api/swap_resource.rs
```diff
@@ -107,21 +108,48 @@ pub enum IncludeState {
...
 }
 
+/// Build siren entity.  For failed swaps do not return any actions.
+pub fn build_rfc003_siren_entity_no_err<S: StateStore>(
+    state_store: &S,
+    swap: Swap,
+    types: SwapTypes,
+    include_state: IncludeState,
+) -> anyhow::Result<siren::Entity> {
+    build_rfc003_siren_entity(state_store, swap, types, include_state, false)
```
D4nte: Ok, here is a proposal that would make things clearer (I believe):
1. Do not add these new wrapper functions `build_rfc003_siren_entity_no_err`/`build_rfc003_siren_entity` and directly use `build_rfc003_siren_entity`
1. Instead of passing a bool to `build_rfc003_siren_entity` pass a new enum:
```
pub enum OnFail {
   Error,
   NoAction
}
```

I think it would make things clearer than having to comment and append `_no_err` and use wrapper and having bool argument.

da-kami: I'd definitely upvote for enum, way more readable than a random boolean.

tcharding:
> Can we test it? Do you aim to add a test as part of this PR?

Discussion offline between @da-kami, @thomaseizinger, and myself led to the conclusion that automated testing of this is not at all trivial and probably not worth doing.  At least one manual test should be done though really.

cnd/src/http_api/swap_resource.rs
tcharding: Thanks, old habits die hard.

cnd/src/http_api/swap_resource.rs
tcharding: Use better function name

cnd/src/http_api/swap_resource.rs
tcharding: Cool, nice design.  Will implement.  This is Rust not C right :)

D4nte: I did not want to comment but I did feel this was coming from a C dev.

D4nte:
Do update the changelog @tcharding :) Mergify only asked 3 times ;)

cnd/src/http_api/swap_resource.rs
```diff
@@ -168,10 +182,14 @@ pub fn build_rfc003_siren_entity<S: StateStore>(
                 .with_type("text/html")
                 .with_class_member("protocol-spec"),
             );
-        let entity = actions.into_iter().fold(entity, |acc, action| {
-            let action = action.to_siren_action(&id);
-            acc.with_action(action)
-        });
+
+        if on_fail == OnFail::NoAction {
```
D4nte: Hum, just do a `match` block? :)

tcharding: This code is buggy, is now:
```
if state.swap_failed() && on_fail == OnFail::NoAction {
```

D4nte: So actually a match block within `if state.swap_failed()` would help clarify the logic. (ok looking at the code now).

cnd/src/swap_protocols/rfc003/bob/mod.rs
```diff
@@ -23,6 +23,7 @@ pub struct State<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> {
     pub beta_ledger_state: LedgerState<BL, BA>,
     #[derivative(Debug = "ignore")]
     pub secret_source: Arc<dyn DeriveIdentities>,
+    pub failed: bool, // Gets set on any error during the execution of a swap.
```
thomaseizinger: I don't like comments like these.

They scream to being outdated because you describe functionality here that happens somewhere else.
If a developer is interested in when this gets set, they can do a dataflow analysis from this field.

tcharding: Interesting point.  I agree that these comments are susceptible to becoming outdated.  Can I explain how I use these comments in code then perhaps you could tell me if you see holes in the method (or ways to improve).

Case 1: with a comment like this

If I'm interested enough in this area of the codebase, I use the comment as a hint of where to look - then if I find that the comment is correct, win, the comment has helped me quickly find what `failed` field is for.  If I find the comment is incorrect, I will typically have well understood the code by now, enough to be sure I'm wasting peoples time on the mailing list with the patch that I then create removing the comment, win - the code is cleaner and the I understand the field. 

Case 2: with out a comment like this

If I come across a field like this, one that its not immediately obvious when/why its set I have to do, as you suggest, data flow analysis of the containing object.  This is typically a larger scope than in case 1, but on the plus side I guess it means more understanding of the code.  I have to check every where though where as with the hint I can be reasonable confident (as long as the field is not being abused somewhere) where I should look and not bother looking at everysingle call site that modifies the field.

So, my argument is, the comment saves more time than it takes to remove it once it goes stale.  Every developer that hits this declaration is going to wonder what its for so why not give a hint and trust other devs to remove it if it gets stale?

Open to your thoughts

tcharding: Adding a mention to this thread in case you don't see it since its on a merged PR @thomaseizinger.

D4nte: Maybe there is a better name to ths field to?
> Gets set on any error during the execution of a swap.

instead of fail it could be `error_encountered` ?

thomaseizinger: Excuse my brevity as I am on my phone:

- Improved naming might make the comment obsolete
- Your option 1 assumes that I have to by default mistrust every comment, which kind of defeats the purpose IMO.
- Without bashing tools, but getting all call sites with write access to this field are two consecutive shortcuts in CLion 🤷‍♂️. That is probably even faster than reading and understanding the comment.

Inline comments always scream at me: Careful! Non-obvious shit going on here, more attention than usual needed.

Rationale is: if only obvious things were happening, there would be no need to point them out.

The increased understanding of the codebase is a plus aswell :)

PR #1970, opened 2020-02-01
Split a comit crate out of cnd

thomaseizinger:
This PR is the first step towards having a comit rust library.

More importantly, it allows us to set enforceable architectural boundaries for which modules can depend on which other modules.

I also tried to merge the libp2p-comit crate into the comit crate but failed :(
Ultimately, the reason why we can't merge it atm is our communication protocol and how we use libp2p swarms.

Even without that merge, I think this is a good step in the right direction.

Tagging everyone for review as a FYI but also ofc to provide input :)

Since this will likely generate a lot of merge conflicts, I'd ask everyone to only block merging this if there are serious concerns.
Improvements can always be landed iteratively!


D4nte:
I would love if #1968 could be merged first as I already spent too much time on the related issue.

cnd/src/http_api/action.rs
```diff
@@ -54,17 +54,17 @@ pub enum ActionResponseBody {
         min_median_block_time: Option<Timestamp>,
     },
     EthereumDeployContract {
-        data: crate::ethereum::Bytes,
+        data: comit::ethereum::Bytes,
```
tcharding: Can we have these all changed to just `ethereum::Bytes` please?  Can be done as follow up PR though (perhaps we should have a ticket to audit the whole codebase for unnecessary long paths).

cnd/src/facade.rs
```diff
@@ -137,14 +146,46 @@ impl<B: ledger::bitcoin::Bitcoin + 'static> HtlcEvents<B, asset::Bitcoin> for Fa
 }
 
 #[async_trait::async_trait]
-impl<A> HtlcEvents<Ethereum, A> for Facade
-where
-    A: Asset + Send + Sync + 'static,
-    ethereum::Cache<Web3Connector>: HtlcEvents<Ethereum, A>,
-{
+impl HtlcEvents<Ethereum, asset::Ether> for Facade {
```
D4nte: Would it be possible to avoid this kind of refactoring as part of this PR?

cnd/src/http_api/routes/rfc003/accept.rs
```diff
@@ -53,15 +51,31 @@ pub struct OnlyRefund<L: Ledger> {
     pub beta_ledger_refund_identity: L::Identity,
 }
 
-impl<B: ledger::Bitcoin + 'static> ListRequiredFields for Accept<B, Ethereum> {
+impl ListRequiredFields for Accept<bitcoin::Mainnet, Ethereum> {
```
D4nte: Does this need to be done now?

1. I am not sure I understand the point of this change.
1. If this is not needed for the extraction to a lib then we should avoid it to make this 168files PR easier to review.

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
```diff
@@ -547,43 +553,68 @@ impl<B: ledger::Bitcoin + 'static> IntoIdentities<B, ledger::Ethereum> for HttpI
...
+    ) -> anyhow::Result<Identities<ledger::Ethereum, bitcoin::Mainnet>> {
+        into_identities(self, secret_source)
+    }
+}
 
-        let beta_ledger_redeem_identity = crate::bitcoin::PublicKey::from_secret_key(
-            &*crate::SECP,
-            &secret_source.derive_redeem_identity(),
-        );
+impl IntoIdentities<ledger::Ethereum, bitcoin::Testnet> for HttpIdentities {
```
D4nte: Is that needed now?

comit/src/btsieve/ethereum/transaction_pattern.rs
```diff
@@ -172,42 +172,12 @@ pub struct Event {
...
-        },
-        quickcheck::Quickcheck,
+    use crate::ethereum::{
+        Address, Block, Bytes, Log, Transaction, TransactionReceipt, H160, H2048, H256,
     };
     use spectral::prelude::*;
     use std::str::FromStr;
 
-    #[test]
-    fn given_pattern_from_arbitrary_address_contract_creation_transaction_matches() {
```
da-kami: ❓ Why was this test removed? Was it re-introduced somewhere or is it just not necessary...?

cnd/src/facade.rs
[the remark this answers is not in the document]
thomaseizinger: 1. Because the other thing is no longer allowed by the compiler now that the type we are abstracting over (`Asset`) lives in a different crate.
2. In the longer run, I'd hope to use the ambassador crate to create this impls and those will always be concrete forwardings.
3. I think it is actually more correct to be explicit here instead of trying to be clever and use generics. I'd rather use something like ambassador to avoid us having to type this code instead of using generics.

cnd/src/http_api/action.rs
thomaseizinger: I don't think having a ticket helps here. It is probably not worth the time talking about but it.

If you feel strongly enough, just go ahead and change it :)
I wanted to get this over the finish line so I didn't bother changing small things like this.

cnd/src/db/load_swaps.rs
```diff
@@ -346,15 +345,15 @@ impl From<EthereumBitcoinEtherBitcoinAcceptedSwap>
                 beta_asset: record.bitcoin_amount.0.into(),
                 hash_function: *record.hash_function,
                 alpha_ledger_refund_identity: record.ethereum_refund_identity.0.into(),
-                beta_ledger_redeem_identity: record.bitcoin_redeem_identity.0.into(),
```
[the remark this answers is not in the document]
thomaseizinger: No I changed the public key type from `rust-bitcoin` to ours or the other way round, whatever made the conversion useless :)

cnd/src/facade.rs
thomaseizinger: Unfortunately not, because I hit a compiler error that forbids to abstract over traits the crate doesn't own.

cnd/src/http_api/routes/rfc003/accept.rs
thomaseizinger: Unfortunately yes, same reason as here: https://github.com/comit-network/comit-rs/pull/1970#discussion_r374798588

comit/src/btsieve/ethereum/transaction_pattern.rs
thomaseizinger: :D

I could not be bothered duplicating the quickcheck stuff in two crates so I just removed this test because I don't think it actually adds value.

thomaseizinger:
As replied to in several threads, I actually didn't have a choice with the refactorings here :(

I tried to do a little as possible but I understand the frustration in reviewing this.
I will close this one and (now that I know what needs to be done) prepare for the changes to be easier.

cnd/src/facade.rs
D4nte: Indeed, I should have spent more time trying to understand.

D4nte:
> As replied to in several threads, I actually didn't have a choice with the refactorings here :(
> 
> I tried to do a little as possible but I understand the frustration in reviewing this.
> I will close this one and (now that I know what needs to be done) prepare for the changes to be easier.

Hi Thomas, I really feel like your effort was somehow wasted? Sorry if we went strong on review, but I am pretty sure we all rooted for this change.
Your answer were satisfying for me to go and continue the review to give it a green tick.
However, as you mentioned, this can lead to a lot of merge conflicts and I only see two practical ways to move forward:
1. Pause other work on comit-rs to get this in
2. Incremental approach of "trying it all" > "targeting smaller refactoring needed" > "get smaller refactoring in a dedicated PR" > rinse & repease

thomaseizinger:
All good, no hurt feelings :)
I opted for the 2nd approach! Will try and send some PRs today!

PR #1977, opened 2020-02-04
Split HtlcEvents

### LINKED ISSUE #1779: Rewrite `HtlcEvents` trait into single function traits
## Problem

The `HtlcEvents` trait is unidiomatic Rust. Unidiomatic usage of the language makes it harder to use it to its fullest potential.

## Goal / Benefits

- Easy onboarding of new devs
- Easy adoption of Rust-patterns

## Recommendation

Split the trait into 4:

- `HtlcFunded`
- `HtlcDeployed`
- `HtlcRedeemed`
- `HtlcRefunded`

Splitting the last method into two will require changes to `rfc003::create_swap` as we will have to watch for redeemed or refunded in parallel. That can be fairly easily achieved with [`futures::select!`](https://docs.rs/futures/0.3.1/futures/macro.select.html )

D4nte:
Resolves #1779

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,2 @@
 [workspace]
 members = ["cnd", "libp2p-comit"]
-
-[replace]
-"http-api-problem:0.14.0" = { git = "https://github.com/thomaseizinger/http-api-problem", branch = "warp-support" }
--- a/cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
+++ b/cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
@@ -11,7 +11,7 @@ use crate::{
         rfc003::{
             self,
             alice::State,
-            events::{HtlcDeployed, HtlcEvents, HtlcFunded},
+            events::{HtlcDeployed, HtlcFunded, HtlcRedeemed, HtlcRefunded},
             state_store::StateStore,
             Accept, Decline, DeriveIdentities, DeriveSecret, Ledger, Request, SecretHash,
         },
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -1,5 +1,2 @@
 [workspace]
 members = ["cnd", "libp2p-comit"]
-
-[replace]
```
thomaseizinger: Please don't do two things in the same PR :clown_face:

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
```diff
@@ -21,6 +24,72 @@ use futures_core::future::TryFutureExt;
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 
+async fn initiate_request<AL, BL, AA, BA>(
```
thomaseizinger: Was that re-ordering necessary?

cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -90,7 +102,11 @@ where
     BL: Ledger,
     AA: Asset,
     BA: Asset,
-    D: HtlcEvents<AL, AA>,
+    D: HtlcFunded<AL, AA>
+        + HtlcFunded<AL, AA>
```
thomaseizinger: Two times the same trait bounds?

Cargo.toml
tcharding: I'm still not sold on this @thomaseizinger, since we talked about how I do small refactorings/cleanup as the initial patches on a series (that becomes a single PR) I have tried _not_ doing this and making separate PRs.  It is a PITA because if the refactor/cleanup turns out to be wrong (when typically I would just rebase and ammend it) one has to go over to the other PR, fix it there, explain why it was wrong in the first place etc etc etc.  Also, any time while reviewing a PR, viewed on Github, if there is anything non-trivial in it I check out the branch and read each patch separately.  I've been doing this more since I noticed you were writing commit messages also :)  So I feel that this change is fine here since it is done as a separate patch at the start of the series.

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
tcharding: Continuing on the discussion above, I'd be fine with the re-ordering if it was in a separate patch.  It adds unnecessary cognitive load to the reviewer having refactor done within a patch that does logic changes also.

Cargo.toml
D4nte: > Please don't do two things in the same PR 🤡

Is that a new rule? It's a separate commit.

cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
D4nte: This re-ordering was absolutely not done in purpose. I have no idea how this happened.

cnd/src/swap_protocols/rfc003/create_swap.rs
D4nte: What is clippy good for? Obvious typo.

Cargo.toml
thomaseizinger: I was joking in reference to you guys picking apart my comit-crate refactoring :D

Back to being serious:

@tcharding 

It depends on the size, you can't come up with a hard rule ...
In this case I was joking (hence the clown), a little change like this should be fine (as long as it is in a separate commit).
Generally, if your refactorings have a bigger impact, you will have an easier life (less rebasing and conflicts) if you send them as separate PRs because they get merged faster.

Whether or not you actually need them for a change will depend on your forecasting abilities :)
If you don't trust them, don't send separate PRs until you are done (so you know).
You can also send them in one PR obviously. As mentioned above, I was only suggesting that as a solution to your complaints about having to constantly rebase your refactorings.
Try and optimize the following at the same time:

- make reviews as pleasant as possible
- make your own life as easy as possible

Issue #1927, opened 2020-01-23
Run the e2e tests in parallel


thomaseizinger:
## Problem

At the moment, all of our e2e tests run sequentially. This takes a long time of our CI build and also when run locally.

## Goal

Fast e2e tests.

## Recommendation

Migrate the test runner from mocha to Jest.

Mocha cannot run tests in parallel, hence we will need to switch our test suite to a different runner, for example `Jest`.
Jest can be run [programmically](https://github.com/facebook/jest/issues/5048) as-well, although the API is not quite as nice as the Mocha one is.

Running Jest programmatically allows us to keep our `harness` entrypoint which can boot up the ledgers for us (if needed). We can then simply pass on the test files to Jest which will run them in parallel.

Replacing mocha with Jest will also allow us to remove all those `setTimeout` hacks we have in every test file.

Similar to how it's done now, replace the call to Mocha with a call to Jest in `harness.ts`.

bonomat:
I see another advantage of this: it tests if cnd is able to deal with a _noisier_ blockchain.

PR #1978, opened 2020-02-04
Convert api dry tests to new format

### LINKED ISSUE #1960: Convert dry tests to our new e2e test framework
## Problem

To fully migrate our e2e test suite, we also need to migrate the dry tests.

## Goal

No longer use the legacy test framework.

## Recommendation

Converting the dry tests is tricky because they need access to a lot of details + they involve more than two actors.
Several extensions to the test framework will be necessary:

- Being able to control, which actors are started
- Getting access to the actual response without failing the tests even if a 400 is returned

bonomat:
Resolves #1960 

The goal was to move to the new test setup so that we can do https://github.com/comit-network/comit-rs/issues/1927. 

The tests might not be the most beautiful ones but most of them will go away once we move to the new communication protocol


D4nte:
GitHub diff rendering is being rubbish, will need to review locally.

bonomat:
> GitHub diff rendering is being rubbish, will need to review locally.

@D4nte : don't overdue the review please. @thomaseizinger pointed out that these tests will go away once we have the new communication protocol. 

The main goal here was to switch to the new e2e test format so that we can get rid of the old test setup, and have faster tests overall.

api_tests/dry/multiple_peers.ts
```diff
@@ -1,207 +1,193 @@
...
+    BitcoinWallet,
+    ComitClient,
+    EthereumWallet,
+    SwapRequest,
+} from "comit-sdk";
+import { Mock } from "ts-mockery";
+import { sleep } from "../lib_sdk/utils";
+import { SwapDetails } from "comit-sdk/dist/src/cnd";
+
+const alpha = {
```
thomaseizinger: Why can we not use our defaults from the test framework here?

api_tests/lib_sdk/actor_test.ts
```diff
@@ -1,6 +1,41 @@
...
 import { timeout } from "./utils";
+import { Actor } from "./actors/actor";
+import { createActor } from "./create_actor";
+/*
+ * Instantiates a new e2e test based on three actors
+ *
+ * This test function will take care of instantiating the actors and tearing them down again
+ * after the test, regardless if the test succeeded or failed.
+ */
+export function threeActorTest(
```
thomaseizinger: No motivation to reduce the duplication between one, two and three actor test? :D

thomaseizinger:
> If your only reason for rewriting this is to allow the tests to be run in parallel, then I feel like we are using the wrong abstractions here.
> 
> Can we not ditch `ComitClient` for using `Cnd` from the SDK directly?
> `ComitClient` allows you to do things like `accept` but you actually never use these?

I am divided between merging this to move forward and solving it in a better way. I feel like this goes a bit too far into the direction "as long as it works, merge it" :sweat_smile:

api_tests/dry/multiple_peers.ts
```diff
@@ -1,207 +1,85 @@
...
-        asset: {
-            name: "ether",
-            quantity: {
-                bob: ethers.utils.parseEther("10").toString(),
-                charlie: ethers.utils.parseEther("20").toString(),
-            },
-        },
-        expiry: new Date("2080-06-11T13:00:00Z").getTime() / 1000,
+import { SwapDetails } from "comit-sdk/dist/src/cnd";
+import {createDefaultSwapRequest} from "../lib_sdk/utils";
```
D4nte: Does not look like it went through prettier.

da-kami:
So, shall I review this or not? :)

api_tests/dry/multiple_peers.ts
bonomat: Yes, missed it. done now.

bonomat:
@D4nte / @da-kami : yes, please review :)

api_tests/dry/peers_using_ip.ts
```diff
@@ -1,149 +1,118 @@
...
-                    },
-                    beta_ledger_redeem_identity:
-                        "0x00a329c0648769a73afac7f9381e08fb43dbea72",
-                    alpha_expiry:
-                        new Date("2080-06-11T23:00:00Z").getTime() / 1000,
-                    beta_expiry:
-                        new Date("2080-06-11T13:00:00Z").getTime() / 1000,
+import { createDefaultSwapRequest } from "../lib_sdk/utils";
+
+async function assertNoPeersAvailable(actor: Actor, message: string) {
```
thomaseizinger: FYI: Chai actually has a way of providing your own assertions :)

api_tests/dry/peers_using_ip.ts
```diff
@@ -1,149 +1,118 @@
...
+        },
+    ]);
+}
+
+setTimeout(async function() {
+    describe("SWAP request with ip address", () => {
+        twoActorTest(
+            "[Alice] Should not yet see Bob's peer id in her list of peers",
+            async function({ alice }) {
+                const res = await request(alice.cndHttpApiUrl()).get("/peers");
```
thomaseizinger: I am fine with a partial migration but then we should record what still needs to be done, f.e. using the `cnd` instance inside the `Actor` to do communication with cnd instead of using `request` here.

api_tests/dry/peers_using_ip.ts
thomaseizinger: The way we currently did this was to create `assert` functions on the `Actor`, which would also work nicely here.

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -614,6 +618,45 @@ export class Actor {
 
         return defaultBetaAssetKind;
     }
+
+    public cndHttpApiUrl() {
```
thomaseizinger: You could avoid this if you would use the `cnd` instance + the `fetch` method.

api_tests/lib_sdk/utils.ts
```diff
@@ -16,3 +20,66 @@ export async function timeout<T>(ms: number, promise: Promise<T>): Promise<T> {
     // Returns a race between our timeout and the passed in promise
     return Promise.race([promise, timeout]);
 }
+
+export const DEFAULT_ALPHA = {
```
thomaseizinger: Hmm, that is exactly the kind of code we wanted to avoid / abstract over with the new testing framework. Unfortunately, we can't reuse the other part well as it is coupled with starting up the ledgers etc.

I guess this should be fine for now as it helps us move forward with the e2e tests but it still bugs me :D 

Can we not just merge all the tests into one file, then we don't have to create something global here that looks like it is supposed to be used across tests.
That's what bugs me: these kind of globals create coupling between code snippets where there shouldn't or doesn't have to be coupling. A variable scoped to a single test file at least doesn't invite re-use in other tests :)

bonomat:
Thomas' comments are tracked in #2013

PR #1984, opened 2020-02-05
1951 e2e lightning nodes

tcharding:
Finally got LND nodes spinning up!

FYI, here is the test I'm working on enabling.  So far have startLnd() working for both actors.

```
import { twoActorTest } from "../lib_sdk/actor_test";

setTimeout(function() {
    twoActorTest("sanity-lnd-alice-pays-bob", async function({ alice, bob }) {
        await alice.startLnd();
        await bob.startLnd();

<--!  Up to here  -->

        await alice.fundLnd();

        await alice.connectLnd(bob);
        await alice.openChannel(bob);

        const invoice = await bob.addInvoice(alice);
        await alice.sendPayment(invoice);

        await alice.assertChannelBalanceSender();
        await bob.assertChannelBalanceReceiver();
    });

    run();
}, 0);
```

### DIFF
--- a/api_tests/lib/bitcoind_instance.ts
+++ b/api_tests/lib/bitcoind_instance.ts
@@ -18,7 +18,10 @@ export class BitcoindInstance implements LedgerInstance {
         private readonly logDir: string,
         public readonly p2pPort: number,
         public readonly rpcPort: number
-    ) {}
+    ) {
+        this.dbDir = tmp.dirSync();
+        this.writeLogFile();
+    }
 
     public async start() {
         const bin = process.env.BITCOIND_BIN
@@ -26,31 +29,14 @@ export class BitcoindInstance implements LedgerInstance {
             : this.projectRoot +
               "/blockchain_nodes/bitcoin/bitcoin-0.17.0/bin/bitcoind";
 
-        this.dbDir = tmp.dirSync();
-
-        this.process = spawn(
-            bin,
-            [
-                `-datadir=${this.dbDir.name}`,
-                "-regtest",
-                "-server",
-                "-printtoconsole",
-                `-bind=0.0.0.0:${this.p2pPort}`,
-                `-rpcbind=0.0.0.0:${this.rpcPort}`,
-                "-rpcallowip=0.0.0.0/0",
-                "-nodebug",
-                "-acceptnonstdtxn=0",
-                "-rest",
+        this.process = spawn(bin, [`-datadir=${this.dbDir.name}`], {
+            cwd: this.projectRoot,
+            stdio: [
+                "ignore", // stdin
+                await openAsync(this.logDir + "/bitcoind.log", "w"), // stdout
+                await openAsync(this.logDir + "/bitcoind.log", "w"), // stderr
             ],
-            {
-                cwd: this.projectRoot,
-                stdio: [
-                    "ignore", // stdin
-                    await openAsync(this.logDir + "/bitcoind.log", "w"), // stdout
-                    await openAsync(this.logDir + "/bitcoind.log", "w"), // stderr
-                ],
-            }
-        );
+        });
 
         this.process.on("exit", (code: number, signal: number) => {
             console.log(`bitcoind exited with ${code || "signal " + signal}`);
@@ -77,5 +63,28 @@ export class BitcoindInstance implements LedgerInstance {
 
     public getUsernamePassword() {
         return { username: this.username, password: this.password };
+    }
+
+    private writeLogFile() {
+        const output = `regtest=1
+server=1
+printtoconsole=1
+bind=0.0.0.0:${this.p2pPort}
+rpcbind=0.0.0.0:${this.rpcPort}
+rpcallowip=0.0.0.0/0
+nodebug=1
+rest=1
+acceptnonstdtxn=0
+zmqpubrawblock=tcp://127.0.0.1:28332
+zmqpubrawtx=tcp://127.0.0.1:28333
+`;
+        const config = this.dbDir.name + "/bitcoin.conf";
+
+        fs.writeFile(config, output, function(err: any) {
+            if (err) {
+                return console.error(err);
+            }
+            console.log("bitcoind config file created: %s", config);
+        });
     }
 }
--- a/api_tests/lib/bitcoind_instance.ts
+++ b/api_tests/lib/bitcoind_instance.ts
@@ -20,7 +20,7 @@ export class BitcoindInstance implements LedgerInstance {
         public readonly rpcPort: number
     ) {
         this.dbDir = tmp.dirSync();
-        this.writeLogFile();
+        this.createConfigFile();
     }
 
     public async start() {
@@ -69,7 +69,7 @@ export class BitcoindInstance implements LedgerInstance {
         return this.dbDir.name;
     }
 
-    private writeLogFile() {
+    private createConfigFile() {
         const output = `regtest=1
 server=1
 printtoconsole=1
--- a/api_tests/lib/bitcoind_instance.ts
+++ b/api_tests/lib/bitcoind_instance.ts
@@ -65,6 +65,10 @@ export class BitcoindInstance implements LedgerInstance {
         return { username: this.username, password: this.password };
     }
 
+    public getDataDir() {
+        return this.dbDir.name;
+    }
+
     private writeLogFile() {
         const output = `regtest=1
 server=1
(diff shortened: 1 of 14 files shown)


D4nte:
> LND might be able to use `bitcoin.conf` to get these? Can we just pass path to
> bitcoind directory?

Yes the [doc](https://github.com/lightningnetwork/lnd/blob/master/docs/INSTALL.md#using-bitcoind-or-litecoind) says:

> The auth parameters rpcuser and rpcpass parameters can typically be determined by lnd for a bitcoind instance running under the same user, including when using cookie auth. In this case, you can exclude them from the lnd options entirely.

D4nte:
I don't think the note is relevant for regtest but let's see :) we won't connect more than 3 instances in any case.

D4nte:
> Two days working on this now: no concrete finish in sight so opening it up as a draft PR to get some help. Here are notes on what I've worked out needs doing.
> # Adding LND nodes to test infrastructure

I am sorry but it is not clear to me where you need help, can you please clarify?

tcharding:
> > Two days working on this now: no concrete finish in sight so opening it up as a draft PR to get some help. Here are notes on what I've worked out needs doing.
> 
> I am sorry but it is not clear to me where you need help, can you please clarify?

Just to confirm I'm on the right track, you've read over the PR message so that's help enough to keep me going - thanks.

api_tests/lib/bitcoind_instance.ts
```diff
@@ -78,4 +64,32 @@ export class BitcoindInstance implements LedgerInstance {
     public getUsernamePassword() {
         return { username: this.username, password: this.password };
     }
+
+    public getDataDir() {
+        return this.dbDir.name;
+    }
+
+    private writeLogFile() {
```
bonomat: That should be something like `createConfigFile()` right?

api_tests/lib/bitcoind_instance.ts
```diff
@@ -18,39 +18,25 @@ export class BitcoindInstance implements LedgerInstance {
         private readonly logDir: string,
         public readonly p2pPort: number,
         public readonly rpcPort: number
-    ) {}
+    ) {
+        this.dbDir = tmp.dirSync();
+        this.writeLogFile();
```
bonomat: What's the reason behind doing this in the constructor? I think this is better done in the `start()` function.

api_tests/sanity/lnd.ts
```diff
@@ -0,0 +1,21 @@
+import { twoActorTest } from "../lib_sdk/actor_test";
+
+setTimeout(function() {
+    twoActorTest("sanity-lnd-alice-pays-bob", async function({ alice, bob }) {
+        await alice.startLnd();
+        await bob.startLnd();
```
bonomat: This could be a performance issue as starting LND nodes takes some time. 
Ideally we can configure the need for LND nodes in the `config.toml` file which would start LND nodes for all tests within a particular folder. 

We can optimize this later though, just wanted to point it out.

api_tests/sanity/lnd.ts
```diff
@@ -0,0 +1,21 @@
...
+        await alice.fundLnd();
+
+        await alice.connectLnd(bob);
+        await alice.openChannel(bob);
+
+        const invoice = await bob.addInvoice(alice);
+        await alice.sendPayment(invoice);
+
+        await alice.assertChannelBalanceSender();
+        await bob.assertChannelBalanceReceiver();
```
bonomat: I think it is more likely that between Alice and Bob there is at least one other party (Charlie), hence I would opt for a sanity test setting up 3 nodes and connect them ala: 
Alice -> Charlie -> Bob

api_tests/sanity/lnd.ts
```diff
@@ -0,0 +1,21 @@
...
+
+        await alice.fundLnd();
+
+        await alice.connectLnd(bob);
+        await alice.openChannel(bob);
+
+        const invoice = await bob.addInvoice(alice);
+        await alice.sendPayment(invoice);
+
+        await alice.assertChannelBalanceSender();
```
bonomat: I think it is better to (additionally) assert that the invoice was actually settled: e.g. 
```
alice.assertInvoiceSettled(invoice);
bob.assertInvoiceSettled(invoice);
```

api_tests/sanity/lnd.ts
```diff
@@ -0,0 +1,21 @@
...
+        await alice.startLnd();
+        await bob.startLnd();
+
+        await alice.fundLnd();
+
+        await alice.connectLnd(bob);
+        await alice.openChannel(bob);
+
+        const invoice = await bob.addInvoice(alice);
+        await alice.sendPayment(invoice);
```
bonomat: Are we testing that the channel setup was successful? Then this is ok. 

Alternatively you can test the proposed process from the spike, e.g.: 

```
const secret = alice.createSecret();
const invoice = bob.createHoldInvoice(hash(secret));
const invoice = bob.addHoldInvoice(invoice);
alice.payInvoice(invoice);
bob.settleInvoice(invoice);
```

api_tests/lib/bitcoind_instance.ts
tcharding: Yes, good catch.  Thanks.

api_tests/lib/bitcoind_instance.ts
tcharding: I don't have a strong opinion on it.  For lnd we don't pass the bitcoind data dir out of the constructor so that's why its done in the constructor.  Happy to re-consider if you have a technical reason why not to do it in the constructor otherwise I'd prefer to see the bitcoind and lnd code done in a uniform manner.

api_tests/sanity/lnd.ts
tcharding: Paraphrasing to confirm I got you, optimise so that all tests in a single directory use the same LND instances.  Nice idea, didn't think of that myself.  FTR LND nodes must appear in the `config.toml`.  Also, I'm not totally understanding what is old test code and what is new test code as far as the harness goes (i.e., what is going away and what is staying).

api_tests/sanity/lnd.ts
tcharding: That's a second, three-party, sanity test IMO.  This may also be important when we start to think about cnd expiry times in relation to lnd expiry times.  I'll raise an issue.

tcharding: https://github.com/comit-network/comit-rs/issues/2023

api_tests/sanity/lnd.ts
tcharding: Right this second I don't see the benefit, I don't see any harm either.  I'll trust you on this one and add the above lines to the current test skeleton.  When I come to implement them I'll get a better understanding :)

api_tests/sanity/lnd.ts
tcharding: For me this goes to far into the proposed protocol.  This issue is just adding basic LND nodes and doing a _normal_ payment between two nodes.  I assume that asserting the channel setup successful is handled by the library code.

tcharding:
Thanks for the review @bonomat!

tcharding:
Closing this branch as we have a [new one](https://github.com/comit-network/comit-rs/pull/2020) as a result of pairing with @D4nte.  @bonomat your suggestions are implemented on that branch.

api_tests/lib/bitcoind_instance.ts
tcharding: Turns out that doing as you suggest resolves the async/sync constructor mix up pointed out over on #2020 by @thomas.  So implementing as suggested.

api_tests/sanity/lnd.ts
bonomat: Got it.

api_tests/sanity/lnd.ts
bonomat: I think it is easier to verify that a invoice was settled than to verify the balance of a state. For the latter one you will need to remember the state at the beginning and at the end of the payment +- payment fees (which annoys me every time :D)

api_tests/sanity/lnd.ts
bonomat: Yes, correct. Two (one for Alice and one for Bob) LND nodes per directory. 
Right now harness.ts it taking care of starting the blockchain nodes which I think we want to change later on.

api_tests/sanity/lnd.ts
tcharding: Cool, thanks for saving me the time of have to work this out :)

PR #1988, opened 2020-02-05
Macrofy everything

tcharding:
Recent work added types for different ledgers i.e., `Mainnet`, `Testnet`, etc.  This has led to a lot of code duplication.  Do we want to macrofy everything?

We should, IMO, weigh the costs vs the benefits.  Some of these are:

### Pro
- Less code duplication
- Reduced maintenance burden because of point above
- Helps us increase our proficiency with macros
- Way less copy/pasta.  Much more enjoyable to work with both during writing and after when modifying

### Cons
- Harder to read
- Harder to write ... at first

Thoughts @comit-network/rust-devs?

### DIFF
--- a/cnd/src/swap_protocols/ledger/bitcoin.rs
+++ b/cnd/src/swap_protocols/ledger/bitcoin.rs
@@ -14,27 +14,17 @@ pub trait Bitcoin:
 {
 }
 
+impl Bitcoin for Mainnet {}
+impl Bitcoin for Testnet {}
+impl Bitcoin for Regtest {}
+
 pub trait Network {
     fn network() -> ::bitcoin::Network;
 }
 
-impl Bitcoin for Mainnet {}
-
-impl From<Mainnet> for LedgerKind {
-    fn from(_: Mainnet) -> Self {
-        LedgerKind::BitcoinMainnet
-    }
-}
 impl Network for Mainnet {
     fn network() -> ::bitcoin::Network {
         ::bitcoin::Network::Bitcoin
-    }
-}
-impl Bitcoin for Testnet {}
-
-impl From<Testnet> for LedgerKind {
-    fn from(_: Testnet) -> Self {
-        LedgerKind::BitcoinTestnet
     }
 }
 impl Network for Testnet {
@@ -42,15 +32,26 @@ impl Network for Testnet {
         ::bitcoin::Network::Testnet
     }
 }
-
-impl Bitcoin for Regtest {}
-impl From<Regtest> for LedgerKind {
-    fn from(_: Regtest) -> Self {
-        LedgerKind::BitcoinRegtest
-    }
-}
 impl Network for Regtest {
     fn network() -> ::bitcoin::Network {
         ::bitcoin::Network::Regtest
+    }
+}
+
+impl From<Mainnet> for LedgerKind {
+    fn from(_: Mainnet) -> Self {
+        LedgerKind::BitcoinMainnet
+    }
+}
+
+impl From<Testnet> for LedgerKind {
+    fn from(_: Testnet) -> Self {
+        LedgerKind::BitcoinTestnet
+    }
+}
+
+impl From<Regtest> for LedgerKind {
+    fn from(_: Regtest) -> Self {
+        LedgerKind::BitcoinRegtest
     }
 }
--- a/cnd/src/swap_protocols/ledger/bitcoin.rs
+++ b/cnd/src/swap_protocols/ledger/bitcoin.rs
@@ -36,20 +36,15 @@ impl_network!(Mainnet, bitcoin::Network::Bitcoin);
 impl_network!(Testnet, bitcoin::Network::Testnet);
 impl_network!(Regtest, bitcoin::Network::Regtest);
 
-impl From<Mainnet> for LedgerKind {
-    fn from(_: Mainnet) -> Self {
-        LedgerKind::BitcoinMainnet
-    }
+macro_rules! impl_from_for_ledgerkind {
+    ($ledger:ty, $ledger_kind:expr) => {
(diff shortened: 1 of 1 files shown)


thomaseizinger:
I am generally open to the use of macros for making our life simpler but I also think we have to be careful.

In particular, not everything that looks the same is _harmful_ code duplication. Rigorously following DRY leads to coupling, which at some stage is even worse than code duplication. It prevents us from evolving parts of the code in the way we need.
Concretely, recent efforts like splitting the networks into structs and splitting traits into single-fn traits are an attempt to enable different designs from what we currently have.

Once we actually settle on a design and still see repetition in various places, we can think about using macros to help with that.
Even then though, I'd be more open to the use of procedural macros instead of declarative ones.

In regards to this particular PR, I am unsure if that would be helpful in the long-run. I don't see how the declaration of 3 structs is (harmful) duplication. IMO they actually _should_ be separate.
Likely, it is going to be orders of magnitude more helpful if we identify functionality that is duplicated across our codebase that can be generalized from our domain. For example, think of something like the `Debug` trait and imagine there would be no custom derive for it.

tcharding:
> I am generally open to the use of macros for making our life simpler but I also think we have to be careful.

Agreed

> In particular, not everything that looks the same is _harmful_ code duplication. Rigorously following DRY leads to coupling, which at some stage is even worse than code duplication. It prevents us from evolving parts of the code in the way we need.

Yes, at first I thought that these macros could be moved up a level and put in `mod.rs` to be re-used for the ethereum side since the code is identical but I realised the concerns are separate.  I agree with you here not all duplication is bad. 

> Concretely, recent efforts like splitting the networks into structs and splitting traits into single-fn traits are an attempt to enable different designs from what we currently have.

Cool, I didn't realise this.

> Once we actually settle on a design and still see repetition in various places, we can think about using macros to help with that.
> Even then though, I'd be more open to the use of procedural macros instead of declarative ones.

I'm up to the section on procedural macros in my book ATM :) 

> In regards to this particular PR, I am unsure if that would be helpful in the long-run. I don't see how the declaration of 3 structs is (harmful) duplication. IMO they actually _should_ be separate.

Adding the macro here does not make the declarations _not_ separate.  The primary benefit (apart from conciseness) is that alterations to the attributes (which admittedly are infrequent) would only need to be done once instead of three times. 

> Likely, it is going to be orders of magnitude more helpful if we identify functionality that is duplicated across our codebase that can be generalized from our domain. For example, think of something like the `Debug` trait and imagine there would be no custom derive for it.

Looking forward to exploring this more.

cnd/src/swap_protocols/ledger/bitcoin.rs
```diff
@@ -1,56 +1,50 @@
...
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-pub struct Mainnet;
-
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-pub struct Testnet;
-
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-pub struct Regtest;
+macro_rules! declare_ledger {
```
D4nte: I vote no on this one. The result is longer than the original code which means to me that it is not beneficial. Or would you want to re-use it for Ethereum?

cnd/src/swap_protocols/ledger/bitcoin.rs
```diff
@@ -1,56 +1,50 @@
...
+    ($ledger:ident) => {
+        #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+        pub struct $ledger;
+    };
+}
+declare_ledger!(Mainnet);
+declare_ledger!(Testnet);
+declare_ledger!(Regtest);
 
 pub trait Bitcoin:
```
D4nte: I think that with @thomaseizinger 's other PR, this trait may go to the trash.

cnd/src/swap_protocols/ledger/bitcoin.rs
```diff
@@ -1,56 +1,50 @@
...
+macro_rules! impl_from_for_ledgerkind {
+    ($ledger:ty, $ledger_kind:expr) => {
+        impl From<$ledger> for LedgerKind {
+            fn from(_: Mainnet) -> Self {
+                $ledger_kind
+            }
+        }
+    };
+}
+impl_from_for_ledgerkind!(Mainnet, LedgerKind::BitcoinMainnet);
```
D4nte: This impl is at the wrong place it should be in `comit_api` as it is where `LedgerKind` is defined.
`swap_protocols` is part of the comit lib, imported by cnd.

da-kami:
Agree with @thomaseizinger for this specific PR.

cnd/src/swap_protocols/ledger/bitcoin.rs
tcharding: lols, you put these here @D4nte :)

cnd/src/swap_protocols/ledger/bitcoin.rs
tcharding: Original total lines of code for this file is 56 lines, macro version is 50.

I would not re-use these macros for the ethereum side just because it makes the code harder to read, easier to have the macro code write were its called I feel.  Also because the macros rely on identifiers in the file its better to have the macro in the file also.

tcharding:
Closing this for now since there are more folks on the 'against' side than the 'for' side.  The underlying problem is still there, we have too much duplication.  Lets find a better solution that more people like.  Cheers.

cnd/src/swap_protocols/ledger/bitcoin.rs
D4nte: I know :)

PR #1992, opened 2020-02-05
Unroll some implementations generic over ledger::Bitcoin

thomaseizinger:
If we want to split a comit crate out of cnd, we cannot abstract over a trait that is not defined in our crate.

This patch removes all the usages of the ledger::Bitcoin that are used in code that would be outside of the comit crate for abstracting over any bitcoin network.

Split out of #1970.

### DIFF
--- a/cnd/src/http_api/routes/rfc003/accept.rs
+++ b/cnd/src/http_api/routes/rfc003/accept.rs
@@ -1,7 +1,6 @@
 use crate::{
     http_api::action::ListRequiredFields,
     swap_protocols::{
-        ledger,
         ledger::{bitcoin, Ethereum},
         rfc003::{
             actions::Accept,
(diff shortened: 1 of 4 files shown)


D4nte:
> If we want to split a comit crate out of cnd, we cannot abstract over a trait that is not defined in our crate.
> 
> This patch removes all the usages of the ledger::Bitcoin that are used in code that would be outside of the comit crate for abstracting over any bitcoin network.
> 
> Split out of #1970.

Why are you doing this change again?

thomaseizinger:
> > If we want to split a comit crate out of cnd, we cannot abstract over a trait that is not defined in our crate.
> > This patch removes all the usages of the ledger::Bitcoin that are used in code that would be outside of the comit crate for abstracting over any bitcoin network.
> > Split out of #1970.
> 
> Why are you doing this change again?

You mean splitting out a comit crate?
To enforce architectural boundaries :)
But it is not high priority, just something I wanted to do for a long time and I think it is going to be beneficial in understanding and extending the codebase.

D4nte:
> > > If we want to split a comit crate out of cnd, we cannot abstract over a trait that is not defined in our crate.
> > > This patch removes all the usages of the ledger::Bitcoin that are used in code that would be outside of the comit crate for abstracting over any bitcoin network.
> > > Split out of #1970.
> > 
> > 
> > Why are you doing this change again?
> 
> You mean splitting out a comit crate?
> To enforce architectural boundaries :)
> But it is not high priority, just something I wanted to do for a long time and I think it is going to be beneficial in understanding and extending the codebase.

No I meant, why is this PR needed to split out a comit crate?

thomaseizinger:
Because you can't do a blanket impl over a trait that is defined in a different crate.
All of the above modified code would be outside the comit crate.

D4nte:
> Because you can't do a blanket impl over a trait that is defined in a different crate.
Oh, I did not know/forgot about that. I just knew you could not implement a foreign trait on a foreign type.

PR #1995, opened 2020-02-06
Use SIGKILL instead of SIGINT to kill parity node

da-kami:
I had problems with multiple parity instances being alive on my Mac.
This solves the problem, should not affect Linux.

I tested this by running single tests, i.e. `yarn run test e2e/rfc003/btc_eth/happy.ts` in `api_tests` folder and with `SIGINT` and `SIGTERM` the instance is almost never killed properly on my machine.

### DIFF
--- a/api_tests/lib/parity_instance.ts
+++ b/api_tests/lib/parity_instance.ts
@@ -53,6 +53,6 @@ export class ParityInstance implements LedgerInstance {
     }
 
     public stop() {
-        this.process.kill("SIGINT");
+        this.process.kill("SIGKILL");
     }
 }
--- a/api_tests/lib/parity_instance.ts
+++ b/api_tests/lib/parity_instance.ts
@@ -4,6 +4,7 @@ import tmp from "tmp";
 import { promisify } from "util";
 import { LedgerInstance } from "./ledger_runner";
 import { LogReader } from "./log_reader";
+import { sleep } from "./util";
 
 const openAsync = promisify(fs.open);
 
@@ -52,7 +53,9 @@ export class ParityInstance implements LedgerInstance {
         return this;
     }
 
-    public stop() {
+    public async stop() {
+        this.process.kill("SIGTERM");
+        await sleep(3000);
         this.process.kill("SIGKILL");
     }
 }


api_tests/lib/parity_instance.ts
```diff
@@ -53,6 +53,6 @@ export class ParityInstance implements LedgerInstance {
     }
 
     public stop() {
-        this.process.kill("SIGINT");
+        this.process.kill("SIGKILL");
```
D4nte:
```suggestion
        this.process.kill("SIGTERM");
```
`SIGKILL` is not really healthy, did you try `SIGTERM`?

da-kami: Unfortunately `SIGTERM` does not kill it properly. Not sure why partiy is that difficult to kill, the only one that does the job for me is `SIGKILL` 😐
I would like to investigate further, but that might be a rabbit-hole...

D4nte: I'd been to find a different strategy then as it only impacts you. Let's chat tomorrow

tcharding:
https://linuxhandbook.com/sigterm-vs-sigkill/

tcharding:
Not sure if its related but yesterday I was also getting parity node process staying up after the btc_eth/happy test ended, I think it was only when the test failed.  Perhaps we are exiting early and not cleaning up correctly?  Or perhaps its something to do with only running a single e2e test.

da-kami:
> Not sure if its related but yesterday I was also getting parity node process staying up after the btc_eth/happy test ended, I think it was only when the test failed. Perhaps we are exiting early and not cleaning up correctly? Or perhaps its something to do with only running a single e2e test.

Well, if I use `SIGKILL` it is always cleaned up properly. Never had issues then. So I am not sure what the relation would be. The only thing that I can guess (from the article you posted) is that parity might run child-processes in some cases? (wild guesses...)

da-kami:
@D4nte I upgraded to latest parity version, but faced the same problem. 
So I opted for the solution we discussed earlier. Send `SIGTERM` and then after three seconds send `SIGKILL`.

PR #1996, opened 2020-02-06
convert under-/overfunded to new e2e model

da-kami:
@thomaseizinger note that I did not opt for overloading `fund()` but added separate `overfund` and `underfund` functions because I feel it reads way better in the test. 

Fixed a small bug in comit-rs, Bob never transitioned from `INCORRECTLY_FUNDED` to `REFUND` 😄

### DIFF
--- a/api_tests/lib_sdk/actors/actor.ts
+++ b/api_tests/lib_sdk/actors/actor.ts
@@ -1,5 +1,5 @@
 import { expect } from "chai";
-import { BigNumber, Cnd, ComitClient, Swap } from "comit-sdk";
+import { BigNumber, Cnd, ComitClient, Field, Swap } from "comit-sdk";
 import { parseEther } from "ethers/utils";
 import getPort from "get-port";
 import { Logger } from "log4js";
@@ -13,6 +13,7 @@ import { sleep } from "../utils";
 import { Wallet, Wallets } from "../wallets";
 import { Actors } from "./index";
 import { HarnessGlobal } from "../../lib/util";
+import { timeoutPromise } from "comit-sdk/dist/src/timeout_promise";
 
 declare var global: HarnessGlobal;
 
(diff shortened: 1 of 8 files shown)


api_tests/lib_sdk/actors/actor.ts
```diff
@@ -241,6 +241,30 @@ export class Actor {
         }
     }
 
+    public async overfund() {
+        const response = await this.swap.tryExecuteAction("fund", {
+            maxTimeoutSecs: 10,
+            tryIntervalSecs: 1,
+        });
+        const amount = response.data.payload.amount;
+        response.data.payload.amount = amount * 1.01;
```
bonomat: Just to be on the save side, what do you think of changing this to this? 
I’m thinking of catching rounding errors in advance.

```suggestion
        response.data.payload.amount = amount + 100 ;
```

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -241,6 +241,30 @@ export class Actor {
...
+        this.logger.debug("Funded swap %s in %s", this.swap.self, txid);
+    }
+
+    public async underfund() {
+        const response = await this.swap.tryExecuteAction("fund", {
+            maxTimeoutSecs: 10,
+            tryIntervalSecs: 1,
+        });
+        const amount = response.data.payload.amount;
+        response.data.payload.amount = amount * 0.01;
```
bonomat: Same as above: 
```suggestion
        response.data.payload.amount = amount - 100;
```

api_tests/lib_sdk/actors/actor.ts
da-kami: nobody put the `no-mergify` label, so it merged :)
But happy to still reply to this (same for the other comment):
I think the multiplication is actually safer, because the amount depends on the currency and the supply. With the multiplication we always add a fraction  relative to the amount rather than a fixed amount.

bonomat: > With the multiplication we always add a fraction relative to the amount rather than a fixed amount.

That is exactly the problem. 
If you are swapping 10 wei/satoshi, this would fail or not have any impact. Which means, we just have to have high enough values.

PR #1997, opened 2020-02-06
Unroll Bitcoin trait

tcharding:
Currently we are using a trait bound bound to `Bitcoin`.  Due to the
    fact that rustc does not use trait bounds when differentiating impl
    blocks this is currently buggy, the impl blocks are actually blanket
    implementations.  This currently works because we only use the
    implementations for the bitcoin side, as soon as we try to add the
    ethereum side this breaks.

    Attempt to unroll the `Bitcoin` trait and use concrete network types.
    This all goes ok until we get to the database layer.  Here we cannot
    unroll it because our current db implementation does not allow different
    types to be in the same column (by design).  With concrete network types
    all the database tables break - we would need to 'unroll' the tables
    also i.e., triple the tables, one for each network.

### DIFF
--- a/cnd/src/swap_protocols/facade.rs
+++ b/cnd/src/swap_protocols/facade.rs
@@ -103,15 +103,13 @@ where
 }
 
 #[async_trait::async_trait]
-impl<B: ledger::bitcoin::Bitcoin + ledger::bitcoin::Network> HtlcFunded<B, asset::Bitcoin>
-    for Facade
-{
+impl HtlcFunded<ledger::bitcoin::Mainnet, asset::Bitcoin> for Facade {
     async fn htlc_funded(
         &self,
-        htlc_params: HtlcParams<B, asset::Bitcoin>,
-        htlc_deployment: &Deployed<B>,
+        htlc_params: HtlcParams<ledger::bitcoin::Mainnet, asset::Bitcoin>,
+        htlc_deployment: &Deployed<ledger::bitcoin::Mainnet>,
         start_of_swap: NaiveDateTime,
-    ) -> anyhow::Result<Funded<B, asset::Bitcoin>> {
+    ) -> anyhow::Result<Funded<ledger::bitcoin::Mainnet, asset::Bitcoin>> {
         self.bitcoin_connector
             .htlc_funded(htlc_params, htlc_deployment, start_of_swap)
             .await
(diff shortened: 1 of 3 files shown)


thomaseizinger:
Hmm, I am not sure if we want to completely unroll it. For everything inside the "library" (currently roughly `swap_protocols`), using the trait should be fine.

I am not 100% sure though.

You mentioned something about doing the same thing for Ethereum breaks things, can you elaborate on that?

tcharding:
> Hmm, I am not sure if we want to completely unroll it. For everything inside the "library" (currently roughly `swap_protocols`), using the trait should be fine.
> 
> I am not 100% sure though.
> 
> You mentioned something about doing the same thing for Ethereum breaks things, can you elaborate on that?

We have the following code in `swap_protocols/rfc003/bitcoin.rs`
```

impl<B: ledger::Bitcoin> Ledger for B {
    type HtlcLocation = OutPoint;
    type Identity = PublicKey;
    type Transaction = Transaction;
}
```

The trait bound on `B` is not used by the compiler when creating concrete implementations of this generic code.  This means we cannot add
```
impl<B: ledger::Etherem> Ledger for E {
    type HtlcLocation = Address;
    type Identity = Address;
    type Transaction = Transaction;
}
```

thomaseizinger:
Did you actually try to compile that and it failed? Interesting, I want to look into that :)

tcharding:
> Did you actually try to compile that and it failed? Interesting, I want to look into that :)

Yes, the error is 'conflicting implementations'.  There is a new feature called 'specialisation' that is working to fix this.

https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md

tcharding:
Close, blocking issue is tracked elsewhere.

D4nte:
> > Did you actually try to compile that and it failed? Interesting, I want to look into that :)
> 
> Yes, the error is 'conflicting implementations'. There is a new feature called 'specialisation' that is working to fix this.
> 
> https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md

Yes but not ready yet: https://github.com/rust-lang/rust/issues/31844 :(

thomaseizinger:
> > > Did you actually try to compile that and it failed? Interesting, I want to look into that :)
> > 
> > 
> > Yes, the error is 'conflicting implementations'. There is a new feature called 'specialisation' that is working to fix this.
> > [rust-lang/rfcs:text/1210-impl-specialization.md@`master`](https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md) [](/rust-lang/rfcs/blob/HEAD@{2020-05-05T00:33:09Z}/text/1210-impl-specialization.md)
> 
> Yes but not ready yet: [rust-lang/rust#31844](https://github.com/rust-lang/rust/issues/31844) :(

It is unlikely that we will get specialization this year if ever from what I know. There are too many unresolved issues with it to land on stable :)

PR #1998, opened 2020-02-06
Remove what seems to be an unused folder

D4nte:


thomaseizinger:
Make sure to get a review from @bonomat on this, he created it initially :)

PR #1985, opened 2020-02-05
Upgrade blockchain-contracts to 0.3.1

D4nte:
Resolves comit-network/blockchain-contracts#40

PR #2000, opened 2020-02-07
Do not match failed eth tx

D4nte:
I think I need #1996 first to be able to test that because rust-web3 lib fails the RPC call if there is not enough gas (meaning, I have difficulties to get a failed tx in the block). Hence I think it'd work if I am able to override the gas limit used to deploy the contract.

I found this bug because with #1985, I had a _out of gas_ issue at the htlc deployment but both Alice and Bob did "see" the deployment and could not see the redeem/refund.

Of course, redeem/refund was totally failing as the contract was not deployed at all.

Yes, this is a critical bug 🐛☠️☹️

Note: I believe this only applies to deploy as I don't think we would get the transaction in the block for a `revert` due to failed redeem/refund.

Note: Bug is proven with the [new test failing](https://github.com/comit-network/comit-rs/pull/2000/commits/623d846f287f89ebe5f105446080d383eee88881) at commit 623d846f287f89ebe5f105446080d383eee88881.

### DIFF
--- a/api_tests/e2e/rfc003/eth_btc/ignore_fail_eth_transactions.ts
+++ b/api_tests/e2e/rfc003/eth_btc/ignore_fail_eth_transactions.ts
@@ -9,7 +9,7 @@ setTimeout(function() {
         await alice.sendRequest(AssetKind.Ether, AssetKind.Bitcoin);
         await bob.accept();
 
-        await alice.fundLowGas();
+        await alice.fundLowGas("0x1b000");
 
         await alice.assertAlphaNotDeployed();
         await bob.assertAlphaNotDeployed();
--- a/api_tests/e2e/rfc003/eth_btc/ignore_fail_eth_transactions.ts
+++ b/api_tests/e2e/rfc003/eth_btc/ignore_fail_eth_transactions.ts
@@ -1,0 +1,21 @@
+import { twoActorTest } from "../../../lib_sdk/actor_test";
+import { AssetKind } from "../../../lib_sdk/asset";
+
+setTimeout(function() {
+    twoActorTest("rfc003-eth-btc-alpha-deploy-fails", async function({
+        alice,
+        bob,
+    }) {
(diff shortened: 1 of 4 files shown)


cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -222,51 +222,55 @@ where
...
-                    )
-                })?;
-
-            Some(receipt)
-        } else {
-            None
-        };
-
-        let result = pattern.matches(&transaction, receipt.as_ref());
+        let receipt = connector
```
tcharding: I don't understand these changes @D4nte, with this applied it seems now that we do not consider the bloom filter results and just fetch every receipt irrespective?

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -222,51 +222,55 @@ where
...
-                            tx_hash
-                        )
-                    })?,
-            };
-
-            return Ok(Some(TransactionAndReceipt {
-                transaction,
-                receipt,
-            }));
+        if receipt.status == Some(1.into()) {
```
tcharding: Can we have a const with descriptive name please instead of `1.into()`, perhaps `const ACCEPTED = 1.into()` (type missing :)

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -222,51 +222,55 @@ where
...
+                pattern,
+                tx_hash,
+                result
+            );
+
+            if result {
+                let receipt = match receipt {
+                    Some(receipt) => receipt,
+                    None => connector
+                        .receipt_by_hash(tx_hash)
```
tcharding: hmmm, I think this is broken mate.  We fetch the receipt by hash again here for the same `tx_hash` as above?

D4nte: Note that this code was already here, we already used to fetch the receipt lazily twice (well, there are 2 blocks fetching it but only one would fetch it).

PR #2008, opened 2020-02-10
Do not match failed Ethereum transaction

D4nte:
_Replaces #2000 because the force pushes were starting to get ugly_

I found this bug because with #1985, I had a _out of gas_ issue at the htlc deployment but both Alice and Bob did "see" the deployment and could not see the redeem/refund.

Of course, redeem/refund was totally failing as the contract was not deployed at all.

Yes, this is a critical bug 🐛☠️☹️

Note: I believe this only applies to deploy as I don't think we would get the transaction in the block for a `revert` due to failed redeem/refund.

Note: Bug is proven with the [new test failing](https://github.com/comit-network/comit-rs/pull/2000/commits/623d846f287f89ebe5f105446080d383eee88881) at commit 623d846f287f89ebe5f105446080d383eee88881: Alpha ledger is marked as funded: https://3843-130005855-gh.circle-artifacts.com/0/api_tests/log/tests/rfc003-eth-btc-alpha-deploy-fails.log

### DIFF
--- a/cnd/src/btsieve/ethereum/mod.rs
+++ b/cnd/src/btsieve/ethereum/mod.rs
@@ -4,7 +4,7 @@ mod web3_connector;
 
 pub use self::{
     cache::Cache,
-    transaction_pattern::{Event, Topic, TransactionPattern},
+    transaction_pattern::{Event, Topic, TransactionPattern, TRANSACTION_STATUS_OK},
     web3_connector::Web3Connector,
 };
 use crate::{
@@ -222,24 +222,18 @@ where
     for transaction in block.transactions.into_iter() {
         let tx_hash = transaction.hash;
 
-        let receipt = if needs_receipt {
-            let receipt = connector
-                .receipt_by_hash(tx_hash)
-                .compat()
-                .await?
-                .ok_or_else(|| {
-                    anyhow::anyhow!(
-                        "Could not get transaction receipt for transaction {:x}",
-                        tx_hash
-                    )
-                })?;
+        let receipt = connector
+            .receipt_by_hash(tx_hash)
+            .compat()
+            .await?
+            .ok_or_else(|| {
+                anyhow::anyhow!(
+                    "Could not get transaction receipt for transaction {:x}",
+                    tx_hash
+                )
+            })?;
 
-            Some(receipt)
-        } else {
-            None
-        };
-
-        let result = pattern.matches(&transaction, receipt.as_ref());
+        let result = pattern.matches(&transaction, &receipt);
 
         tracing::debug!(
             "matching {:?} against transaction {:x} yielded {}",
@@ -249,20 +243,6 @@ where
         );
 
         if result {
-            let receipt = match receipt {
-                Some(receipt) => receipt,
-                None => connector
-                    .receipt_by_hash(tx_hash)
-                    .compat()
-                    .await?
-                    .ok_or_else(|| {
-                        anyhow::anyhow!(
-                            "Could not get transaction receipt for transaction {:x}",
-                            tx_hash
-                        )
-                    })?,
-            };
-
             return Ok(Some(TransactionAndReceipt {
                 transaction,
                 receipt,
--- a/cnd/src/swap_protocols/rfc003/actions/ether.rs
+++ b/cnd/src/swap_protocols/rfc003/actions/ether.rs
@@ -17,7 +17,15 @@ impl FundAction<Ethereum, asset::Ether> for (Ethereum, asset::Ether) {
     type FundActionOutput = DeployContract;
 
     fn fund_action(htlc_params: HtlcParams<Ethereum, asset::Ether>) -> Self::FundActionOutput {
-        htlc_params.into()
+        let htlc = EtherHtlc::from(htlc_params.clone());
+        let gas_limit = EtherHtlc::deploy_tx_gas_limit();
+
+        DeployContract {
+            data: htlc.into(),
+            amount: htlc_params.asset.clone(),
+            gas_limit: gas_limit.into(),
+            chain_id: htlc_params.ledger.chain_id,
+        }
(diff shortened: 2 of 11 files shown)


cnd/src/swap_protocols/rfc003/actions/ether.rs
```diff
@@ -17,7 +17,15 @@ impl FundAction<Ethereum, asset::Ether> for (Ethereum, asset::Ether) {
     type FundActionOutput = DeployContract;
 
     fn fund_action(htlc_params: HtlcParams<Ethereum, asset::Ether>) -> Self::FundActionOutput {
-        htlc_params.into()
+        let htlc = EtherHtlc::from(htlc_params.clone());
+        let gas_limit = htlc.deployment_gas_limit();
```
da-kami:
```suggestion
        let gas_limit = htlc.fund_tx_gas_limit();
```

da-kami: 📣 I think that's supposed to be the one for fund.

D4nte: Ha-ha. Good point. Well actually for Ether there is only one action and it is simply incorrectly named on Blockchain contracts side so I can't do the change here.

D4nte: Also note the code is moved so doing a request change on that is a bit harsh.

cnd/tests/ethereum_transaction_pattern_e2e.rs
```diff
@@ -12,6 +12,18 @@ use web3::{
     Web3,
 };
 
+pub fn new_web3_client<D: Docker, E: Image>(
```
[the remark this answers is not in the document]
D4nte: It's a small optimisation to follow the common test conventions: set-up, execute, assert.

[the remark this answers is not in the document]
thomaseizinger: Not disputing the change but wanted to clarify that usually this pattern only refers to the code inside the test function, not to the order of fns inside the file :)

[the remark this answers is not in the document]
D4nte: Happy to drop the commit. It is an artefact from the time when I tried to test this via rust.

cnd/src/swap_protocols/rfc003/ethereum/mod.rs
```diff
@@ -82,32 +81,3 @@ impl HtlcParams<Ethereum, asset::Erc20> {
         Erc20Htlc::from(self).into()
     }
 }
-
-impl From<HtlcParams<Ethereum, asset::Ether>> for DeployContract {
```
thomaseizinger: I can't seem to find the difference of the From impl and what you are doing. How does this change contribute to the bugfix?

D4nte: This change helps understanding what is going on by moving all the code converting from htlc to actions in one file. It's a simple move see commit.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -263,10 +265,12 @@ where
                     })?,
             };
 
-            return Ok(Some(TransactionAndReceipt {
-                transaction,
-                receipt,
-            }));
+            if receipt.status == Some(TRANSACTION_STATUS_OK.into()) {
```
thomaseizinger: Would this maybe be better to do in the `matches` function?

This fn here is more about composing existing functionality instead of acting on this level of detail.

D4nte: I hesitated. Happy to move it.

cnd/src/swap_protocols/rfc003/ethereum/mod.rs
thomaseizinger: Didn't realize it is just moved :)

I guess I was just overly thorough in the review because it is a bug fix, hence I was expecting only changes relevant to the bug :)

cnd/tests/ethereum_transaction_pattern_e2e.rs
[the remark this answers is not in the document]
thomaseizinger: I like to think about the structure of source code similarly to that of a book.

The first thing you read (ToC) should give you and overview about the software: directory and file names.
With a file (chapter / section), things make sense if you read from top to bottom.
You probably want to start with what is most important and keep refining the story you are telling.
Mapped to source code, that would be the most important function at the top of the file with functions that are being used in there beneath it in the order they are used.

That is why I originally placed this fn at the bottom: it is an implementation detail for the test (which is the most important thing in this file) and you probably don't even care about the details when you read over the test fn so you don't even have to scroll down.

If you agree, drop the commit, if you disagree, leave it :)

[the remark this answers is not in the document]
thomaseizinger: Interestingly, `use` statements are at the top of the file and literally that last thing you are usually interested in :joy:

cnd/src/btsieve/ethereum/mod.rs
thomaseizinger: Up to you, we have the feature now, can always improve that later :)

cnd/tests/ethereum_transaction_pattern_e2e.rs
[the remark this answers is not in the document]
tcharding: > Interestingly, `use` statements are at the top of the file and literally that last thing you are usually interested in

That might be because 'every other language ever does this so just to not be too crazy lets just do it to'.  Rubbish reason if it is, I believe they are only at the top by convention, they can actually be placed anywhere.  Totally happy to start putting them at the bottom :)

Issue #1781, opened 2019-12-17
`TransactionPatterns` are unnecessarily generic and complex


thomaseizinger:
## Problem

The current design of `TransactionPattern` is very generic. It supports matching on all kinds of aspects of a transaction. In reality, we only use a fraction of this design.

The overly generic design is a legacy of times when btsieve was running in its own process and we had ideas of hosting it for users etc.

The problem with this design is that it has been and is a source of bugs and a lot of developer time has already been spent on debugging this part of the code.

## Goal

Propose a design that fits our purpose. That is, it completely reflects the way the patterns are used and turns invalid usage of them compile-errors.

Aspects the spike should address:

- The design should still allow us to use the main parts of btsieve (looking in the past, etc) for future protocols which will likely require different patterns
- Invalid usages of patterns should be compile-errors
- The output of a specific pattern should be linked to the usage.
For example, for the pattern of sending bitcoin to an address, we are interested in the specific `Outpoint`. Currently, we only return the full transaction from `MatchingTransactions` and look for the `Outpoint` again inside `HtlcEvents`. This is an unnecessary indirection. Similarly, we don't always need the `TransactionReceipt` for `TransactionPattern`s on Ethereum. Yet the current design forces us to always return a `TransactionAndReceipt` because some patterns need to return the receipt.

## Recommendation

Avoid to abstract above both Bitcoin and Ethereum as it would introduce unnecessary complication.

## The following issues have been closed in favor of this redesign:

- https://github.com/comit-network/comit-rs/issues/1726
- https://github.com/comit-network/comit-rs/issues/1459
- https://github.com/comit-network/comit-rs/issues/1551
- https://github.com/comit-network/comit-rs/issues/826

tcharding:
Ethereum side is done in PR: https://github.com/comit-network/comit-rs/pull/2039
Tracking issue for the Bitcoin side of this issue: https://github.com/comit-network/comit-rs/issues/2083

D4nte:
Done with #2039

Issue #1878, opened 2020-01-17
Adapt implementation of bitcoin `matching_transaction` to what the Ethereum side does


thomaseizinger:
## Problem

The implementation of `matching_transaction` for Bitcoin is still on the first iteration of the initial design based on async/await. We have since figured out a better way of doing this and have successfully done so in `matching_transactions` for Ethereum.

The current implementation of `matching_transaction` for Bitcoin is also blocking the final steps of the overall async/await migration.

## Goal

- Unblock the async/await migration
- Simplify the code

## Recommendation

Look at what `matching_transaction` for Ethereum does and do the same / something similar (it is easier for Bitcoin because Bitcoin doesn't have receipts).
This ticket also includes getting rid of `TransactionPattern` for Bitcoin (see here https://github.com/comit-network/comit-rs/issues/2083).

Note: Tests need to be adapted of course keep them (similar to what is done is Ethereum). Some unit tests are going away (pattern tests).

thomaseizinger:
The PR was reverted: https://github.com/comit-network/comit-rs/pull/2126

tcharding:
This was resolved in: https://github.com/comit-network/comit-rs/pull/2142

PR #2039, opened 2020-02-17
Re-design the matching_transaction logic for Ethereum

tcharding:
Currently the logic for matching transactions is super generic.  We would like an API that accepts only what is needed to match a transaction and returns what we need from the matched transaction. From the COMIT protocol level i.e., rfc003 this is a 'per action' API i.e., funded/deployed/redeem/refunded.

The API is now
```
matching_create_contract(...) -> Result<(Transaction, Address)>
matching_event(...) -> Result<(Transaction, Log)>
```

All the matching logic is contained within the `btsieve/ethereum` module.

Makes a start at resolving: #1781, Bitcoin side to do.

Issue https://github.com/comit-network/comit-rs/issues/2083 tracks the Bitcoin side.


tcharding:
Please do 10% review @thomaseizinger

cnd/src/btsieve/bitcoin/mod.rs
```diff
@@ -194,7 +232,7 @@ mod tests {
         let transaction = r#"02000000014135047eff77c95bce4955f630bc3e334690d31517176dbc23e9345493c48ecf000000004847304402200da78118d6970bca6f152a6ca81fa8c4dde856680eb6564edb329ce1808207c402203b3b4890dd203cc4c9361bbbeb7ebce70110d4b07f411208b2540b10373755ba01feffffff02644024180100000017a9142464790f3a3fddb132691fac9fd02549cdc09ff48700a3e1110000000017a914c40a2c4fd9dcad5e1694a41ca46d337eb59369d78765000000
 "#.to_owned();
 
-        let bytes = decode_response::<bitcoin::Transaction>(transaction);
```
thomaseizinger: Unrelated to the 10% review:

You are removing the most important 7 characters from this line: bitcoin.

Knowing, _which_ transaction is decoded here is massively important because you will be able to immediately narrow down further possible paths/areas in your brain of what to think about.
Having to scroll to the top of the `tests` block to do that is a huge burden IMO.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,11 +24,110 @@ use std::collections::HashSet;
...
+        matching_transaction(blockchain_connector, start_of_swap, |_transaction| {
+            // TODO: Add logic for matching a deployed transaction.
+            (false, None)
+        })
+        .await?;
+
+    Ok(transaction)
+}
+
+pub async fn matching_funded_transaction<'b, C>(
```
thomaseizinger: Bear in mind that funding for erc20 is different than for ether :)

The erc20 one needs to look for a Transfer event where as the ether one only needs to look for a transaction.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,11 +24,110 @@ use std::collections::HashSet;
...
+where
+    C: LatestBlock<Block = Option<Block>>
+        + BlockByHash<Block = Option<Block>, BlockHash = Hash>
+        + ReceiptByHash<Receipt = Option<TransactionReceipt>, TransactionHash = Hash>
+        + Clone,
+{
+    match matching_transaction(blockchain_connector, start_of_swap, |_transaction| {
+        // TODO: Add logic for matching a redeemed transaction.
+        //
+        // Includes using the bloom filter to conditionally fetch receipts
```
thomaseizinger: This is a very important!

How are you going to solve the issue of needing the receipt for refunded and redeemed?
In order for the underlying code to know that it has to fetch the receipt, you will kind of have to give it the event you are looking for.

The crucial code is here: https://github.com/comit-network/comit-rs/blob/64840879fae1d087d697a8c7530b527c757463a1/cnd/src/btsieve/ethereum/transaction_pattern.rs#L93-L95

Maybe it is actually enough to pass down the topic to matching_transactions? The underlying code can then filter out all blocks where the bloom filter doesn't match and call you for those that does match with a tuple of (`Transaction`, `Receipt`).

cnd/src/btsieve/bitcoin/mod.rs
tcharding: Will leave the paths in.

cnd/src/btsieve/ethereum/mod.rs
tcharding: Yep, I'm across that, will however keep it mind.  Thanks

cnd/src/btsieve/ethereum/mod.rs
tcharding: Cheers, I won't be 100% sure until I flesh it out some more but I think the design can accommodate the bloom filter correctly - time will tell.

tcharding:
> Interesting idea, I think you are on a good track!
> 
> Make sure you resolve the issue with the receipt before continuing, to validate that the design actually works.
> 
> Side-note:
> What to look for correlates with the individual protocols. This is why we have `HtlcDeployed` etc.
> Adding another layer that also uses the `deployed` terminology seems unnecessary.
> It would be nice if the next layer, i.e. the implementation of `HtlcDeployed` would use a more generic layer that no longer knows about those steps.
> 
> That is sort of what we have now, it is just bad.
> Maybe you can build something out of what you have here that exposes more primitive but useful matching logic like:
> 
> * `matching_contract_deployed_transactions`
> * `matching_event_emitted_transactions`
> * ...?

This is a nice idea, I might keep working with the functions I've got, make sure they work then see if I can generalise like you suggest.  I'd like the final functions to be defined by what they accept and what they return and _also_ not have conditions within the matching logic that require domain knowledge to understand i.e., the matching is explicit in the function names.  Will keep this in mind, thanks man.  Good review.

tcharding:
Ready for 50% review please @thomaseizinger, @luckysori, @D4nte

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,231 @@ use std::collections::HashSet;
...
+            let blockchain_connector = blockchain_connector.clone();
+            let events = events.clone();
+            async move {
+                let block_hash = block
+                    .hash
+                    .ok_or_else(|| anyhow::anyhow!("block without hash"))?;
+
+                if !block
+                    .logs_bloom
+                    .contains_input(Input::Raw(events[0].topics[0].unwrap().0.as_ref()))
```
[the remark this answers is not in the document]
thomaseizinger: Would also greatly simplify `events_exist_in_receipt` the method below.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -43,12 +173,48 @@ where
...
+            // By matching against the never type explicitly, we assert that the `Ok` value of the
+            // result is actually the never type and has not been changed since this
+            // line was written. The never type can never be constructed, so we cannot
+            // reach this line never anyway.
+            GeneratorState::Complete(Ok(never)) => match never {},
+        }
+    }
+}
+
+async fn matching_transaction_receipt<C, F>(
```
thomaseizinger:
```suggestion
async fn matching_transaction_and_receipt<C, F>(
```

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -104,25 +96,15 @@ async fn htlc_redeemed<A: Asset>(
...
-        transaction_data_length: None,
-        events: Some(vec![Event {
-            address: Some(htlc_deployment.location),
-            data: None,
-            topics: vec![Some(Topic(*REDEEM_LOG_MSG))],
-        }]),
-    };
+    let events = vec![Event {
+        address: Some(htlc_deployment.location),
+        data: None,
```
thomaseizinger: Do we ever use `data`?
Maybe you can remove the `Event` struct altogether and only pass the address and a list of topics to the function?

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,150 @@ use std::collections::HashSet;
...
+    events: Vec<Event>,
+    action: &'static str,
+) -> anyhow::Result<(Transaction, TransactionReceipt)>
+where
+    C: LatestBlock<Block = Option<Block>>
+        + BlockByHash<Block = Option<Block>, BlockHash = Hash>
+        + ReceiptByHash<Receipt = Option<TransactionReceipt>, TransactionHash = Hash>
+        + Clone,
+{
+    matching_transaction_receipt(blockchain_connector.clone(), start_of_swap, |block| {
```
thomaseizinger: You could go ahead and pass the `topics` already to `matching_transaction_receipt`. This would allow you to do the bloom filter matching inside and call the closure with a tuple of transaction and receipt.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,231 @@ use std::collections::HashSet;
...
+    location: Address,
+    token_contract: Address,
+) -> anyhow::Result<(Transaction, TransactionReceipt)>
+where
+    C: LatestBlock<Block = Option<Block>>
+        + BlockByHash<Block = Option<Block>, BlockHash = Hash>
+        + ReceiptByHash<Receipt = Option<TransactionReceipt>, TransactionHash = Hash>
+        + Clone,
+{
+    let events = vec![Event {
```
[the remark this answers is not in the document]
tcharding: Nice question, I thought I answered this to myself but I've mixed up Topic's and Events, we look for multiple Topic's but not multiple Events - will refactor.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,231 @@ use std::collections::HashSet;
...
+    bytecode: Bytes,
+    create_contract: CreateContract,
+) -> anyhow::Result<Transaction>
+where
+    C: LatestBlock<Block = Option<Block>>
+        + BlockByHash<Block = Option<Block>, BlockHash = Hash>
+        + ReceiptByHash<Receipt = Option<TransactionReceipt>, TransactionHash = Hash>
+        + Clone,
+{
+    matching_transaction(blockchain_connector, start_of_swap, |block| {
```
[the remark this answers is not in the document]
tcharding: Your observation is correct, `matching_transaction` could iterate over the txs but `matching_transaciton_receipt` cannot do this because it needs access to the block.  I'm inclined to keep the two functions inline and leave the iteration in the closure.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,231 @@ use std::collections::HashSet;
...
+        matching_events_transaction(blockchain_connector, start_of_swap, events, "refunded")
+            .await?;
+    Ok(transaction)
+}
+
+async fn matching_events_transaction<C>(
+    blockchain_connector: C,
+    start_of_swap: NaiveDateTime,
+    events: Vec<Event>,
+    action: &'static str,
```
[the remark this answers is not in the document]
tcharding: Yes, open to a cleaner idea.  I did not put that much thought into it.

cnd/src/btsieve/ethereum/mod.rs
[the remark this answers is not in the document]
tcharding: Have not even touched `events_exist_in_receipt` yet.  Now the 50% review is in I'll do some more refactoring, thanks for the reminder.

cnd/src/btsieve/ethereum/mod.rs
tcharding: cool

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
tcharding: I played with this a little and the conclusion I came to was we were trying to imitate web3 with the whole `Event` struct.  Is this assumption correct?

cnd/src/btsieve/ethereum/mod.rs
tcharding: Nice idea, will consider.

tcharding:
Thanks @thomaseizinger, super helpful review.

cnd/src/btsieve/ethereum/mod.rs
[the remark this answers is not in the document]
tcharding: Misunderstood you at first here; refactored `events_exist_in_receipt` to accept a single Event.  Did no other refactoring - I believe that's what you meant :)

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
tcharding: Answering my own question; on further investigation we can remove this, update the comment that refers to web3 and still maintain clarity while reducing complexity - WIN.  Thanks

cnd/src/btsieve/ethereum/mod.rs
thomaseizinger: It would also invalidate the assumption here: https://github.com/comit-network/comit-rs/pull/2039#discussion_r380997282

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
thomaseizinger: I don't think we need to imitate something. The original design might have been inspired by that but the original design was also trying to be general purpose (which turned out to be a really bad idea :sweat_smile:).

Question everything my friend :)

cnd/src/btsieve/ethereum/mod.rs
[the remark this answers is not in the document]
thomaseizinger: I think this is the kind of problem that tracing is trying to solve. If you open a new span at the call-site that contains this action as a context, every message inside will be scoped to this span and hence doesn't need to include the action name.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,134 @@ use std::collections::HashSet;
...
+            let blockchain_connector = blockchain_connector.clone();
+            let bytecode = bytecode.clone();
+            async move {
+                // transaction.to address is None if, and only if, the transaction
+                // creates a contract.
+                if transaction.to.is_none() && transaction.input == bytecode {
+                    // We can get a transaction that matches even for failed attempts
+                    // at contract creation e.g., low gas.
+                    let receipt = fetch_receipt(blockchain_connector, transaction.hash).await?;
+                    if receipt.status == Some(TRANSACTION_STATUS_OK.into()) {
```
thomaseizinger: Oh, nice one! I didn't think of that!

This got me thinking:
We should probably never return any transaction that failed right?

So the whole check should probably move into `matching_transaction` and run _after_ the matcher returned true. This will make this matcher non-async again :)
We probably then want to log a big warning that says:

"We found the transaction we were looking for but the execution failed!"

tcharding:
Can I get 90% review please, all that is left to do is integration testing (yes I am omitting unit tests, long gone are the days of TDD :(

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -50,21 +50,13 @@ impl HtlcDeployed<Ethereum, asset::Ether> for Cache<Web3Connector> {
...
-            events: None,
-        };
-        let TransactionAndReceipt { transaction, .. } =
-            matching_transaction(connector, pattern, start_of_swap)
-                .await
-                .context("failed to find transaction for htlc deployment")?;
+        let transaction =
+            matching_create_contract_transaction(connector, start_of_swap, htlc_params.bytecode())
+                .await?;
+        let location = calculate_contract_address_from_deployment_transaction(&transaction);
```
thomaseizinger: The contract address is actually in the receipt!

`matching_create_contract_transaction` could return us the location from the receipt, then we would not need to calculate it and could throw away a whole bunch of code (and some dependencies!) :)

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,134 @@ use std::collections::HashSet;
...
+        blockchain_connector.clone(),
+        start_of_swap,
+        event.topics.clone(),
+        action,
+        |transaction| {
+            Box::new({
+                let blockchain_connector = blockchain_connector.clone();
+                let event = event.clone();
+                async move {
+                    let receipt = fetch_receipt(blockchain_connector, transaction.hash).await?;
```
thomaseizinger: Moving this one layer down would make the matcher non-async (and hence a lot more concise).

cnd/src/btsieve/ethereum/mod.rs
thomaseizinger: This would also reduce some code duplication between the matchers.

tcharding:
@thomaseizinger, implemented your recent suggestions - this is getting seriously sexy.  Force pushed, check out the last two patches!  (CI will fail due to integration tests, working on that now.)

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -1,55 +1,247 @@
...
+            anyhow::anyhow!(
+                "Could not get transaction receipt for transaction {:x}",
+                hash
+            )
+        })?;
+    Ok(receipt)
+}
+
+fn event_exists_in_receipt(event: &Event, receipt: &TransactionReceipt) -> bool {
+    match event {
```
thomaseizinger: I think it would be the right time to refactor this method to use defensive programming:

```rust
let { topics, address } = event;

if topics.is_empty() {
   return false;
}

// etc ...
```

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -1,55 +1,247 @@
...
+                tracing::trace!(
+                    "bloom filter indicates that we should check the block for {} transactions: {:x}",
+                    action,
+                    block_hash,
+                );
+                for transaction in block.transactions.into_iter() {
+                    if let Some((transaction, receipt)) = matcher(transaction.clone()).await? {
+                        // We can get a transaction that matches even for failed attempts
+                        // at contract creation e.g., low gas.
+                        if receipt.status != Some(TRANSACTION_STATUS_OK.into()) {
```
thomaseizinger: We are doing this check in two different places. Might be worth defining a trait for it an implementing it on `TransactionReceipt`.

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -94,35 +85,20 @@ impl HtlcRefunded<Ethereum, asset::Ether> for Cache<Web3Connector> {
...
-    let TransactionAndReceipt {
-        transaction,
-        receipt,
-    } = matching_transaction(connector, pattern, start_of_swap)
-        .await
-        .context("failed to find transaction to redeem from htlc")?;
+    let (transaction, receipt) =
+        matching_event_transaction(connector, start_of_swap, event, "redeemed").await?;
+
     let log = receipt
```
thomaseizinger: As you can see, we iterate over all the logs and try to find the one again that we passed to `matching_event_transaction`.

We can do a similar optimization as with the contract address and return the event already from the `matching_event_transaction` function. That would save us this, in practice, unreachable error path of trying to find the one log again.

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -94,35 +85,20 @@ impl HtlcRefunded<Ethereum, asset::Ether> for Cache<Web3Connector> {
...
     };
 
-    let TransactionAndReceipt {
-        transaction,
-        receipt,
-    } = matching_transaction(connector, pattern, start_of_swap)
-        .await
-        .context("failed to find transaction to redeem from htlc")?;
+    let (transaction, receipt) =
+        matching_event_transaction(connector, start_of_swap, event, "redeemed").await?;
```
thomaseizinger: With the below optimization, a better name for this could be `matching_event`?

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -1,55 +1,247 @@
...
+where
+    C: LatestBlock<Block = Option<Block>>
+        + BlockByHash<Block = Option<Block>, BlockHash = Hash>
+        + ReceiptByHash<Receipt = Option<TransactionReceipt>, TransactionHash = Hash>
+        + Clone,
+{
+    matching_transaction_bloom_filter(
+        blockchain_connector.clone(),
+        start_of_swap,
+        event.topics.clone(),
```
thomaseizinger: With the matcher being non-async, this could turn into a reference.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -1,55 +1,247 @@
...
+    action: &'static str,
+) -> anyhow::Result<(Transaction, TransactionReceipt)>
+where
+    C: LatestBlock<Block = Option<Block>>
+        + BlockByHash<Block = Option<Block>, BlockHash = Hash>
+        + ReceiptByHash<Receipt = Option<TransactionReceipt>, TransactionHash = Hash>
+        + Clone,
+{
+    matching_transaction_bloom_filter(
+        blockchain_connector.clone(),
```
thomaseizinger: Similarly, with the matcher being non-async, you should be able to avoid the clone because you don't need the connector out here.

cnd/tests/ethereum_create_contract_matching.rs
```diff
@@ -12,14 +14,14 @@ use web3::{
...
-/// from the parity dev account. Afterwards we verify that the tx hash of
-/// the sent tx equals the one that we found through btsieve.
+const VALUE: u32 = 2_000_000_000;
+
+lazy_static::lazy_static! {
+    static ref REDEEM_LOG_MSG: H256 = blockchain_contracts::ethereum::rfc003::REDEEMED_LOG_MSG.parse().expect("to be valid hex");
+}
+
 #[tokio::test]
-async fn ethereum_transaction_pattern_e2e_test() {
```
thomaseizinger: I don't agree with repurposing this test.

The original idea was:

- have fast (i.e. non-IO) integration tests of the whole btsieve module by using a connector mock and real data (that's the tests you deleted for now). they verify things like missing blocks, etc but are unconcerned about the actual matching (they don't need to as it is orthogonal)
- have unit tests for the actual matching logic (with the new design, that might be obsolete but I guess it wouldn't hurt if we have some basic tests for edge-case logic). You can pass function pointers instead of closures and unit test the functions you are pointing to.
- have a **single**, slow e2e test that verifies that we glued things together correctly (this test). Note that this test should be as simple as possible and not use complicated contracts like HTLCs or something.

All these three topics are orthogonal and together form a good test coverage without introducing unnecessary dependencies and levels of detail.

The added value of these tests is very little because we have the e2e tests anyway and they cover the same thing (they also use actual HTLCs).
Yet, they are super heavy and slow.

What we need to test at this level of the code are edgecases that are _not_ covered by the e2e tests:

- missing blocks
- go back in the past
- etc

but those are missing with your testing strategy.
I'd recommend to revert this and just bring the other tests back.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -1,55 +1,247 @@
...
+                    if matcher(transaction.clone()) {
+                        let receipt = fetch_receipt(connector.clone(), transaction.hash).await?;
+                        // We can get a transaction that matches even for failed attempts
+                        // at contract creation e.g., low gas.
+                        if receipt.status != Some(TRANSACTION_STATUS_OK.into()) {
+                            tracing::warn!(
+                                "transaction matched {:x} but status was NOT OK",
+                                transaction.hash,
+                            );
+                            // A failed transaction could not be in the same block as
```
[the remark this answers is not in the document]
tcharding: Cool, thanks.

cnd/src/btsieve/ethereum/mod.rs
tcharding: Oh, you think that is nicer?  Interesting, that's how I had it for a lot time I was working on this.  I refactored it to use the 'if guard' within the match.  Unless I'm missing the meaning of your comment?
```
    match event {
        Event { topics, .. } if topics.is_empty() => false,
```
Is the same as :
```
let { topics, address } = event;

if topics.is_empty() {
   return false;
}
```

cnd/src/btsieve/ethereum/mod.rs
tcharding: Had a go but the `event` being used in the closure as well as here means we still need to clone.

cnd/src/btsieve/ethereum/mod.rs
tcharding: This ones right, rustc caught this one.  On a separate note, I would be interested in going through the code base and trying to remove clones at some stage when we are playing around - bit early for optimisations I know but this would be a good way to learn more about Rust life times, which is something I'd like to learn more about.

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
tcharding: Changed API functions to
```
matching_create_contract() -> (transaction, location)
matching_event() -> (transaction, log)
```

cnd/src/btsieve/ethereum/mod.rs
thomaseizinger: Yes it is the same functionally :)
But the last time I did that, you were like: "that is just using the language features for the sake of it! :D"

And then you suggested a defensive programming approach which is also quite neat.
I am not sure what is better in this case honestly. The defensive one maps clearer to the algorithm that one would have in their mind I guess :man_shrugging: 

Up to you - no strong feelings on this part :)

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
thomaseizinger: Nice, looks good :)

cnd/src/btsieve/ethereum/mod.rs
thomaseizinger: Only if the closure `move`s it right? If it is not async, there shouldn't be a need to move it. My brain borrow checker is weak though, so could very easily be that I am wrong :D

cnd/src/btsieve/ethereum/mod.rs
thomaseizinger: Cloning things per se is not bad, even though the Rust community is religious about it :sweat_smile: 

What is important is thinking about _what_ you are cloning here. The connector is not data, it is functionality that we are passing around. Maybe we should make that explicit and always take an `Arc<C>` to make it totally clear that cloning it is:

a) cheap
b) not harmful in any way

cnd/tests/ethereum_create_contract_matching.rs
tcharding: Ok, I think I've got you.  This is what I've understood to do

- Add unit tests that test the private matching functions i.e., use different matching closures
- Add integration tests that test the API, use static data for these (you didn't say this but as a precursor to the next point I think this is useful)
- Add integration tests that test 'go back in the past' and 'missing blocks'
- Remove the parity container integration test (you didn't say this but the description of point 3 for the old design is no longer valid/useful with the new design)

@thomaseizinger I've pushed a couple of commits that add skeleton functions.  I'm going to go ahead and work on these, if you feel like taking a look at the direction I'm taking please do so.  I do not think that the added integration test (using parity) adds much although it is still included because its done and working already, can probably be removed before this PR merges.  Thanks

tcharding: In case you saw the message above mid-edit, I had this message sitting open for an hour and during which I wrote the skeleton test functions.  Seems I contradicted myself.

thomaseizinger:
@tcharding do you plan on doing the bitcoin side also in this PR?
If no, please open an issue for it so we can track the remaining work!

cnd/src/btsieve/ethereum/mod.rs
tcharding: > @tcharding do you plan on doing the bitcoin side also in this PR?
> If no, please open an issue for it so we can track the remaining work!

I have not reached a decision yet, I was planning on finishing the Ethereum side then once I know exactly whats involved looking into the Bitcoin code and seeing if we should do this now or if we should make the Bitcoin code use Generators etc (all the improvements that have been done only on the Ethereum side).  Open to suggestions, will definitely create an issue if we go that way.

thomaseizinger: There is already a ticket for adopting the use of generators (https://github.com/comit-network/comit-rs/issues/1878). Happy to do that one first and then follow-up with adopting the model proposed in this PR for the Bitcoin side.

cnd/Cargo.toml
```diff
@@ -59,21 +59,11 @@ uuid = { version = "0.8", features = ["serde", "v4"] }
 void = "1.0.2"
 warp = { version = "0.2", default-features = false }
 
-# These versions need to be "in sync".
-# web3 0.8 gives us primitive-types 0.3.0
-# primitive-types 0.3.0 with the "rlp" feature gives us "rlp" version 0.4.2
 [dependencies.web3]
```
thomaseizinger: `web3` was only declared separately so I could add this comment. Feel free to declare it next to all the other dependencies now!

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,119 @@ use std::collections::HashSet;
 type Hash = H256;
 type Block = crate::ethereum::Block<Transaction>;
 
-pub async fn matching_transaction<C>(
+pub const TRANSACTION_STATUS_OK: u32 = 1;
```
thomaseizinger: This is now unused because the trait defines it's own const.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -192,71 +376,57 @@ fn seen_block_or_predates_start_of_swap(
...
-                receipt,
-            }));
-        }
-    }
+#[derive(Clone, Copy, Default, Eq, PartialEq, serde::Serialize, serdebug::SerDebug)]
+#[serde(transparent)]
+pub struct Topic(pub H256);
 
-    Ok(None)
+/// Event works similar to web3 filters:
```
thomaseizinger: Should update or delete these docs.

Also it would be nice to not `ignore` them if we keep them but just fix the compile errors in there.

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -192,71 +376,57 @@ fn seen_block_or_predates_start_of_swap(
...
+pub struct Event {
+    pub address: Address,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub topics: Vec<Option<Topic>>,
 }
 
-impl Predates for Block {
-    fn predates(&self, timestamp: NaiveDateTime) -> bool {
-        let unix_timestamp = timestamp.timestamp();
+trait TransactionStatusOk {
```
thomaseizinger: This could actually live in the general `ethereum` module.

It would also be nice if the function naming would give away that the fn returns a bool. Something like:

```rust
fn is_status_ok() -> bool {

}
```

tcharding: Fixed as suggested.

cnd/src/btsieve/ethereum/mod.rs
tcharding: Had a play with this, did not remove the 'ignore'.  Seems to me that adding all the type stuff to satisfy the compiler just makes the comment more cluttered with no real added advantage (since we are only parsing static strings into Rust types).

cnd/src/btsieve/ethereum/mod.rs
tcharding: Nice, removed.

cnd/Cargo.toml
tcharding: Cool, fixed.  Thanks

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,148 @@ use std::collections::HashSet;
...
+
+            log.topics.iter().enumerate().all(|(index, tx_topic)| {
+                let topic = &topics[index];
+                topic.as_ref().map_or(true, |topic| tx_topic == &topic.0)
+            })
+        }),
+    }
+}
+
+pub async fn matching_transaction_and_receipt<C, F>(
```
luckysori: :question: Is this public just so that we can test it?

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,148 @@ use std::collections::HashSet;
...
+                        }
+                        tracing::trace!("transaction matched {:x}", transaction.hash,);
+                        return Ok((transaction, receipt));
+                    }
+                }
+            }
+            GeneratorState::Complete(Err(e)) => return Err(e),
+            // By matching against the never type explicitly, we assert that the `Ok` value of the
+            // result is actually the never type and has not been changed since this
+            // line was written. The never type can never be constructed, so we cannot
```
luckysori: :wrench: 
```suggestion
            // line was written. The never type can never be constructed, so we can
```

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,148 @@ use std::collections::HashSet;
...
+            // reach this line never anyway.
```
luckysori: :wrench: 
```suggestion
            // never reach this line anyway.
```

cnd/src/btsieve/ethereum/mod.rs
```diff
@@ -24,16 +20,148 @@ use std::collections::HashSet;
 type Hash = H256;
 type Block = crate::ethereum::Block<Transaction>;
 
-pub async fn matching_transaction<C>(
+pub async fn matching_create_contract<C>(
```
luckysori: :thinking: Probably for a follow-up PR, but what do you think about renaming this function to `look_for_contract_creation` and doing something similar for the other ones? I find the term matching *so* confusing because it's both and adjective and a verb.

thomaseizinger: I am onboard with this! Alternative ways could also be:

- `wait_for_contract_creation`
- `watch_for_contract_created`

cnd/src/btsieve/ethereum/mod.rs
tcharding: Yes, funny you should ask that ...

cnd/src/btsieve/ethereum/mod.rs
tcharding: Well spotted @lucas, choosing 'watching' as a nice verb for what we are doing.

tcharding: Renamed:
- `matching_create_contract` -> `watch_for_contract_creation`
- `matching_event` -> `watch_for_event`

But that leaves the functions
- `matching_transaction_and_log`
- `matching_transaction_and_receipt`

Also the `matcher` predicate function will need to be renamed to 'match' the functions (pun intended :)

Any ideas?  Otherwise I'll have another think tomorrow and submit follow up PR.

thomaseizinger: I vote for follow-up PR!

Let's get this in! :tada:

Issue #1951, opened 2020-01-28
Extend our e2e test suite with Lightning nodes


thomaseizinger:
## Problem

In order to verify that our Lightning integration works, we need to test it. This in turn requires us to extend our e2e test suite with at least two Lightning nodes.

## Goal

Test the Lightning integration.
Have one test that:
- requires lnd
- Bob adds an invoice
- Alice sends a payment to Bob
- Check balances

## Recommendation

Lightning channels, compared to the whole blockchain, are per user. Even though several tests could share the same lnd instance, I feel it would be better if each tests starts their own lnd instance. Contrary to blockchain nodes, we definitely need one lnd instance per actor. We do start one cnd instance per actor, hence starting one lnd instance per actor seems natural as-well.

In the case of a LN swap, we need to:

- send money to lnd's wallet
- open a channel (should both parties have money in the beginning to make this part independent of the swap combination?)

Note: lnd needs to be compiled with the correct feature flag: `make tags=invoicesrpc`. As part of this issue you need to provide such binary (docker?)

Note: upon starting, lnd scans the blockchain to find any htlc or transactions related to its wallet/channels. This may take some time and make explain a delay in seeing money sent to its wallet.

thomaseizinger:
This is blocked by https://github.com/comit-network/comit-rs/issues/1927 to make sure our e2e test suite is ready handle more combinations of swaps without the total test time scaling linearly.

bonomat:
I remove the block of https://github.com/comit-network/comit-rs/issues/1927. 

It's more a _soft block_ in the way that we should not have two people working on both tickets at the same time. It is not a functional block, i.e. nothing in this setup is blocked by the other one. 

If we want to prevent 2 people working on these issues in parallel, we could turn the block around?

thomaseizinger:
> If we want to prevent 2 people working on these issues in parallel, we could turn the block around?

I don't want https://github.com/comit-network/comit-rs/issues/1927 to fall under the table :)

I am fine with not having a dependency as long as it still gets done.
Certainly, I would not put the dependency in the other direction.

tcharding:
Putting this back to the start of the pipeline after working on it for a sprint.  Draft PR: https://github.com/comit-network/comit-rs/pull/1984 adds the LND nodes, a sanity test, and skeleton `Actor` methods to glue it all together.

[...]

PR #2052, opened 2020-02-18
e2e lnd: mint, addPeer

D4nte:
_99%_
Added minting and the connection to a peer as part of the infrastructure setup.

Works towards #1951.
next PR(s): Open channel, add invoice, send payment, assert payment.

### DIFF
--- a/api_tests/lib_sdk/actors/actor.ts
+++ b/api_tests/lib_sdk/actors/actor.ts
@@ -57,7 +57,7 @@ export class Actor {
             JSON.stringify(actorConfig.generateCndConfigFile(ledgerConfig))
         );
 
-        return new Actor(logger, cndInstance, logRoot, actorConfig);
+        return new Actor(actorConfig, logger, logRoot, cndInstance);
     }
 
     public actors: Actors;
@@ -79,10 +79,10 @@ export class Actor {
     public lnd: Lnd;
 
     private constructor(
+        private readonly config: E2ETestActorConfig,
         private readonly logger: Logger,
-        private readonly cndInstance: CndInstance,
         private readonly logRoot: string,
-        private readonly actorConfig: E2ETestActorConfig
+        private readonly cndInstance: CndInstance
     ) {
         this.wallets = new Wallets({});
         const { address, port } = cndInstance.getConfigFile().http_api.socket;
@@ -354,6 +354,90 @@ export class Actor {
         }
     }
 
+    /**
+     * This method uses the cnd HTTP API routes introduced to support lightning.
+     * It should ultimately replace [sendRequest] once the previous route format
+     * is deprecated.
+     * @param alphaLedgerKind
+     * @param alphaAssetKind
+     * @param betaLedgerKind
+     * @param betaAssetKind
+     */
+    public async sendRequestOnLightningRoute(
+        alphaLedgerKind: LedgerKind,
+        alphaAssetKind: AssetKind,
+        betaLedgerKind: LedgerKind,
+        betaAssetKind: AssetKind
+    ) {
+        // By default, we will send the swap request to bob
+        const to = this.actors.bob;
+
+        this.logger.info("[WIP] Sending lighting swap request");
+
+        this.alphaLedger = defaultLedgerDescriptionForLedger(alphaLedgerKind);
+        this.alphaAsset = defaultAssetDescriptionForAsset(alphaAssetKind);
+        to.alphaLedger = this.alphaLedger;
+        to.alphaAsset = this.alphaAsset;
+
+        this.logger.debug(
+            "Alpha Ledger: %o; Alpha Asset: %o",
+            this.alphaLedger,
+            this.alphaAsset
+        );
+
+        this.betaLedger = defaultLedgerDescriptionForLedger(betaLedgerKind);
+        this.betaAsset = defaultAssetDescriptionForAsset(betaAssetKind);
+        to.betaLedger = this.betaLedger;
+        to.betaAsset = this.betaAsset;
+
+        this.logger.debug(
+            "Beta Ledger: %o; Beta Asset: %o",
+            this.betaLedger,
+            this.betaAsset
+        );
+
+        await this.initializeDependencies();
+        await to.initializeDependencies();
+
+        await this.setStartingBalance([
+            this.alphaAsset,
+            { name: this.betaAsset.name, quantity: "0" },
+        ]);
+        await to.setStartingBalance([
+            { name: to.alphaAsset.name, quantity: "0" },
+            to.betaAsset,
+        ]);
+
+        this.expectedBalanceChanges.set(
+            betaAssetKind,
+            new BigNumber(this.betaAsset.quantity)
+        );
+        to.expectedBalanceChanges.set(
+            alphaAssetKind,
+            new BigNumber(to.alphaAsset.quantity)
+        );
+
+        if (
+            this.alphaLedger.name === "lightning" ||
+            this.betaLedger.name === "lightning"
+        ) {
+            await this.wallets
+                .getWalletForLedger("lightning")
+                .addPeer(to.wallets.getWalletForLedger("lightning"));
+        }
+
+        this.expectedBalanceChanges.set(
+            betaAssetKind,
+            new BigNumber(this.betaAsset.quantity)
+        );
+        to.expectedBalanceChanges.set(
+            alphaAssetKind,
+            new BigNumber(to.alphaAsset.quantity)
+        );
+
+        // Here is where we would send the swap request to cnd HTTP API
+    }
+
     public async currentSwapIsAccepted() {
         let swapEntity;
 
@@ -528,17 +612,6 @@ export class Actor {
         return entity.properties.role;
     }
 
-    public async startLnd() {
-        const bitcoind = global.bitcoind.getDataDir();
-        this.lnd = new Lnd(
-            this.logger,
-            this.logRoot,
-            this.actorConfig,
-            bitcoind
-        );
-        await this.lnd.start();
-    }
-
     private async waitForAlphaExpiry() {
         const swapDetails = await this.swap.fetchDetails();
 
@@ -637,18 +710,31 @@ export class Actor {
     }
 
     private async initializeDependencies() {
+        let lightning = false;
         for (const ledgerName of [
             this.alphaLedger.name,
             this.betaLedger.name,
         ]) {
-            await this.wallets.initializeForLedger(ledgerName);
+            if (ledgerName === "lightning") {
+                lightning = true;
+            }
+            await this.wallets.initializeForLedger(
+                ledgerName,
+                this.logger,
+                this.logRoot,
+                this.config
+            );
         }
 
-        this.comitClient = new ComitClient(
-            this.wallets.getWalletForLedger("bitcoin").inner,
-            this.wallets.getWalletForLedger("ethereum").inner,
-            this.cnd
-        );
+        // Once `ComitClient` can be built with lightning, this hack should
+        // be removed and ComitClient built with the wallets available
+        if (!lightning) {
+            this.comitClient = new ComitClient(
+                this.wallets.getWalletForLedger("bitcoin").inner,
+                this.wallets.getWalletForLedger("ethereum").inner,
+                this.cnd
+            );
+        }
     }
 
     private getComitClient(): ComitClient {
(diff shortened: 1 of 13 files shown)


api_tests/lib_sdk/lnd.ts
```diff
@@ -131,8 +133,26 @@ export class Lnd {
         await sleep(1);
     }
 
-    public async fund() {
-        await sleep(1);
+    public async createChainAddress(): Promise<string> {
```
bonomat: My understanding is that this is needed to open the channel later?

api_tests/lib_sdk/wallets/lightning.ts
```diff
@@ -0,0 +1,102 @@
...
+declare var global: HarnessGlobal;
+
+export class LightningWallet implements Wallet {
+    public static async newInstance(
+        bitcoinNodeConfig: BitcoinNodeConfig,
+        logger: Logger,
+        logDir: string,
+        actorConfig: E2ETestActorConfig
+    ) {
+        const bitcoinRpcClient = new BitcoinRpcClient({
```
bonomat: Can we reuse `bitcoinWallet.ts` for this? They should be connected to the same chain right? 

you could then define LightningWallet like this: 

```
public static async newInstance(
        bitcoinNodeConfig: BitcoinNodeConfig,
        logger: Logger,
        logDir: string,
        actorConfig: E2ETestActorConfig
    ) {
        const bitcoinWallet = BitcoinWallet.newInstance({...
```

and then further down you don't have to reimplement the logic for minting but can refer to:

```
bitcoinWallet.mint(...)
```

or different idea: Do you need this function in here? Can't you use: 

```
Actor.fund(ln.createChainAddress())
```

My assumption is no, because `mint` is somewhere called automatically based on the interface `Wallet`.

D4nte:
Thank you for the review @bonomat. It looks like I have not been able to frame this 50% in a productive way as my description was apparently unclear/uncomplete. I'll move to 99% and try to think more about this 50% review concept.

api_tests/lib_sdk/lnd.ts
```diff
@@ -131,8 +133,26 @@ export class Lnd {
...
+        return response.address;
+    }
+
+    public async getChainBalance(): Promise<number> {
+        return (
+            await lnService.getChainBalance({ lnd: this.authenticatedLndGrpc })
+        ).chain_balance;
+    }
+
+    public async getChannelBalance(): Promise<number> {
```
thomaseizinger: Shouldn't this take an identifier of the channel or something?

api_tests/lib_sdk/wallets/lightning.ts
thomaseizinger: `mint` is part of the wallet interface yes.

I am in favor of this proposal, reusing the bitcoin wallet sounds like a good idea to reduce the code duplication.

api_tests/lib_sdk/wallets/lightning.ts
```diff
@@ -0,0 +1,102 @@
...
+            startingBalance.plus(minimumExpectedBalance),
+            asset
+        );
+    }
+
+    public async address(): Promise<string> {
+        return this.inner.createChainAddress();
+    }
+
+    public async getBalanceByAsset(asset: Asset): Promise<BigNumber> {
```
thomaseizinger: I think this should only return the channel balance.

api_tests/lib_sdk/wallets/lightning.ts
```diff
@@ -0,0 +1,102 @@
...
+
+        const channelBalance = await this.inner.getChannelBalance();
+        console.log("channelBalance", channelBalance);
+
+        return new BigNumber(chainBalance).plus(channelBalance);
+    }
+
+    // This functions does not have its place on a Wallet
+    public async getBlockchainTime(): Promise<number> {
+        throw new Error(
```
thomaseizinger: Not sure if this will not cause problems. Maybe best to just return the current system time?

api_tests/lib_sdk/lnd.ts
D4nte: Nope.

api_tests/lib_sdk/wallets/lightning.ts
D4nte: Let's fix that once I assert paid invoice.

api_tests/lib_sdk/wallets/lightning.ts
D4nte: Let's leave it cause a problem and we can see how to handle it.

api_tests/lib_sdk/lnd.ts
D4nte: It's needed to mint.

api_tests/lib_sdk/wallets/bitcoin.ts
```diff
@@ -37,6 +37,23 @@ export class BitcoinWallet implements Wallet {
...
+        minimumExpectedBalance: BigNumber,
+        toAddress: string
+    ): Promise<void> {
+        const blockHeight = await this.bitcoinRpcClient.getBlockCount();
+        if (blockHeight < 101) {
+            throw new Error(
+                "unable to mint bitcoin, coinbase transactions are not yet spendable"
+            );
+        }
+
```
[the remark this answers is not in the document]
D4nte: I am not quite sure, this is not new code :)

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -637,18 +660,31 @@ export class Actor {
...
         }
 
-        this.comitClient = new ComitClient(
-            this.wallets.getWalletForLedger("bitcoin").inner,
-            this.wallets.getWalletForLedger("ethereum").inner,
-            this.cnd
-        );
+        // Once `ComitClient` can be built with lightning, this hack should
+        // be removed and ComitClient built with the wallets available
+        if (!lightning) {
```
thomaseizinger: Ditch the mess with the mutable boolean var and just check it here:

```suggestion
        if (this.alphaLedger.name != "lightning" && this.betaLedger.name != "lightning") {
```

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -745,21 +801,36 @@ export class Actor {
...
+            return LedgerKind.Ethereum;
+        case AssetKind.Erc20:
+            return LedgerKind.Ethereum;
+    }
+}
+
+/**
+ * WIP as the cnd REST API routes for lightning are not yet defined.
+ * @param ledger
+ * @returns The ledger formatted as needed for the request body to cnd HTTP API on the lightning route.
```
thomaseizinger: They are gonna be different though if you think back to what the spikes says. The whole point of making new routes is to avoid these nested objects.

api_tests/lib_sdk/lnd.ts
```diff
@@ -127,16 +131,55 @@ export class Lnd {
...
+        return (
+            await lnService.getChannelBalance({
+                lnd: this.authenticatedLndGrpc,
+            })
+        ).channel_balance;
+    }
+
+    public addPeer(peer: Lnd): Promise<void> {
+        this.logger.debug(
+            `Connecting ${this.publicKey}@${this.getGrpcSocket()} to ${
```
thomaseizinger: Log statement still contains the gRPC socket but you are using the LN socket.

api_tests/lib_sdk/wallets/bitcoin.ts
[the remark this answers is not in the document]
thomaseizinger: Not really. We are mining 101 blocks in one go outside in the harness.
If the blockheight is not 101, there is a bug.

Waiting would only cause the test suite to hang for a very long time because we only mine 1 block per second.

api_tests/lib_sdk/wallets/index.ts
```diff
@@ -42,16 +50,36 @@ export class Wallets {
...
         switch (name) {
             case "ethereum":
                 this.wallets.ethereum = new EthereumWallet(
                     global.ledgerConfigs.ethereum
                 );
                 break;
             case "bitcoin":
-                this.wallets.bitcoin = await BitcoinWallet.newInstance(
-                    global.ledgerConfigs.bitcoin
+                if (!this.wallets.bitcoin) {
```
thomaseizinger: Why do we suddenly need this check?

api_tests/lib_sdk/wallets/lightning.ts
```diff
@@ -0,0 +1,89 @@
...
+declare var global: HarnessGlobal;
+
+export class LightningWallet implements Wallet {
+    public static async newInstance(
+        bitcoinWallet: BitcoinWallet,
+        logger: Logger,
+        logDir: string,
+        actorConfig: E2ETestActorConfig
+    ) {
+        const bitcoindDataDir = global.bitcoind.getDataDir();
```
thomaseizinger: :mega: Don't access globals in here please.

api_tests/lib_sdk/wallets/index.ts
```diff
@@ -42,16 +50,36 @@ export class Wallets {
...
+                }
+                break;
+            case "lightning":
+                if (!this.wallets.bitcoin) {
+                    this.wallets.bitcoin = await BitcoinWallet.newInstance(
+                        global.ledgerConfigs.bitcoin
+                    );
+                }
+                this.wallets.lightning = await LightningWallet.newInstance(
+                    this.wallets.bitcoin,
```
thomaseizinger: The LN wallet could just have its own wallet, no need to reuse them.

api_tests/sanity/config.toml
```diff
@@ -1,2 +1,2 @@
-ledgers = ["bitcoin"]
+ledgers = ["bitcoin", "lightning"]
```
thomaseizinger: :mega: We shouldn't put this here if the ledgerrunner is not starting it.

api_tests/sanity/basic_lnd_payment.ts
```diff
@@ -1,24 +1,30 @@
...
-
-        await alice.startLnd();
-        await bob.startLnd();
+        await alice.sendRequest(
+            { ledger: LedgerKind.Lightning, asset: AssetKind.Bitcoin },
+            { ledger: LedgerKind.Bitcoin, asset: AssetKind.Bitcoin }
+        );
 
-        await alice.lnd.fund(); // Send to address: bitcoin wallet -> LND wallet
+        const alicePeers = await alice.wallets
```
thomaseizinger: Reaching this deep into objects makes the tests hard to read and brittle to refactoring of the internal structure.

I'd suggest to provide an accessor for the LN peers on `Actor`.

api_tests/lib_sdk/wallets/index.ts
D4nte: Because few lines below we may have already instantiated the bitcoin wallet for lightning purposes.

api_tests/lib_sdk/wallets/lightning.ts
D4nte: I am doing a separate PR to remove `bitcoind` from `GlobalHarness`.

api_tests/lib_sdk/wallets/index.ts
D4nte: I don't see the point in instantiated 2 different bitcoin wallets.

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -79,10 +79,10 @@ export class Actor {
     public lnd: Lnd;
 
     private constructor(
+        private readonly config: E2ETestActorConfig,
         private readonly logger: Logger,
-        private readonly cndInstance: CndInstance,
         private readonly logRoot: string,
-        private readonly actorConfig: E2ETestActorConfig
+        private readonly cndInstance: CndInstance
```
bonomat: Why did the order change?

api_tests/lib_sdk/wallets/index.ts
thomaseizinger: The LN wallet could just instantiate it's own wallet, and this would be gone then :man_shrugging:

api_tests/sanity/basic_lnd_payment.ts
D4nte: These are only temporary assertions to show that the PR does what it says it does.

Such assertions should be remove at the next PR that does the `addInvoice` and `sendPayments`.

api_tests/lib_sdk/wallets/index.ts
thomaseizinger: Avoiding conditionals doesn't convince you? :(

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -637,18 +660,31 @@ export class Actor {
...
     private async initializeDependencies() {
+        let lightning = false;
         for (const ledgerName of [
             this.alphaLedger.name,
             this.betaLedger.name,
         ]) {
-            await this.wallets.initializeForLedger(ledgerName);
+            if (ledgerName === "lightning") {
+                lightning = true;
+            }
```
bonomat:
```suggestion
            lightning = ledgerName === "lightning";
```

api_tests/lib_sdk/wallets/lightning.ts
```diff
@@ -0,0 +1,89 @@
...
+            );
+        }
+
+        const chainBalance = await this.inner.getChainBalance();
+        console.log("chainBalance", chainBalance);
+
+        const channelBalance = await this.inner.getChannelBalance();
+        console.log("channelBalance", channelBalance);
+
+        return new BigNumber(chainBalance).plus(channelBalance);
```
bonomat: Why is the on-chain balance relevant in this case?

D4nte: Because `getBalanceByAsset` is used to confirm the initial funding (line 38 of this file).

PR #2053, opened 2020-02-19
1844 simpler config

### LINKED ISSUE #1844: Minimal viable ledger configuration strategy

bonomat:
Resolves #1844

As discussed in the ticket, the goal is to 
* let cnd come up with some smart defaults: 
  * right now I default to regtest, in the future we might want to change this to mainnet
* let the user configure everything
  * you can find a full example in the e2e tests (ignoring blockchain explorers right now)
* I propose to manually call verify once the configs were loaded 
  * for this I'll create a follow-up ticket once this is merged.

What do you think?

### DIFF
--- a/cnd/src/config/file.rs
+++ b/cnd/src/config/file.rs
@@ -1,6 +1,10 @@
-use crate::config::{Bitcoin, Data, Ethereum, Network, Socket};
+use crate::{
+    config::{Bitcoind, Data, Network, Parity, Socket},
+    swap_protocols::ledger::ethereum,
+};
 use config as config_rs;
 use log::LevelFilter;
+use serde::{Deserialize, Serialize};
 use std::{ffi::OsStr, path::Path};
 
 /// This struct aims to represent the configuration file as it appears on disk.
@@ -16,6 +20,19 @@ pub struct File {
     pub logging: Option<Logging>,
     pub bitcoin: Option<Bitcoin>,
     pub ethereum: Option<Ethereum>,
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct Bitcoin {
+    #[serde(with = "crate::config::serde_bitcoin_network")]
+    pub network: bitcoin::Network,
+    pub bitcoind: Option<Bitcoind>,
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct Ethereum {
+    pub chain_id: ethereum::ChainId,
+    pub parity: Option<Parity>,
 }
 
 impl File {
@@ -112,7 +129,11 @@ pub enum None {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::config::Settings;
+    use crate::{
+        config::{Bitcoind, Parity, Settings},
+        swap_protocols::ledger::ethereum,
+    };
+    use reqwest::Url;
     use spectral::prelude::*;
     use std::{
         net::{IpAddr, Ipv4Addr},
@@ -182,11 +203,16 @@ dir = "/tmp/comit/"
 level = "Debug"
 
 [bitcoin]
-network = "mainnet"
-node_url = "http://example.com/"
+network = "regtest"
+
+[bitcoin.bitcoind]
+node_url = "http://localhost:18443/"
 
 [ethereum]
-node_url = "http://example.com/"
+chain_id = 17
+
+[ethereum.parity]
+node_url = "http://localhost:8545/"
 "#;
 
         let file = File {
@@ -209,11 +235,16 @@ node_url = "http://example.com/"
                 level: Some(Level::Debug),
             }),
             bitcoin: Some(Bitcoin {
-                network: bitcoin::Network::Bitcoin,
-                node_url: "http://example.com".parse().unwrap(),
+                network: bitcoin::Network::Regtest,
+                bitcoind: Some(Bitcoind {
+                    node_url: "http://localhost:18443".parse().unwrap(),
+                }),
             }),
             ethereum: Some(Ethereum {
-                node_url: "http://example.com".parse().unwrap(),
+                chain_id: ethereum::ChainId::regtest(),
+                parity: Some(Parity {
+                    node_url: "http://localhost:8545".parse().unwrap(),
+                }),
             }),
         };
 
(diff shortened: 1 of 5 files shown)


api_tests/dry/config.ts
```diff
@@ -0,0 +1,83 @@
...
+listen = ["/ip4/0.0.0.0/tcp/9939"]
+[http_api.socket]
+address = "0.0.0.0"
+port = 8000
+
+[http_api.cors]
+allowed_origins = "none"
+
+[data]
+dir = "/Users/bonomat/Library/Application Support/comit"
```
tcharding: Username and OSX specific path here mate.

api_tests/dry/config.ts
```diff
@@ -0,0 +1,83 @@
...
+level = "Info"
+
+[bitcoin]
+network = "regtest"
+
+[bitcoin.connector.bitcoind]
+node_url = "http://localhost:18443/"
+
+[ethereum]
+chain_id = "0x07"
```
tcharding: Did you mean this to be "0x11" (17 in hex), like in our code or is this intentionally an uncommon chain number?

api_tests/dry/config.ts
thomaseizinger: yeah that is going to fail on every other machine :sweat_smile:.

Not sure how to solve that best. so I am gonna just ask a few questions:

What exactly is the goal of this tests?

1. Testing that `--dump-config` exists?
2. Testing whether the output of `--dump-config` is a valid config? It is a good idea to do so, we currently lack tests for our CLI interface. But for that, it would be good enough to treat the config as opaque, write it to a temp file and test if cnd starts up correctly with its own defaults.
3. Testing the default values of the config file? We do that in Rust already, not sure if we want to test this level of detail on the outside, the problem with OS and user specific paths stems from doing that.

cnd/src/config/file.rs
```diff
@@ -167,26 +167,30 @@ mod tests {
...
-address = "127.0.0.1"
+address = "0.0.0.0"
 port = 8000
 
 [http_api.cors]
-allowed_origins = "all"
+allowed_origins = "none"
 
 [data]
-dir = "/tmp/comit/"
```
thomaseizinger: Why are test values updated? They shouldn't be.

cnd/src/config/mod.rs
```diff
@@ -28,11 +30,37 @@ pub struct Socket {
...
+            network: bitcoin::Network::Regtest,
+            bitcoind: Bitcoind::defualt(),
+        }
+    }
+}
+
+impl Default for Bitcoind {
+    fn default() -> Self {
+        Self {
+            node_url: "http://localhost:18443",
```
thomaseizinger: Shouldn't this depend on the network?

I think you probably can't use the `Default` trait here. If the user says `Network::Mainnet` but leaves this out, it should default to `http://localhost:8443`.

cnd/src/config/sanity.rs
```diff
@@ -0,0 +1,31 @@
+use crate::config::{Bitcoind, Parity};
+
+pub trait Sanity {
```
thomaseizinger: The verification is going to require network communication, which is async code. This trait doesn't account for that at the moment.

Also, it uses types from the `config` folder, which are meant to only act as the serialization/deserialization layer of the configuration system.

Currently, we sometimes re-use types from the `file` module in the `settings` module. This only happens for types that are literally the same, so we don't have to duplicate them unncessarily.

With this ticket, I think you will have to break that re-use and introduce specific types for each layer (with smarter defaults, that re-use gets impossible):

- `file` should model the config file as closely as possible. Use `Option` types for optional stuff etc, to have an easy time deriving `Deserialize`.
- `settings` should model the settings we want in our application, i.e. no unnecessary `Option`s etc

cnd/src/config/settings.rs
```diff
@@ -257,4 +257,21 @@ mod tests {
...
+            ..File::default()
+        };
+
+        let settings = Settings::from_config_file_and_defaults(config_file);
+
+        assert_that(&settings)
+            .is_ok()
+            .map(|settings| &settings.bitcoin.connectors)
+            .is_equal_to(Network {
+                listen: vec!["/ip4/0.0.0.0/tcp/9939".parse().unwrap()],
```
thomaseizinger: There seems to be a mixup here in what is compared. Does this even compile? :sweat_smile:

config.toml
```diff
@@ -0,0 +1,23 @@
+[network]
```
thomaseizinger: What is this file doing here?

cnd/src/config/mod.rs
```diff
@@ -28,11 +30,37 @@ pub struct Socket {
 pub struct Bitcoin {
     #[serde(with = "crate::config::serde_bitcoin_network")]
     pub network: bitcoin::Network,
+    pub bitcoind: Bitcoind,
```
thomaseizinger: This structure doesn't represent your config. I think the `connectors` is missing here.

cnd/src/main.rs
```diff
@@ -43,6 +43,9 @@ fn main() -> anyhow::Result<()> {
 
     let settings = read_config(&options).and_then(Settings::from_config_file_and_defaults)?;
 
+    settings.bitcoin.bitcoind.verify_connection_details();
```
thomaseizinger: What is the point of the `Sanity` trait abstraction, if you are not making use of it? (By that I mean, there is not function that abstracts over `Sanity` as in: `fn verify<C: Sanity>(connection: C)`)

These could just be functions :)

But I'd suggest a **less** object-oriented design where you don't delegate to the data but have separate functions here in `main` that allow you to pass the connection details of each ledger connector and it verify the connection:

```rust
async fn verify_bitcoind_connection(network: Network, url: Url) -> Result<...> {
   // do bitcoind verification
}

async fn verify_parity_connection(chain_id: u16, url: Url) -> Result<...> {
   // do parity verification
}
```

and then just call them sequentially:

```rust
verify_bitcoind_connection(settings.bitcoin.network, settings.bitcoin.connectors.bitcoind.node_url).await?;
verify_parity_connection(settings.ethereum.network, settings.ethereum.connectors.parity.node_url).await?;
```

api_tests/dry/config.ts
bonomat: > yeah that is going to fail on every other machine 😅.

I'm going to force everyone to create this folder or e2e tests will fail on your machine 
(PR to change config generation in cnd will follow) 🤪

> What exactly is the goal of this tests?

I realized that we don't test `--dump-config` at all. 
The goal was to test that cnd is able to dump a complete config file. It was only afterwards that I realized that we have extensive tests in rust to test that. 

I'll remove this test again as it does not add that much value. 

@thomaseizinger , regarding your point 2:
I think this would not test if cnd generated a complete config file.

cnd/src/config/mod.rs
bonomat: I opted for removing the `connectors`  thingy if you look at the example above (see your comment https://github.com/comit-network/comit-rs/pull/2053/files#r381681455)

cnd/src/config/sanity.rs
bonomat: I'm afraid that the this verification would broaden the scope of this ticket unnecessarily. 
I'll create a follow-up ticket for this and focus on the smart-default generation.

cnd/src/config/mod.rs
bonomat: Good point. I'll see how I can resolve this.

cnd/src/config/settings.rs
bonomat: 🙈🙉🙊

api_tests/dry/config.ts
thomaseizinger: >  I think this would not test if cnd generated a complete config file.

True, but I also didn't claim that :)

thomaseizinger: To test that, I would write a test in TS that parses the toml and verifies that certain sections are there.

cnd/src/config/mod.rs
thomaseizinger: Ah okay, I thought I still saw it somewhere. The reason I added it in my proposal in the ticket is because `bitcoin.bitcoind` reads a bit weird IMO. But I am fine either way, less noisy also has its advantages.

cnd/src/config/settings.rs
```diff
@@ -257,4 +308,38 @@ mod tests {
...
+            .is_equal_to(Bitcoin {
+                network: bitcoin::Network::Regtest,
+                bitcoind: Bitcoind {
+                    node_url: "http://localhost:18443".parse().unwrap(),
+                },
+            })
+    }
+
+    #[test]
+    fn ethereum_defaults() {
```
thomaseizinger: It would be good to have tests here for the case where the user specifies the network but not the URL.

They would catch the bug you have in `derive_url_ethereum` ;)

thomaseizinger:
Plus, you seemed to have branched of master because the first commit is the merge of the release branch into `master`.

Please rebase against `dev`!

While you are at it, please squash the two commits together. I usually run build every commit during rebase (with clippy) and having existing commits in the history that don't build hurts that. Thanks :)

Workflow tip: Setup a run configuration in CLion that executes clippy and use that instead of `check` while developing. Allows you to fix the problems as they happen and CLion's shortcuts make it a breeze to navigate through them.

cnd/src/config/settings.rs
bonomat: Please have a look at the new tests.

cnd/src/config/file.rs
```diff
@@ -237,4 +268,104 @@ node_url = "http://example.com/"
...
+            Bitcoin {
+                network: bitcoin::Network::Testnet,
+                bitcoind: Some(Bitcoind {
+                    node_url: Url::parse("http://example.com:18332").unwrap(),
+                }),
+            },
+            Bitcoin {
+                network: bitcoin::Network::Regtest,
+                bitcoind: Some(Bitcoind {
+                    node_url: Url::parse("http://example.com:18443").unwrap(),
```
tcharding: Is there any relevance to these port numbers?  Are they different for a reason?  Not super important though.

tcharding: Oh, these are the default Bitcoin ports per network.

thomaseizinger: Yes, those are the defaults used by `bitcoind`.

thomaseizinger:
@bonomat do you mind squashing the last commit into the previous one as-well? :)

Otherwise feel free to get it merged anytime by just removing the label!

bonomat:
@thomaseizinger : done :)

PR #2054, opened 2020-02-19
Remove lib folder from api_tests

da-kami:
Move code that is still used to `lib_sdk` folder. Remove unused code.

I put a separate commit that renames `lib_sdk` to `lib`.


api_tests/lib_sdk/wallets/ethereum.ts
```diff
@@ -2,15 +2,17 @@ import { BigNumber, EthereumWallet as EthereumWalletSdk } from "comit-sdk";
...
 import { pollUntilMinted, Wallet } from "./index";
-import { HarnessGlobal } from "../../lib/util";
 import { TransactionRequest } from "ethers/providers";
 import * as fs from "fs";
-import { sleep } from "../utils";
+import { HarnessGlobal, sleep } from "../utils";
 
 declare var global: HarnessGlobal;
 
+export interface EthereumNodeConfig {
```
da-kami: That should probably be moved to an `ethereum.ts` file inside the `ledgers` folder rather than moving it in here.

D4nte:
Go for incremental steps please.

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -3,17 +3,16 @@ import { BigNumber, Cnd, ComitClient, Swap } from "comit-sdk";
 import { parseEther } from "ethers/utils";
 import getPort from "get-port";
 import { Logger } from "log4js";
-import { E2ETestActorConfig } from "../../lib/config";
-import { LedgerConfig } from "../../lib/ledger_runner";
-import "../../lib/setup_chai";
+import { E2ETestActorConfig } from "../ledgers/config";
```
D4nte: Should `E2ETestActorConfig` really be in `./ledgers/config` ? I'd see it more somewhere in the actor module. Or maybe we need to rename?

api_tests/lib_sdk/utils.ts
```diff
@@ -4,6 +4,19 @@ import { SwapRequest } from "comit-sdk";
 import * as fs from "fs";
 import { promisify } from "util";
```
D4nte: Can you merge all the imports together by removing this new line and putting the comment on the same line than the import (ie, on the right).

api_tests/lib_sdk/ledgers/config.ts
```diff
@@ -1,7 +1,7 @@
 import * as tmp from "tmp";
```
D4nte: I am not sure I agree with the `config.ts` file going under `ledgers`. For example it contains the `E2EActorConfig` I already mentioned but it also contains the `CndConfigFile` which is not related to a ledger.

I'd put the file in `lib_sdk/` for now and maybe do a 2nd PR where you split the config file (as it looks like something you'd like to do).

da-kami: good point, thanks

PR #2055, opened 2020-02-19
e2e setup for lnd: open channel, add invoice, pay invoice, assert settlement

### LINKED ISSUE #1951: Extend our e2e test suite with Lightning nodes

D4nte:
1. Automatically opens a channel in the right direction when doing `Actor.sendRequest()` with a `Lightning` ledger.
1. Bob adds an invoice
1. Alice pays the invoice
1. Bob verifies the invoice is paid

Resolves #1951.

### DIFF
--- a/api_tests/lib_sdk/actors/actor.ts
+++ b/api_tests/lib_sdk/actors/actor.ts
@@ -6,7 +6,7 @@ import { Logger } from "log4js";
 import { E2ETestActorConfig } from "../../lib/config";
 import { LedgerConfig } from "../../lib/ledger_runner";
 import "../../lib/setup_chai";
-import { Asset, AssetKind } from "../asset";
+import { Asset, AssetKind, toKey, toKind } from "../asset";
 import { CndInstance } from "../cnd_instance";
 import { Lnd } from "../lnd";
 import { Ledger, LedgerKind } from "../ledger";
@@ -73,8 +73,8 @@ export class Actor {
     private betaLedger: Ledger;
     private betaAsset: Asset;
 
-    private readonly startingBalances: Map<AssetKind, BigNumber>;
-    private readonly expectedBalanceChanges: Map<AssetKind, BigNumber>;
+    private readonly startingBalances: Map<string, BigNumber>;
+    private readonly expectedBalanceChanges: Map<string, BigNumber>;
 
     public lnd: Lnd;
 
@@ -115,7 +115,10 @@ export class Actor {
         }
 
         this.alphaLedger = defaultLedgerDescriptionForLedger(alphaLedgerKind);
-        this.alphaAsset = defaultAssetDescriptionForAsset(alphaAssetKind);
+        this.alphaAsset = defaultAssetDescription(
+            alphaAssetKind,
+            alphaLedgerKind
+        );
         to.alphaLedger = this.alphaLedger;
         to.alphaAsset = this.alphaAsset;
 
@@ -144,7 +147,7 @@ export class Actor {
         }
 
         this.betaLedger = defaultLedgerDescriptionForLedger(betaLedgerKind);
-        this.betaAsset = defaultAssetDescriptionForAsset(betaAssetKind);
+        this.betaAsset = defaultAssetDescription(betaAssetKind, betaLedgerKind);
         to.betaLedger = this.betaLedger;
         to.betaAsset = this.betaAsset;
 
@@ -164,10 +167,18 @@ export class Actor {
 
         await this.setStartingBalance([
             this.alphaAsset,
-            { name: this.betaAsset.name, quantity: "0" },
+            {
+                name: this.betaAsset.name,
+                ledger: this.betaLedger.name,
+                quantity: "0",
+            },
         ]);
         await to.setStartingBalance([
-            { name: to.alphaAsset.name, quantity: "0" },
+            {
+                name: to.alphaAsset.name,
+                ledger: this.alphaLedger.name,
+                quantity: "0",
+            },
             to.betaAsset,
         ]);
 
@@ -204,11 +215,11 @@ export class Actor {
         }
 
         this.expectedBalanceChanges.set(
-            betaAssetKind,
+            toKey(this.betaAsset),
             new BigNumber(this.betaAsset.quantity)
         );
         to.expectedBalanceChanges.set(
-            alphaAssetKind,
+            toKey(this.alphaAsset),
             new BigNumber(to.alphaAsset.quantity)
         );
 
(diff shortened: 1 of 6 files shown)


api_tests/types/ln-service/index.d.ts
```diff
@@ -123,4 +123,202 @@ declare module "ln-service" {
...
+    interface CreateInvoiceParams {
+        cltv_delta?: number;
+        description?: string;
+        expires_at?: string; // ISO 8601
+        is_fallback_included?: boolean;
+        is_fallback_nested?: boolean;
+        is_including_private_channels?: boolean;
+        lnd: AuthenticatedLndGrpc;
+        log?: any; // Required when WSS is passed
+        secret?: string; // hex
```
tcharding:
```suggestion
        secret?: string; // Hex
```

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -846,27 +878,26 @@ function defaultLedgerDescriptionForLedger(ledger: LedgerKind): Ledger {
...
-
-function defaultAssetDescriptionForAsset(asset: AssetKind): Asset {
+function defaultAssetDescription(asset: AssetKind, ledger: LedgerKind): Asset {
     switch (asset) {
         case AssetKind.Bitcoin: {
             return {
                 name: AssetKind.Bitcoin,
-                quantity: "100000000",
+                ledger,
+                quantity: "10000000",
```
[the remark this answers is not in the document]
D4nte: > Can we use "10_000_000" here? (Not sure what %review this)

We did not say all reviews need to be flagged 10/50/99 so if there is no flag and it is not a draft PR you can assume it's a normal review.

api_tests/lib_sdk/asset.ts
```diff
@@ -10,3 +12,20 @@ export enum AssetKind {
...
+        case "bitcoin-on-bitcoin":
+            return { asset: AssetKind.Bitcoin, ledger: LedgerKind.Bitcoin };
+        case "bitcoin-on-lightning":
+            return { asset: AssetKind.Bitcoin, ledger: LedgerKind.Lightning };
+        case "ether-on-ethereum":
+            return { asset: AssetKind.Ether, ledger: LedgerKind.Ethereum };
+        case "erc20-on-ethereum":
+            return { asset: AssetKind.Erc20, ledger: LedgerKind.Ethereum };
+    }
+}
```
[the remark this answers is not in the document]
D4nte: I based myself on the existing code. Undefined will be returned and something will break down the line which will be enough for us to investigate.

api_tests/lib_sdk/actors/actor.ts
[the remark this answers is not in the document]
D4nte: No, we cannot:
```
  1) rfc003-btc-eth-erc20-alice-redeems-bob-redeems:
     TypeError: toBitcoin must be called on a whole number or string format whole number
      at Object.toBitcoin (node_modules/satoshi-bitcoin/index.js:42:13)
      at BitcoinWallet.<anonymous> (lib_sdk/wallets/bitcoin.ts:53:13)
      at Generator.next (<anonymous>)
      at fulfilled (lib_sdk/wallets/bitcoin.ts:5:58)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

api_tests/lib_sdk/lnd.ts
```diff
@@ -182,31 +188,49 @@ export class Lnd {
...
+        );
+        return lnService.createInvoice({
+            lnd: this.authenticatedLndGrpc,
+            tokens: quantity,
+        });
     }
 
-    public async assertChannelBalanceReceiver() {
-        await sleep(1);
+    public pay(invoice: string) {
```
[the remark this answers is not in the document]
D4nte: Because I did not name the return type of `ln-service.pay` but I named the return type of the other function you are looking at.
Return types are not needed in typescripts and as a team we haven't chosen a convention about it.

api_tests/lib_sdk/wallets/lightning.ts
```diff
@@ -82,4 +82,73 @@ export class LightningWallet implements Wallet {
...
+            .is_synced_to_chain;
+        const toIsSynced = (await toWallet.inner.getWalletInfo())
+            .is_synced_to_chain;
+
+        while (!thisIsSynced || !toIsSynced) {
+            this.logger.info(
+                `One of the lnd node is not yet synced, waiting. this: ${thisIsSynced}, to: ${toIsSynced}`
+            );
+            await sleep(500);
+        }
```
bonomat: 📣
I think you have to to recheck if it is synched in the loop? Otherwise this will just keep looping.

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -177,17 +188,38 @@ export class Actor {
...
+
+            await thisLightningWallet.addPeer(toLightningWallet);
+
+            if (this.alphaLedger.name === "lightning") {
+                // Alpha Ledger is lightning so Alice will be sending assets over lightning
+                const quantity = parseInt(this.alphaAsset.quantity, 10);
+                await thisLightningWallet.openChannel(
+                    toLightningWallet,
+                    quantity
+                );
```
bonomat: 📣
My LN knowledge is a bit rusty so please excuse if this is wrong: 
I think that you can't fully _empty_ a LN channel but there has always to remain some satoshis left. This means, you will need to open the channel with a higher balance than what you want to send later on.

(There is an attack vector if this would be possible).

bonomat: Ignore this if your tests run.

D4nte: Yes good point, it is not an issue right now as we choose an arbitrary number for the invoice (20000) but it would have been one once we actually implement the swap.Well done.

D4nte: > Ignore this if your tests run.

Yes see my comment above, it works now but it would have been an issue once someone try to actually use this for a swap.

PR #2060, opened 2020-02-20
LND sanity test follow-up proposal

thomaseizinger:
See commit messages.

https://github.com/comit-network/comit-rs/commit/b4881718bbc704061ba514d3098f04236ae4b7ca may be controversial because I am removing the comparison of the `tokens`. I couldn't really figure out, what these tokens are.

@D4nte What was the reason for asserting them? I left them out because they would bloat the interface of `createInvoice` if I would return them as-well.
My assumption was: if I give an ID to LND and it tells me it is confirmed, I don't need to check anything else?

### DIFF
--- a/api_tests/lib/actors/actor.ts
+++ b/api_tests/lib/actors/actor.ts
@@ -15,7 +15,6 @@ import { Wallet, Wallets } from "../wallets";
 import { Actors } from "./index";
 import { Entity } from "../../gen/siren";
 import { SwapDetails } from "comit-sdk/dist/src/cnd";
-import { CreateInvoiceResponse } from "ln-service";
 
 declare var global: HarnessGlobal;
 
@@ -843,12 +842,15 @@ export class Actor {
         return this.wallets.lightning.createInvoice(sats);
     }
 
-    public async payLnInvoice(invoice: CreateInvoiceResponse) {
-        return this.wallets.lightning.pay(invoice);
+    public async payLnInvoice(request: string) {
+        return this.wallets.lightning.pay(request);
     }
 
-    public async assertLnInvoiceSettled(invoice: CreateInvoiceResponse) {
-        await this.wallets.lightning.assertInvoiceSettled(invoice);
+    public async assertLnInvoiceSettled(id: string) {
+        const resp = await this.wallets.lightning.getInvoice(id);
+        if (!resp.is_confirmed) {
+            throw new Error(`Invoice ${id} is not confirmed}`);
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


D4nte:
> See commit messages.
> 
> [b488171](https://github.com/comit-network/comit-rs/commit/b4881718bbc704061ba514d3098f04236ae4b7ca) may be controversial because I am removing the comparison of the `tokens`. I couldn't really figure out, what these tokens are.
> 
> @D4nte What was the reason for asserting them? I left them out because they would bloat the interface of `createInvoice` if I would return them as-well.
> My assumption was: if I give an ID to LND and it tells me it is confirmed, I don't need to check anything else?

IMHO, this is bikeshedding because we will not assert whether an invoice is paid in the final version but only if cnd consider the invoice paid and if the balance of the ln wallet is as expected, as we do today for the other ledgers & assets.
This assertion is only temporary here to prove that issue #1951 is done.

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -838,6 +838,21 @@ export class Actor {
             return await this.pollSwapDetails(swapUrl, iteration);
         }
     }
+
+    public async createLnInvoice(sats: number) {
```
D4nte: IMHO this is also bikeshedding because in the final version this will become useless/to be removed as we expect the SDK to do the creation of a hodl invoice. The e2e should not be calling this function.

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -838,6 +838,21 @@ export class Actor {
...
+        return this.wallets.lightning.createInvoice(sats);
+    }
+
+    public async payLnInvoice(request: string) {
```
D4nte: 🚲🏠

api_tests/lib_sdk/actors/actor.ts
```diff
@@ -838,6 +838,21 @@ export class Actor {
...
+        return this.wallets.lightning.pay(request);
+    }
+
+    public async assertLnInvoiceSettled(id: string) {
```
D4nte: 🚲🏠

thomaseizinger:
> I apologies if the comments sounds harsh but all this code is throw away and hence I am surprised you'd spend time on it.

Fair point, I didn't consider that it is going to be thrown away. I've noticed that this test doesn't fit into how we do the other tests so I opted for sending a follow-up to align it.

To be fair, it is not as use-less as you are making it sound: for example even in the actual LN-swaps tests, we are going to assert if an invoice is paid and IMO, that belongs in the actor class and not the wallet.

D4nte:
> even in the actual LN-swaps tests, we are going to assert if an invoice is paid 

I don't think we are going to make such assertions, the same way that today we do not assert whether funding/redeem transactions are confirmed and successful.

Issue #2037, opened 2020-02-17
Relax trait bounds on structs


thomaseizinger:
## Problem

Traits in Rust are used to abstract over functionality.
`impl` blocks in Rust are used to define functionality.

We have many places in the code base where we define a trait bound on the declaration site of a struct. This makes the trait bound bubble up to all the usage sites of the struct, even if they don't use it.

If you try to split rfc003 in han/herc20 then you'd need 2 ledger traits which then makes things hard.
In several places, we use the `Ledger` trait to define the types of fields using its associated types (such as `L::Identity`).

We keep the association of `Ledger` to `Identity` but we move it from type association to implementation.

## Goal

Less trait bounds and as a result of that, hopefully simpler code.

## Recommendation

We cannot get rid of all usages of `Ledger` on the struct level but we can get rid of some. Specifically, I recommend the following scope for _this_ issue:

Remove the `Ledger` trait bounds from all structs that _only_ use it access associated types. For example:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
pub struct Funded<L: Ledger, A: Asset> {
    pub transaction: L::Transaction,
    pub asset: A,
}
```

should turn into:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
pub struct Funded<T, A: Asset> {
    pub transaction: T,
    pub asset: A,
}
```

D4nte:
Please see the linked PR to see what has been done.
Note that the following `struct`/`enum` are yet to be done and will be handled with `2065`

- `Request`
- `SwapCommunication` (depends on `Request` being done)
- `alice::actions::State` and `bob::actions::State` (depends on `SwapCommunication` being done)
- `OngoingSwap` (see attempt #2069)

Finally, I think it would be worth considering the `ActorState` trait too.

PR #2064, opened 2020-02-20
Relax trait bounds on structs

D4nte:
Time for a new PR.

`Request` needs to keep the `Ledger` trait bounds due to the fact that it contains the ledger.
Hopefully, with the REST API routes we introduce with lightning, this will go away.
This will then free a bunch of other structs.

Works Towards #2037

### DIFF
--- a/cnd/src/db/save.rs
+++ b/cnd/src/db/save.rs
@@ -371,21 +371,12 @@ struct InsertableEthereumBitcoinAcceptMessage {
     bitcoin_refund_identity: Text<bitcoin::PublicKey>,
 }
 
-#[impl_template]
 #[async_trait]
-impl
-    Save<
-        Accept<
-            Ethereum,
-            ((
-                ledger::bitcoin::Mainnet,
-                ledger::bitcoin::Testnet,
-                ledger::bitcoin::Regtest,
-            )),
-        >,
-    > for Sqlite
-{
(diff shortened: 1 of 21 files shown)


cnd/src/db/save.rs
```diff
@@ -371,21 +371,12 @@ struct InsertableEthereumBitcoinAcceptMessage {
...
-            ((
-                ledger::bitcoin::Mainnet,
-                ledger::bitcoin::Testnet,
-                ledger::bitcoin::Regtest,
-            )),
-        >,
-    > for Sqlite
-{
-    async fn save(&self, message: Accept<Ethereum, __TYPE0__>) -> anyhow::Result<()> {
+impl Save<Accept<crate::ethereum::Address, crate::bitcoin::PublicKey>> for Sqlite {
```
thomaseizinger: OMG SO NICE!

thomaseizinger: This is exactly the kind of benefits I expected from these changes.
Basically impossible to see upfront but things turn out super nicely :)

thomaseizinger:
So, who are you pair programming this with Franck? :D

thomaseizinger:
> Request needs to keep the Ledger trait bounds due to the fact that it contains the ledger.

I have an idea of how to work around that!

PR #2068, opened 2020-02-21
Relax trait bound `SwapEvent`

### LINKED ISSUE #2037: Relax trait bounds on structs

D4nte:
Doing a separate PR as this commit may be seen as controversial.

The complexity can be decreased by introducing types.

I think this is the way forward as ultimately, we want to remove the `Ledger` trait bound as with han/herc20 we will have separate (or no) `Ledger` traits.

Resolves #2037.

### DIFF
--- a/cnd/src/swap_protocols/rfc003/create_swap.rs
+++ b/cnd/src/swap_protocols/rfc003/create_swap.rs
@@ -101,12 +101,9 @@ pub async fn create_swap<D, A: ActorState>(
 /// Returns a future that waits for events on alpha ledger to happen.
 ///
 /// Each event is yielded through the controller handle (co) of the coroutine.
-#[allow(clippy::type_complexity)]
 async fn watch_alpha_ledger<D, AL, AA, BL, BA>(
     dependencies: &D,
-    co: &Co<
-        SwapEvent<AL::HtlcLocation, AL::Transaction, BL::HtlcLocation, BL::Transaction, AA, BA>,
-    >,
+    co: &Co<SwapEventOnLedger<AL, BL, AA, BA>>,
     htlc_params: HtlcParams<AL, AA, AL::Identity>,
     start_of_swap: NaiveDateTime,
 ) -> anyhow::Result<()>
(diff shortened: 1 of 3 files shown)


cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -289,34 +305,34 @@ impl<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> OngoingSwap<AL, BL, AA, BA> {
 }
 
 #[derive(Debug, Clone, PartialEq, strum_macros::Display)]
-pub enum SwapEvent<AL, BL, AA, BA>
+pub enum SwapEvent<AH, AT, BH, BT, AA, BA>
```
thomaseizinger: This looks ugly but should ultimately get better as soon as it is going to be split into two!

cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -138,9 +151,12 @@ where
 /// Returns a future that waits for events on beta ledger to happen.
 ///
 /// Each event is yielded through the controller handle (co) of the coroutine.
+#[allow(clippy::type_complexity)]
 async fn watch_beta_ledger<D, AL, AA, BL, BA>(
```
thomaseizinger: As this is a private fn, you could potentially introduce more type parameters. Not sure if it helps.

D4nte: Similar to #2069 this would need to link associated types of one type param with another type param:
```
172 |     co.yield_(SwapEvent::BetaDeployed(deployed.clone())).await;
    |                                       ^^^^^^^^^^^^^^^^ expected type parameter `BT`, found associated type
```

I'll go for a type alias instead and stop here.

Issue #2065, opened 2020-02-20
Remove the `Into<LedgerKind>` trait bounds from `Ledger`


thomaseizinger:
## Problem

Traits bounds on structs are bad as they bubble up through the whole call chain and cause unnecessarily complex code. https://github.com/comit-network/comit-rs/pull/2064 shows some good examples of how things can get massively simpler if the trait bounds on structs are removed.

The PR also identified that one of the problems is the `rfc003::Request` struct where you cannot remove the bounds at the moment.

## Goal

Remove the `Into<LedgerKind>` trait bounds from `Ledger`.

## Recommendation

The reason why this struct needs the trait bounds is because of this function:

https://github.com/comit-network/comit-rs/blob/96241ea8209aa92332461967aa28a8ec765213fe/cnd/src/network/mod.rs#L871-L898

In the body, we call `into()` on the ledger, which calls the `Into<LedgerKind>` trait that allows us to convert the struct into an `OutgoingRequest`.

To solve this, we need to employ a different strategy of creating the `OutboundRequest` instance.

This [tweet](https://twitter.com/sgrif/status/1227044885770821632?s=20) tells us how to do that. I hereby propose the following:

Instead of having our own function, `build_outbound_request`, we are going to make use of the `TryInto` trait.

This line:

https://github.com/comit-network/comit-rs/blob/96241ea8209aa92332461967aa28a8ec765213fe/cnd/src/network/mod.rs#L730

needs to change into:

```rust
let request = request.try_into().expect("...");
```

The function signature needs to be extended with:

```rust
where rfc003::Request<AL, BL, AA, BA>: TryInto<OutboundRequest>
```

This is the only thing that needs to "bubble" up, which is only in two places:

- facade
- post_swap

After that, you can hit compile and the compiler will tell you, which impl blocks are missing.
Add an impl-block of:

```rust
impl TryInto<OutboundRequest> for rfc003::Request<bitcoin::Mainnet, ...> {

}
```
for every combination that you need.

Finally, you should be able to remove the trait bounds from the `rfc003::Request` struct.
Note: If you can remove the `Ledger` trait bounds from `rfc003::Request` then it's great but not compulsory.
Note: The removal of `Ledger` trait is tracked with #2081.

D4nte:
Can you please clarify if we do as part of this scope or in a new issue the following structs and traits? See https://github.com/comit-network/comit-rs/issues/2037#issuecomment-589486805 for details.

- `SwapCommunication`
[...]

PR #2069, opened 2020-02-21
Remove `Ledger` trait bound from `OngoingSwap` struct

D4nte:
Cannot be merged until #2068 is merged and the base branch is set to dev.

Works towards #2037.

### DIFF
--- a/cnd/src/swap_protocols/rfc003/create_swap.rs
+++ b/cnd/src/swap_protocols/rfc003/create_swap.rs
@@ -22,6 +22,8 @@ use genawaiter::{
     sync::{Co, Gen},
     GeneratorState,
 };
+use serde::{de::DeserializeOwned, Serialize};
+use std::{fmt::Debug, hash::Hash};
 
 /// Returns a future that tracks the swap negotiated from the given request and
 /// accept response on both ledgers.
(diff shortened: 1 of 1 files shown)


cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -251,20 +251,45 @@ where
...
-impl<AL: Ledger, BL: Ledger, AA: Asset, BA: Asset> OngoingSwap<AL, BL, AA, BA> {
-    pub fn new(
-        request: Request<AL, BL, AA, BA>,
-        accept: Accept<AL::Identity, BL::Identity>,
-    ) -> Self {
+impl<AL, BL, AI, BI, AA, BA> OngoingSwap<AL, BL, AI, BI, AA, BA>
+where
+    AL: Ledger<Identity = AI>,
+    BL: Ledger<Identity = BI>,
+    AI: Clone
```
thomaseizinger: Ouch! :D

Should we stop here and do https://github.com/comit-network/comit-rs/issues/2065 first? That would allow you to omit these trait bounds here ...

D4nte: If you think #2065 helps (I haven't checked) then I agree to stop here.

D4nte:
As per @thomaseizinger's comment. Better do #2065 first.

PR #2072, opened 2020-02-21
Update libp2p to version 0.16

### LINKED ISSUE #1941: Update libp2p to latest version
## Problem

Libp2p has been updated to version 0.15.
This release includes the migration to `std::future::Future` as-well as several, bigger changes to APIs.

## Goal

No outdated dependencies.

## Recommendation

Update our use of libp2p to version 0.15 (or whatever is the latest version when this is picked up).
There are several things that need to be considered according to the [changelog](https://github.com/libp2p/rust-libp2p/blob/master/CHANGELOG.md#version-0150-2020-01-24):

- migration to async/await (meaning we need to remove all usages of the compatibility layer)
- the `SwarmBuilder` now has an `executor` function to replace the executor that is in use by default. our async stack is [tokio](https://github.com/comit-network/spikes/blob/master/0018-async-stack.adoc), hence we should plug a tokio executor here.
- we should also activate the correct feature flags for the `libp2p-tcp` crate to use the tokio futures instead of the `async-std` ones.

Note that the goal here is to avoid bringing `async-std` into our dependency tree (check with `cargo tree`). It seems like it is not possible to completely achieve that as mdns [still depends on `async-std` without providing a feature-flag](https://github.com/libp2p/rust-libp2p/pull/1402#issuecomment-578407139).

thomaseizinger:
Resolves #1941.


libp2p-comit/src/handler.rs
```diff
@@ -201,442 +190,52 @@ impl<TSubstream: AsyncRead + AsyncWrite> ProtocolsHandler for ComitHandler<TSubs
...
         if let Some(event) = event {
-            return Some(Ok(Async::Ready(event)));
+            return Some(Poll::Ready(event));
         }
     }
     None
 }
-
-#[cfg(test)]
-mod tests {
```
thomaseizinger: Check the commit messages to learn why this is gone.

libp2p-comit/src/substream/inbound.rs
```diff
@@ -5,50 +5,52 @@ use crate::{
...
+};
 
 #[derive(strum_macros::Display)]
 #[allow(missing_debug_implementations)]
 /// States of an inbound substream i.e. from peer node to us.
-pub enum State<TSubstream> {
+pub enum State {
     /// Waiting for a request from the remote.
-    WaitingMessage { stream: Frames<TSubstream> },
+    WaitingMessage { stream: Pin<Box<Frames>> },
```
thomaseizinger: We have to `Pin` the stream to a specific location in memory.

The current code here does what async/await would do under the hood: It represents the state as an enum and each variant is a check (await) point.

async functions automatically use `Pin` if necessary.
We have to use it here, to be able to access the `poll` APIs on the stream.

thomaseizinger:
> Can you comment on this note in the original issue?

Sure!
I didn't manage to do this but I opened an issue on the libp2p repository for resolving this (https://github.com/libp2p/rust-libp2p/issues/1464).

PR #2089, opened 2020-02-25
Add Lightning routes

### LINKED ISSUE #2038: Add skeleton routes for Lightning swaps
## Problem

For cnd to support LN swaps, we need to provide an API. The current API model of the `/swaps/rfc003` endpoint is fairly complex because it used nested objects and the fields in the payload are not static; they depend on _values_ of other fields (the type of `alpha_ledger_redeem_identity` depends on the value of `alpha_ledger.name`).
Extending the current endpoint with support for LN will only make this more complicated.

## Goal

- Add support for LN to cnd
- Not make stuff more complicated

## Recommendation

1. Add routes as proposed in the spike (https://github.com/comit-network/spikes/blob/master/0022-lnd-plan-of-attack.adoc#215-our-current-rest-api-endpoints-are-not-capable-of-expressing-the-concept-of-a-protocol-per-ledger) to cnd.
1. Return a 400 BAD REQUEST with an error that says it is not supported at the moment.
1. Write e2e tests that assert this error.

Note: The current `rfc003` route will exist temporarily even with the new `han` and `herc20` routes to be added in a different issue.
Note: This is blocked on comit-network/comit-js-sdk#117 as it needs to be tested but it's not possible until the SDK is extended.
Note: Keep the body empty at this stage.

D4nte:
Resolves #2038.

### DIFF
--- a/api_tests/dry/lightning_routes.ts
+++ b/api_tests/dry/lightning_routes.ts
@@ -1,0 +1,63 @@
+// These are stateless tests -- they don't require any state of the cnd and they don't change it
+// They are mostly about checking invalid request responses
+import "chai/register-should";
+import "../lib/setup_chai";
+import { oneActorTest } from "../lib/actor_test";
+import { expect } from "chai";
+
+setTimeout(async function() {
+    describe("Lightning routes", () => {
+        oneActorTest(
+            "lightning-routes-post-eth-lnbtc-return-400",
+            async function({ alice }) {
+                const promise = alice.cnd.postHanEthereumEtherHalightLightningBitcoin();
+                return expect(promise).to.eventually.be.rejected.then(error => {
+                    expect(error).to.have.property(
+                        "message",
+                        "Request failed with status code 400"
+                    );
+                });
+            }
+        );
+
+        oneActorTest(
+            "lightning-routes-post-erc20-lnbtc-return-400",
+            async function({ alice }) {
+                const promise = alice.cnd.postHerc20EthereumErc20HalightLightningBitcoin();
+                return expect(promise).to.eventually.be.rejected.then(error => {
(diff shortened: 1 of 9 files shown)


api_tests/dry/lightning_routes.ts
```diff
@@ -0,0 +1,67 @@
...
+import { expect } from "chai";
+
+setTimeout(async function() {
+    describe("Lightning routes", () => {
+        oneActorTest(
+            "lightning-routes-post-eth-lnbtc-return-400",
+            async function({ alice }) {
+                return expect(
+                    alice.cnd.postHanEthereumEtherHalightLightningBitcoin()
+                ).to.eventually.be.rejected.then(error => {
```
thomaseizinger: This is actually not the assertion we want but we can't assert what we want until https://github.com/comit-network/comit-js-sdk/issues/122 is implemented.

I am leaning towards making this assertion do less so that we have more reasons to do the above ticket.

api_tests/dry/lightning_routes.ts
```diff
@@ -0,0 +1,67 @@
...
+import { oneActorTest } from "../lib/actor_test";
+import { expect } from "chai";
+
+setTimeout(async function() {
+    describe("Lightning routes", () => {
+        oneActorTest(
+            "lightning-routes-post-eth-lnbtc-return-400",
+            async function({ alice }) {
+                return expect(
+                    alice.cnd.postHanEthereumEtherHalightLightningBitcoin()
```
thomaseizinger: Nit: Can we have that on a separate line please? :see_no_evil: 

So that it is separated from the assertion..

D4nte: Yes. It had that initially but it took me many attempts to hack chai

PR #2093, opened 2020-02-26
Update logo to version without TM

da-kami:
Update logo to latest version from notion


da-kami:
> Slightly off-topic:
> Do we have to embed the logo as a file in the repo?
> 
> Could we not just serve that from the website or something?

I thought about this, the tough part is, that I extended the logo with text for other repos. In comit-rs we literally use the logo from notion, but for SDK, create-comit-app I added text. Not sure that was the best idea in the world. But my hope would be to fix this once and then we don't have to touch it again.

thomaseizinger:
bors retry

Issue #1952, opened 2020-01-28
Extend cnd's configuration to account for an lnd instance


thomaseizinger:
## Problem

In order to do Lightning swaps, cnd needs to talk to an lnd instance. It should therefore be possible to configure:

- which port lnd is listening on
- where to find the macaroon that is used for authentication

## Goal

A minimalistic configuration interface which allows Lightning to be configured.

## Recommendation

According to the Lightning spike we are only [going to use HTTP](https://github.com/comit-network/spikes/blob/lnd-plan-of-attack/0022-lnd-plan-of-attack.adoc#L387) for communicating with lnd.

- The configuration MUST make it explicit that we need the HTTP port.
- The configuration MUST have a default value being the default lnd HTTP port.

What about the macaroon?

We MUST must maintain the zero-config approach of cnd.
To get access to the macaroon, we have two options:

- The user places it in cnd's config directory.
- We try to find a specific maracroon in lnd's config directory.

Both are well-known locations and therefore could act as the default value for the macaroon path.

I'd **recommend** the following algorithm:

1. Check if our config file explicitly defines a macaroon
2. Check if we have a macaroon with a default name inside cnd's config directory
3. Check if `LND_DIR` is defined
a. If yes, try and read the macaroon from the config directory with a specific name (decide on it)
b. If no, then default to default lnd config dir and look for the macaroon
c. If everything fails, tell the user which options they have to configure the macaroon.

thomaseizinger:
This is blocked by https://github.com/comit-network/comit-rs/issues/1844 because we first need to sort out the existing ledger configuration to resolve the problem of knowing which network we want to connect to.

LND, similarly to bitcoind maintains a data directory per network. Hence, we need to resolve https://github.com/comit-network/comit-rs/issues/1844 first before this ticket can be implemented as the results apply here as-well.

tcharding:
Does this issue assume that `lnd` is running on the same machine as `cnd`?

thomaseizinger:
> Does this issue assume that `lnd` is running on the same machine as `cnd`?

You mean in regards to the default values? Yes you can assume both run on the same machine :)

tcharding:
> You mean in regards to the default values? Yes you can assume both run on the same machine :)

In the end I elected to use a `Socket` instead of just a port number so that the user can run `lnd` where ever they please.

Issue #2082, opened 2020-02-24
Complete the LN connector implementation


thomaseizinger:
## Problem

In order to execute LN swaps, we need to monitor the LN node.

## Goal

A working implementation of the interface defined in https://github.com/comit-network/comit-rs/issues/1954.

## Recommendation

https://github.com/comit-network/comit-rs/issues/1954 defines an interface for a LndCconnector.
The scope of this ticket is to implement this interface by calling the API endpoints described in the spike (https://github.com/comit-network/spikes/blob/master/0022-lnd-plan-of-attack.adoc#224-which-lnd-apis-do-we-need-to-target).

Note: (it's in the spike), REST API used to target lnd (only clients will need gRPC).
Note: we do not expect this to be tested until further down the line with e2e tests. Do not do testcontainers-rs setup to test this, only test manually once the code is done. Which is fine because there is a very little processing/filter logic to be applied here comparing to Ethereum/Bitcoin block/transaction matching logic.
Note: The API is likely to be async functions that resolve after several polling.

thomaseizinger:
This is blocked by https://github.com/comit-network/comit-rs/issues/1952 otherwise the connector to LND requires this configuration.

PR #2097, opened 2020-02-26
Enable lnd configuration

tcharding:
Enable cnd to be able to be configured to talk to an lnd instance.  With this applied we have:

```
[lightning]
network = "regtest"

[lightning.lnd]
rest_api_socket = "127.0.0.1:8080"
dir = "~/.lnd"
```

Add for the macaroon location logic we have two functions, one library function and a default caller on top of that
```
locate_macaroon()       \\ takes a vector of directories to look in
locate_macaroon_in_default_places()
```

The second function looks in the following places in an OS agnostic manner.  For Linux this would be:
```
    1. ~/.local/share/comit/
    2. ~/.lnd/data/chain/bitcoin/regtest/
```
The macaroon file name used is `readonly.macaroon`.

Fixes: #1952

### DIFF
--- a/api_tests/lib/config.ts
+++ b/api_tests/lib/config.ts
@@ -11,7 +11,7 @@ export interface CndConfigFile {
 }
 
 export interface HttpApi {
-    socket: { address: string; port: number };
+    socket: string;
 }
 
 export class E2ETestActorConfig {
@@ -36,10 +36,7 @@ export class E2ETestActorConfig {
     public generateCndConfigFile(ledgerConfig: LedgerConfig): CndConfigFile {
         return {
             http_api: {
-                socket: {
-                    address: "0.0.0.0",
-                    port: this.httpApiPort,
-                },
+                socket: `0.0.0.0:${this.httpApiPort}`,
             },
             data: {
                 dir: this.data,
--- a/cnd/Cargo.toml
+++ b/cnd/Cargo.toml
@@ -18,6 +18,7 @@ derivative = "1"
 diesel = { version = "1.4", features = ["sqlite", "chrono"] }
 diesel_migrations = "1.4.0"
 directories = "2.0"
+dirs = "2.0"
 ethbloom = "0.6.4"
 fern = { version = "0.5", features = ["colored"] }
 futures = { version = "0.1" }
--- a/cnd/Cargo.toml
+++ b/cnd/Cargo.toml
@@ -18,7 +18,6 @@ derivative = "1"
 diesel = { version = "1.4", features = ["sqlite", "chrono"] }
 diesel_migrations = "1.4.0"
 directories = "2.0"
-dirs = "2.0"
 ethbloom = "0.6.4"
 fern = { version = "0.5", features = ["colored"] }
 futures = { version = "0.1" }
--- a/cnd/src/config/file.rs
+++ b/cnd/src/config/file.rs
@@ -1,5 +1,5 @@
 use crate::{
-    config::{Bitcoind, Data, Network, Parity, Socket},
+    config::{Bitcoind, Data, Lnd, Network, Parity, Socket},
     swap_protocols::ledger::ethereum,
 };
 use config as config_rs;
@@ -20,6 +20,7 @@ pub struct File {
     pub logging: Option<Logging>,
     pub bitcoin: Option<Bitcoin>,
     pub ethereum: Option<Ethereum>,
+    pub lnd: Option<Lnd>,
 }
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
@@ -44,6 +45,7 @@ impl File {
             logging: Option::None,
             bitcoin: Option::None,
             ethereum: Option::None,
+            lnd: Option::None,
         }
     }
 
@@ -213,6 +215,13 @@ chain_id = 17
 
 [ethereum.parity]
 node_url = "http://localhost:8545/"
+
+[lnd]
+macaroon = "/home/jim/.lnd/data/chain/bitcoin/simnet/readonly.macaroon"
+
+[lnd.http_rpc_socket]
+address = "127.0.0.1"
+port = 443
 "#;
 
         let file = File {
@@ -245,6 +254,15 @@ node_url = "http://localhost:8545/"
                 parity: Some(Parity {
                     node_url: "http://localhost:8545".parse().unwrap(),
                 }),
+            }),
+            lnd: Some(Lnd {
+                http_rpc_socket: Socket {
+                    address: IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
+                    port: 443,
+                },
+                macaroon: Some(PathBuf::from(
+                    "/home/jim/.lnd/data/chain/bitcoin/simnet/readonly.macaroon",
+                )),
             }),
         };
 
@@ -364,6 +382,36 @@ node_url = "http://localhost:8545/"
             .into_iter()
             .map(toml::from_str)
             .collect::<Result<Vec<Ethereum>, toml::de::Error>>()
+            .unwrap();
+
+        assert_eq!(actual, expected);
+    }
+
+    #[test]
+    fn lnd_deserializes_correctly() {
+        let file_contents = vec![
+            r#"
+            macaroon = "/home/jim/.lnd/data/chain/bitcoin/simnet/readonly.macaroon"
+            [http_rpc_socket]
+            address = "127.0.0.1"
+            port = 443
+            "#,
+        ];
+
+        let expected = vec![Lnd {
+            http_rpc_socket: Socket {
+                address: IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
+                port: 443,
+            },
+            macaroon: Some(PathBuf::from(
+                "/home/jim/.lnd/data/chain/bitcoin/simnet/readonly.macaroon",
+            )),
+        }];
+
+        let actual = file_contents
+            .into_iter()
+            .map(toml::from_str)
+            .collect::<Result<Vec<Lnd>, toml::de::Error>>()
             .unwrap();
 
         assert_eq!(actual, expected);
--- a/cnd/src/config/file.rs
+++ b/cnd/src/config/file.rs
@@ -1,11 +1,11 @@
 use crate::{
-    config::{Bitcoind, Data, Lnd, Network, Parity, Socket},
+    config::{Bitcoind, Data, Lnd, Network, Parity},
     swap_protocols::ledger::ethereum,
 };
 use config as config_rs;
 use log::LevelFilter;
 use serde::{Deserialize, Serialize};
-use std::{ffi::OsStr, path::Path};
+use std::{ffi::OsStr, net::SocketAddr, path::Path};
 
 /// This struct aims to represent the configuration file as it appears on disk.
 ///
@@ -99,7 +99,7 @@ impl From<Level> for LevelFilter {
 
 #[derive(Clone, Debug, serde::Serialize, serde::Deserialize, PartialEq)]
 pub struct HttpApi {
-    pub socket: Socket,
+    pub socket: SocketAddr,
     pub cors: Option<Cors>,
 }
 
@@ -138,7 +138,7 @@ mod tests {
     use reqwest::Url;
     use spectral::prelude::*;
     use std::{
-        net::{IpAddr, Ipv4Addr},
+        net::{IpAddr, Ipv4Addr, SocketAddr},
         path::PathBuf,
     };
 
@@ -191,9 +191,8 @@ mod tests {
 [network]
 listen = ["/ip4/0.0.0.0/tcp/9939"]
 
-[http_api.socket]
-address = "127.0.0.1"
-port = 8000
+[http_api]
+socket = "127.0.0.1:8000"
 
 [http_api.cors]
 allowed_origins = "all"
@@ -217,11 +216,8 @@ chain_id = 17
 node_url = "http://localhost:8545/"
 
 [lnd]
+rest_api_socket = "127.0.0.1:8080"
 macaroon = "~/.lnd/data/chain/bitcoin/simnet/readonly.macaroon"
-
-[lnd.rest_api_socket]
-address = "127.0.0.1"
-port = 8080
 "#;
 
         let file = File {
@@ -229,10 +225,7 @@ port = 8080
                 listen: vec!["/ip4/0.0.0.0/tcp/9939".parse().unwrap()],
             }),
             http_api: Some(HttpApi {
-                socket: Socket {
-                    address: IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
-                    port: 8000,
-                },
+                socket: SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8000),
                 cors: Some(Cors {
                     allowed_origins: AllowedOrigins::All(All::All),
                 }),
@@ -256,10 +249,7 @@ port = 8080
                 }),
             }),
             lnd: Some(Lnd {
-                rest_api_socket: Socket {
-                    address: IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
-                    port: 8080,
-                },
+                rest_api_socket: SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8080),
                 macaroon: Some(PathBuf::from(
                     "~/.lnd/data/chain/bitcoin/simnet/readonly.macaroon",
                 )),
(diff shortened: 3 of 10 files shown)


cnd/src/config/file.rs
```diff
@@ -246,6 +255,15 @@ node_url = "http://localhost:8545/"
                     node_url: "http://localhost:8545".parse().unwrap(),
                 }),
             }),
+            lnd: Some(Lnd {
+                http_rpc_socket: Socket {
+                    address: IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
+                    port: 443,
+                },
+                macaroon: Some(PathBuf::from(
+                    "/home/jim/.lnd/data/chain/bitcoin/simnet/readonly.macaroon",
```
D4nte: Can we use a `~` instead?

cnd/src/config/file.rs
```diff
@@ -246,6 +255,15 @@ node_url = "http://localhost:8545/"
                     node_url: "http://localhost:8545".parse().unwrap(),
                 }),
             }),
+            lnd: Some(Lnd {
+                http_rpc_socket: Socket {
+                    address: IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
+                    port: 443,
+                },
+                macaroon: Some(PathBuf::from(
```
D4nte: Isn't the path to the tls cert needed too?

cnd/src/config/mod.rs
```diff
@@ -91,6 +94,28 @@ pub struct Parity {
...
+            http_rpc_socket: default_lnd_http_rpc_socket(),
+            macaroon: None,
+        }
+    }
+}
+
+fn default_lnd_http_rpc_socket() -> Socket {
+    Socket {
+        address: IpAddr::V4(Ipv4Addr::UNSPECIFIED),
+        port: 443,
```
D4nte:
```suggestion
        port: 8080,
```
I think the default port is actually 8080.

cnd/src/config/mod.rs
```diff
@@ -91,6 +94,28 @@ pub struct Parity {
     pub node_url: reqwest::Url,
 }
 
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct Lnd {
+    pub http_rpc_socket: Socket,
```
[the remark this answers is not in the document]
thomaseizinger: It is a bit tedious to require the user to specify the host if all they want to configure is the port. Should we make the address in `Socket` optional?

cnd/src/config/file.rs
thomaseizinger: Yes! Good point! We need the TLS certificate too :)

cnd/src/config/settings.rs
```diff
@@ -17,6 +20,7 @@ pub struct Settings {
     pub logging: Logging,
     pub bitcoin: Bitcoin,
     pub ethereum: Ethereum,
+    pub lnd: Option<Lnd>,
```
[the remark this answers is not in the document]
thomaseizinger: Note that at the moment, you are using the **bitcoin** network in the algorithm that looks for the macaroon but actually, you should be using the network that is configured here.

Example case:

A user might connect their Lightning and Ethereum nodes to mainnet and Bitcoin to testnet and only to LN to Ethereum swaps.

cnd/src/config/mod.rs
tcharding: TLS is on by default so I assumed the default was 443.  The default config suggests both.  I also thought using 443 makes it explicit that LND use TLS (perhaps with this in mind I should have noticed the TLS cert was missing :)  The LND REST API docs use 8080 as the example port so I will do the same.

thomaseizinger: TLS == 443 only applies to web servers I believe :)

cnd/src/config/settings.rs
```diff
@@ -393,4 +450,46 @@ mod tests {
                 })
         }
     }
+
+    #[test]
+    fn http_lnd_section_defaults_to_none() {
```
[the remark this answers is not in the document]
tcharding: Because we want to default to spinning up cnd without a LND node, right?

cnd/src/config/file.rs
tcharding: Done, thanks.

cnd/src/config/mod.rs
tcharding: > TLS == 443 only applies to web servers I believe :)

Cool, I did not know that.

cnd/src/lib.rs
```diff
@@ -81,15 +81,19 @@ pub fn data_dir() -> Option<PathBuf> {
     ProjectDirs::from("", "", "comit").map(|proj_dirs| proj_dirs.data_dir().to_path_buf())
 }
 
+#[cfg(target_os = "macos")]
+static LND_DIR_STEM: &str = "Lnd";
+#[cfg(target_os = "linux")]
+static LND_DIR_STEM: &str = ".lnd";
+#[cfg(target_os = "windows")]
```
thomaseizinger: We are not supporting windows anywhere else and I think also lnd is not supporting it. I would just opt to remove this.

cnd/src/lib.rs
```diff
@@ -81,15 +81,19 @@ pub fn data_dir() -> Option<PathBuf> {
...
+pub fn lnd_dir() -> Option<PathBuf> {
     match env::var("LND_DIR") {
-        Ok(dir) => PathBuf::from(dir),
-        Err(_) => {
-            // Just unwrap, if we don't have $HOME we are hosed anyway.
-            let home = dirs::home_dir().unwrap();
-            home.join(".lnd")
-        }
+        Ok(dir) => Some(PathBuf::from(dir)),
+        Err(_) => ProjectDirs::from("", "", LND_DIR_STEM)
```
thomaseizinger: Doesn't this evaluate to `$XDG_DATA_HOME/.lnd` or `$HOME/.local/share/.lnd` for Linux? According to the [docs](https://api.lightning.community/#lnd-grpc-api-reference), it should be `$HOME/.lnd` for Linux.

It looks like we'll need to do some messy conditional compiliation here ...
I would also recommend to add tests (that also have `cfg` applied) to make sure, we are actually resolving to the folders stated in the documentation.

cnd/Cargo.toml
```diff
@@ -18,6 +18,7 @@ derivative = "1"
 diesel = { version = "1.4", features = ["sqlite", "chrono"] }
 diesel_migrations = "1.4.0"
 directories = "2.0"
+dirs = "2.0"
```
thomaseizinger: I think we are no longer using that?

tcharding:
> Also, I noticed that you are not yet parsing / looking for the TLS certificate? I checked and realized that we actually didn't mention that in the issue, sorry about that!
> 
> We could do that as a follow-up PR if you want to get this one merged. Just make sure to record it in an issue then :)

With the move to using `dir` for lnd I didn't think it necessary to add the code for TLS cert.  It will likely involve doing the same thing as the macaroon, but since this PR already adds that (as dead code) and I already spend a bunch of time on it I feel its better to just leave TLS for when we actually use it.

cnd/src/config/settings.rs
```diff
@@ -218,51 +218,60 @@ impl Settings {
...
-        if let Some(lnd) = &self.lnd {
-            if let Some(macaroon) = &lnd.macaroon {
-                if macaroon.exists() {
-                    return Some(macaroon.clone());
-                }
-            }
+    /// Order is (using Linux as an example):
+    ///
+    ///  1. ~/.config/comit/
+    ///  2. ~/.config/comit/lnd/
```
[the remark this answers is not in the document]
tcharding: Fail!  The docs are wrong before they even merge, this should be `.local/share/comit`.

cnd/src/config/settings.rs
```diff
@@ -218,51 +218,60 @@ impl Settings {
...
+    ///  3. ~/.lnd/
```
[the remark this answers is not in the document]
tcharding: Not sure I understand the question.  This is the default directory used by lnd.  If the user wishes to use a macaroon path that is _not_ in the default place i.e., in the chain data, tree then this seems like an obvious place to put it?

If you prefer we can just look in `.local/cache/comit` and `~/.lnd/data/chain/bitcoin/regtest`.  This is actually probably enough options.

cnd/src/config/settings.rs
```diff
@@ -452,29 +455,32 @@ mod tests {
...
         assert_that(&settings)
             .is_ok()
-            .map(|settings| &settings.lnd)
-            .is_equal_to(None)
+            .map(|settings| &settings.lightning)
+            .is_equal_to(Lightning {
+                network: bitcoin::Network::Regtest,
+                lnd: Some(Lnd {
+                    rest_api_socket: Some(SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 8080)),
+                    dir: Some(PathBuf::from("~/.lnd")),
```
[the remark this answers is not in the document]
tcharding: My bad, tests should be portable too.  Will fix.

cnd/src/config/settings.rs
```diff
@@ -483,13 +489,73 @@ mod tests {
...
+                network: bitcoin::Network::Bitcoin,
+                lnd: Some(Lnd {
+                    rest_api_socket: Some(SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 8080)),
+                    dir: Some(PathBuf::from("~/.lnd")),
+                }),
+            })
+    }
+
+    #[test]
+    fn lnd_dir_defaults() {
```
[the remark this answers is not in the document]
tcharding: I agree they are verbose.  Each test however only has three statements, set up the `File` object, set up the `Settings` object, and assert they match.  I wanted to test all the combinations of the nested options because the code to write them, although quite simple,  took some thought to get the logic 'right'.

cnd/src/lib.rs
tcharding: cool.

cnd/src/lib.rs
tcharding: Man, some morning.  You are correct, this code is totally broken.

tcharding: So, the confusion is; on OSX lnd dir _is_ in the `$XDG_DATA_HOME` directory and on linux it is not.  Will fix.

cnd/Cargo.toml
tcharding: Nice, cheers.  Removed.

cnd/src/config/mod.rs
```diff
@@ -151,4 +161,86 @@ mod tests {
...
+
+    #[test]
+    fn lnd_deserializes_correctly() {
+        let file_contents = vec![
+            r#"
+            rest_api_socket = "127.0.0.1:8080"
+            dir = "~/.lnd"
+            "#,
+            r#"
+            rest_api_socket = "0.0.0.0:8080"
```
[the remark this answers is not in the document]
tcharding: Of course, that's correct.  Sloppy work today.  Thanks for catching this.

tcharding:
This PR introduces a dependency in the e2e tests between `lnd` and `cnd`, the configuration of `lnd` must be known _before_ starting `cnd` because some of this config info is currently dynamic and generated by `lnd` on startup.  This should not be a problem yet because we don't actually talk to `lnd` from `cnd` in the e2e tests.

cnd/src/config/settings.rs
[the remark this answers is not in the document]
thomaseizinger: >  This is actually probably enough options.

Yeah that is what I was thinking:

Either the user leaves it in the default place, puts it in the comit config folder or directly specifies a path.

cnd/src/config/settings.rs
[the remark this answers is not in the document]
thomaseizinger: Fair enough :)

cnd/src/lib.rs
thomaseizinger: Yeah it is a mess unfortunately :/
But we can't really do anything about that (except maybe sending a PR to lnd :D)

thomaseizinger:
> > Also, I noticed that you are not yet parsing / looking for the TLS certificate? I checked and realized that we actually didn't mention that in the issue, sorry about that!
> > We could do that as a follow-up PR if you want to get this one merged. Just make sure to record it in an issue then :)
> 
> With the move to using `dir` for lnd I didn't think it necessary to add the code for TLS cert. It will likely involve doing the same thing as the macaroon, but since this PR already adds that (as dead code) and I already spend a bunch of time on it I feel its better to just leave TLS for when we actually use it.

Fair enough!
We are going to need the TLS cert for https://github.com/comit-network/comit-rs/issues/2082 so either the scope of that ticket needs to be extended or we need a new one :)
What I would like to avoid is a surprising scope creep in https://github.com/comit-network/comit-rs/issues/2082.

thomaseizinger:
> This PR introduces a dependency in the e2e tests between `lnd` and `cnd`, the configuration of `lnd` must be known _before_ starting `cnd` because some of this config info is currently dynamic and generated by `lnd` on startup. This should not be a problem yet because we don't actually talk to `lnd` from `cnd` in the e2e tests.

Good catch!
Same as with the TLS cert, if work remains to be done after this PR, please capture it somewhere to minimize the surprises later.

api_tests/lib/config.ts
```diff
@@ -11,7 +11,7 @@ export interface CndConfigFile {
 }
 
 export interface HttpApi {
-    socket: { address: string; port: number };
```
thomaseizinger: Don't forget to write a changelog update for this breaking change :)

cnd/src/config/mod.rs
```diff
@@ -112,18 +116,15 @@ pub struct Lnd {
...
     }
 }
 
-fn default_lnd_rest_api_socket() -> SocketAddr {
-    SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 8080)
-}
-
 fn default_lnd_dir() -> PathBuf {
-    PathBuf::from("~/.lnd")
+    // Only fails if we don't have a home directory.
```
thomaseizinger: Can use `.expect` for that :)

cnd/src/lib.rs
```diff
@@ -81,20 +81,25 @@ pub fn data_dir() -> Option<PathBuf> {
     ProjectDirs::from("", "", "comit").map(|proj_dirs| proj_dirs.data_dir().to_path_buf())
 }
 
+/// Returns `/Users/[username]/Library/Application Support/Lnd/`.
```
thomaseizinger: Should we just inline this maybe?
If we use `cfg = macos` anyway, I don't really see a point in using generic code if we already know what we want it to be :)

tcharding: We use these functions in the tests now.

api_tests/lib/config.ts
tcharding: Added, thanks.

cnd/src/config/mod.rs
tcharding: Done as suggested, thanks.

cnd/src/config/settings.rs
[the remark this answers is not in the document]
tcharding: Done.

cnd/src/lib.rs
thomaseizinger: Inline was probably the wrong word, I meant literally constructing the above path manually instead of delegating to `ProjectDirs`.

tcharding: I don't see the benefit, we already depend on the library that provides `ProjectDirs` so why not use it?  If you mean remove that dependency all together that's different, I can create a ticket if you want to do that.

cnd/src/config/file.rs
tcharding: Boy, blindly following questions burnt me on this PR.  `~/.lnd` is not valid on OSX for lnd data directory \me cries quietly   :)

cnd/src/config/file.rs
tcharding: Please see https://github.com/comit-network/comit-rs/pull/2131 (final patch) which implements functions to get the TLS cert as well as the macaroon.

cnd/src/lib.rs
```diff
@@ -78,4 +81,16 @@ pub fn data_dir() -> Option<PathBuf> {
     ProjectDirs::from("", "", "comit").map(|proj_dirs| proj_dirs.data_dir().to_path_buf())
 }
 
+/// Returns the LND data directory: $LND_DIR if it is set or else "~/.lnd".
+pub fn lnd_data_dir() -> PathBuf {
+    match env::var("LND_DIR") {
+        Ok(dir) => PathBuf::from(dir),
+        Err(_) => {
+            // Just unwrap, if we don't have $HOME we are hosed anyway.
+            let home = dirs::home_dir().unwrap();
```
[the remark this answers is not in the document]
tcharding: Retro'ing my appalling amount of mistakes on this PR.  Confusion here, like above, caused by me blindly following suggestions.

cnd/src/config/settings.rs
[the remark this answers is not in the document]
tcharding: So in the end @thomaseizinger, with follow up PR I went back to looking in CND_DATA_DIR/lnd so we could use the default names that lnd uses for its files i.e., `tls.cert`, `tls.key`.  These are so generic that its nice to have them under a sub-directory that says what they are.

cnd/src/lib.rs
tcharding: Should we be considering removing the comments in `cnd/src/lib.rs` that reference Windows so as not to falsely imply that we _do_ support windows?

PR #2098, opened 2020-02-26
Bitcoin matching transactions refactor

da-kami:
WIP - If you are in the mood @thomaseizinger feel free to have a glance - 10% :)

Fixes: https://github.com/comit-network/comit-rs/issues/1878


cnd/src/swap_protocols/rfc003/bitcoin/htlc_events.rs
```diff
@@ -85,15 +78,16 @@ impl<Bitcoin: bitcoin::Bitcoin + bitcoin::Network> HtlcRedeemed<Bitcoin, asset::
...
-        let pattern = TransactionPattern {
-            to_address: None,
-            from_outpoint: Some(htlc_deployment.location),
-            unlock_script: Some(vec![vec![1u8]]),
-        };
 
-        let transaction = matching_transaction(connector, pattern, start_of_swap)
-            .await
-            .context("failed to find transaction to redeem from htlc")?;
+        let transaction = watch_for_transaction(
```
tcharding: How about adding a layer of abstraction on top of this function, like what we do in ethereum with `watch_for_contract_creation` -> `matching_transaction_receipt`.  That way the `unlock_script` (const) would not need to be known to `htlc_events` and the function could return the secret directly.

da-kami: I implemented it the same way as for Ethereum - the secret is extracted in `ethereum/htlc_events.rs`, not in `watch_for_event`. I actually like it that way - extracting the secret is a different concern than finding the transaction/event for me.

cnd/src/btsieve/bitcoin/mod.rs
```diff
@@ -8,142 +8,228 @@ pub use self::{
...
-    pattern: TransactionPattern,
+type Hash = bitcoin::BlockHash;
+type Block = bitcoin::Block;
+
+// TODO: reevaluate naming of function
+pub async fn watch_for_transaction<C>(
+    blockchain_connector: C,
     start_of_swap: NaiveDateTime,
+    from_outpoint: OutPoint,
+    unlock_script: Vec<Vec<u8>>,
```
thomaseizinger: Thinking ahead for miniscript, we might want to change this to be less generic and only provide a public key as the "spending with" condition.

i.e. we want to match whether it was spent with the redeem or the refund public key. To do that, I think we just have to check for the presence of the public key on the witness stack.

You can do this in a follow-up PR though if you want to keep the scope small.

cnd/src/btsieve/bitcoin/mod.rs
```diff
@@ -8,142 +8,228 @@ pub use self::{
...
-    let mut oldest_block: Option<bitcoin::Block> = Some(latest_block.clone());
-    prev_blockhashes.insert(latest_block.bitcoin_hash());
+    let transaction = matching_transaction(blockchain_connector, start_of_swap, |transaction| {
+        pattern.matches(&transaction)
+    })
+    .await?;
+
+    let (vout, _txout) = transaction
+        .find_output(&compute_address)
+        .expect("Deployment transaction must contain outpoint described in pattern");
```
thomaseizinger: It would be good to completely eliminate this error path but for that, you will need to change the API of `matching_transaction` to return you the Outpoint right away.

cnd/src/btsieve/bitcoin/mod.rs
```diff
@@ -8,142 +8,228 @@ pub use self::{
...
+/// This function uses the `connector` to find blocks relevant to a swap.  To do
+/// this we must get the latest block, for each latest block we receive we must
+/// ensure that we saw its parent i.e., that we did not miss any blocks between
+/// this latest block and the previous latest block we received.  Finally, we
+/// must also get each block back until the time that the swap started i.e.,
+/// look into the past (in case any action occurred on chain while we were not
+/// watching).
+///
+/// It yields those blocks as part of the process.
+async fn find_relevant_blocks<C>(
```
thomaseizinger: I can already see us sharing this code between Bitcoin and Ethereum, nice :)

cnd/src/btsieve/bitcoin/mod.rs
```diff
@@ -8,142 +8,228 @@ pub use self::{
...
 use reqwest::{Client, Url};
 use std::collections::HashSet;
 
-pub async fn matching_transaction<C>(
-    mut blockchain_connector: C,
-    pattern: TransactionPattern,
+type Hash = bitcoin::BlockHash;
+type Block = bitcoin::Block;
+
+// TODO: reevaluate naming of function
```
thomaseizinger: `watch_for_outpoint_spent`?

cnd/src/btsieve/bitcoin/mod.rs
```diff
@@ -8,142 +8,228 @@ pub use self::{
...
+        pattern.matches(&transaction)
+    })
+    .await?;
+
+    Ok(transaction)
+}
 
-    let mut prev_blockhashes: HashSet<bitcoin::BlockHash> = HashSet::new();
-    let mut missing_block_futures: Vec<_> = Vec::new();
+// TODO: reevaluate naming of function
```
thomaseizinger: Could create some nice symmetric with the above suggestion and name this: `watch_for_outpoint_created`.

thomaseizinger: Rationale: Sending bitcoin to an address creates an (unspent) outpoint :)

cnd/src/swap_protocols/rfc003/bitcoin/htlc_events.rs
da-kami: Discussed this further with Thomas and understand better what you meant Tobin :)

tcharding:
Nit: This is not a 'refactor', this is a total re-write of the matching logic for Bitcoin.  I just stumbled on this looking through PR titles, it may happen in the future if we ever are searching for this.

thomaseizinger:
> Nit: This is not a 'refactor', this is a total re-write of the matching logic for Bitcoin. I just stumbled on this looking through PR titles, it may happen in the future if we ever are searching for this.

What, if not this is a refactoring?

Hopefully, there is no observable change from the outside of the system :)

PR #2099, opened 2020-02-26
Remove the `Asset` trait

### LINKED ISSUE #2036: Remove the `Asset` trait
## Problem

Traits in Rust help to abstract over functionality. The `Asset` trait doesn't define any functionality.

## Goal

No unnecessary abstractions

## Recommendation

Delete the `Asset` trait.
Some functionality in our codebase relies on some of the super traits that are defined by `Asset`.
Those trait bounds should be moved to the impl block _that actually needs them_.

thomaseizinger:
I recommend going through this PR commit by commit as the messages include useful information on why I did certain things.

In general, I made the following decisions and learnings:

### Say what you actually mean:

This refers to settings trait bounds like: 

```rust
rfc003::Request<AL, BL, AA, BA>: Clone
```
instead of 
```rust
AL: Clone,
BL: Clone,
AA: Clone,
BA: Clone
```
From the latter, the compiler can infer that `Request` is also Clone but it is much more verbose.
The form only tells the compiler that you would like `Request` to be `Clone` _because you actually call `.clone()` on it_.
Whether or not that is valid for a _specific_ set of type parameters is then figured out later.

### Allow auto traits to bubble up:

The last point (say what you actually mean) is great for traits that provide functionality (`Clone`, `Serialize`, etc) but doesn't work that well for auto-traits like `Send` or `Sync`.

Those are (among other rules) automatically implemented by the compiler if all the inner fields implement them.

At first, the same reasoning seems to apply as in: 
```rust
rfc003::Request<AL, BL, AA, BA>: Send
```

Unfortunately, it doesn't scale that well because `Send` is not actually a function you call on a struct. As soon as you disassemble the original struct (like we do with `Request`) and spread the data over different ones (`OngoingSwap`, `Funded`, etc) the trait bound you stated no longer works and you would have to restate them:

```rust
rfc003::Request<AL, BL, AA, BA>: Send,
OngoingSwap<AL, BL, AA, BA>: Send,
Funded<AL::Transaction, AA>: Send,
...
```

For situations like this, I then opted for just stating:

```rust
AA: Send + Sync + 'static
```

That way, anything that contains `AA` will be `Send` as-long as everything else is `Send`.

It only applies to a few functions anyway. Hopefully we can minimize that in the future by having an even flatter call-hierachy of functions with type variables (as soon as we fill in type variables, the compiler can verify the bound and shuts up).

Fixes #2036.

### DIFF
--- a/cnd/src/asset/mod.rs
+++ b/cnd/src/asset/mod.rs
@@ -6,9 +6,8 @@ pub use self::{
 };
 use crate::asset;
 use derivative::Derivative;
-use std::fmt::Debug;
 
-pub trait Asset: Clone + Debug + Send + Sync + 'static + Ord {}
+pub trait Asset: Clone + Send + Sync + 'static + Ord {}
 
 impl Asset for Bitcoin {}
 
--- a/cnd/src/asset/mod.rs
+++ b/cnd/src/asset/mod.rs
@@ -7,7 +7,7 @@ pub use self::{
 use crate::asset;
 use derivative::Derivative;
 
-pub trait Asset: Clone + Send + Sync + 'static + Ord {}
+pub trait Asset: Send + Sync + 'static + Ord {}
 
 impl Asset for Bitcoin {}
 
--- a/cnd/src/asset/mod.rs
+++ b/cnd/src/asset/mod.rs
@@ -7,14 +7,6 @@ pub use self::{
 use crate::asset;
 use derivative::Derivative;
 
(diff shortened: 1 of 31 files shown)


thomaseizinger:
@luckysori @da-kami @bonomat @D4nte @tcharding 

I requested a review from all of you because I think it is crucial to understand the learnings from this PR in order for us to efficiently collaborate on this!

PR #2096, opened 2020-02-26
Use `where` clause everywhere to express trait bounds

thomaseizinger:
My OCD triggered :sweat_smile: 

#CLIONPOWER

PR #2103, opened 2020-02-26
Use where clauses

tcharding:
We would like to move the code base to using 'where clauses' instead of putting
the trait bound where the type is first mentioned.  This helps reduce the number
of merge conflicts when multiple developers work on the same area in the code
base.


tcharding:
@thomaseizinger already did this [here](https://github.com/comit-network/comit-rs/pull/2096#pullrequestreview-365256227).

PR #2104, opened 2020-02-26
Remove unused trait bound `Into<AssetKind>` from `Asset`

tcharding:
Recently we removed the requirement for this trait bound and forgot to
remove the trait bound itself from the `Asset` trait.

Remove it now its unused.


thomaseizinger:
Cool, I did this as part of my PR aswell but can't hurt to get it in earlier :)

PR #2105, opened 2020-02-27
Add `identity` module

tcharding:
Our domain is cross chain.  We should have domain specific terms for
things in various chains.  We currently have the `asset` module and the
`ledger` module that do just this.

Add and `identity` module that defines an identity/address on a chain.
With this applied we can write

```
	let address = identity::Ethereum::new(...)
```
Instead of
```
	let address = ethereum::Address::new(...)
```
which is actually just a re-export of:
```
	let address = web3::types::Address::new(...)
```

Similarly we have `identity::Bitcoin` for the Bitcoin addresses.

### DIFF
--- a/cnd/src/http_api/action.rs
+++ b/cnd/src/http_api/action.rs
@@ -309,10 +309,7 @@ impl IntoResponsePayload for Infallible {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::{
-        ethereum::{Address as EthereumAddress, U256},
-        swap_protocols::ledger::ethereum::ChainId,
-    };
+    use crate::{ethereum::U256, identity, swap_protocols::ledger::ethereum::ChainId};
     use bitcoin::Address as BitcoinAddress;
     use std::str::FromStr;
 
@@ -340,7 +337,8 @@ mod test {
 
     #[test]
     fn call_contract_serializes_correctly_to_json_with_none() {
-        let addr = EthereumAddress::from_str("0A81e8be41b21f651a71aaB1A85c6813b8bBcCf8").unwrap();
+        let addr =
+            identity::Ethereum::from_str("0A81e8be41b21f651a71aaB1A85c6813b8bBcCf8").unwrap();
         let chain_id = ChainId::from(3);
         let contract = ActionResponseBody::EthereumCallContract {
             contract_address: addr,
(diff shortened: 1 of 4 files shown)


tcharding:
Check this out!  I had this idea while trying to raise the Identity (Ledger associated type) up in preparation for removing it.  This makes adding all the concrete types to the db layer way less taxing to write and read because one can just glance and see the `Bitcoin`/`Ethereum` stuff for ledgers, assets, and now identities match instead of having to mentally map `PublicKey` to a Bitcoin identity.

tcharding:
If this is approved we can then also add `src/cnd/transaction.rs` in a similar fashion so that we can write `transaction::Bitcoin` instead of `bitcoin::Transaction`.  This at first seems like no benefit but I would argue that we should be cross-chain first i.e., we are primarily concerned with transactions _not_ with Bitcoin so we should sort on what we are primarily concerned with instead of sorting by the underlying thing.

thomaseizinger:
> If this is approved we can then also add `src/cnd/transaction.rs` in a similar fashion so that we can write `transaction::Bitcoin` instead of `bitcoin::Transaction`. This at first seems like no benefit but I would argue that we should be cross-chain first i.e., we are primarily concerned with transactions _not_ with Bitcoin so we should sort on what we are primarily concerned with instead of sorting by the underlying thing.

That is a good observation! I like it :)

The only thing to consider is that **if** we ever want to provide compile-time feature-flags for the support of certain ledgers, we have to put them in several places.
I don't see us doing that in the near future so it shouldn't matter!

tcharding:
Added the `transaction` module and put them both in `lib.rs`.  Adding `no-mergify` label to give reviewers (@thomaseizinger, and @D4nte) a chance to look at it again.  Perhaps one of you could remove the label if you think its ok to merge this, one new extra opinion should be enough?

thomaseizinger:
ACK from my side but I would like the modules to be used as-well. Otherwise it is just dead code :/

tcharding:
> ACK from my side but I would like the modules to be used as-well. Otherwise it is just dead code :/

Will do, thanks.

D4nte:
Still looks good to me. I would not mind several follow-up PRs to sort it out.

tcharding:
Removed the transaction module (since it adds dead code).  Squashed commits while I was at it since force push is required anyways.  This now just adds the `identity` module, and uses it in a few places.  It will be heavily used in the 'remove Ledger trait' work.

PR #2125, opened 2020-03-02
Remove the Identity associated type from Ledger

tcharding:
In an effort to remove the `Ledger` trait first remove the associated type `Identity`.  While doing this use the new `identity` modules.

Replace useages of `crate::ethereum::Address` with `identity::Ethereum`.  We have to be a bit more careful with `PublicKey` because we have a wrapper by the same name.

Works towards: #2081

Issue #2081, opened 2020-02-23
Remove the `Ledger` trait


thomaseizinger:
## Problem

The `Ledger` trait only serves the purpose of having associated types. However, this usages of traits is blocking us from splitting the [rfc003 protocol into two](https://github.com/comit-network/comit-rs/issues/1953).

## Goal

Being able to split the rfc003 protocol into two separate protocols while remaining backwards compatible on our APIs (REST, COMIT and DB).

## Recommendation

Do not start by removing the `Ledger` trait definition but instead remove the `Ledger` trait bound module by module (or file by file) for a smooth incremental change. You are likely to add some new trait bound to `impl` blocks (`Clone`, `Copy`, `Sync`, `'static`, `Send`, etc). Commit when it compiles and then continue until you hit a major road block or you are done.
Please aim for small incremental PRs.

https://github.com/comit-network/comit-rs/issues/2037 already did a lot of work towards removing instances of the `Ledger` trait bound on structs. This allows a more liberal use of those structs (e.g. for other protocols and not just for rfc003).

After https://github.com/comit-network/comit-rs/issues/2065 is done, we should be able to delete the `Ledger` trait but also removing the bounds from the structs mentioned in this comment: https://github.com/comit-network/comit-rs/issues/2065#issuecomment-589487172

Potentially, we need to apply similar techniques as used in https://github.com/comit-network/comit-rs/issues/2065.

PR #2127, opened 2020-03-02
Remodel Fund/Deplop/Redeem/Refund-Action traits to not have type parameters

thomaseizinger:
Draft because it builds on top of https://github.com/comit-network/comit-rs/pull/2125.

Work towards #2081 because I removes the `Ledger` trait bound from those traits.


cnd/src/http_api/routes/rfc003/handlers/post_swap.rs
```diff
@@ -34,10 +34,10 @@ where
         Save<Request<AL, BL, AA, BA, AI, BI>> + Save<Accept<AI, BI>> + Save<Swap> + Save<Decline>,
     AL: Ledger,
     BL: Ledger,
-    AA: Ord + Send + Sync + 'static,
-    BA: Ord + Send + Sync + 'static,
-    AI: Send + Sync + 'static,
-    BI: Send + Sync + 'static,
+    AA: Clone + Ord + Send + Sync + 'static,
```
thomaseizinger: As a note for reviewers: This clone appears here now because I had to move from references inside `HtlcParams` to owned values.
There would be a way of keeping references, so we don't need to clone stuff but it would spread lifetimes everywhere. I opt that having Clone here is the simpler approach.

cnd/src/swap_protocols/rfc003/alice/actions/erc20.rs
```diff
@@ -17,15 +17,17 @@ use std::convert::Infallible;
 impl<BL, BA, BI> Actions for alice::State<Ethereum, BL, asset::Erc20, BA, identity::Ethereum, BI>
 where
     BL: Ledger,
-    (BL, BA): RedeemAction<BL, BA, BI>,
+    BA: Clone,
+    BI: Clone,
+    (BL, BA): RedeemAction<HtlcParams = HtlcParams<BL, BA, BI>, HtlcLocation = BL::HtlcLocation>,
```
thomaseizinger: This change is unfortunate, I would have loved if I could have just not specified the associated types here. Maybe there is still a way to do this but for now, this allows us to move forward!

cnd/src/swap_protocols/rfc003/bob/actions/erc20.rs
```diff
@@ -55,7 +57,7 @@ where
                     HtlcParams::new_alpha_params(request, response),
                     htlc_location.clone(),
                     &*self.secret_source, // Derive identities with this.
-                    *secret,              /* Bob uses the secret learned from Alice's redeem
+                    *secret,              /* Bob uses the secret learned from Aliceredeem
```
thomaseizinger: Ups, this change shouldn't be here. Do you mind fixing this @tcharding ? :)

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -31,7 +31,7 @@ lazy_static::lazy_static! {
 impl HtlcFunded<Ethereum, Ether, identity::Ethereum> for Cache<Web3Connector> {
```
thomaseizinger: Note: We can apply the same idea of this PR to these traits.

PR #2136, opened 2020-03-02
Use modules for domain terms

tcharding:
Recently we added `identity::Bitcoin` and `transaction::Bitcoin`. Add `htlc_location::Bitcoin`.

Use these modules across the codebase.

Done while working on removing Ledger trait.

### DIFF
--- a/cnd/src/swap_protocols/rfc003/bitcoin/htlc_events.rs
+++ b/cnd/src/swap_protocols/rfc003/bitcoin/htlc_events.rs
@@ -68,7 +68,7 @@ where
             .expect("Deployment transaction must contain outpoint described in pattern");
 
         Ok(Deployed {
-            location: ::bitcoin::OutPoint {
+            location: htlc_location::Bitcoin {
                 txid: transaction.txid(),
                 vout,
             },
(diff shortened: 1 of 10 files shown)


cnd/src/swap_protocols/rfc003/bitcoin/htlc_events.rs
```diff
@@ -68,7 +68,7 @@ where
             .expect("Deployment transaction must contain outpoint described in pattern");
 
         Ok(Deployed {
-            location: OutPoint {
+            location: ::bitcoin::OutPoint {
```
D4nte:
```suggestion
            location: htlc_location::Bitcoin {
```
What do you think?

thomaseizinger: Yes!

At some point we may even create our own wrapper types instead of just aliasing :)

tcharding: Nice, my brain only got to seeing that `Self::HtlcLocation` would not work.  Thanks

PR #2141, opened 2020-03-03
Replaced web3 types without own ones #1938

### LINKED ISSUE #1938: Replace remaining web3 types with our own ones
## Problem

We are still depending on web3 for several for some of the types in our repository:

https://github.com/comit-network/comit-rs/blob/ad8d7d34aa227777c3eb7b8b57f1cc0624fd6bbf/cnd/src/ethereum/mod.rs#L5-L8

## Goal

Not depending on web3.

## Recommendation

- check which of the types re-exported from web3 are actually in use
- inline types like `Block`, `Transaction` etc (at this stage, we might also remove the annoying `Option` wrappers around certain fields)
- remove web3 from our production dependencies

rishflab:
Moved the web3 dependency to dev-dependencies and created copies of
those types in the ethereum module.

Fixes #1938.

### DIFF
--- a/cnd/Cargo.toml
+++ b/cnd/Cargo.toml
@@ -20,12 +20,14 @@ diesel_migrations = "1.4.0"
 directories = "2.0"
 ethbloom = "0.6.4"
 fern = { version = "0.5", features = ["colored"] }
+fixed-hash = { version = "0.5", default-features = false }
 futures = { version = "0.1" }
 futures-core = { version = "0.3", features = ["compat", "async-await"], default-features = false, package = "futures" }
 genawaiter = "0.99"
 hex = "0.4"
 http-api-problem = { version = "0.15", features = ["with_warp"] }
 impl-template = "1.0.0-alpha"
+impl-serde = { version = "0.3.0", default-features = false }
 lazy_static = "1"
 libp2p = { version = "0.16" }
 libp2p-comit = { path = "../libp2p-comit" }
@@ -35,6 +37,7 @@ lru = "0.4.3"
 num = "0.2"
 paste = "0.1"
 pem = "0.7"
+primitive-types = { version = "0.6", features = ["serde"] }
 rand = "0.7"
 reqwest = { version = "0.10", default-features = false, features = ["json", "native-tls"] }
 serde = { version = "1", features = ["derive"] }
@@ -59,7 +62,6 @@ url = { version = "2", features = ["serde"] }
 uuid = { version = "0.8", features = ["serde", "v4"] }
 void = "1.0.2"
 warp = { version = "0.2", default-features = false }
-web3 = { version = "0.8", default-features = false, features = ["http"] }
 
 [dev-dependencies]
 base64 = "0.11"
@@ -71,3 +73,4 @@ serde_urlencoded = "0.6"
 spectral = { version = "0.6", default-features = false }
 tempfile = "3.1.0"
 testcontainers = "0.8"
+web3 = { version = "0.10", default-features = false, features = ["http"] }
--- a/cnd/src/ethereum/mod.rs
+++ b/cnd/src/ethereum/mod.rs
@@ -1,10 +1,63 @@
 #![warn(unused_extern_crates, missing_debug_implementations, rust_2018_idioms)]
 #![forbid(unsafe_code)]
 
-pub use web3::types::{
-    Address, Block, BlockId, BlockNumber, Bytes, Log, Transaction, TransactionReceipt,
-    TransactionRequest, H160, H2048, H256, U128, U256,
+pub use primitive_types::{
+    H160, H256, U128, U256
 };
+
+pub use ethbloom::{Bloom as H2048, Input};
+
+use fixed_hash::*;
+use impl_serde::impl_fixed_hash_serde;
+
+use hex::{FromHex, ToHex};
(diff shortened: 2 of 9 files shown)


cnd/src/ethereum/mod.rs
```diff
@@ -1,11 +1,78 @@
...
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt;
+
+pub type Index = U128;
+pub type Address = H160;
+
+construct_fixed_hash! {
+    pub struct H64(8);
+}
+impl_fixed_hash_rlp!(H64, 8);
```
thomaseizinger: I don't think we need the rlp encoding as we are never deserializing/serializing to the binary format of Ethereum transactions.

cnd/src/ethereum/mod.rs
```diff
@@ -1,11 +1,78 @@
...
+impl_fixed_hash_serde!(H64, 8);
+
+construct_uint! {
+	/// Unsigned 64-bit integer.
+	pub struct U64(1);
```
thomaseizinger: Can we avoid this if we just use a U256 for the status for now?

I think it is not used anywhere else.

cnd/src/ethereum/mod.rs
thomaseizinger: Would be nice to avoid the `impl_rlp` dependency.

cnd/src/ethereum/mod.rs
```diff
@@ -1,11 +1,78 @@
...
+use rustc_hex::{FromHex, ToHex};
+use serde::de::{Error, Visitor};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt;
+
+pub type Index = U128;
+pub type Address = H160;
+
+construct_fixed_hash! {
+    pub struct H64(8);
```
thomaseizinger: This is only used for the nonce if I see this correctly? We could also use a bigger one here and then get rid of the `fixed-hash` dependency.

cnd/src/ethereum/mod.rs
```diff
@@ -16,3 +83,243 @@ impl IsStatusOk for TransactionReceipt {
...
+    pub transaction_log_index: Option<u64>,
+    /// Log Type
+    #[serde(rename = "logType")]
+    pub log_type: Option<String>,
+    /// Removed
+    pub removed: Option<bool>,
+}
+
+/// The block type returned from RPC calls.
+/// This is generic over a `TX` type.
```
thomaseizinger: This doc is now invalid :)

cnd/src/ethereum/mod.rs
```diff
@@ -16,3 +83,243 @@ impl IsStatusOk for TransactionReceipt {
...
+    pub size: Option<U256>,
+    /// Mix Hash
+    #[serde(rename = "mixHash")]
+    pub mix_hash: Option<H256>,
+    /// Nonce
+    pub nonce: Option<H64>,
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum BlockNumber {
```
thomaseizinger: Can we simplify this API based on our usage? I think we only ever use `Latest` here.

cnd/src/ethereum/mod.rs
```diff
@@ -16,3 +83,243 @@ impl IsStatusOk for TransactionReceipt {
...
+            BlockNumber::Latest => serializer.serialize_str("latest"),
+            BlockNumber::Earliest => serializer.serialize_str("earliest"),
+            BlockNumber::Pending => serializer.serialize_str("pending"),
+        }
+    }
+}
+
+/// Block Identifier
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum BlockId {
```
thomaseizinger: Same for this, I think we statically use this in only one function.

cnd/Cargo.toml
```diff
@@ -37,6 +42,7 @@ paste = "0.1"
 pem = "0.7"
 rand = "0.7"
 reqwest = { version = "0.10", default-features = false, features = ["json", "native-tls"] }
+rustc-hex = "2.1.0"
```
thomaseizinger: We already depend on the `hex` crate, would be good to use that one instead of bringing in a 2nd dependency.

cnd/src/ethereum/mod.rs
rishflab: Tried and it doesn't pass tests, cant deserialise the test data. I would have to write a custom deserialiser?

thomaseizinger: I think that would be preferable over taking another dependency. According to https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt, the field is either 1 or 0. You could write a custom deserializer that turns it into a boolean :)

rishflab: The H64 is used for nonce field in block. I have replaced the status field with U256. 
`pub nonce: Option<H64>`
Throws an error when trying to deserialise the H64 into a H256. I was thinking I could modify the deserialiser for H256 to accept H64 but afaik it is generated by macros

rishflab: actually i think ill just create our own H64 type, all it needs is deserialise/serialise

Issue #1953, opened 2020-01-28
Finish re-modelling COMIT's core to account for HAN and HErc20


thomaseizinger:
## Problem

The protocol naming spike, with foresight for the Lightning integration identified that instead of a single swap protocol, we should be talking about a swap protocol _per ledger_.
This is a requirement for integration Lightning that allows us to re-use our existing swap protocols on the other ledger if we are doing a Lightning swap.

## Goal

Build on top of the work done in #2195 (and #2198) in order replace `rfc003` with `han` and `herc20` in the state and core. `rfc003` shall remain on DB, HTTP and COMIT APIs.

## Recommendation

I can't foresee a clear strategy until the work blocking this issue is done. That work used to be part of this issue, so it can only be easier to tackle this once all that is done, but it will still be a very major refactoring with unknowns. For that reason I would suggest this one should be treated as exploratory, with at least 2 people working on it at the same time.

You should not be touching the DB, HTTP API, COMIT API. The rfc003 request swap message has to stay as it is the one used on all APIs. Below that, then rfc003 should be gone.

D4nte:
Ok for those interested, after reading the code and trying to understand what could be done, I decided to split the `rfc003/events` trait in 2 `han/events` and `herc20/events` and see where it leads.....

D4nte:
Please note this is blocked as @thomaseizinger is currently doing a major refactoring that is creating merge conflicts.

D4nte:
Learnings on how to do this issue so far:
- Backward compatibility with HTTP/COMIT API seems possible by having some superseeding traits (e.g. Ledger) at `swap_protocols` level to be deleted later
- Backward compatibility with DB seems a waste of time. Possible options:
   - Remodel sql db from `rfc003_` to `han_bitcoin_..herc20_ethereum`: seems like waste of effort
   - Keep `rfc003_` compability: huge work, waste of effrt
   - Migrate to nosql first
- In any case, it seems like extracting comit as a lib first, then do the han/herc20 changes and make the comit-lib crate compile to finally integrate it with the cnd crate by working on the various APIs sounds easier and more strategic

D4nte:
Marked blocked until strategy is reviewed.

thomaseizinger:
Thanks for reporting that! It is a little hard to judge remotely on how to proceed.
I'll think about it and have a look at the draft PR later.

D4nte:
Assigning to @thomaseizinger as he wants to have a go.

[...]

PR #2142, opened 2020-03-03
1878 bitcoin matching transactions refactor

da-kami:


cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -33,7 +33,7 @@ use genawaiter::{
 ///
 /// It is highly unlikely for Bob to fund the HTLC now, yet the current
 /// implementation is still waiting for that.
-pub async fn create_swap<D, A, AI, BI>(
+pub async fn create_alpha_watcher<D, A, AI, BI>(
```
tcharding: These two functions `create_alpha_watcher` and `create_beta_watcher` are now identical except for a single function call (and alpha/beta in the function comment).  Can we have some abstraction over this code please?  One idea would be to pass the function called (i.e., `watch_foo_ledger`) as an argument, not sure if this works with the turbo fish syntax needed but off the top of my head I don't see why not.

da-kami: Yeah, I discussed this with @thomaseizinger and we would go for a follow up PR it that's OK.

thomaseizinger: The split is possible but I think not trivial because of the `SwapEvent` type still requiring `AL, BL, AA, BA` and its use in `StateStore`.

In any case, the dupliation is an implementation detail and not that critical I would say :)

This is a nice step towards #1953, where we are going to need such functions, albeit they have to be ledger independent for that.

More importantly, this fixes a crucial runtime bug where we shared the instance of the Coroutine between the two watchers and is a blocker for @da-kami's feature. Hence, we opted not to scope creep to much and try to land the actual ticket that is motivating this :)

As always, if you have the time and will, you are happy to send a follow-up PR ;)

cnd/src/init_swap.rs
```diff
@@ -50,23 +51,37 @@ where
...
 
-            tokio::task::spawn(
-                create_swap::<D, bob::State<AL, BL, AA, BA, AI, BI>, AI, BI>(
-                    dependencies.clone(),
-                    accepted,
-                ),
-            );
+            tokio::task::spawn(create_alpha_watcher::<
+                D,
+                alice::State<AL, BL, AA, BA, AI, BI>,
```
da-kami: hahaha, that's a fun merge mistake here :P 
Bob listening on Alice' state ;)

Fixing this...

da-kami:
> Nicely done @da-kami :)
> 
> As discussed, there is a bunch of follow-up work possible now. At this stage, I would recommend to create tickets for it and no longer stretch the scope of this issue.
> 
> I see the following follow-up work:
> 
> * make `btsieve::bitcoin` and `btsieve::ethereum` use the same `find_relevant_blocks` function, including all the child functions like `walk_back_until` etc. Please mention in the ticket that we want to generalize the bitcoin one as the algorithm has been improved there.
> * remove the potential error path in `watch_for_created_outpoint` by making the matcher return the outpoint directly
> * inline the logic for `spends_from_with` inside `watch_for_spend_outpoint` and delete the whole `TransactionExt` trait + module
> * change the API of `watch_for_spent_outpoint` to take a public key instead of the spending script and adapt the matching to check, whether the transaction spends the outpoint using this key

Yes, thanks for summing it up! I would create follow-up PRs for all the issues mentioned, starting with the first one you mention (make Bitcoin and Ethereum use the same logic for finding relevant blocks).

thomaseizinger:
> > Nicely done @da-kami :)
> > As discussed, there is a bunch of follow-up work possible now. At this stage, I would recommend to create tickets for it and no longer stretch the scope of this issue.
> > I see the following follow-up work:
> > 
> > * make `btsieve::bitcoin` and `btsieve::ethereum` use the same `find_relevant_blocks` function, including all the child functions like `walk_back_until` etc. Please mention in the ticket that we want to generalize the bitcoin one as the algorithm has been improved there.
> > * remove the potential error path in `watch_for_created_outpoint` by making the matcher return the outpoint directly
> > * inline the logic for `spends_from_with` inside `watch_for_spend_outpoint` and delete the whole `TransactionExt` trait + module
> > * change the API of `watch_for_spent_outpoint` to take a public key instead of the spending script and adapt the matching to check, whether the transaction spends the outpoint using this key
> 
> Yes, thanks for summing it up! I would create follow-up PRs for all the issues mentioned, starting with the first one you mention (make Bitcoin and Ethereum use the same logic for finding relevant blocks).

You can also create follow-up PR yes :)
There are 3 days left in the sprint, where IIRC 2 of them you reserved for community (Fr + Mo). If you don't finish all of the above, please record them as tickets so stuff doesn't get lost.

cnd/src/btsieve/bitcoin/mod.rs
```diff
@@ -1,152 +1,217 @@
...
+        transaction.spends_to(&compute_address)
+    })
+    .await?;
 
-    loop {
-        // Delay so that we don't overload the CPU in the event that
-        // latest_block() and block_by_hash() resolve quickly.
+    let (vout, _txout) = transaction
+        .find_output(&compute_address)
+        .expect("Deployment transaction must contain outpoint described in pattern");
```
tcharding: Stale string (refers to 'pattern').

thomaseizinger: Good catch!
Luckily, that whole error path is going away soon :)

da-kami:
bors retry

da-kami:
> * make `btsieve::bitcoin` and `btsieve::ethereum` use the same `find_relevant_blocks` function, including all the child functions like `walk_back_until` etc. Please mention in the ticket that we want to generalize the bitcoin one as the algorithm has been improved there.

done in https://github.com/comit-network/comit-rs/pull/2172

> * remove the potential error path in `watch_for_created_outpoint` by making the matcher return the outpoint directly
> * inline the logic for `spends_from_with` inside `watch_for_spend_outpoint` and delete the whole `TransactionExt` trait + module
> * change the API of `watch_for_spent_outpoint` to take a public key instead of the spending script and adapt the matching to check, whether the transaction spends the outpoint using this key

WIP, https://github.com/comit-network/comit-rs/tree/1878-follow-up-bitcoin-matching

Issue #2132, opened 2020-03-02
Flaky dry test


thomaseizinger:
## Problem

We have a flaky dry test: https://app.circleci.com/jobs/github/comit-network/comit-rs/4310

## Goal

No flaky tests

## Recommendation

Fix it.

PR #2146, opened 2020-03-04
Properly wait for cnds to start up and accept requests

thomaseizinger:
Found this while debugging #2132. 

It is not fully going to solve it but help resilience of the test suite in general.

### DIFF
--- a/api_tests/lib/cnd/cnd_instance.ts
+++ b/api_tests/lib/cnd/cnd_instance.ts
@@ -5,7 +5,7 @@ import tempWrite from "temp-write";
 import { promisify } from "util";
 import { CndConfigFile, E2ETestActorConfig } from "../config";
 import { LedgerConfig } from "../ledgers/ledger_runner";
-import { HarnessGlobal } from "../utils";
+import { HarnessGlobal, sleep } from "../utils";
 import path from "path";
 import { LogReader } from "../ledgers/log_reader";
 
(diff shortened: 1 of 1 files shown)


api_tests/lib/cnd/cnd_instance.ts
```diff
@@ -74,7 +75,11 @@ export class CndInstance {
             }
         });
 
-        await sleep(1000); // allow the nodes to start up
+        const logReader = new LogReader(logFile);
+        await logReader.waitForLogMessage("Starting HTTP server on");
+
+        // we emit the log _before_ we start the http server, let's make sure it actually starts up
+        await sleep(1000);
```
D4nte: Can we put less than 1 sec?  Shall we update cnd to log actual readiness?

bonomat: I would leave it with 1 sec with this PR as this is an attempt to fix flaky e2e tests. But I like your proposal to print _actual readiness_. 
@thomaseizinger : can you create a follow-up ticket for that please?

thomaseizinger: I tried but it is not really possible because the future that warp gives us never resolves.

This is just an attempt to give warp _some_ time to actually bind the port etc.
We are waiting for a log line further up already.

A second is probably an overkill but it is hard to say, how much is enough.

You are obviously free to give it a try yourself :)
Not sure if its worth a ticket.
Once all tests run in parallel, this should be barely noticeable.

PR #2150, opened 2020-03-04
Replaced web3 types without own ones #1938

rishflab:
Moved the web3 dependency to dev-dependencies and created copies of
those types in the ethereum module

I tried to remove the ethbloom and primitive type dependencies as well but it will require a fair bit of work to replace H256, ethbloom, H160, U128, U256.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -545,6 +545,7 @@ dependencies = [
  "num 0.2.1",
  "paste",
  "pem",
+ "primitive-types",
  "quickcheck",
  "rand 0.7.3",
  "regex",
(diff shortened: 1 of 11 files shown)


cnd/src/hex_serde/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+use std::marker::PhantomData;
+
+/// A serializer that first encodes the argument as a hex-string
+pub fn serialize<T, S>(value: &T, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+        T: AsRef<[u8]>,
+{
+    let output: String = value.encode_hex();
+    serializer.serialize_str(&output)
```
thomaseizinger: Doesn't this need `0x` prefixing?

PR #2152, opened 2020-03-04
Libp2p clean up

tcharding:
This clean up PR is a result of reading our libp2p code in preparation for the LN communication protocol epic.

### DIFF
--- a/libp2p-comit/src/frame/codec.rs
+++ b/libp2p-comit/src/frame/codec.rs
@@ -54,12 +54,12 @@ impl Decoder for JsonFrameCodec {
 mod tests {
 
     use super::*;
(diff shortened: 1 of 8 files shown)


tcharding:
> Happy to merge this but I just wanted to point out that none of our existing libp2p code is relevant for the new protocol. We are gonna start from scratch here! :)
>
> Not by deleting everything but by building it next to it. If you haven't already, I suggest you read the tickets (+ links in there) I already created. They are fairly detailed and should get you on the right track.

Oh, I did not realise that we were going to start from scratch.  I've read all the tickets, and now I'm familiar with the current libp2p code we have.  Onward and upwards.

PR #2168, opened 2020-03-05
Upgrade node to version 12

bonomat:
@thomaseizinger found an issue describing incompatibility of the node version we are using on circleCI and jest: https://github.com/facebook/jest/issues/9453.

The tests are running locally. So let's upgrade circleCI as well.


thomaseizinger:
> Does someone remember why we locked at version 10? I believe there was a reason.

Stuff didn't work. Doesn't seem to be the case anymore :)

PR #2172, opened 2020-03-05
Generic `find_relevant_blocks` for Bitcoin and Ethereum

da-kami:
This is a follow up of #1878  

You can take a look at the commits, I would leave these three steps and don't squash them together (all of them are clean). If you don't like that please comment.

### DIFF
--- a/cnd/src/btsieve/bitcoin/mod.rs
+++ b/cnd/src/btsieve/bitcoin/mod.rs
@@ -6,24 +6,33 @@ pub use self::{
     bitcoind_connector::BitcoindConnector, cache::Cache, transaction_ext::TransactionExt,
 };
 use crate::{
-    btsieve::{BlockByHash, LatestBlock, Predates},
-    transaction, Never,
+    btsieve::{
+        find_relevant_blocks, BlockByHash, BlockHash, LatestBlock, Predates, PreviousBlockHash,
+    },
+    transaction,
 };
 use bitcoin::{
     consensus::{encode::deserialize, Decodable},
     BitcoinHash, OutPoint,
 };
 use chrono::NaiveDateTime;
-use futures_core::compat::Future01CompatExt;
-use genawaiter::{
-    sync::{Co, Gen},
-    GeneratorState,
-};
+use genawaiter::{sync::Gen, GeneratorState};
 use reqwest::{Client, Url};
-use std::collections::HashSet;
 
 type Hash = bitcoin::BlockHash;
 type Block = bitcoin::Block;
+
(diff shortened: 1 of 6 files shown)


cnd/src/btsieve/bitcoin/mod.rs
```diff
@@ -6,25 +6,30 @@ pub use self::{
...
-    GeneratorState,
-};
+use genawaiter::{sync::Gen, GeneratorState};
 use reqwest::{Client, Url};
-use std::collections::HashSet;
 
 type Hash = bitcoin::BlockHash;
 type Block = bitcoin::Block;
 
+impl BlockHashes<Hash> for Block {
```
thomaseizinger: I understand your motivation of putting these in one trait but it is not very idiomatic Rust to do that.

Traits are not interfaces that describe instances of objects.
They abstract over functionality and for that, a good rule of thumb is to have one trait per functionality.

cnd/src/btsieve/mod.rs
```diff
@@ -42,3 +46,126 @@ pub trait ReceiptByHash: Send + Sync + 'static {
...
+/// This function uses the `connector` to find blocks relevant to a swap.  To do
+/// this we must get the latest block, for each latest block we receive we must
+/// ensure that we saw its parent i.e., that we did not miss any blocks between
+/// this latest block and the previous latest block we received.  Finally, we
+/// must also get each block back until the time that the swap started i.e.,
+/// look into the past (in case any action occurred on chain while we were not
+/// watching).
+///
+/// It yields those blocks as part of the process.
+pub async fn find_relevant_blocks<C, B, H>(
```
thomaseizinger: So nice! Finally we found an abstraction that covers the core functionality of btsieve, is not opinionated about what it does with the blocks and even generic across blockchains.

Today is a good day!

cnd/src/btsieve/mod.rs
```diff
@@ -42,3 +46,126 @@ pub trait ReceiptByHash: Send + Sync + 'static {
...
+    mut connector: C,
+    co: &Co<B>,
+    start_of_swap: NaiveDateTime,
+) -> anyhow::Result<Never>
+where
+    C: LatestBlock<Block = B> + BlockByHash<Block = B, BlockHash = H> + Clone,
+    B: Predates + BlockHashes<H> + Clone,
+    H: Eq + std::hash::Hash + Copy,
```
thomaseizinger: Within this scope, you should be able to remove that.
```suggestion
    H: Eq + Hash + Copy,
```

PR #2176, opened 2020-03-05
Make `btsieve` traits use async/await 

tcharding:
We want async/await all over, make it so in `btsieve`.

This PR makes the following traits async: `ReceiptByHash`, `LatestBlock`, `BlockByHash`.

Removes the `impl_block_by_hash` macro.

Fixes: #1879


cnd/src/btsieve/bitcoin/cache.rs
```diff
@@ -28,35 +24,55 @@ impl<C> Cache<C> {
...
         }
-        .boxed()
-        .compat();
 
-        Box::new(future)
+        Ok(block)
     }
 }
 
-impl_block_by_hash!();
```
luckysori: Quick look on my phone, but it looks like you stopped using the macro yet you didn't delete it and actually updated it to be async. Am I wrong?

tcharding: Thanks @luckysori, you are right.  Deleted it now and force pushed.  (unrelated note: Are we aiming for correct discreet patches on PRs yet?)

thomaseizinger: Aim for patches that yield a Rust binary :)

cnd/src/btsieve/bitcoin/cache.rs
```diff
@@ -28,35 +24,55 @@ impl<C> Cache<C> {
...
+#[async_trait]
+impl<C> BlockByHash for Cache<C>
+where
+    C: BlockByHash<Block = Block, BlockHash = Hash> + Clone,
+{
+    type Block = Block;
+    type BlockHash = BlockHash;
+
+    async fn block_by_hash(&mut self, block_hash: Self::BlockHash) -> anyhow::Result<Self::Block> {
+        let mut connector = self.connector.clone();
```
thomaseizinger: You shouldn't need to clone the connector :)

PR #2178, opened 2020-03-06
Remove duplication in the TryFrom<Request> impl blocks

thomaseizinger:
We can collapse all those impls into a single one that is bounded by the correct trait bounds for the individual items.

cc @tcharding you will want to see this :D


tcharding:
Oh sick, how many other places are we going to be able to do this @thomaseizinger?

thomaseizinger:
> Oh sick, how many other places are we going to be able to do this @thomaseizinger?

Can't think of similar duplication somewhere off the top of my head. If you have some, point it out!
The key thing is "say what you actually mean": We said: `OutgoingRequest: TryFrom<Request<...>>`. The big advantage of that is that the validation of the trait bounds that are placed on _that_ impl are implicitly carried along by Rust. We should apply this knowledge throughout the codebase :)

PR #2200, opened 2020-03-09
Make StateStore trait generic over the state and events

luckysori:
Pre-work towards re-modelling COMIT's core to account for `HAN` and `HErc20`.

The goal is to make it easier to split the state store into different ledgers (and eventually different single ledger protocols such as HAN and HErc20) in the near future.

### DIFF
--- a/cnd/src/network/mod.rs
+++ b/cnd/src/network/mod.rs
@@ -4,7 +4,10 @@ pub use transport::ComitTransport;
 
 use crate::{
     asset::AssetKind,
-    btsieve::{bitcoin, bitcoin::BitcoindConnector, ethereum, ethereum::Web3Connector},
+    btsieve::{
+        bitcoin::{self, BitcoindConnector},
+        ethereum::{self, Web3Connector},
+    },
     comit_api::LedgerKind,
     config::Settings,
     db::{Save, Sqlite, Swap},
@@ -15,9 +18,10 @@ use crate::{
         ledger,
         rfc003::{
             self, bob,
+            create_swap::SwapEvent,
             messages::{Decision, DeclineResponseBody, Request, RequestBody, SwapDeclineReason},
-            state_store::{InMemoryStateStore, StateStore},
         },
+        state_store::{InMemoryStateStore, StateStore},
         HashFunction, Role, SwapId, SwapProtocol,
     },
     transaction,
@@ -40,8 +44,7 @@ use libp2p::{
 };
 use libp2p_comit::{
     frame::{OutboundRequest, Response, ValidatedInboundRequest},
-    handler,
-    handler::{ComitHandler, ProtocolInEvent, ProtocolOutEvent},
+    handler::{self, ComitHandler, ProtocolInEvent, ProtocolOutEvent},
     BehaviourOutEvent, Comit, PendingInboundRequest,
 };
 use serde::{de::DeserializeOwned, Serialize};
@@ -730,8 +733,8 @@ async fn insert_state_for_bob<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT, DB>(
 where
     AL: Send + 'static,
     BL: Send + 'static,
-    AA: Send + 'static,
-    BA: Send + 'static,
+    AA: Ord + Send + 'static,
+    BA: Ord + Send + 'static,
     AH: Send + 'static,
     BH: Send + 'static,
     AI: Send + 'static,
@@ -740,6 +743,7 @@ where
     BT: Send + 'static,
     DB: Save<Request<AL, BL, AA, BA, AI, BI>> + Save<Swap>,
     Request<AL, BL, AA, BA, AI, BI>: Clone,
+    bob::State<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT>: Clone + Sync,
 {
     let id = swap_request.swap_id;
     let seed = seed.derive_swap_seed(id);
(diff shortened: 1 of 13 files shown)


cnd/src/network/mod.rs
```diff
@@ -1,10 +1,14 @@
 pub mod transport;
 
+use crate::swap_protocols::rfc003::create_swap::SwapEvent;
```
thomaseizinger: :mega: Put these in the same block as the others please so the imports get merged.

cnd/src/network/mod.rs
```diff
@@ -728,16 +731,16 @@ async fn insert_state_for_bob<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT, DB>(
...
-    BL: Send + 'static,
-    AA: Send + 'static,
-    BA: Send + 'static,
-    AH: Send + 'static,
-    BH: Send + 'static,
-    AI: Send + 'static,
-    BI: Send + 'static,
-    AT: Send + 'static,
-    BT: Send + 'static,
+    AL: Clone + Send + Sync + 'static,
```
thomaseizinger: Why do all of these have to be Clone suddenly? What is the type that you are actually cloning?
Putting a type constraint on this fn for what we are actually cloning would reduce the noise drastically. See line 745 here.

cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -135,7 +135,7 @@ pub async fn create_beta_watcher<D, A, AI, BI>(
     A::BT: Clone,
     AI: Clone,
     BI: Clone,
-    A: ActorState,
+    A: ActorState + Clone + Send,
```
thomaseizinger: What is the source of this `Clone` bound? Are we actually cloning the ActorState somewhere down here?

cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -393,15 +393,6 @@ where
     }
 }
 
-pub type SwapEventOnLedger<A> = SwapEvent<
```
thomaseizinger: :tada: Nice one!

cnd/src/network/mod.rs
```diff
@@ -728,16 +731,16 @@ async fn insert_state_for_bob<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT, DB>(
...
+    BL: Clone + Send + Sync + 'static,
+    AA: Clone + Ord + Send + Sync + 'static,
+    BA: Clone + Ord + Send + Sync + 'static,
+    AH: Clone + Send + Sync + 'static,
+    BH: Clone + Send + Sync + 'static,
+    AI: Clone + Send + Sync + 'static,
+    BI: Clone + Send + Sync + 'static,
+    AT: Clone + Send + Sync + 'static,
+    BT: Clone + Send + Sync + 'static,
     DB: Save<Request<AL, BL, AA, BA, AI, BI>> + Save<Swap>,
```
D4nte:
```suggestion
	SwapEvent<AA, BA, AH, BH, AT, BT>: Clone + Sync 
	...
```
Where does the trait bound requirement come from?
Would the following work instead?

cnd/src/swap_protocols/state_store.rs
```diff
@@ -144,9 +131,10 @@ mod tests {
     fn insert_and_get_state() {
         let state_store = InMemoryStateStore::default();
 
-        let bitcoin_pub_key = "02c2a8efce029526d364c2cf39d89e3cdda05e5df7b2cbfc098b4e3d02b70b5275"
-            .parse()
-            .unwrap();
+        let bitcoin_pub_key: identity::Bitcoin =
+            "02c2a8efce029526d364c2cf39d89e3cdda05e5df7b2cbfc098b4e3d02b70b5275"
+                .parse()
+                .unwrap();
```
D4nte:
```suggestion
        let bitcoin_pub_key = identity::Bitcoin::from_str("02c2a8efce029526d364c2cf39d89e3cdda05e5df7b2cbfc098b4e3d02b70b5275").unwrap();
```

I feel there is not much point using `.parse()` if you have to hint the type. Personal preference.

cnd/src/network/mod.rs
luckysori: Good catch! And very didactic of you to let me figure out the answer rather than telling me what to do :relaxed: 

I'm still not used to this new style (which is obviously superior) because the compiler directs you towards doing it the "wrong" way.

luckysori: > Why do all of these have to be Clone suddenly? What is the type that you are actually cloning?

Just for clarity, this is the bound that it was looking for:
```rust
bob::State<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT>: Clone
```

cnd/src/swap_protocols/rfc003/create_swap.rs
luckysori: A clonable state is required for the `StateStore`. Here we only use the `update` method which doesn't need `Clone` (`get` does), but that's the problem with traits that define more than one method: the constraints on the individual methods cannot be more restrictive than those on the trait itself.

cnd/src/network/mod.rs
luckysori: Almost! It's

```rust
bob::State<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT>: Clone + Sync,
```

Thank you!

cnd/src/swap_protocols/state_store.rs
luckysori: I agree :smile:

PR #2201, opened 2020-03-09
Don't clone blockchain connectors

thomaseizinger:
Initially, we thought it was a good idea to make the traits BlockByHash and LatestBlock take &mut self. It turns out that we don't need this functionality for anything other than the tests and it actually forces us to clone the connector in many places.

Removing the mutable reference to self allows us to stop cloning the connector and simply pass references into all the functions.

### DIFF
--- a/cnd/tests/ethereum_helper/connector_mock.rs
+++ b/cnd/tests/ethereum_helper/connector_mock.rs
@@ -4,24 +4,20 @@ use cnd::{
     btsieve::{ethereum::ReceiptByHash, BlockByHash, LatestBlock},
     ethereum::{Block, TransactionReceipt, H256},
 };
-use std::{
-    collections::HashMap,
-    time::{Duration, Instant},
-};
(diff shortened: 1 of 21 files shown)


cnd/tests/ethereum_helper/connector_mock.rs
```diff
@@ -31,39 +27,34 @@ impl EthereumConnectorMock {
...
-        let latest_blocks = latest_blocks.into_iter().collect();
-
         EthereumConnectorMock {
             all_blocks,
-            latest_blocks,
-            latest_time_return_block: Instant::now(),
-            current_latest_block_index: 0,
             receipts: receipts.into_iter().collect(),
+            latest_blocks: Mutex::new(
+                throttle(Duration::from_secs(1), stream::iter(latest_blocks)).boxed(),
```
da-kami: cool!

luckysori: yes!

PR #2210, opened 2020-03-10
Simplify the `walk_back_until` algorithm

thomaseizinger:
This patch improves the walk_back_until algorithm by removing the duplication at the beginning the function. With some simple re-orderings and re-assignments, we can remove a call to `clone` on the block and remove the duplication in the `stop_condition` check which is now named `should_stop_here` to be more descriptive.

### DIFF
--- a/cnd/src/btsieve/mod.rs
+++ b/cnd/src/btsieve/mod.rs
@@ -67,7 +67,7 @@ where
 
     // Look back in time until we get a block that predates start_of_swap.
     let mut seen_blocks =
-        walk_back_until(predates_start_of_swap(start_of_swap), connector, &co, block).await?;
+        walk_back_until(predates_start_of_swap(start_of_swap), block, connector, &co).await?;
 
     // Look forward in time, but keep going back for missed blocks
     loop {
@@ -75,9 +75,9 @@ where
 
         let missed_blocks = walk_back_until(
             seen_block_or_predates_start_of_swap(&seen_blocks, start_of_swap),
+            block,
             connector,
             &co,
-            block,
         )
         .await?;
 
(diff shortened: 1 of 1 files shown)


cnd/src/btsieve/mod.rs
```diff
@@ -88,39 +88,37 @@ where
...
+    let mut current_block = starting_block;
 
     loop {
-        let block = connector.block_by_hash(blockhash).await?;
-        co.yield_(block.clone()).await;
-        seen_blocks.insert(blockhash);
+        seen_blocks.insert(current_blockhash);
 
-        if stop_condition(&block) {
+        // compute variables before consuming the block by yielding it
```
tcharding:
```suggestion
        // yielding the block consumes it
```

tcharding: Use comments that say _why_ we do something not _what_ we are doing :)

cnd/src/btsieve/mod.rs
```diff
@@ -88,39 +88,37 @@ where
...
+        co.yield_(current_block).await;
+
+        if should_stop_here {
             return Ok(seen_blocks);
-        } else {
-            blockhash = block.previous_block_hash();
         }
+
+        current_block = connector.block_by_hash(current_blockhash).await?
     }
```
tcharding: Its not immediately obvious from this diff as to why we don't have to call `previous_block_hash()`?  Just flagging it with this comment, leaving the judgement up to you.

thomaseizinger: We don't have to because it is already computed before :)

cnd/src/btsieve/mod.rs
thomaseizinger: Well, the comment says _why_ we are doing this here and not further down. I am not sure how reducing the comment to your suggestion makes that more obvious as it removes the link between the comment and where it is placed.

If I would read: "yielding the block consumes it", my first thought would be:

If the comment concerns the yielding, why is it not on top of the yield statement?

And the second one would be: So what? Moving is the default in Rust.

The reason I added the comment is because it is not immediately obvious that we have to compute these values before we yield even though we only need them afterwards.

cnd/src/btsieve/mod.rs
thomaseizinger: That is exactly what the comment above is supposed to explain but I failed apparently ..

tcharding: Writing clear documentation is non-trivial :)

thomaseizinger: I pushed an update to the comment before it merged, have a look if that is more understandable!

Issue #2111, opened 2020-02-27
Implement all "oneshot" protocols listed in the communication layer spike


thomaseizinger:
## Problem

The new communication layer involves several new libp2p protocols.

## Goal

A tested and working implementation of the "oneshot" protocols described in the [spike](https://github.com/comit-network/spikes/blob/master/0023-communication-protocol.adoc#recommendation) using `rust-libp2p`.

## Recommendation

The scope of this ticket are all "oneshot" protocols described in the spike:

- `/comit/swap/secrethash/1.0.0`
- `/comit/swap/identity/bitcoin/1.0.0`
- `/comit/swap/identity/ethereum/1.0.0`
- `/comit/swap/identity/lightning/1.0.0`
- `/comit/swap/finalize/1.0.0`

Take inspiration from the way, other oneshot protocols in libp2p are implemented:

https://github.com/libp2p/rust-libp2p/blob/96cc7912a3f885ac9a09ed6eb31ae8854c278eee/protocols/identify/src/protocol.rs#L136-L166

As you can see, the parsing already happens within the implementation of `OutboundUpgrade`.
This is completely different from our current implementation where we only upgrade the incoming stream to a `Framed` one that allows us to read frames.

The above listed approach is much simpler and should be preferred for our new implementation.

Inspire yourself from the libp2p lib to see how they test it.

Use JSON via serde.

PR #2215, opened 2020-03-10
Add secret hash protocol

tcharding:
Add a oneshot protocol for sharing the secret hash.  This implementation is based on the `/ipfs/id/1.0.0` protocol.  We push the data from the dialer to the listener (i.e., from Alice to Bob).

Makes a start on: #2111


cnd/src/main.rs
```diff
@@ -178,11 +193,19 @@ fn dump_config(settings: Settings) -> anyhow::Result<()> {
 }
 
 /// Throws an error if the connected connected blockchain network is invalid
-fn handle_invalid_network_validation<T: std::fmt::Debug>(validation: NetworkValidationResult<T>) -> anyhow::Result<(), ConfigValidationError<T>> {
+fn handle_invalid_network_validation<T: std::fmt::Debug>(
```
thomaseizinger: What is going on here?
Are we not validating rustfmt in CI anymore or why is this changed?

libp2p-comit/Cargo.toml
```diff
@@ -18,4 +18,5 @@ thiserror = "1"
 tracing = "0.1"
 
 [dev-dependencies]
+async-std = "1.0"
```
thomaseizinger: Please use tokio for testing!

libp2p-comit/src/protocol.rs
```diff
@@ -48,3 +50,173 @@ impl OutboundUpgrade<NegotiatedSubstream> for Config {
...
+        async move { upgrade::write_one(&mut self.inner, &message).await }
+    }
+}
+
+impl UpgradeInfo for SecretHashProtocolConfig {
+    type Info = &'static [u8];
+    type InfoIter = iter::Once<Self::Info>;
+
+    fn protocol_info(&self) -> Self::InfoIter {
+        iter::once(b"/comit/swap/secret_hash/1.0.0") // FIXME: add a helper
```
thomaseizinger: This is fine, no need to change anything here :)

libp2p-comit/src/protocol.rs
```diff
@@ -48,3 +50,173 @@ impl OutboundUpgrade<NegotiatedSubstream> for Config {
...
+    C: AsyncRead + AsyncWrite + Unpin + Send + 'static,
+{
+    type Output = SecretHashInfo;
+    type Error = upgrade::ReadOneError;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
+
+    fn upgrade_outbound(self, mut socket: C, _: Self::Info) -> Self::Future {
+        trace!("Upgrading outbound connection");
+        Box::pin(async move {
+            let msg = upgrade::read_one(&mut socket, 4096).await?; // FIXME: how much data to read?
```
thomaseizinger: Given we know the size of the message (UUID for swapid and secret hash are always the same size, we can calculate this upfront. What is to be investigated is whether this is a buffer or actually a fixed amount. We cannot read more than the sender is trying to send.

rust-toolchain
```diff
@@ -1 +1 @@
-1.41.1
+1.41.0
```
thomaseizinger: Nope.

libp2p-comit/src/protocol.rs
```diff
@@ -48,3 +50,173 @@ impl OutboundUpgrade<NegotiatedSubstream> for Config {
...
+                                                     // function to generate
+                                                     // this.
+    }
+}
+
+impl<C> InboundUpgrade<C> for SecretHashProtocolConfig
+where
+    C: AsyncRead + AsyncWrite + Unpin,
+{
+    type Output = ReplySubstream<C>;
```
thomaseizinger: I've seen this design in the libp2p codebase and always wondered what the motivation is.

Do you think we can figure this out?
More concretely, what is the reason that `InboundUpgrade` cannot mirror `OutboundUpgrade` and just write to the socket directly instead of reading?

The consequence would be that `SecretHashProtocolConfig` would have to have the `SecretHashInfo` as a field so it can send it. That is obviously useless for the `OutboundUpgrade` trait. But those two don't necessarily have to be implemented on the same struct. There could be one protocol config for inbound and one for outbound.

We should experiment with this idea to understand the need for this particular design because at the moment, I don't see a specific one.

libp2p-comit/src/protocol.rs
```diff
@@ -48,3 +50,173 @@ impl OutboundUpgrade<NegotiatedSubstream> for Config {
...
+        Transport,
+    };
+    use libp2p_tcp::TcpConfig;
+
+    #[test]
+    fn correct_transfer() {
+        // We open a server and a client, send info from the server to the client, and
+        // check that they were successfully received.
+
+        let send_swap_id = "fixme";
```
thomaseizinger: Would be good to use values that reflect the actual length in the test.

rust-toolchain
tcharding: lol, I wondered last night when I was going to sleep if I'd left this in :)

libp2p-comit/src/protocol.rs
tcharding: > I've seen this design in the libp2p codebase and always wondered what the motivation is.
>
> Do you think we can figure this out?

I've got a feeling it will work out nicer if we follow libp2p's example but use an immediately ready future for the outbound upgrade that resolves to a type that we can call send on to send the data (instead of putting it in the config object).  I've not got a fully formed argument for this so I'll explore the idea further then either comment more or post a follow up PR.

thomaseizinger: I would like us to figure out why we are doing this indirection before going for it. Opening a substream is going to require us to create a `ProtocolsHandlerEvent::OutboundSubstreamRequest` like here: https://github.com/comit-network/comit-rs/blob/76c7e4421fba55c903256bf69cb4be6e8edd2b43/libp2p-comit/src/handler.rs#L189-L192

Here we instantiate the config and pass some info along. The info is passed back to us in `inject_fully_negotiated_outbound`:

https://github.com/comit-network/comit-rs/blob/76c7e4421fba55c903256bf69cb4be6e8edd2b43/libp2p-comit/src/handler.rs#L141-L157

I don't see the point of calling `send` on the first argument with the `info` that is passed along. The upside I see is that we can have the same `Config` implementation for both, upgrading inbound and outbound.

I feel like it should be symmetric: Either we only upgrade and return a substream in both cases or we send/receive in both cases.

I would be keen to do the latter because it means we have libp2p-core driving the future to completion instead of us having to store it inside the `ProtocolsHandler` implementation and manually poll it.

libp2p-comit/Cargo.toml
thomaseizinger: You should be able to do `#[tokio::test]` and make the fn async.

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,207 @@
...
+pub struct SecretHashProtocolConfig;
+
+/// Data sent to peer in secret hash protocol.
+#[derive(Deserialize, Debug, Serialize)]
+pub struct SecretHashInfo {
+    swap_id: SwapId,
+    secret_hash: SecretHash,
+}
+
+lazy_static::lazy_static! {
```
thomaseizinger: The sizes are always the same. Just add a test if you want to be sure, no need to pay this runtime overhead.

libp2p-comit/src/protocol.rs
tcharding: Ok, I missed this comment until now and have not done as suggested while asking for 50% review.  Will reconsider now I've read this.

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,207 @@
...
+    C: AsyncRead + AsyncWrite + Unpin + Send + Sync + 'static,
+{
+    type Output = SecretHashInfo;
+    type Error = upgrade::ReadOneError;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
+
+    fn upgrade_inbound(self, mut socket: C, _: Self::Info) -> Self::Future {
+        trace!("Upgrading inbound connection");
+        Box::pin(async move {
+            let msg = upgrade::read_one(&mut socket, *JSON_INFO_LENGTH).await?;
```
thomaseizinger: Did you do research/testing we actually need to pass the exact same length here? I wouldn't want to have this complexity if this is just the max buffer size and also works if the message is shorter.

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,207 @@
...
+        let default_swap_id = format!("{}", default_swap_id);
+
+        let want = static_swap_id.len();
+        let got = default_swap_id.len();
+
+        assert_eq!(want, got)
+    }
+
+    #[test]
+    fn correct_transfer() {
```
thomaseizinger: I think your test will abort too early because you are using spawn two times ...

You need to block/await on something.

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,207 @@
...
+
+impl<T> PushSubstream<T>
+where
+    T: AsyncWrite + Unpin,
+{
+    /// Sends the information we wish to push on the substream.
+    ///
+    /// Consumes the substream, returning a future that resolves
+    /// when the info has been sent on the underlying connection.
+    pub async fn send(mut self, info: SecretHashInfo) -> Result<(), io::Error> {
```
thomaseizinger: As mentioned in the comment, I am not a big fan of this design as it will force us to manually drive this future to completion in the `poll` of the protocols handler. Or how is rust-libp2p doing this?

tcharding:
Ok, updated to use the config struct to hold the secret hash info.  Added a couple of constructors for creating the `sender` and `receiver` config struct and used them in the tests.  This actually works quite nicely, thanks for keeping on pushing for this.

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,180 @@
...
+            let (listener_upgrade, _) = listener_event.unwrap().into_upgrade().unwrap();
+            let conn = listener_upgrade.await.unwrap();
+            let config = SecretHashProtocolConfig::sender(SecretHashInfo {
+                swap_id: send_swap_id,
+                secret_hash: send_secret_hash,
+            });
+            apply_outbound(conn, config, Version::V1).await.unwrap();
+        });
+
+        tokio::task::spawn_blocking(|| async move {
```
thomaseizinger: I don't think `spawn_blocking` does what you think it is doing. It is meant for spawning blocking code, not blocking until it is done!

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,180 @@
...
+    C: AsyncRead + AsyncWrite + Unpin + Send + Sync + 'static,
+{
+    type Output = ();
+    type Error = anyhow::Error;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
+
+    fn upgrade_outbound(self, mut socket: C, _: Self::Info) -> Self::Future {
+        trace!("Upgrading outbound connection");
+        Box::pin(async move {
+            let message = serde_json::to_vec(&self.info)?;
```
thomaseizinger: You are serializing an option here. I think we should either unwrap this because it is a bug if it is not set or refactor to have two different structs for the config to make it impossible to call `OutboundUpgrade` if we didn't supply the info.

cnd/src/network/protocol.rs
thomaseizinger: We are going to write several of these tests. I suggest to refactor something reusable out of this as soon as we write the 2nd one to not duplicate too much.

cnd/src/network/protocol.rs
thomaseizinger: https://docs.rs/tokio/0.2.13/tokio/task/fn.spawn_blocking.html

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,180 @@
...
+        let send_secret_hash = secret.hash();
+        let send_swap_id = SWAP_ID;
+
+        let (tx, rx) = oneshot::channel();
+
+        let mem_addr = multiaddr![Memory(thread_rng().gen::<u64>())];
+        let mut listener = MemoryTransport.listen_on(mem_addr).unwrap();
+
+        let listener_addr =
+            if let Some(Some(Ok(ListenerEvent::NewAddress(a)))) = listener.next().now_or_never() {
```
thomaseizinger: Lol, this is nice: https://docs.rs/futures/0.3.4/futures/future/trait.FutureExt.html#method.now_or_never

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,180 @@
...
+        let got = default_swap_id.len();
+
+        assert_eq!(want, got)
+    }
+
+    #[test]
+    fn correct_transfer() {
+        let bytes = b"hello world, you are beautiful!!";
+        let secret = Secret::from(*bytes);
+        let send_secret_hash = secret.hash();
```
thomaseizinger: Would prefer less setup noise here and just take any 32byte hex string as hash.

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// Receiver receives the secret hash from the peer.
+    pub fn receiver() -> Self {
+        SecretHashProtocolConfig { info: None }
+    }
+}
+
+/// Data sent to peer in secret hash protocol.
+#[derive(Clone, Deserialize, Debug, Serialize)]
+pub struct SecretHashInfo {
+    swap_id: SwapId,
```
thomaseizinger: You could make these fixed length byte arrays, then the length is going to be guaranteed at compile-time and you can chuck all these tests out.

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,180 @@
...
+    C: AsyncRead + AsyncWrite + Unpin + Send + Sync + 'static,
+{
+    type Output = SecretHashInfo;
+    type Error = anyhow::Error;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
+
+    fn upgrade_inbound(self, mut socket: C, _: Self::Info) -> Self::Future {
+        trace!("Upgrading inbound connection");
+        Box::pin(async move {
+            let message = upgrade::read_one(&mut socket, *JSON_INFO_LENGTH).await?;
```
thomaseizinger: Did you get to the bottom of whether we need this length to be exact?

cnd/src/network/protocol.rs
tcharding: oh, cheers.

thomaseizinger: Take a look at the `tokio::test` attribute :)

cnd/src/network/protocol.rs
tcharding: I didn't look into it, it worked both reading a 4096 byte buffer and also using lazy_static! to read just the right amount of bytes.

cnd/src/network/protocol.rs
tcharding: done, nice.

cnd/src/network/protocol.rs
tcharding: Fixed, not entirely happy with the code though.
```
       if self.info.is_none() {
            return Box::pin(future::err(anyhow::anyhow!(
                "cannot upgrade without secret hash info"
            )));
        }

        let info = self.info.unwrap();
```

.circleci/config.yml
```diff
@@ -87,8 +87,8 @@ commands:
           name:  "Install Rust"
           command: |
             curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain $(< rust-toolchain) && source $HOME/.cargo/env
-            rustup install nightly-2019-07-31
-            rustup component add rustfmt --toolchain nightly-2019-07-31
+            rustup install nightly-2020-01-15
```
thomaseizinger: That is already merged by now, feel free to rebase :)

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,160 @@
+use futures::{future, AsyncRead, AsyncWrite, Future};
+use libp2p::core::{upgrade, InboundUpgrade, OutboundUpgrade, UpgradeInfo};
+use serde::{Deserialize, Serialize};
+use std::{iter, pin::Pin};
+use tracing::trace;
+
+const LENGTH: usize = 32;
+
+lazy_static::lazy_static! {
```
thomaseizinger: Why do we still have this? I find this very confusing.

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub fn receiver() -> Self {
+        SecretHashProtocolConfig { info: None }
+    }
+}
+
+/// Data sent to peer in secret hash protocol.
+#[derive(Clone, Deserialize, Debug, Serialize)]
+pub struct SecretHashInfo {
+    swap_id: [u8; LENGTH],
+    secret_hash: [u8; LENGTH],
```
thomaseizinger: These two fields just happen to have the same length but that is not semantically related. Using the same const for that feels wrong to me.

tcharding: And its wrong altogether :)  one of them is 16 bytes and one is 32.  I've fixed this, sorry for leaving an old version open for review.  Same goes for the `lazy_static` comment above.

.circleci/config.yml
tcharding: Cool, cheers.

tcharding:
> The spike describes the protocol as push: Alice sends the secret hash to Bob (she generates it and Bob needs to learn about it).

Just to clarify; by 'push' vs 'pull' I mean within the libp2p substream communication for data to get from Alice to Bob _either_ Alice or Bob can be the dialer and either can be the listener.  In other words, Bob can dial Alice to 'pull' the secret hash OR Alice can dial Bob to 'push' the secret.  I have read all the protocols in the libp2p, the `OneShotHandler` code, and the docs and I'm still not sure how to tie the protocol configs (In/Out) together with the `InEvent`.  And I'm not sure if `OneShotHandler` expects the dialer or the listener to send the data ???

cnd/src/main.rs
tcharding: This was a tooling issue, resolved by: https://github.com/comit-network/comit-rs/pull/2218

libp2p-comit/src/protocol.rs
tcharding: Note: @thomaseizinger discovered that the size is just a maximum to prevent DOS attacks against a node.  Using 1024.

cnd/src/network/behaviour.rs
```diff
@@ -0,0 +1,160 @@
...
+        NetworkBehaviourAction<
+            <Self::ProtocolsHandler as ProtocolsHandler>::InEvent,
+            Self::OutEvent,
+        >,
+    > {
+        if let Some(event) = self.events.pop_front() {
+            return Poll::Ready(event);
+        }
+
+        Poll::Pending
```
thomaseizinger: If you do this, you will have to get a reference to the current waker (`context.waker()`) and store it inside `self`.

Whenever you change the `events` queue (inside `inject_node_event`) you will then have to check if you currently have a waker `if let Some(waker) ...` and call `waker.wake()` to tell the runtime that you would like to be polled again :)

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,162 @@
...
+pub struct OutboundProtocolConfig {
+    msg: Message,
+}
+
+impl UpgradeInfo for OutboundProtocolConfig {
+    type Info = &'static [u8];
+    type InfoIter = iter::Once<Self::Info>;
+
+    fn protocol_info(&self) -> Self::InfoIter {
+        iter::once(b"/comit/swap/secret_hash/1.0.0")
```
thomaseizinger: I think it would be good to share this protocol string in a const between the two configs.

cnd/src/network/behaviour.rs
```diff
@@ -0,0 +1,160 @@
...
+pub struct Behaviour {
+    /// Events that need to be yielded to the outside when polling.
+    events: VecDeque<NetworkBehaviourAction<OutboundProtocolConfig, OutEvent>>,
+}
+
+/// Event generated by the NetworkBehaviour and that the swarm will report back.
+#[derive(Clone, Copy, Debug)]
+pub enum OutEvent {
+    Rx(Message), // OutEvent containing secret hash message.
+    Sent,        // Empty/nil OutEvent i.e., `()`.
```
thomaseizinger: Who is generating the `Sent` event, do we really need this?

cnd/src/network/protocol.rs
tcharding: Lols, see this comment above: https://github.com/comit-network/comit-rs/pull/2215#discussion_r390159147

cnd/src/network/behaviour.rs
tcharding: I'll be damned if I can work out when events get generated but from what I can tell this event gets generated when we are the dialer and we do the OutboundUpgrade ???  At least the compiler requires us to have this and also the From<()> impl.  Also the Floodsub protocol as this like this also.  Please do school me if you work this out.

cnd/src/network/behaviour.rs
tcharding: The floodsub implementation does not use a waker?  I cut'n'pasta the `poll` and copied the ideas for the `inject_node_event`.  Are you sure we need it?

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,163 @@
...
+        trace!(
+            "Upgrading outbound connection: {}",
+            String::from_utf8_lossy(info)
+        );
+        Box::pin(async move {
+            let bytes = serde_json::to_vec(&self.msg).unwrap();
+            upgrade::write_one(&mut socket, &bytes).await.unwrap();
+
+            Ok(())
+        })
```
thomaseizinger:
```suggestion
        async move {
            let bytes = serde_json::to_vec(&self.msg).unwrap();
            upgrade::write_one(&mut socket, &bytes).await.unwrap();

            Ok(())
        }.boxed()
```

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,163 @@
...
+    type Error = Error;
+    type Future = BoxFuture<'static, Result<Self::Output, Self::Error>>;
+
+    fn upgrade_outbound(self, mut socket: C, info: Self::Info) -> Self::Future {
+        trace!(
+            "Upgrading outbound connection: {}",
+            String::from_utf8_lossy(info)
+        );
+        Box::pin(async move {
+            let bytes = serde_json::to_vec(&self.msg).unwrap();
```
thomaseizinger: This is still missing error handling!

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,163 @@
...
+
+            Ok(info)
+        })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("read: ")]
+    Read(#[from] upgrade::ReadOneError),
```
thomaseizinger: With the added error handling for the write part, this is a bit weird because write just emits and `io::Error` and the `io::Error` is part of this one.

Can you figure something out here that makes sense? Potentially flattening the error cases maybe?

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,163 @@
...
+            let info = Message::deserialize(&mut de)?;
+
+            Ok(info)
+        })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("read: ")]
```
thomaseizinger: Please write some more descriptive messages here.

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,163 @@
...
+        let mut listener = MemoryTransport.listen_on(mem_addr).unwrap();
+
+        let send_msg = Message {
+            swap_id: SwapId::from_str("ad2652ca-ecf2-4cc6-b35c-b4351ac28a34").unwrap(),
+            secret_hash: SecretHash::from_str(
+                "bfbfbfbfbfbfbfbfbfbfbfbfbfbfbfbf\
+                 bfbfbfbfbfbfbfbfbfbfbfbfbfbfbfbf",
+            )
+            .unwrap(),
+        };
```
thomaseizinger: The initialization order here is a bit weird because the message is in the middle of creating a listener and interacting with it.

```suggestion
		let send_msg = Message {
            swap_id: SwapId::from_str("ad2652ca-ecf2-4cc6-b35c-b4351ac28a34").unwrap(),
            secret_hash: SecretHash::from_str(
                "bfbfbfbfbfbfbfbfbfbfbfbfbfbfbfbf\
                 bfbfbfbfbfbfbfbfbfbfbfbfbfbfbfbf",
            )
            .unwrap(),
        };
        
        let mem_addr = multiaddr![Memory(thread_rng().gen::<u64>())];
        let mut listener = MemoryTransport.listen_on(mem_addr).unwrap();
```

cnd/src/network/protocol.rs
```diff
@@ -0,0 +1,163 @@
...
+                    .unwrap();
+            }
+        });
+
+        let conn = MemoryTransport.dial(listener_addr).unwrap().await.unwrap();
+
+        let config = InboundProtocolConfig {};
+        let msg = upgrade::apply_inbound(conn, config).await.unwrap();
+
+        assert_eq!(send_msg, msg)
```
thomaseizinger: I think the order is "actual", "expected".

```suggestion
        assert_eq!(received_msg, sent_msg)
```

cnd/src/swap_protocols/swap_id.rs
```diff
@@ -5,6 +5,12 @@ use uuid::Uuid;
 #[derive(Serialize, Deserialize, Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct SwapId(pub Uuid);
 
+impl SwapId {
+    pub fn as_bytes(&self) -> &[u8] {
```
thomaseizinger: This seems to be a relic from using byte arrays in the message.

tcharding: Nice catch, thanks.

cnd/src/network/behaviour.rs
```diff
@@ -0,0 +1,165 @@
...
+        swarm::Swarm,
+        yamux,
+    };
+    use rand::random;
+    use std::{io, str::FromStr, u64};
+
+    type TestSwarm = Swarm<Behaviour>;
+
+    /// Builds 2 swarms, each listening on a port.
+    fn build_nodes(num: usize) -> (Vec<PeerId>, Vec<TestSwarm>) {
```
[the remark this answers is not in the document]
tcharding: All these tests have been removed.

tcharding:
> LGTM, I trust that the terminology used here matches what is done in libp2p (Especially regarding the `INFO` const).

Yep, that's what they call it
```
impl UpgradeInfo for OutboundProtocolConfig {
    type Info = &'static [u8];
    type InfoIter = iter::Once<Self::Info>;
    ...
```

PR #2218, opened 2020-03-11
Use new rustfmt toolchain

tcharding:
Run make format with the newly updated toolchain

Recently we updated the toolchain to 1.41.1, and also the nightly toolchain to `nightly-2020-01-15`.  CI is configured to run a different version of `rustfmt` so this got past CI.  However, this upgrade introduce a bunch of new formatting to our codebase.

- Run `make format`
- Configure CI to use the same nightly toolchain as the makefile

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,8 +87,8 @@ commands:
           name:  "Install Rust"
           command: |
             curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain $(< rust-toolchain) && source $HOME/.cargo/env
-            rustup install nightly-2019-07-31
-            rustup component add rustfmt --toolchain nightly-2019-07-31
+            rustup install nightly-2020-01-15
+            rustup component add rustfmt --toolchain nightly-2020-01-15
 
             # Define variables that need interpolation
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -87,8 +87,8 @@ commands:
           name:  "Install Rust"
           command: |
             curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain $(< rust-toolchain) && source $HOME/.cargo/env
-            rustup install nightly-2019-07-31
-            rustup component add rustfmt --toolchain nightly-2019-07-31
+            rustup install nightly-2020-01-15
```
thomaseizinger: Ah yeah, we should fix this by not install something here at all :/

The makefile will actually install Rust correctly if rustup is available. Do you think you can change the CI scripts to just use that so we don't forget that in the future and the makefile is the single source of truth?

Can be a follow-up PR though, this should get in ASAP.

thomaseizinger: It is a bit tricky because of caching, we should discard the cache on update but we can't get the knowledge out of the makefile ..

PR #2219, opened 2020-03-11
Remove some noise from the e2e test declarations

thomaseizinger:
We have several layers of function calls here that can be reduced.
Unfortunately, it does not reduce the indentation yet but I think
this is a step in the right direction. Anyway, we can remove a layer
of indirection that is unnecessary.

### DIFF
--- a/api_tests/lib/actor_test.ts
+++ b/api_tests/lib/actor_test.ts
@@ -2,59 +2,68 @@ import { Actors } from "./actors";
 import { createActors } from "./create_actors";
 import JasmineSmacker from "smack-my-jasmine-up";
 import { timeout } from "./utils";
+import ProvidesCallback = jest.ProvidesCallback;
 
 /*
  * This test function will take care of instantiating the actors and tearing them down again
  * after the test, regardless if the test succeeded or failed.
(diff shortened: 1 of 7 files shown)


D4nte:
Are you saying it's a step in the right direction because we are decrease functions calls in favour of returning functions themselves?

thomaseizinger:
> Are you saying it's a step in the right direction because we are decreasing functions calls in favour of returning functions themselves?

Yes basically. Overall, anything that decreases noise in tests is usually good :)

D4nte:
> > Are you saying it's a step in the right direction because we are decreasing functions calls in favour of returning functions themselves?
> 
> Yes basically. Overall, anything that decreases noise in tests is usually good :)

How does it decreases noise?

thomaseizinger:
> > > Are you saying it's a step in the right direction because we are decreasing functions calls in favour of returning functions themselves?
> > 
> > 
> > Yes basically. Overall, anything that decreases noise in tests is usually good :)
> 
> How does it decrease noise?

The patch removes one instance of the `await`, `async` and `function` keyword per test.

Issue #2214, opened 2020-03-10
Cnd returns the internal error when trying to GET /swaps


xcchain:
## Problem

We’ve updated our cnd to the latest version from master and took the following steps:

1. Deployed 2 cnd nodes on a local machine running on localhost:8000 and localhost:8001 (to do a test swap across two nodes from our DEX app)

2. We send POST localhost:8000/swaps/rfc003 with the following:

```
{
    "alpha_ledger": {
        "name": "bitcoin",
        "network": "testnet"
    },
    "alpha_asset": {
        "name": "bitcoin",
        "quantity": "1000"
    },
    "beta_ledger": {
        "name": "ethereum",
        "chain_id": 3
    },
    "beta_asset": {
        "name": "ether",
        "quantity": "100"
    },
    "alpha_expiry": 1000,
    "beta_expiry": 1000,
    "beta_ledger_redeem_identity": "0xcb46c0e906950CaD69906815e21a84794D5da07a",
    "peer": {
        "peer_id": "QmPea6iEGU5gKSd6JCkPirDXerabYewAHWBY73ujswKYrL",
        "address_hint": "/ip4/127.0.0.1/tcp/9940"
    }
}
```

3. We send GET localhost:8001/swaps and receive this error:

`[2mMar 09 11:46:40.066 [0m  [31mERROR [0m cnd::http_api::problem: internal error occurred: state store did not contain an entry for 48bf8f7f-b8d5-4187-a476-411815a2f414`

**Summary**: the second cnd node does see an ID of the swap request (created on the first cnd node) but cannot retrieve its details, giving us the internal error instead. 

## Goal

Get a list of swaps known by cnd, no error message.

## Proposed solution

1. Investigation
2. Fix

thomaseizinger:
Hi @xcchain,

thanks for the report! Would you be able to turn the log level to trace within the config file (use `cnd --dump-config` to explore the format) and send us the logs of both `cnd` when this error happens? That would be really useful.

Thanks!

D4nte:
@xcchain `dev` branch is the branch with latest change. `master` only (since recently) contains merge commit for the latest release.
I was not able to pre-reproduce the issue on the `dev` branch. Please try out this branch.

@thomaseizinger can you please proceed with a release now that we have fixed the docker image build?

thomaseizinger:
On it: https://github.com/comit-network/comit-rs/issues/2224

llilakoblock:
@D4nte
[...]

PR #2222, opened 2020-03-11
Reproduce #2214

D4nte:
It actually works on dev.


D4nte:
Fault was due to 2 cnds using same data dir.

PR #2226, opened 2020-03-11
Swap Digest

### LINKED ISSUE #2112: Implement the swap digest algorithm
## Problem

The "announce" protocol described in the communication protocol describes an algorithm for computing a swap digest. We need an implementation of this algorithm.

See the related spike: https://github.com/comit-network/spikes/blob/master/0023-communication-protocol.adoc

## Goal

A working and tested implementation of the described algorithm.
We don't yet exactly know, how to things will fall into place, so the implementation should **not** build on existing data structures like `Request` or `ledger::Bitcoin` but instead be independent of the domain.

## Recommendation

Create a generic and Rust-idiomatic implementation of the algorithm using traits and functions. Ideally, in the end we should only need to implement the trait on the types we want to include in the digest and delegate to a function that implements the necessary parts of the algorithm.

### What is in scope of this ticket?

- Implement the interfaces (traits) for the algorithm
- Implement crucial details of the algorithm like the lexical sorting and hashing
- Define a way by giving examples in tests for how the implemented algorithm is used on:

1. structs with several fields
2. structs with a single field
3. enums

We are going to encounter all of the above when we are using the implementation on the actual structs so the implementation created in this ticket should be prepared to handle them.

The code will be isolated and not used at this stage. Rust tests should be here to prove it is working of course.

It is a good candidate for TDD or design/API first development.

### What is not in scope of this ticket?

- **Implementing** the conversion of a specific swap-related item like a Ledger to binary as the input for the algorithm.

D4nte:
Resolves #2112.

_99% Review_

If the current code and approach looks good then let's merge it and do a second PR to show case enums and nested structs.


cnd/src/swap_digest.rs
```diff
@@ -0,0 +1,30 @@
...
+
+enum Error {
+    SerdeCbor(()),
+}
+
+trait Digest {
+    fn digest(&self) -> Multihash;
+}
+
+impl Digest for String {
```
tcharding: I'm happy with this name, `Digest` is a noun and this function returns the digest and _does not_ change program state.  (Comment added here as preamble to discussion on names of `DigestField` and `DigestRoot`.)

digest/src/lib.rs
```diff
@@ -0,0 +1,37 @@
+use multihash::Multihash;
+
+pub fn digest(bytes: &[u8]) -> Multihash {
+    // Time the tests and take fastest hash?
+    multihash::Sha3_256::digest(bytes)
+}
+
+pub trait DigestRoot {
```
tcharding: This trait needs renaming IMO.  It took me a while to workout exactly _why_.  As commented before `Digest` trait is good - its a noun.  But `DigestRoot` is using the word 'digest' as a verb, in crypto 'digest' is a noun not a verb - it is the output of a hash function.  Can we re-name this to be a noun?  I'm struggling to find something nice and short; 'FieldSortedDigest', 'OrderedDigest', 'FieldOrderedDigest'

digest/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+    fn digest_root(self) -> Multihash;
+}
+
+pub trait DigestField {
+    fn digest_field(self, field_name: String) -> Multihash;
+}
+
+impl DigestField for String {
```
tcharding: With the above discussion on noun/verb usage of 'digest', perhaps we could re-name this `FieldDigest` - then its a noun.

digest/src/lib.rs
tcharding: Oh, this could just become `RootDigest` and then 'digest' is a noun.

api_tests/tests/dry/multiple_peers.ts
```diff
@@ -21,57 +21,65 @@ function toMatch(swapDetail: SwapDetails): MatchInterface {
     };
 }
 
+async function createSwapRequest(counterParty: Actor) {
```
tcharding: Why is this in this PR please?

cnd/src/swap_digest.rs
```diff
@@ -0,0 +1,170 @@
+use digest::{digest, DigestField, DigestRoot};
+use digest_macro_derive::DigestRootMacro;
+
+use multihash::Multihash;
+
+struct NewType(String);
+
+impl DigestRoot for NewType {
```
tcharding: I reviewed patch by patch initially, now going through 'files changed'.  Please excuse the double posts.

This whole file is testing right?  Can it all go under a conditionally compiled module please i.e., `cfg(test)`.  Perhaps it should all be in the `digest` crate?

digest-macro-derive/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+        if let Fields::Named(fields) = &data.fields {
+            let idents = fields
+                .named
+                .iter()
+                .map(|field| field.ident.as_ref().expect("Named field"));
+
+            let types = fields.named.iter().map(|field| &field.ty);
+
+            let gen = quote! {
+            impl DigestRoot for #name
```
tcharding: This is using types that it does not import, perhaps we need `use DigestRoot` here otherwise there is an added burden on the macro user to add these use statements?

digest-macro-derive/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+                            res
+                        });
+
+                        digest(&res)
+                    }
+                }
+            };
+            gen.into()
+        } else {
+            panic!("DigestRootMacro does not support new types.");
```
tcharding: I don't quite understand this error path.  This matches on `Fields` and is ok for `Fields::Named` so this error path is for `Fields::Unnamed` or `Fields::Unit`.  Perhaps this string could be:

"DigestRootMacro only supports named fields"

D4nte:
> Nice Franck, proc macro is sexy. My comments are all mixed up I'm afraid because I reviewed by patch first then all together. I'm still learning how to work with GitHub to review :)
> 
> One suggestion for improving your, if you are open to it; if you are going to push PRs with multiple commits, it implies, at least to me, that review can be done per commit. If this is the case the commits should be discreet steps towards the goal _not_ arbitary steps one took while developing i.e., patch 3 should not make major changes to code added in patch one. Does this make sense? Otherwise reviewer time is spent commenting on things that change later in the series ('patch series' == series of patches making up the PR). Requesting changes just for the re-naming discussion of `DigestField` and `DigestRoot`.

Hi @tcharding, I am definitely happy to be more mindful on the patch series. I feel like this is not really engrained in our culture and hence I'd suggest we should talk more about it as a team.

Also the first patch was a mistake due to branching for the wrong branch. Sorry about that.

bonomat:
> Hi @tcharding, I am definitely happy to be more mindful on the patch series. I feel like this is not really engrained in our culture and hence I'd suggest we should talk more about it as a team.

sorry for off-topic but I think this is important to point out. We haven't used this approach of reviewing in our team and I would be careful with applying it @tcharding. 
For example in this PR it looks like you added a comment which which addressed outdated code (see below). Imho best is to review the final stage only. 
Let's discuss this in the team and not in this PR though :)


image

p.s. sorry for posting this and killing the discussion immediately.

api_tests/tests/dry/multiple_peers.ts
D4nte: Mistake indeed, now fixed.

cnd/src/swap_digest.rs
D4nte: Excellent point.

digest-macro-derive/src/lib.rs
D4nte: "named field" did not mean much to me, so I wanted to use terminology I am more familiar such as `new types`. But indeed, the current message does not cover `tuples` and `unit types`.

digest/src/lib.rs
```diff
@@ -0,0 +1,40 @@
+pub use multihash;
+use multihash::Multihash;
+
+const SEPARATOR: &[u8; 1] = b":";
+
+pub fn digest(bytes: &[u8]) -> Multihash {
+    // Time the tests and take fastest hash?
+    multihash::Sha3_256::digest(bytes)
```
D4nte: @thomaseizinger @bonomat do we have any strong feeling regarding the hashing function to use for the digest?

bonomat: Not really. I think we can roll with sha256 as you propose and move to something else in the future if needed.

digest/tests/swap_digest.rs
```diff
@@ -0,0 +1,165 @@
+use digest::{digest, FieldDigest, RootDigest};
+use digest_macro_derive::DigestRootMacro;
+
+use digest::multihash::Multihash;
+
+struct NewType(String);
+
+impl RootDigest for NewType {
+    fn digest_root(self) -> Multihash {
+        self.0.digest_field("".into())
```
thomaseizinger: For tuple structs, we should take into account the position, similarly how we use the field name for regular structs.

digest/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "digest"
```
thomaseizinger: I think the standard approach for proc macros with additional dependencies is to re-export the proc-macro from the other crate so people only have to depend on one crate.

digest/tests/swap_digest.rs
D4nte: Do we want to support tuples and new types? Maybe the best is to just remove this.

D4nte: Actually I changed it so I think it makes sense now.

digest/src/lib.rs
```diff
@@ -0,0 +1,39 @@
+pub use multihash;
+use multihash::Multihash;
+
+const SEPARATOR: &[u8; 1] = b":";
```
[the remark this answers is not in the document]
D4nte: I think it's neater.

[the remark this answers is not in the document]
thomaseizinger: Just thinking ahead for the documentation of this. We have to be precise about every byte here. As long as we achieve that, we should be fine :)

digest/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+    fn digest_root(self) -> Multihash;
+}
+
+pub trait FieldDigest {
+    fn digest_field(self, field_name: String) -> Multihash;
+}
+
+impl FieldDigest for String {
+    fn digest_field(self, field_name: String) -> Multihash {
+        let mut bytes = field_name.into_bytes();
```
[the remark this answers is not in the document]
D4nte: I get your point regarding encoding however I don't see how this helps with the hash of the actual value.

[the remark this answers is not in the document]
thomaseizinger: If we have full control over what is hashed by providing the exact bytes that act as the identifier for this field, we are less prone to bugs due to renaming etc.

It will also allows us for better interoperability at some point where someone in a different language has to implement this algorithm and be complaint with our implementation.

Regarding the actual hash of the value:

Maybe we should **not** be implementing digest on any primitive values. I think it is reasonable to expect consumers of `Digest` to implement the trait themselves and not use primitive types.

D4nte:
> Good stuff, looking forward to this :)
> 
> Put some challenging questions there! I would also like to push again for some through towards a single-trait design. If you can make that happen, I think the solution would be a lot nicer :)

A single trait design would result in something extremely similar to serde's `Serializer` trait. Such trait would contain several function, one for each basic type. I am not convinced this is a step forward though.

Note that the `FieldDigest` trait is only local to the `digest` trait and we could just implement it for basic types in this crate so that the `cnd` crate does not have to be aware of it.

We could also expand the derive macro to support more types of structures so that `cnd` would only need to apply the macro.

digest/src/lib.rs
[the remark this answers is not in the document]
D4nte: Thinking more about it, I still do not understand why you are pushing for a hardcoded byte value for fields.

The problematic that you are raising is the same for field names and `String` values as we are processing field names as `String`.

If the lack of encoding specification is an issue then this issue affect both sides. Because of that, I don't see how a hardcoded byte value per field would help as it would not solve the issue on the value side.

`String::as_bytes` returns the bytes for UTF-8. I think it is fair to just specify that the digest of a field is the hash of `<name of the field as bytes, utf-8 encoded>:<value as bytes, utf-8 encoded>`.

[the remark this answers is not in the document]
D4nte: > Maybe we should **not** be implementing digest on any primitive values. I think it is reasonable to expect consumers of `Digest` to implement the trait themselves and not use primitive types.

I see.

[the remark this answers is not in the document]
D4nte: How do you expect the `FieldDigest` implementation of bitcoin mainnet network to look like? I was going with the utf-8 bytes of `mainnet` string but you seem to be thinking something else.

digest/src/lib.rs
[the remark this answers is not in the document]
D4nte: I don't see how this is going to impact the documentation effort. Yes, precise that it's `<field name bytes>:<value bytes>`

digest/src/lib.rs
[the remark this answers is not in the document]
thomaseizinger: Well, the bitcoin mainnet network actually has an unambigous byte representation: the 4 magic bytes in front of every message. That is something I would have used.

In the context of Ethereum, probably the bytes of the chain id in big/little endian, whatever we specify.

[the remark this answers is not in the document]
D4nte: > Maybe we should **not** be implementing digest on any primitive values. I think it is reasonable to expect consumers of `Digest` to implement the trait themselves and not use primitive types.

We are adding a level of complexity here where the consumer would need to implement a foreign trait on possible foreign types.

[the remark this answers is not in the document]
thomaseizinger: I would say the opposite: Computing the digest correctly is such a crucial thing that you should be forced to implement this on your own structs and not use foreign ones.

thomaseizinger:
> > Good stuff, looking forward to this :)
> > Put some challenging questions there! I would also like to push again for some through towards a single-trait design. If you can make that happen, I think the solution would be a lot nicer :)
> 
> A single trait design would result in something extremely similar to serde's `Serializer` trait. Such trait would contain several function, one for each basic type. I am not convinced this is a step forward though.
> 
> Note that the `FieldDigest` trait is only local to the `digest` trait and we could just implement it for basic types in this crate so that the `cnd` crate does not have to be aware of it.
> 
> We could also expand the derive macro to support more types of structures so that `cnd` would only need to apply the macro.

As mentioned further up, I am not sure if supporting several primitives types is such a good idea. Happy to support `Vec<u8>` and `[u8]` but anything else, I am not sure ..

Going in that direction, the trait would not be local. How would your current design work with nested structs? Can that even be supported?

D4nte:
> Going in that direction, the trait would not be local. How would your current design work with nested structs? Can that even be supported?

Looks like I need to add a nested struct and an enum example as the next step.

digest/src/lib.rs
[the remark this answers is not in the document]
D4nte: Proposal implemented.

digest-macro-derive/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+            });
+
+            let gen = quote! {
+            impl ::digest::RootDigest for #name
+                 where #(#types: ::digest::FieldDigest),*
+                 {
+                    fn root_digest(self) -> Multihash {
+                        use ::digest::hex;
+                        let mut digests = vec![];
+                        #(digests.push(self.#idents.field_digest(hex::decode(#bytes_str).unwrap())););*
```
thomaseizinger: Can we avoid this runtime error here?

D4nte: Happy to work on it but it's tricky as per commit message 1b5a3449de8d7ed4f52a83addf2386b64c479018

D4nte: Yes! Took some times but got it. Thanks @tcharding for being my rubber duck.

D4nte:
> At this stage (with attributes), you might want to take another look at "darling", it makes these kind of things easier to implement!

I feel like I already invested a lot of time to understand `syn`, `quote` and `proc_macro`. Not really keen to rewrite everything again for the sake of adding a dependency.

D4nte:
> LGTM. nit: `digest()` function would be nicer at the bottom of the file (we are not writing C anymore :)

Are you talking about
```
pub fn digest(bytes: &[u8]) -> Multihash {
    multihash::Sha3_256::digest(bytes)
}
```
?

If so, I disagree, this function basically tells you "This is the hash function we are using". It's like a `const` and if it could be a const, it would, but I can't make it so, so it shan't :)

PR #2229, opened 2020-03-11
Make config validation function generic and condense error handling

rishflab:
The Ethereum and Bitcoin node config validation functions were
replaced with a generic one. A trait was added to provide a
cleaner API to the caller and reduce the amount of imports
required to call the the validation function in main. Error
handling was condensed and cnd will crash if a unspecified error
occurs when trying to contact the blockchain node to validate
the config.

### DIFF
--- a/cnd/src/btsieve/bitcoin/bitcoind_connector.rs
+++ b/cnd/src/btsieve/bitcoin/bitcoind_connector.rs
@@ -5,6 +5,7 @@ use async_trait::async_trait;
 use bitcoin::{BlockHash, Network};
 use reqwest::{Client, Url};
 use serde::Deserialize;
+use crate::config::validation::FetchNetworkId;
 
 #[derive(Copy, Clone, Debug, Deserialize)]
 pub struct ChainInfo {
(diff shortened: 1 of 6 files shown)


cnd/src/btsieve/bitcoin/bitcoind_connector.rs
```diff
@@ -76,20 +77,23 @@ impl BlockByHash for BitcoindConnector {
...
+        let client = self.client.clone();
+        let chaininfo_url = self.chaininfo_url.clone();
 
-    let chain_info = client
-        .get(chaininfo_url)
-        .send()
-        .await?
-        .json::<ChainInfo>()
-        .await?;
+        let chain_info: ChainInfo = client
```
D4nte: Is it really needed, I believe that `.json::<ChainInfo>()` already hints the compiler.

rishflab: The compiler figures it out fine but the IDE plugin struggles. Thought I would leave it in for developer ergonomics.

PR #2236, opened 2020-03-12
Small follow-up PR for the validation of the blockchain connectors

thomaseizinger:
This PR contains two patches with separate ideas for the config validation.

Not sure if the combinators are more readable than what was there before. To me, they seem to be superior because we don't have to manually `match` on the result but can just `?` as usual.

This works because `downcast` returns an error if the downcast fails, which in our case means the error was not a connection error but something else.

### DIFF
--- a/cnd/src/main.rs
+++ b/cnd/src/main.rs
@@ -74,16 +74,16 @@ fn main() -> anyhow::Result<()> {
         ))
     };
 
-    match runtime.block_on(validate_blockchain_config(&bitcoin_connector.connector, settings.bitcoin.network)) {
-        Ok(_) => (),
-        Err(e) => {
-            if e.is::<reqwest::Error>() {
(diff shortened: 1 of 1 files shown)


cnd/src/main.rs
```diff
@@ -74,16 +74,16 @@ fn main() -> anyhow::Result<()> {
...
-        }
-    };
+    runtime.block_on(async {
+        validate_blockchain_config(&bitcoin_connector.connector, settings.bitcoin.network)
+            .await
+            .or_else::<anyhow::Error, _>(|e| {
+                let conn_error = e.downcast::<reqwest::Error>()?;
+                tracing::warn!("Could not validate Bitcoin node config: {}", conn_error);
+
+                Ok(())
```
rishflab: will this terminate the program?

PR #2243, opened 2020-03-13
Do not keep track of Cargo.toml files to avoid missing new ones

D4nte:
Because I already almost did the mistake with PR #2226.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@ INSTALLED_COMPONENTS = $(shell $(RUSTUP) component list --installed --toolchain 
 INSTALLED_NIGHTLY_COMPONENTS = $(shell $(RUSTUP) component list --installed --toolchain $(NIGHTLY_TOOLCHAIN))
 AVAILABLE_CARGO_COMMANDS = $(shell $(CARGO) --list)
 
+CARGO_TOML_FILES = $(shell ls **/Cargo.toml)
+
 # All our targets go into .PHONY because none of them actually create files
 .PHONY: init_git_hooks default install_rust install_rust_nightly install_clippy install_rustfmt install_tomlfmt install clean all ci build clippy test doc e2e check_format format check_rust_format check_toml_format check_ts_format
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -90,9 +92,7 @@ MODIFIED_TYPESCRIPT_FILES = $(filter %.ts %.json %.yml,$(MODIFIED_FILES))
 
 format: install_rustfmt install_tomlfmt
 	$(CARGO_NIGHTLY) fmt -- --files-with-diff | xargs -I{} git add {}
-	RUST_LOG=error $(CARGO) tomlfmt -p Cargo.toml && git add Cargo.toml
-	RUST_LOG=error $(CARGO) tomlfmt -p cnd/Cargo.toml && git add cnd/Cargo.toml
-	RUST_LOG=error $(CARGO) tomlfmt -p libp2p-comit/Cargo.toml && git add libp2p-comit/Cargo.toml
+	@$(foreach file,$(CARGO_TOML_FILES),$(CARGO) tomlfmt -p $(file) && git add $(file);)
```
thomaseizinger: :mega: I think you dropped the `RUST_LOG` now right? Can you add that back please?

D4nte: Can you explain the utility of the `RUST_LOG` here?

thomaseizinger: To silence the output of tomlfmt.

D4nte: Have you tested locally? I achieve the wanted result with the `@` at the start of the line.

thomaseizinger: I did not!

I trust you that you figured out this Makefile voodoo :)

Makefile
```diff
@@ -15,6 +15,8 @@ INSTALLED_COMPONENTS = $(shell $(RUSTUP) component list --installed --toolchain
 INSTALLED_NIGHTLY_COMPONENTS = $(shell $(RUSTUP) component list --installed --toolchain $(NIGHTLY_TOOLCHAIN))
 AVAILABLE_CARGO_COMMANDS = $(shell $(CARGO) --list)
 
+CARGO_TOML_FILES = $(shell ls **/Cargo.toml)
```
thomaseizinger: Actually, I think you shouldn't even need this `ls`. The Makefile should expand the glob directly no?

PR #2255, opened 2020-03-15
Implement remaining oneshot communication protocols

tcharding:
Draft because builds on top of [2111-oneshot](https://github.com/comit-network/comit-rs/pull/2215).  A bunch of new changes have been done to 2111-oneshot.  Will not integrate them until it merges.

Adds the other one shot communication protocols by cut'n'pasta the code from the secret hash protocol (added in PR linked above).

This PR introduces a lot of duplicate code.

Resolves: #2111


tcharding:
> There are some changes missing in these now that were applied to the other branch.
> 
> Please apply those here as-well.
> 
> After that, we can merge this if you want but I think it would be better if we would make a generic NetworkBehaviour. It should be pretty easy to extract the message type as a type parameter.

Will have a go at this.  Cheers.

> My comment in Slack was in regards to using proc-macros. Proc-macros are the wrong tool for this job, they are too powerful.
> You can solve this duplication with regular Rust generics.
> 
> The tests also have quite some duplication. It is not that important but would be nice to get rid of that as-well.

Will take a look at this also.

> Also, I would name the modules for the identity protocols bitcoin_identity, etc.

Can do.

tcharding:
Closing this in order to implement the remaining oneshots on top of https://github.com/comit-network/comit-rs/pull/2260

PR #2265, opened 2020-03-16
Remove non-functional mergify rule

thomaseizinger:
Dependabot doesn't allow to recreate PRs from users that don't have write-access. Mergify doesn't have this so this always fails anyway. In addition, I checked with dependabot and it actually has a feature to automatically rebase on conflicts, which I now enabled.


bonomat:
Had to trigger it manually as mergify would not comment due to config changes./

PR #2266, opened 2020-03-16
Split swap state into multiple stores

### LINKED ISSUE #2195: Split up rfc003 implementation of StateStore into alpha and beta
## Problem
The core of `cnd` uses the `StateStore` trait to push updates on the state of the two ledgers involved in a swap, so that another layer (in our case `cnd`'s HTTP API) can access it on demand. Since the `StateStore` was only ever implemented for the soon-to-be-gone `rfc003` protocol, it is tightly coupled to its requirements, to the point where it is actually defined inside the `rfc003` module.

Assuming that we still want to use this `StateStore` abstraction with `HAN`, `HErc20` and `HALight`, the way it's defined and implemented needs to change.

## Goal
To update the way the core provides updates on the state of a swap, accounting for single-ledger protocols.

## Recommendation
Some work has been done in this direction with #2200 already.

Instead of introducing the new protocols straight away, I would suggest to split up `rfc003`'s implementation of `StateStore` into alpha and beta ledger. Introducing the new protocols is **out of the scope** of this ticket and will be done with #1953.

This will involve splitting up `SwapEvents` and `ActorState` accordingly; pushing updates to two different state stores from `create_swap.rs`; combining these states at a higher level in order to derive actions from them.

Don't get scared by the type parameters 🤡

luckysori:
### Straight from the commit description

COMIT's core provides updates on the state of the ledgers during a swap. These updates used to be applied directly on a struct which held all the information pertaining to the swap: offer, communication state, ledger state and error state. Storing everything together is convenient but creates dependencies which cause pain.

The code can be simplified by splitting up this state into multiple stores. Primarily, a generic ledger store (instantiated twice, for alpha and beta) can be used to fulfil the core of COMIT's principal
task.

The communication state also needs to be saved and updated given the current communication protocol. Since this protocol is going away soon, it is not that appealing to invest a lot of effort into
improving the design, so it is simply stored in a separate `HashMap`. This is not ideal because once the swap is accepted the information can also be obtained from the database layer, meaning that we are storing redundant information. Since it is easier to add another `HashMap` than adding APIs to the database to store and load proposed swaps, and this communication protocol is on the way out, incurring in this technical debt is justified.

In terms of the error state, it turns out that we were never setting the state of the swap to failed if watching a particular ledger had resulted in an error. Since this was the apparent intention when
introducing these APIs, the functionality has been kept. It has also been implemented using a `HashMap`, but a better design may come in the future.

### Some more information

I made an effort to change as little as possible outside the core, which is why I kept `alice::State` and `bob::State` around. I discussed with @da-kami and @thomaseizinger about how to better model the entire state of the swap, but I decided against tackling that here as it would just make the PR more complex to review. 

Also, maybe controversially, I decided to model the incorrect funding of a HTLC at the swap event level. I think it makes things simpler, since we already have the expected asset amount in the HTLC parameters needed to construct the event we are watching for. I could be wrong though.

---

Fixes #2195


cnd/src/db/with_swap_types.rs
```diff
@@ -9,12 +9,12 @@ macro_rules! _match_role {
...
             Role::Alice => {
                 #[allow(dead_code)]
-                type ROLE = alice::State<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT>;
+                type ROLESTATE = alice::State<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT>;
                 $fn
             }
             Role::Bob => {
                 #[allow(dead_code)]
-                type ROLE = bob::State<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT>;
+                type ROLESTATE = bob::State<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT>;
```
tcharding:
```suggestion
                type ROLE_STATE = bob::State<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT>;
```

cnd/src/http_api/swap_resource.rs
```diff
@@ -91,21 +89,42 @@ pub fn build_rfc003_siren_entity(
     let id = swap.swap_id;
 
     with_swap_types!(types, {
-        let state: ROLE = state_store
-            .get(&id)?
-            .ok_or_else(|| anyhow!("state store did not contain an entry for {}", id))?;
+        let swap_has_failed = dependencies.swap_error_state.get(&id);
```
tcharding: I have some re-naming ideas for you.  I've a patch, will push to another branch and link you to it in the comments on 'Finish your review'.  You can take a look and see what you think.

cnd/src/swap_protocols/ledger_states.rs
```diff
@@ -0,0 +1,126 @@
+use crate::swap_protocols::{
+    rfc003::{create_swap::SwapEvent, LedgerState},
+    state::{Get, Insert, Update},
+    swap_id::SwapId,
+};
+use std::{
+    any::Any,
+    collections::HashMap,
+    ops::{Deref, DerefMut},
+    sync::Mutex,
```
tcharding: This is the wrong `Mutex`.  We need to use an async mutex, in other places we use `tokio::sync::Mutex` (see `db/mod.rs`).  Ref: https://rust-lang.github.io/async-book/03_async_await/01_chapter.html (search for 'Mutex').

cnd/src/swap_protocols/ledger_states.rs
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+}
+
+impl<A, H, T, L> Update<LedgerState<A, H, T>, SwapEvent<A, H, T>> for L
+where
+    L: DerefMut<Target = Mutex<HashMap<SwapId, Box<dyn Any + Send>>>> + Send + Sync + 'static,
+    LedgerState<A, H, T>: 'static,
+{
+    fn update(&self, key: &SwapId, event: SwapEvent<A, H, T>) {
```
tcharding: Can we have a comment on this function please, the return value is not immediately obvious.

cnd/src/http_api/swap_resource.rs
luckysori: Naming improvements are very welcome!

cnd/src/http_api/swap_resource.rs
```diff
@@ -82,7 +80,7 @@ pub enum OnFail {
 // `with_swap_types!` macro and can be iteratively improved
 #[allow(clippy::cognitive_complexity)]
 pub fn build_rfc003_siren_entity(
-    state_store: &InMemoryStateStore,
+    dependencies: Facade,
```
thomaseizinger: Would a reference to the `Facade` be sufficient here?

luckysori:
> I have pushed three follow up patches to `2195-tobins-suggestions` for you to take a look at, feel free to use them or not as you see fit. Good work.

Thank you for your review and the follow-up patches! I left some comments on the commits, but will likely take them all :smile:

cnd/src/swap_protocols/ledger_states.rs
luckysori: Sorry, auto-import!

cnd/src/swap_protocols/facade.rs
```diff
@@ -51,43 +51,62 @@ use std::{convert::TryInto, fmt::Debug, sync::Arc};
...
     pub seed: RootSeed,
     pub swarm: Swarm,
     pub db: Sqlite,
 }
 
-impl<S> Insert<S> for Facade
+impl<AL, BL, AA, BA, AI, BI> state::Insert<SwapCommunication<AL, BL, AA, BA, AI, BI>> for Facade
 where
-    S: Send + 'static,
+    AL: Clone + Send + 'static,
```
thomaseizinger: :mega: I think these should not be necessary (esp. `Clone`) and if, then it should be enough to put the bounds on `SwapCommunication<AL, BL, AA, BA, AI, BI>`.

cnd/src/swap_protocols/ledger_states.rs
thomaseizinger: There is no return value.

cnd/src/swap_protocols/ledger_states.rs
```diff
@@ -0,0 +1,126 @@
...
+    sync::Mutex,
+};
+
+#[derive(Default, Debug)]
+pub struct AlphaLedgerState {
+    states: Mutex<HashMap<SwapId, Box<dyn Any + Send>>>,
+}
+
+#[derive(Default, Debug)]
+pub struct BetaLedgerState {
```
thomaseizinger: :mega: The two structs are exactly the same ?! We can we not just have two instances of the same struct?

cnd/src/swap_protocols/rfc003/bob/mod.rs
```diff
@@ -15,10 +18,23 @@ pub struct State<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT> {
...
+    ) -> Self {
+        Self {
+            swap_communication,
+            alpha_ledger_state,
+            beta_ledger_state,
+            secret_source: Arc::new(secret_source),
+        }
+    }
+
     pub fn proposed(
```
thomaseizinger: Do we still need those functions?

cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -20,136 +20,45 @@ use genawaiter::{
...
 /// account situations where it is clear that no more events will happen even
 /// though in theory, there could. For example:
-/// - alpha funded
-/// - alpha refunded
+/// - funded
+/// - refunded
 ///
 /// It is highly unlikely for Bob to fund the HTLC now, yet the current
 /// implementation is still waiting for that.
-pub async fn create_alpha_watcher<D, A, AI, BI>(
```
thomaseizinger: You are a legend Lucas :D

cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -20,136 +20,45 @@ use genawaiter::{
...
+pub async fn create_watcher<D, S, L, A, H, I, T>(
```
thomaseizinger: We should leave some easter eggs in the codebase where we write funny things with these type parameters xD

cnd/src/swap_protocols/rfc003/create_swap.rs
```diff
@@ -20,136 +20,45 @@ use genawaiter::{
...
     dependencies: D,
-    accepted: AcceptedSwap<A::AL, A::BL, A::AA, A::BA, AI, BI>,
+    ledger_state: Arc<S>,
```
thomaseizinger: So I guess the problem why you are passing this separately is because you wouldn't know which instance of `LedgerState` to dispatch to if you just implemented it on the `Facade`? Would it make sense to assume that every caller of this function knows about the `Facade` itself and not just about `D`?

Differently asked, is every caller of this function something a component of `cnd` or do we need to ever call this from within the `COMIT` core?

If it is the former (which is what I expect) then we can get rid of the `From<>` trait bounds with `Arc`s and just reach into the facade and get the correct field out: `facade.alpha_ledger_states`.

I can send a follow-up PR about this :)

cnd/src/swap_protocols/rfc003/ethereum/htlc_events.rs
```diff
@@ -41,14 +42,26 @@ impl
...
     ) -> anyhow::Result<Funded<asset::Ether, transaction::Ethereum>> {
-        Ok(Funded {
-            asset: Ether::from_wei(deploy_transaction.transaction.value),
-            transaction: deploy_transaction.transaction.clone(),
-        })
+        let expected_asset = &htlc_params.asset;
+
+        let asset = Ether::from_wei(deploy_transaction.transaction.value);
+
+        let event = match expected_asset.cmp(&asset) {
```
thomaseizinger: At first I liked this because we no longer have to specify the `Ord` bound but it is annoying that we have to do this for every asset now :(

Can you please record this as an issue so we don't forget to deal with this. I feel this should really be done in a more generic place.

cnd/src/swap_protocols/swap_communication_state.rs
```diff
@@ -0,0 +1,36 @@
+use crate::swap_protocols::{
+    state::{Get, Insert},
+    swap_id::SwapId,
+};
+use std::{any::Any, clone::Clone, collections::HashMap, sync::Mutex};
+
+#[derive(Default, Debug)]
+pub struct SwapCommunicationState {
```
thomaseizinger: WDYT of naming these in plural because they contain all of them?

```suggestion
pub struct SwapCommunicationStates {
```

cnd/src/swap_protocols/facade.rs
```diff
@@ -51,43 +51,62 @@ use std::{convert::TryInto, fmt::Debug, sync::Arc};
...
+    }
+}
+
+impl From<Facade> for Arc<BetaLedgerState> {
+    fn from(facade: Facade) -> Self {
+        facade.beta_ledger_state
+    }
+}
+
+impl InsertSwapError for Facade {
```
thomaseizinger: In a perfect world, the facade delegates through to the same interface, i.e. it calls `insert_swap_error` on the inner element.

cnd/src/swap_protocols/rfc003/create_swap.rs
da-kami: `<S, H, I, T, L, A, D>`?

thomaseizinger: How long did you work on this @da-kami ?

cnd/src/swap_protocols/ledger_states.rs
thomaseizinger: You will need to make these `async_trait`s with the tokio mutex.

cnd/src/swap_protocols/facade.rs
luckysori: Good catch, `Clone` is not needed for `Insert`.

luckysori: `Send` and `'static` are needed, but can be applied to the combined type.

cnd/src/swap_protocols/rfc003/create_swap.rs
luckysori: I'll leave this one to you. I wasn't sure about making this function aware of the `Facade`, but it seems fine from your argument.

cnd/src/swap_protocols/swap_communication_state.rs
luckysori: Makes more sense. All these new names are a bit odd still.

cnd/src/swap_protocols/rfc003/bob/mod.rs
luckysori: Nice catch! `request` is still in use but the others can go.

cnd/src/swap_protocols/swap_communication_state.rs
thomaseizinger: Pretty sure we will figure something out as we iterate :)

cnd/src/swap_protocols/facade.rs
luckysori: True, I didn't feel great about this. Tobin has dealt with this in one of his follow-up patches.

cnd/src/swap_protocols/rfc003/create_swap.rs
da-kami: 30sec

cnd/src/swap_protocols/rfc003/create_swap.rs
luckysori: Never mind, I'm tackling this as part of handling one of your other comments: https://github.com/comit-network/comit-rs/pull/2266#discussion_r393368253.

cnd/src/swap_protocols/ledger_states.rs
da-kami: I thought it was on purpose, as it is part of this ticket to achieve the split between alpha and beta. Given that there might be specific parameters for each ledger we might need different structs. I see how it could be one here, but would it make sense to keep it for later changes on top of this?
Additionally I wonder how this will go together with introducing new protocols.

luckysori: I believe it was a mistake on my part. If we continue using this model, there won't be any difference between how we store alpha and beta ledger (protocol) state in memory: it will always be `Any`. It's the implementation of `Update` that will vary depending on the protocol.

luckysori:
@thomaseizinger: all done I think! Asking for re-review and keeping the `no-mergify` label so that I can squash everything if I get your approval.

cnd/src/init_swap.rs
```diff
@@ -51,51 +39,39 @@ where
...
-            tokio::task::spawn(create_alpha_watcher::<
-                D,
-                bob::State<AL, BL, AA, BA, AH, BH, AI, BI, AT, BT>,
-                AI,
-                BI,
-            >(dependencies.clone(), accepted.clone()));
+    tokio::task::spawn(
+        create_watcher::<_, _, _, _, AH, _, AT>(
+            dependencies.clone(),
+            dependencies.alpha_ledger_state.clone(),
```
thomaseizinger: Nice! Yeah that is how I imagined it! Good work :)

Issue #2170, opened 2020-03-05
Add skeleton swarm API for announcing swaps


tcharding:
## Problem

We are implementing a new communication API.  In order for this to be used by the rest of cnd we need an API to use.

## Goal

Have an `unimplemented!` function that provides the API we want the libp2p layer to expose to the rest of cnd.

Scope of this ticket is just the API function definitions, you can leave the bodies `unimplemented!()`.  This includes defining the data structure that contains the minimum data required by the swarm in order to complete the required communication protocols.  Some consideration will also be needed for data required to call the spawn functionality once communication with the peer node has completed.

## Recommendation

Please read https://github.com/comit-network/comit-rs/issues/2171 to see what functionality is required once this API is fully implemented.  Designing the data required by the API will require this understanding.  Although this ticket is split into skeleton/implementation this is only done to facilitate working in parallel on other parts of the codebase. 

Inspire yourself from the data structure used by the controller in: https://github.com/comit-network/comit-rs/issues/2158.  This is a soft dependency.

PR #2268, opened 2020-03-16
Add skeleton API function for controller/swarm interface

tcharding:
In order to be able to utilise the new communication protocol we need an interface between the controller and the swarm.

Add an API function that allows the controller (acting as Alice) to call the swarm (libp2p layer) and trigger execution of the new communication protocols (by calling through to the underlying swarm) in order to announce the created swap, send the known parameters, and receive the remaining required swap parameters back from Bob.

Fixes: #2170

### DIFF
--- a/cnd/src/network/mod.rs
+++ b/cnd/src/network/mod.rs
@@ -136,6 +136,30 @@ impl Swarm {
             local_peer_id,
         })
     }
+
+    /// This is the API for Alice to call in order to execute the appropriate
+    /// communication protocols to announce a swap to Bob (i.e., to send the
+    /// known swap parameters and receive the remaining swap parameters) in
+    /// order to finalize this swap.
+    ///
+    /// `id` is an identifier use to access the database to get the known swap
+    /// parameters.
+    // Identifier type will be defined in ticket: https://github.com/comit-network/comit-rs/issues/2173
+    pub async fn finalize_announce_swap(&mut self, _id: String) -> anyhow::Result<()> {
+        // 1. Load parameters for the swap from the database and call down to
+        // the swarm to execute the required communication protocols.
+        //
+        // 2. Write the remaining parameters to the database (keyed by swap_id).
+        //
+        // 3. Spawn the swap using `swap_id` as an argument.  `spawn()` can then
+        // load the swap from the database and spawn it.
+
+        unimplemented!()
+    }
+
+    // On Bob's side, when an announce message is received execute the required
+    // communication protocols and write the finalized swap to the database.  Then
+    // spawn the same as is done for Alice.
 }
 
 struct SwarmWorker {


cnd/src/network/mod.rs
```diff
@@ -128,6 +128,36 @@ impl Swarm {
...
+    /// Takes all the data we know about a 'created' swap.  Executes the
+    /// appropriate communication protocols in order to send data to the peer
+    /// and receive data from the peer (remaining swap parameters) in order to
+    /// finalize this swap.
+    pub fn finalize_created_swap<A, B>(
+        &mut self,
+        swap_id: SwapId,
+        swap_digest: SwapDigest,
+        secret_hash: SecretHash,
+        alpha_identity: A,
```
thomaseizinger: What are we going to need these for?

cnd/src/network/mod.rs
```diff
@@ -128,6 +128,36 @@ impl Swarm {
...
+        let mut swarm = self.swarm.lock().await;
+        // Assume new NetworkBehaviour is called comit still i.e., the
+        // NetworkBehaviour that implements the new communication protocols.
+
+        // Assumptions:
+        // - This works for Han-Han, Han-HALight, HALight-Han
+
+        // Questions:
+        // - Is this all the data we need for a Han-Herc20 swap?
+        // - How are we going to handle swap protocol pairs, Foo-Bar vs Bar-Foo.
```
thomaseizinger: You will either need to pass enough information into this function so you know this or make more functions per combination.

cnd/src/network/mod.rs
```diff
@@ -128,6 +128,36 @@ impl Swarm {
...
+        //   For symmetrical swaps it does not matter but for asymmetrical swaps
+        //   it does i.e., what we send and what we expect to receive is
```
thomaseizinger: This logic will either have to be in there or in the NetworkBehaviour. I'd say preferably the NetworkBehaviour as this should only be glue code and not contain crucial logic like this.

cnd/src/network/mod.rs
```diff
@@ -128,6 +128,36 @@ impl Swarm {
...
+    ) -> anyhow::Result<()> {
+        let mut swarm = self.swarm.lock().await;
+        // Assume new NetworkBehaviour is called comit still i.e., the
+        // NetworkBehaviour that implements the new communication protocols.
+
+        // Assumptions:
+        // - This works for Han-Han, Han-HALight, HALight-Han
+
+        // Questions:
+        // - Is this all the data we need for a Han-Herc20 swap?
```
thomaseizinger: To answer this, I would recommend a bottom-up thinking approach: Look at the protocols we have, think through all the use-cases that we want to achieve and write down what the data is that we are going to need to implement those.

cnd/src/network/mod.rs
tcharding: This is the identities sent using the identity_protocl i.e., bitcoin_identity, ethereum_identity, lightning_identity.

cnd/src/network/mod.rs
tcharding: I think this is resolved with the comments in the new version.  Resolving.

cnd/src/network/mod.rs
tcharding: Good tip, thanks.  Please see new version (force pushed, it is a total rewrite).

cnd/src/network/mod.rs
```diff
@@ -128,6 +131,68 @@ impl Swarm {
...
+///
+/// For Han-HALight
+/// - alpha_identity: refund identity for Alice (i.e., Ethereum address)
+/// - beta_identity: lightning identity for Alice
+///
+/// For HALight-Han
+/// - alpha_identity: lightning identity for Alice
+/// - beta_identity: redeem identity for Alice (i.e., Ethereum address)
+#[derive(Debug)]
+pub struct CreatedSwapParamaters<A, B> {
```
thomaseizinger: I've got some design-challenging questions for you:

- How will Bob be able to call this API? He doesn't have knowledge of the secret hash before the communication protocol is finished. Thinking this further, we are we forcing the caller to generate a secret hash? We can we not do this inside?
- The SwapId is only established as part of the communication protocol, which SwapId is this referring to?
- The identities are tricky. We need to generate a transient one for Bitcoin due to the nature of our HTLC and the complicated redeem/refund process. For Ethereum, we can pass the one through that the user gave us on the API. Can/should we take some of the caller's work (generating a transient Bitcoin identity) away by doing this inside? What does the API look like if we don't explicitly pass identities in?

thomaseizinger: > How will Bob be able to call this API?

If Bob should not call this API, then it the one he should be calling is missing in this PR.

tcharding: > I've got some design-challenging questions for you:

Thought provoking review, thanks.
 
> * How will Bob be able to call this API? He doesn't have knowledge of the secret hash before the communication protocol is finished. 

I totally did not think about the Bob side of things, forgetting that Bob would need to call _back_ into the swarm after receiving the `swap_digest`.  Please see new function and improved stucts.

> Thinking this further, we are we forcing the caller to generate a secret hash? We can we not do this inside?

The secret hash will need to be accessible from outside the swarm, will it not?  So I think it is better to pass it in than pass it out.  

> * The SwapId is only established as part of the communication protocol, which SwapId is this referring to?

Touche

> * The identities are tricky. We need to generate a transient one for Bitcoin due to the nature of our HTLC and the complicated redeem/refund process. For Ethereum, we can pass the one through that the user gave us on the API. Can/should we take some of the caller's work (generating a transient Bitcoin identity) away by doing this inside? What does the API look like if we don't explicitly pass identities in?

As with the SwapId I feel that the implementation details (generating the transient Bitcoin identity) should be done outside of the network module to keep that module cleaner.

Thanks for your effort @thomaseizinger , I did not think this through fully and hence loaded your brain up more than was necessary.

cnd/src/network/mod.rs
```diff
@@ -136,6 +138,81 @@ impl Swarm {
...
+        // swarm.behaviour.finalize_announced_swap(_announced).await?
+
+        unimplemented!()
+    }
+}
+
+/// Holds the data that needs to be sent by Alice to Bob to announce a swap
+/// and take it from 'created' to 'finalized'.
+///
+/// For Han-Han, Han-Herc20, Herc20-Han
```
D4nte: I believe it does not matter at this stage.

cnd/src/network/mod.rs
```diff
@@ -136,6 +138,81 @@ impl Swarm {
...
+/// - alpha_identity: refund identity for Alice
+/// - beta_identity: redeem identity for Alice
+///
+/// For Han-HALight
+/// - alpha_identity: refund identity for Alice (i.e., Ethereum address)
```
D4nte: Hum, what about cltv expiry? Do we know whether it's part of the `start` or `finalized` swap?

thomaseizinger: I would assume it is part of the start because it is price-relevant, same as the expiries for existing protocols.

cnd/src/network/mod.rs
thomaseizinger: > I totally did not think about the Bob side of things, forgetting that Bob would need to call back into the swarm after receiving the swap_digest. Please see new function and improved stucts.

Why does he need to call _back_ into it? Have you considered that the the controller may not (really) care whether they are Alice or Bob and just call the same function on the swarm and let it figure out what to do next, i.e. for Alice sending an announce and for Bob waiting for an announce?
The role information would obviously need to be communicated somehow but that can also be a parameter instead of two different function calls.

> The secret hash will need to be accessible from outside the swarm, will it not? So I think it is better to pass it in than pass it out.

So will all the other parameters that are received from the other side, no?
Why should Alice pass it in and Bob only receive it as an output? In the controller, I would like to avoid this kind of conditional logic and just call the Swarm with the minimally necessary data to start the communication protocol, regardless of whether it is Alice or Bob.

> As with the SwapId I feel that the implementation details (generating the transient Bitcoin identity) should be done outside of the network module to keep that module cleaner.

Let's pass them in for now and see what it evolves too. I feel like in the long-term, it would be better to have this module handle that logic though. It looks cleaner to me to output all the protocol-specific information in the end instead of doing parts upfront and parts in the end.

thomaseizinger: > Thanks for your effort @thomaseizinger , I did not think this through fully and hence loaded your brain up more than was necessary.

No worries man :)

cnd/src/network/mod.rs
```diff
@@ -136,6 +138,81 @@ impl Swarm {
...
+        _announce: AnnounceSwapParameters<A, B>,
+    ) -> anyhow::Result<AnnouncedSwapParameters<A, B>> {
+        // let mut _swarm = self.swarm.lock().await;
+        // swarm.behaviour.finalize_announce_swap(_digest, _announce).await?
+
+        unimplemented!()
+    }
+
+    /// This is the API for Bob to call in order to execute the appropriate
+    /// communication protocols when an announced swap is received from Alice
```
thomaseizinger: I would prefer if the whole communication-cycle happens within the Swarm and does not require two calls.

cnd/src/network/mod.rs
```diff
@@ -136,6 +138,81 @@ impl Swarm {
             local_peer_id,
         })
     }
+
+    /// This is the API for Alice to call in order to execute the appropriate
+    /// communication protocols to announce a swap to Bob (i.e., to send the
+    /// known swap parameters and receive the remaining swap parameters) in
+    /// order to finalize this swap.
+    ///
+    /// This works for all supported combinations on Han, HALight, and Herc20.
```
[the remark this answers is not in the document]
tcharding: oh, it's `HErc20` - my bad.

cnd/src/network/mod.rs
tcharding: To clarify @thomaseizinger comment - it is passed between the nodes _before_ the communication protocols start i.e., before we even announce a swap.

cnd/src/network/mod.rs
tcharding: @rishflab and I were just discussing this now.  We could not work it out.
 
Assumption: conversion from `swap_digest` to `swap_id` during the announce protocol requires some logic i.e., to confirm or not to confirm the announced swap requires some logic.  

Question: What does this logic look like?  Does the swarm know everything that is needed by this logic? 

Depending on the answer to the above question, it seems either

1. The swarm **does not** know everything required: Then the inbound event (arrival of the `swap_digest` to Bob's node) has to be propagated up _out_ of the swarm (e.g., by polling the swarm).
2. The swarm **does** know everything and this logic can all be handled within the swarm.  In this case there are also two possibilities, ether the logic is handled in the swarm or in the NetworkBehaviour i.e, either in the swarm (as seen by cnd) or within the `ExpandedSwarm` as seen by the network module.

cnd/src/network/mod.rs
```diff
@@ -73,6 +73,8 @@ type ExpandedSwarm = libp2p::swarm::ExpandedSwarm<
     EitherError<libp2p_comit::handler::Error, void::Void>,
 >;
 
+type SwapDigest = Vec<u8>;
```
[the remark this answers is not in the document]
tcharding: Cool, will fix.  Thanks.

cnd/src/network/mod.rs
thomaseizinger: There is no layer above the swarm for incoming events? Or where would you put that logic?

Same as currently, the swarm will need access to the database and other things to spawn the swap later.
We can use the same dependencies to also fetch the necessary data in case of an incoming announce and act accordingly.

cnd/src/network/mod.rs
tcharding: Comments removed, thanks.

cnd/src/network/mod.rs
tcharding: > There is no layer above the swarm for incoming events? Or where would you put that logic?

I have no better place to suggest, so in the swarm it goes :)

> Same as currently, the swarm will need access to the database and other things to spawn the swap later.
> We can use the same dependencies to also fetch the necessary data in case of an incoming announce and act accordingly.

Cool, I think we are on the same page.

tcharding:
Merge me baby.  bors r+

PR #2269, opened 2020-03-16
Swap digest: enum & nested structs

### LINKED ISSUE #2112: Implement the swap digest algorithm

D4nte:
## Interface 

3 traits:
- `IntoDigestInput`: The consumer must implement this trait on their basic data types. It basically allow them to tell the lib how they want such type to be transformed in a byte array
- `Digest`: The consumer must use `MacroDigest` derive macro to implement this trait on their data structures, using the `digest_prefix` attribute to specify the prefix applied to fields before hashing in hex
- `FieldDigest`: This is an internal sealed type, the consumer should not touch it.

## Nested structs digest algorithm

1. Hash each field,
	a. If the field is a basic data type: concatenate prefix + result of `into_digest_input`, hash it,
	b. If the field is a data structure: concatenate prefix + result of `digest` (of the struct),
2. Order hash results,
3. Concatenate ordered hash results,
4. Hash them.

Note, if the field of an enum is a unit variant, then only the matching prefix is hashed.

Resolves #2112.

### DIFF
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -12,39 +12,41 @@ pub trait Digest {
     fn digest(self) -> Multihash;
 }
 
-pub trait FieldDigest {
+pub trait IntoDigestInput {
+    fn into_digest_input(self) -> Vec<u8>;
+}
+
+#[doc(hidden)]
+pub trait FieldDigest: private::Sealed {
     fn field_digest(self, prefix: Vec<u8>) -> Multihash;
+}
+
+impl<T> IntoDigestInput for T
+where
+    T: Digest,
+{
+    fn into_digest_input(self) -> Vec<u8> {
+        let field_digest = self.digest();
+
+        field_digest.into_bytes()
+    }
 }
 
 impl<T> FieldDigest for T
 where
-    T: Digest,
+    T: IntoDigestInput,
 {
     fn field_digest(self, prefix: Vec<u8>) -> Multihash {
         let mut bytes = prefix;
-        let field_digest = self.digest();
-        bytes.append(&mut field_digest.into_bytes());
-
-        digest(&bytes)
-    }
-}
-
-impl FieldDigest for String {
-    fn field_digest(self, prefix: Vec<u8>) -> Multihash {
-        let mut bytes = prefix;
-        // String::into_bytes return the bytes for UTF-8 encoding
-        let mut value = self.into_bytes();
+        let mut value = self.into_digest_input();
         bytes.append(&mut value);
 
         digest(&bytes)
     }
 }
 
-impl FieldDigest for Vec<u8> {
-    fn field_digest(mut self, prefix: Vec<u8>) -> Multihash {
-        let mut bytes = prefix;
-        bytes.append(&mut self);
+mod private {
+    pub trait Sealed {}
 
-        digest(&self)
-    }
+    impl<T> Sealed for T where T: super::IntoDigestInput {}
 }
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -16,6 +16,19 @@ pub trait FieldDigest {
     fn field_digest(self, prefix: Vec<u8>) -> Multihash;
 }
 
+impl<T> FieldDigest for T
+where
+    T: Digest,
+{
+    fn field_digest(self, prefix: Vec<u8>) -> Multihash {
+        let mut bytes = prefix;
+        let field_digest = self.digest();
+        bytes.append(&mut field_digest.into_bytes());
+
+        digest(&bytes)
+    }
+}
+
 impl FieldDigest for String {
     fn field_digest(self, prefix: Vec<u8>) -> Multihash {
         let mut bytes = prefix;
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -1,4 +1,4 @@
-/// This crate brings two traits: `Digest` and `FieldDigest`
+/// This crate brings two traits: `Digest` and `IntoDigestInput`
 ///
 /// `Digest` should be implemented on data structures using the `Digest`
 /// derive macro.
@@ -6,8 +6,8 @@
 /// role of an identifier to ensure that fields with same data but different
 /// meaning do not result to the same digest.
 ///
-/// Elementary data types should implement `IntoDigestInput`, this allows you to
-/// control how the data is transformed to a byte array.
+/// Data types within the data structure should implement `IntoDigestInput`,
+/// this allows you to control how the data is transformed to a byte array.
 ///
 /// ```
 /// use digest::{Digest, IntoDigestInput};
@@ -30,23 +30,15 @@
 /// ```
 ///
 /// The digest algorithm goes as follow:
-/// 1. Compute `field_digest(prefix)` for all fields of the struct,
-/// 2. Lexically order the field digests,
-/// 3. Concatenate the result,
+/// 1. For each field in the struct:
+///     a. Concatenate `digest_prefix` + `self.into_digest_input`,
+///     b. Hash the result.
+/// 2. Lexically order the resulting field digests,
+/// 3. Concatenate the list,
 /// 4. Hash the result.
 ///
-/// The field digest algorithm goes as follow:
-/// For elementary types:
-///     1. Transform the data in a byte array (if there is
-///   any data) using `IntoDigestInput` trait,
-///     2. Concatenate prefix and the resulting byte array,
-///     3. Hash the result.
-/// For data structures:
-///     1. Calculate the root digest of the struct,
-///     2. Concatenate prefix and resulting root digest,
-///     3. Hash the result.
-///
 /// For unit variants of enums, only the prefix as input to the hash function.
+/// Note that Nested structures are not supported.
 pub use digest_macro_derive::Digest;
 pub use hex;
 
@@ -65,37 +57,10 @@ pub trait IntoDigestInput {
     fn into_digest_input(self) -> Vec<u8>;
 }
 
-#[doc(hidden)]
-pub trait FieldDigest: private::Sealed {
-    fn field_digest(self, prefix: Vec<u8>) -> Multihash;
-}
+pub fn field_digest<T: IntoDigestInput>(field: T, prefix: Vec<u8>) -> Multihash {
+    let mut bytes = prefix;
+    let mut value = field.into_digest_input();
+    bytes.append(&mut value);
 
-impl<T> IntoDigestInput for T
-where
-    T: Digest,
-{
-    fn into_digest_input(self) -> Vec<u8> {
-        let field_digest = self.digest();
-
-        field_digest.into_bytes()
-    }
-}
-
-impl<T> FieldDigest for T
-where
-    T: IntoDigestInput,
-{
-    fn field_digest(self, prefix: Vec<u8>) -> Multihash {
-        let mut bytes = prefix;
-        let mut value = self.into_digest_input();
-        bytes.append(&mut value);
-
-        digest(&bytes)
-    }
-}
-
-mod private {
-    pub trait Sealed {}
-
-    impl<T> Sealed for T where T: super::IntoDigestInput {}
+    digest(&bytes)
 }
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -1,4 +1,4 @@
-pub use digest_macro_derive::RootDigestMacro;
+pub use digest_macro_derive::DigestMacro;
 pub use hex;
 
 pub use multihash;
@@ -8,8 +8,8 @@ pub fn digest(bytes: &[u8]) -> Multihash {
     multihash::Sha3_256::digest(bytes)
 }
 
-pub trait RootDigest {
-    fn root_digest(self) -> Multihash;
+pub trait Digest {
+    fn digest(self) -> Multihash;
 }
 
 pub trait FieldDigest {
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -1,3 +1,52 @@
+/// This crate brings two traits: `Digest` and `FieldDigest`
+///
+/// `Digest` should be implemented on data structures using the `Digest`
+/// derive macro.
+/// The attribute `digest_prefix` must be applied on each field. It plays the
+/// role of an identifier to ensure that fields with same data but different
+/// meaning do not result to the same digest.
+///
+/// Elementary data types should implement `IntoDigestInput`, this allows you to
+/// control how the data is transformed to a byte array.
+///
+/// ```
+/// use digest::{Digest, IntoDigestInput};
+///
+/// struct MyType(Vec<u8>);
+///
+/// impl IntoDigestInput for MyType {
+///     fn into_digest_input(self) -> Vec<u8> {
+///         self.0
+///     }
+/// }
+///
+/// #[derive(Digest)]
+/// struct MyStruct {
+///     #[digest_prefix = "00AA"]
+///     foo: MyType,
+///     #[digest_prefix = "1122"]
+///     bar: MyType,
+/// }
+/// ```
+///
+/// The digest algorithm goes as follow:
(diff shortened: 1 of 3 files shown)


D4nte:
> Nice stuff! :)
> 
> Regarding the algorithm:
> 
> Does:
> 
> > Hash all fields, sort hashes, concatenate prefix and sorted hashes, hash.
> 
> mean we only sort once or do we sort + hash several times?

I do not understand your question so I'll reword my description.

1. Hash each field separately using the trait, passing a prefix
2. Sort hashes
3. Hash(prefix + hashes)

> I think we should only sort and hash once. The fact that we are using nested structs should be an implementation detail to someone outside of the system.

So you approve the solution I went with?

> Regarding your first question:
> 
> Maybe we can solve that by making the trait have two functions?
> 
>     * one that allows you to add a hash to the list, given some context (your current fn)
>
>     * one that calculates the root hash and does the sorting + hashing stuff

I am not sure I understand the proposal. Are you talking about some builder trait?

1. First function: cumulate hashes in a list
2. Second function: sort hashes from the list of 1., concatenate prefix and hashes, hash them
 
> 
> I think that would combine the best of both worlds:
> 
>     * The end-consumer doesn't have to pass an empty vec
> 
>     * We only sort and hash once, hence nested structs are an implementation detail

tcharding:
Not sure how this combines with Thomas' comments but in regards to the `prefix` parameter one idea is to have `digest()` accept the vector and have another trait method `digest_no_prefix()` that does not.  Unfortunately we cannot use a default implementation though because this would (according to my compiler) require the implementor to be `Sized` and I do not think we want this.  Here is a patch that does this for you to look at
```
diff --git a/digest/src/lib.rs b/digest/src/lib.rs
index 5931c0fa..1ad0bc7c 100644
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -10,6 +10,10 @@ pub fn digest(bytes: &[u8]) -> Multihash {

 pub trait Digest {
     fn digest(self, prefix: Vec<u8>) -> Multihash;
+
+    /// We cannot have a default implementation here because it would
+    /// need to be Sized (String and Vec[u8] are not Sized).
+    fn digest_no_prefix(self) -> Multihash;
 }

 impl Digest for String {
@@ -21,6 +25,10 @@ impl Digest for String {

         digest(&bytes)
     }
+
+    fn digest_no_prefix(self) -> Multihash {
+        self.digest(vec![])
+    }
 }

 impl Digest for Vec<u8> {
@@ -30,4 +38,8 @@ impl Digest for Vec<u8> {

         digest(&self)
     }
+
+    fn digest_no_prefix(self) -> Multihash {
+        self.digest(vec![])
+    }
 }
```

tcharding:
`digest_sans_prefix()` could be ok too.

D4nte:
> Not sure how this combines with Thomas' comments but in regards to the `prefix` parameter one idea is to have `digest()` accept the vector and have another trait method `digest_no_prefix()` that does not. Unfortunately we cannot use a default implementation though because this would (according to my compiler) require the implementor to be `Sized` and I do not think we want this. Here is a patch that does this for you to look at
> 
> ```
> diff --git a/digest/src/lib.rs b/digest/src/lib.rs
> index 5931c0fa..1ad0bc7c 100644
> --- a/digest/src/lib.rs
> +++ b/digest/src/lib.rs
> @@ -10,6 +10,10 @@ pub fn digest(bytes: &[u8]) -> Multihash {
> 
>  pub trait Digest {
>      fn digest(self, prefix: Vec<u8>) -> Multihash;
> +
> +    /// We cannot have a default implementation here because it would
> +    /// need to be Sized (String and Vec[u8] are not Sized).
> +    fn digest_no_prefix(self) -> Multihash;
>  }
> 
>  impl Digest for String {
> @@ -21,6 +25,10 @@ impl Digest for String {
> 
>          digest(&bytes)
>      }
> +
> +    fn digest_no_prefix(self) -> Multihash {
> +        self.digest(vec![])
> +    }
>  }
> 
>  impl Digest for Vec<u8> {
> @@ -30,4 +38,8 @@ impl Digest for Vec<u8> {
> 
>          digest(&self)
>      }
> +
> +    fn digest_no_prefix(self) -> Multihash {
> +        self.digest(vec![])
> +    }
>  }
> ```

So basically the option 2 I proposed but with different names?

tcharding:
> So basically the option 2 I proposed but with different names?

On further consideration I think we should do what you propose in option 2 (this is _not_ what I posted in the diff above).

> Sorry I don't understand, are you saying we should or should not have a default > implementation of one of the traits?

I tried to do it with a default implementation and it did not work.

thomaseizinger:
> I do not understand your question so I'll reword my description.
> 1. Hash each field separately using the trait, passing a prefix
> 2. Sort hashes
> 3. Hash(prefix + hashes)

Does that happen for each nested struct? If yes then I think we should **not** do this.
The resulting digest should be the same, regardless of the layers of nesting because that is an implementation detail.

Maybe I've sent you on a goose-chase with the whole nesting idea and it is actually not possible :grimacing: 

Maybe we need two traits for this: `DigestItem` and `DigestRoot` with the following signatures:

```rust
trait DigestItem {
	fn append_digest(&self, context: Vec<u8>, hashes: &mut Vec<MultiHash>);
}

trait DigestRoot {
	fn digest(&self) -> MultiHash;
}
```

Then we could implement `DigestItem` also on structs and they just call `append_digest` for each of their fields.

Although it would also be possible to have those two functions on a single trait.

> I am not sure I understand the proposal. Are you talking about some builder trait?

Not some builder trait but basically what you mentioned in proposal (2). Maybe it is exactly proposal 2 and I just didn't get it :)

My motivation was to allow the actual end-user of the trait to only call a single function without parameters, something like:

```rust
let swap: Swap = todo!();

let digest = swap.digest();
```

`digest` would accumulate all the hashes of each field, sort them and hash them again.

D4nte:
> > I do not understand your question so I'll reword my description.
> > 
> > 1. Hash each field separately using the trait, passing a prefix
> > 2. Sort hashes
> > 3. Hash(prefix + hashes)
> 
> Does that happen for each nested struct? If yes then I think we should **not** do this.
> The resulting digest should be the same, regardless of the layers of nesting because that is an implementation detail.
> 
> Maybe I've sent you on a goose-chase with the whole nesting idea and it is actually not possible 😬
> 
> Maybe we need two traits for this: `DigestItem` and `DigestRoot` with the following signatures:
> 
> ```rust
> trait DigestItem {
> 	fn append_digest(&self, context: Vec<u8>, hashes: &mut Vec<MultiHash>);
> }
> 
> trait DigestRoot {
> 	fn digest(&self) -> MultiHash;
> }
> ```
> 
> Then we could implement `DigestItem` also on structs and they just call `append_digest` for each of their fields.
> 
> Although it would also be possible to have those two functions on a single trait.
> 
> > I am not sure I understand the proposal. Are you talking about some builder trait?
> 
> Not some builder trait but basically what you mentioned in proposal (2). Maybe it is exactly proposal 2 and I just didn't get it :)
> 
> My motivation was to allow the actual end-user of the trait to only call a single function without parameters, something like:
> 
> ```rust
> let swap: Swap = todo!();
> 
> let digest = swap.digest();
> ```
> 
> `digest` would accumulate all the hashes of each field, sort them and hash them again.

If I understand correctly you are saying that for a nested struct, we should basically flatten it and get the hashes of all fields and sub-fields.

However, this is harder to do than a recursive approach (IMHO).

Regarding the `implementation detail` argument, it means that we should keep 2 separate traits, so that the consumer can only import and use one trait, the other one only used to build the macro.

I'll move forward now taking in account your proposal, I am starting to have a better idea of what is the best solution.

thomaseizinger:
> If I understand correctly you are saying that for a nested struct, we should basically flatten it and get the hashes of all fields and sub-fields.
> 
> However, this is harder to do than a recursive approach (IMHO).
> 
> Regarding the `implementation detail` argument, it means that we should keep 2 separate traits, so that the consumer can only import and use one trait, the other one only used to build the macro.
> 
> I'll move forward now taking in account your proposal, I am starting to have a better idea of what is the best solution.

Yes pretty much! Sorry if this only coming out now but I wasn't sure how this is going to unfold :)

I am excited for what you will come up with :)

D4nte:
> > If I understand correctly you are saying that for a nested struct, we should basically flatten it and get the hashes of all fields and sub-fields.
> > However, this is harder to do than a recursive approach (IMHO).
> > Regarding the `implementation detail` argument, it means that we should keep 2 separate traits, so that the consumer can only import and use one trait, the other one only used to build the macro.
> > I'll move forward now taking in account your proposal, I am starting to have a better idea of what is the best solution.
> 
> Yes pretty much! Sorry if this only coming out now but I wasn't sure how this is going to unfold :)
> 
> I am excited for what you will come up with :)

Look now :) I just pushed. I tried to document but not sure how clear it is.
I am keeping the previously proposed method. There are now 3 traits but it is clearer that the user only need to implement:
- `IntoDigestInput` for basic data types,
- `Digest` using `#[derive(DigestMacro)]` and `#[digest_prefix = ...]` for data structures.

I believe it gives the consumer just the right amount of control.

Regarding the digest implementation, I have kept the previously proposed one because it is just easier to write this kind of thing recursively.

Currently working on supporting nested enums and we should be good to go.

digest/src/lib.rs
```diff
@@ -8,30 +58,45 @@ pub fn digest(bytes: &[u8]) -> Multihash {
...
-impl FieldDigest for Vec<u8> {
-    fn field_digest(mut self, suffix: Vec<u8>) -> Multihash {
-        let mut bytes = suffix;
-        bytes.append(&mut self);
+mod private {
+    pub trait Sealed {}
 
-        digest(&self)
-    }
+    impl<T> Sealed for T where T: super::IntoDigestInput {}
```
thomaseizinger: Shouldn't the `Sealed` trait be implemented manually so you can actually seal things with it? :)

If someone implements `IntoDigestInput` now, they automatically get the `Sealed` trait as-well?

digest/tests/swap_digest.rs
```diff
@@ -52,10 +48,66 @@ impl RootDigest for OtherStruct {
     }
 }
 
+#[derive(DigestMacro)]
```
thomaseizinger: Nit: Would be nice to not have the "Macro" in the name in the end version. I think convention is to name this after the trait is derived.

thomaseizinger:
On the upside, good job on iterating on this! It starts to come out nicely :)

digest/src/lib.rs
D4nte: Well yes that's the idea.

digest/tests/swap_digest.rs
D4nte: The proc macro doc actually show it like that but happy to remove :)

D4nte:
> > Regarding the digest implementation, I have kept the previously proposed one because it is just easier to write this kind of thing recursively.
> 
> I totally understand that it is easier to implement but I would really question if this is the functionality we want?
> 
> Doesn't that result in either not making use of nested data structures or having to carefully document how to exactly calculate the the swap digest with the the intermediary hashing of certain data?
> 
> I can also see quite some risk here in terms of refactoring. What if we one day decide that a certain amount of nesting no longer makes sense and we want to change it? That is totally an implementation detail of comit-rs and yet it affects the API :/
> 
> That is what I meant above with: Maybe we actually don't want/need nested data structure support here because we would always just go for a flat one anyway to avoid this detail to leak into the API?
> 
> If the answer to that is yes, then I am sorry for pushing for this, I didn't think about this particular case and the consequences.

If a refactor change the nesting of the struct then indeed, it will provide different values. But isn't what we want?

If the message change drastically then maybe it should be a new message that is not backward compatible.

If the order of fields or the name of fields change, then the digest will remain the same.

I see your point but at this stage I feel like either way, we are just guessing what could things be.

I believe at this stage I demonstrated what can be done. Once we go and use this trait we can do small changes to the algorithm to finally document it. If we decide that all messages should be flat, it will be extremely easy to update the macro :)

D4nte:
One more step would he to implement enums with c-like structs variants:
```
enum MyEnum {
	Point { x: u32, y: u32 },
	...
}
```

But I think we should see whether we have the need for such enums before wasting the effort.

digest/src/lib.rs
D4nte: Without the `Sealed` trait, this compiles:

`digest/tests/swap_digest.rs`
```
struct MyStruct;

impl FieldDigest for MyStruct {
    fn field_digest(self, _prefix: Vec<u8>) -> Multihash {
        unimplemented!()
    }
}
```

However, with the `Sealed` trait, compilation fails (as desired):

```
error[E0277]: the trait bound `MyStruct: digest::Digest` is not satisfied
   --> digest/tests/swap_digest.rs:289:6
    |
289 | impl FieldDigest for MyStruct {
    |      ^^^^^^^^^^^ the trait `digest::Digest` is not implemented for `MyStruct`
    |
    = note: required because of the requirements on the impl of `digest::IntoDigestInput` for `MyStruct`
    = note: required because of the requirements on the impl of `digest::private::Sealed` for `MyStruct`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: could not compile `digest`.
warning: build failed, waiting for other jobs to finish...
```

digest/src/lib.rs
```diff
@@ -1,4 +1,57 @@
...
+/// impl IntoDigestInput for MyType {
+///     fn into_digest_input(self) -> Vec<u8> {
+///         self.0.revert()
+///     }
+/// }
+/// ```
+///
+/// The digest algorithm goes as follow:
+/// 1. Compute `field_digest(prefix)` for all fields of the struct.
+/// 2. Lexically order the fields
```
tcharding:
```suggestion
/// 2. Lexically order the field digests.
```

digest/src/lib.rs
```diff
@@ -1,4 +1,57 @@
...
+/// 3. Concatenate the result
+/// 4. Hash the result
```
tcharding: nit: either all with ending period or none.

digest/src/lib.rs
```diff
@@ -1,4 +1,57 @@
...
+///
+/// The field digest algorithm goes as follow:
+/// For elementary types:
+///     1. Transform the data in a byte array (if there is
+///   any data) using `IntoDigestInput` trait.
+///     2. Concatenate prefix and the resulting byte array 3. Hash the
```
tcharding: Missing new line.

digest/src/lib.rs
```diff
@@ -1,4 +1,57 @@
...
+///   result
+/// For data structures:
+///     1. Calculate the root digest of the struct
+///     2. concatenate prefix and resulting root digest
```
tcharding: nit: All capitalised first character or none with caps.

PR #2273, opened 2020-03-17
Don't generate siren type definitions automatically

thomaseizinger:
We now have types in the SDK for Siren for quite a while. We can use them in our e2e tests instead of generating the types from the schema.


thomaseizinger:
bors retry

PR #2282, opened 2020-03-19
Fix e2e tests on Mac

D4nte:
Follow-up of #2275.


api_tests/src/ledgers/ledger_lock.ts
```diff
@@ -16,7 +16,7 @@ export default async function ledgerLock(
         retries: {
             retries: 10,
             minTimeout: 200,
-            maxTimeout: 2000,
```
thomaseizinger: This should only be the max timeout _per iteration_ and we do 10 retries :o

D4nte: Oops.

api_tests/src/actor_test.ts
```diff
@@ -24,7 +24,7 @@ function nActorTest(
         const actors = await createActors(name, actorNames);
 
         try {
-            await timeout(60_000, testFn(actors));
+            await timeout(120_000, testFn(actors));
```
bonomat: Is it really necessary to be that high? So far the tests ran just fine on my machine and with these high values we have to wait quite long if something goes wrong.

I assume that there is 1 or 2 tests which take longer than expected and these are probably the LN sanity tests. What do you think of making this value configurable and defaulting to `60_000` if not defined?

thomaseizinger:
@D4nte can you please reduce the changes in this PR to what is essential for the tests to work properly? I am trying to fix the problems on my end and I am unsure, what to pick.

api_tests/tests/ethereum_bitcoin.ts
```diff
@@ -41,6 +42,9 @@ describe("E2E: Ethereum/ether - Bitcoin/bitcoin", () => {
 
             await alice.fundLowGas("0x1b000");
 
+            // It is meaningless to assert before cnd processes a new block
+            await sleep(3000);
```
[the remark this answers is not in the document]
D4nte: Please note that I removed it from actor to transform a 12s wait to 3s.

api_tests/src/actor_test.ts
D4nte: Happy to look at making it configurable. Tests do not run "just fine" on my machine with lower value.

thomaseizinger:
> > @D4nte can you please reduce the changes in this PR to what is essential for the tests to work properly? I am trying to fix the problems on my end and I am unsure, what to pick.
> 
> Each change introduced in this PR is a fix to a test that failed locally during the numerous runs.

I am sorry but how do log messages make a difference for whether the tests pass or fail?

D4nte:
Closing until actual issue is found on `dev` branch.

PR #2289, opened 2020-03-19
Parallel e2e tests: Only start one instance of Parity, Bitcoind and one lnd per actor

thomaseizinger:
By utilizing file-locks on directories, we can ensure that any number
of parallel test suites only ever get to run one instance of
the ledger. The remaining test suites will take the config and
pass that to the tests.

We store the PID of each ledger in a file inside the lock directory,
this also includes a sub-process that we spawn for the Bitcoin miner.
This is important, because same as with the other nodes, this process
must keep going until the last test executed.

To cleanup our environment, we kill all processes with PIDs that we
find in those PID files before and after we execute the tests, just
in case if we for some reason did not exit cleanly and failed to
cleanup after ourselves.

## What works:

- Interrupting the test suite with CTRL+C cleans the enviornment
- Running tests in parallel
- Running an individual test
- Nodes shut down after a successful run

## What doesn't work

- A failing test will not clean up the environment

PR #2307, opened 2020-03-24
Download blockchain nodes in TS instead of with shell scripts 

thomaseizinger:
This is a follow-up PR after https://github.com/comit-network/comit-rs/pull/2289 and should only be reviewed either after that one merged or starting from commit https://github.com/comit-network/comit-rs/commit/c4631e8fe193f7c011498c18195bcbc9a10b0880.

This PR removes the `blockchain_nodes` directory from the repository (it setting there has been bugging me for a while).

We now download the necessary version of the blockchain nodes automatically when trying to start an instance of it. See the commit message for more details.
Apart from it being a better design, this has the main advantage that we don't need to tell everyone to reset their blockchain_nodes folder when we update the version of a node for the e2e tests.


api_tests/src/ledgers/bitcoind_instance.ts
```diff
@@ -100,14 +93,54 @@ export class BitcoindInstance implements BitcoinInstance {
...
+        return binary;
+    }
 
-        this.process.kill("SIGINT");
+    private static downloadUrl() {
+        switch (platform()) {
+            case "darwin":
+                return "https://bitcoincore.org/bin/bitcoin-core-0.17.0/bitcoin-0.17.0-osx64.tar.gz";
+            case "linux":
+                return "https://bitcoincore.org/bin/bitcoin-core-0.17.0/bitcoin-0.17.0-x86_64-linux-gnu.tar.gz";
```
da-kami: ❓  So whenever we want to update to the newest version we have to check what the current version number is and change both links accordingly? Is that by design?

da-kami: An update would mean changing it in code I suppose. Once merged everybody would have the update. Correct me if I'm wrong.

thomaseizinger: This is already the case if you look at the shell scripts that currently download the blockchain nodes.

thomaseizinger: In fact, the whole point of this is (if you read the commit messages) that we can now in fact smoothly update these versions instead of everyone having to delete the nodes from the directory to actually trigger the re-download of the newer version.

With these scripts, they are now stored in a folder inside `node_modules` that contains the version number as-well and the nodes are reliably downloaded if we want a different version.

thomaseizinger: Read the commit messages, they explain it it detail :)

thomaseizinger: Do you see an alternative way of doing this that gives us stable builds?

thomaseizinger: Do you see an alternative way of doing this that gives us stable builds?

api_tests/src/ledgers/bitcoind_instance.ts
```diff
@@ -100,14 +93,54 @@ export class BitcoindInstance implements BitcoinInstance {
...
+
+            return envOverride;
+        }
+
+        const cacheDir = findCacheDir({
+            name: "bitcoin-core-0.17.0",
+            create: true,
+            thunk: true,
+        });
+        const binary = cacheDir("bitcoin-0.17.0", "bin", "bitcoind");
```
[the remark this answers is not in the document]
thomaseizinger: This is unfortunately an artifact of the directory structure inside the archive we are downloading.

The archive is named `bitcoin-core-0.17.0` which results in an equally named directory once extracted.
This directory then contains another directory `bitcoin-0.17.0` which contains a directory `bin` which finally contains the binary `bitcoind`.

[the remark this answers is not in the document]
thomaseizinger: Yes I can introduce variables to make this clearer :)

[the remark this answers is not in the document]
thomaseizinger: Done!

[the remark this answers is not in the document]
bonomat: Got it, thanks :)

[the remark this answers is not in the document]
bonomat: awesome, thanks.

PR #2314, opened 2020-03-25
Simplify states and traits

thomaseizinger:
Some things I came across while we are hacking on the LN implementation :)


thomaseizinger:
> I think the idea was to keep this generic, so we can reuse it for other things than `LedgerStates`. Happy to make it specific for now, if it is only used for `LedgerStates`.
> 
> @luckysori

If we want to make it generic, then the inner thing should be generic and reused (for example in `SwapCommunicationStates`)

Issue #2297, opened 2020-03-23
[Lightning Squad][24Mar-7Apr] Rush 1


D4nte:
### Goal

Have one working e2e test happy path:
- Alpha: Ether
- Beta: Bitcoin lightning

Should work as per [specs](https://github.com/comit-network/spikes/blob/master/0022-lnd-plan-of-attack.adoc).

Use of `unimplemented!()` are authorised.

### Out of scope

- Other combinations
- Refund
- SDK
- DB

Issue #1954, opened 2020-01-29
Add a model-skeleton for the HALight protocol


thomaseizinger:
## Problem

In order to execute Lightning swaps, our software needs to model the protocol that is described in the according spike: https://github.com/comit-network/spikes/blob/lnd-plan-of-attack/0022-lnd-plan-of-attack.adoc

## Goal

Define a skeleton around the core concepts of the protocol:

- Ledger states
- Events
- `Lightning` ledger

## Recommendation

Inspire yourself from the existing protocols: `HAN` and `HErc20`.
This is only about implementing a _skeleton_, hence having `unimplemented!()` in several places is totally fine!

Implementing the event traits (analogous to `HtlcDeployed` for HErc20 for example) of HALight, will uncover, what the API of a future LN connector is.
It is within the scope of this ticket to define a stubbed `LndConnector` that already has the API needed for our event traits.

D4nte:
Done as part of #2297.

PR #2315, opened 2020-03-25
Add a walking skeleton for the halight-based swaps

thomaseizinger:
This is just for people to follow along of what the mob is doing :)

Things to do before this is merged:

## Pull out

- [ ] Pull out tracing improvements of the Ethereum module  (#2379)
- [ ] Pull out extension of cnd's config file with lnd values (TS) (#2382)
- [ ] Pull out cnd's configuration handling (default lnd path) (#2378)
- [ ] Pull out changes to announce protocol (#2383)

## Before merge

- [x] Don't use `Actions` trait for han-halight but create one trait per action, i.e. `FundAction` with a fn `fund_action(&self) -> Option<Self::Output>` @thomaseizinger 
- [ ] Implement Swap Digest properly @D4nte 
- [x] Refactor `halight::create_watcher` to return a Stream instead (i.e. remove the dependency on `InvoiceStates` from this fn)

## Related issues

Resolves https://github.com/comit-network/comit-rs/issues/1954.
Resolves https://github.com/comit-network/comit-rs/issues/2297.


cnd/src/lnd.rs
```diff
@@ -57,27 +57,68 @@ struct Payment {
...
+            Path(path) => {
+                let mut buf = Vec::new();
+                std::fs::File::open(path)?.read_to_end(&mut buf)?;
+                let certificate = reqwest::Certificate::from_pem(&buf)?;
+                Ok(LazyCertificate::Certificate(certificate))
+            }
+        }
+    }
+
+    pub fn certificate(&self) -> Result<&Certificate, Error> {
```
thomaseizinger: @D4nte 

Instead of erroring here, you could just do the read logic then you don't have to care about the error case :)

D4nte: Do you mean read the file in this function?

Yes it is what I have done initially however, I did not want to read the file "every time".
To not have to read the file everytime I have to have a `&mut self` and it just bubbled up too much.

Hence the current solution where I do not read the current file has often.

thomaseizinger:
Replaced by https://github.com/comit-network/comit-rs/pull/2426.

D4nte:
@thomaseizinger "Implement Swap Digest properly @D4nte" we have an issue tracking: #2385.

PR #2329, opened 2020-03-26
Lock datadir

### LINKED ISSUE #2278: Ensure only one cnd uses a given data dir
## Problem

It is not obvious to the user that if they are running several cnds locally for development purposes, they need to add a `data.dir` parameter to their configuration to ensure the various cnds do not use the same seed and database.

See https://github.com/comit-network/comit-rs/issues/2214#issuecomment-600347964 for example.

## Goal

Making it easy for interested developer to build on top of COMIT.

If a cnd is already started on the given datadir, starting a new cnd should not be possible. This second cnd should die and tell the user that the datadir is already in use.

The user should not have to worry about this lock mechanism. Even if a cnd dies unexpectedly/harshly, the user should not have to do any cleanup.
Be sure to (manually) test with `kill -9`.

## Recommendation

Look if there are any crate that can help with handling of lock files.

rishflab:
Place a lock on the datadir specified in config. cnd aborts with
an error message if the specified datadir is already being used
by another cnd process.

Closes #2278

### DIFF
--- a/cnd/src/main.rs
+++ b/cnd/src/main.rs
@@ -53,9 +53,11 @@ fn main() -> anyhow::Result<()> {
 
     crate::trace::init_tracing(settings.logging.level)?;
 
-    let _locked_datadir = &settings.data.dir.try_lock_exclusive()?;
+    let database = Sqlite::new_in_dir(&settings.data.dir)?;
 
     let seed = RootSeed::from_dir_or_generate(&settings.data.dir, OsRng)?;
+
+    let _locked_datadir = &settings.data.dir.try_lock_exclusive()?;
 
     let mut runtime = runtime::Builder::new()
         .enable_all()
@@ -114,8 +116,6 @@ fn main() -> anyhow::Result<()> {
     let swap_communication_states = Arc::new(SwapCommunicationStates::default());
 
     let swap_error_states = Arc::new(SwapErrorStates::default());
-
-    let database = Sqlite::new_in_dir(&settings.data.dir)?;
 
     let swarm = Swarm::new(
         &settings,
--- a/cnd/src/main.rs
+++ b/cnd/src/main.rs
@@ -21,18 +21,18 @@ use cnd::{
     },
     config::{self, validation::validate_blockchain_config, Settings},
     db::Sqlite,
+    file_lock::TryLockExclusive,
     http_api::route_factory,
     jsonrpc, load_swaps,
     network::Swarm,
     seed::RootSeed,
     swap_protocols::{Facade, LedgerStates, SwapCommunicationStates, SwapErrorStates},
 };
-use fs2::FileExt;
+
 use rand::rngs::OsRng;
 use std::{process, sync::Arc};
 use structopt::StructOpt;
 use tokio::runtime;
-
 mod cli;
 mod trace;
 
@@ -53,11 +53,7 @@ fn main() -> anyhow::Result<()> {
 
     crate::trace::init_tracing(settings.logging.level)?;
 
-    let path = &settings.data.dir.to_path_buf();
-
-    let file = std::fs::File::open(&path)?;
-    file.try_lock_exclusive()?;
-    file.lock_exclusive()?;
+    let _locked_datadir = &settings.data.dir.try_lock_exclusive()?;
 
     let seed = RootSeed::from_dir_or_generate(&settings.data.dir, OsRng)?;
 
--- a/cnd/src/main.rs
+++ b/cnd/src/main.rs
@@ -27,6 +27,7 @@ use cnd::{
     seed::RootSeed,
     swap_protocols::{Facade, LedgerStates, SwapCommunicationStates, SwapErrorStates},
 };
+use fs2::FileExt;
 use rand::rngs::OsRng;
 use std::{process, sync::Arc};
 use structopt::StructOpt;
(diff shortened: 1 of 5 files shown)


rishflab:
I wrapped the logic into a function to improve the error message

```

fn lock_datadir(path: &PathBuf) -> anyhow::Result<()> {
    let file = File::open(path)?;
    match file.try_lock_exclusive() {
        Err(e) => {
            tracing::error!("Could not acquire lock on datadir: {:?}", path);
            Err(anyhow::Error::from(e))
        }
        Ok(_) => Ok(file.lock_exclusive()?),
    }
}
```

When i use the ? operator on this function in main 
```
lock_datadir(&path)?;
```
it doesn't abort the program

am i missing something really obvious?

cnd/src/main.rs
```diff
@@ -52,6 +53,12 @@ fn main() -> anyhow::Result<()> {
 
     crate::trace::init_tracing(settings.logging.level)?;
 
+    let path = &settings.data.dir.to_path_buf();
+
+    let file = std::fs::File::open(&path)?;
+    file.try_lock_exclusive()?;
+    file.lock_exclusive()?;
```
tcharding: Unless I'm missing something, this is not correct.  You should not be locking the file twice.  Just using `try_lock_exclusive()?`.  Also, can we have this all in a function please, so in main it just reads something like `lock_data_dir(&settings.data.dir)?`.  Thanks

rishflab: managed to move into a lock_datadir function. It was wasn't working earlier because you need to keep the file object in scope to maintain the lock

rishflab: also was able to remove the redundant lock statement succesfully

config1.toml
```diff
@@ -0,0 +1,5 @@
+[data]
+dir = "/Users/rishflab/datadir"
```
thomaseizinger: I guess these files should not be part of the PR? :D

thomaseizinger:
That `Settings` idea might not be so gooed because the settings and the file lock are actually separate concerns 🤔

cnd/src/main.rs
```diff
@@ -200,3 +203,14 @@ fn dump_config(settings: Settings) -> anyhow::Result<()> {
     println!("{}", serialized);
     Ok(())
 }
+
+fn lock_datadir(path: &PathBuf) -> anyhow::Result<File> {
+    let file = File::open(path)?;
+    match file.try_lock_exclusive() {
+        Err(e) => {
+            tracing::error!("Could not acquire lock on datadir: {:?}", path);
+            Err(anyhow::Error::from(e))
```
thomaseizinger: Logging and passing on the error is a bit odd, you should be able to use `.context` here to add this information :)

https://docs.rs/anyhow/1.0.27/anyhow/trait.Context.html

thomaseizinger: So basically doing this:

```rust
File::open(path)?.try_lock_exclusivly().with_context(|| format!("Could not acquire lock on data dir {}", path.display()))?
```

cnd/src/main.rs
```diff
@@ -52,6 +53,8 @@ fn main() -> anyhow::Result<()> {
 
     crate::trace::init_tracing(settings.logging.level)?;
 
+    let _file = lock_datadir(&settings.data.dir)?;
```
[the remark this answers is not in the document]
rishflab: > That `Settings` idea might not be so gooed because the settings and the file lock are actually separate concerns 🤔

Yeh wasn't completely sure about this. I think we should leave it main for now until there is an obvious and fitting place where to keep it.

cnd/src/main.rs
thomaseizinger: Could also do an extension trait for `PathBuf` to put this code somewhere else and not having it inside `main.rs` :)

cnd/src/main.rs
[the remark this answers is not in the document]
rishflab: > If you would change some of the later function calls to use this file, returning it would not be useless :)
> 
> Otherwise, name it something like `_file_lock_to_ensure_we_have_exclusive_access_to_the"_config_until_this_is_dropped`.

The file lock ensures exclusive access to the datadir containing the seed.pem and the sqlite db file not the config.

cnd/src/main.rs
rishflab: > So basically doing this:
> 
> ```rust
> File::open(path)?.try_lock_exclusivly().with_context(|| format!("Could not acquire lock on data dir {}", path.display()))?
> ```

ok ill check his out, seems much cleaner

cnd/src/main.rs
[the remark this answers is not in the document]
D4nte: Well, the config also needs to be exclusive otherwise you will have conflict on the ports cnd binds to. But that will be solved as cnd will just crash because it cannot bind to said ports.

D4nte: Does it print an error on stderr on failure?

rishflab: No idea tbh. @thomaseizinger  do you know where this error will be printed?

rishflab: Should I put an exclusive lock on the config file as well or not? You said it will crash anyway due to a port binding failure?

D4nte: Lock on data dir is enough.
Just try the behaviour locally ;)

PR #2330, opened 2020-03-26
Close all handles during Jest teardown

### LINKED ISSUE #2211: Investigate e2e test error `worker process has failed to exit gracefully`
## Problem

The new e2e test setup prints this `error` upon execution on some machines ( @D4nte and me had this issue on Mac):

```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks.
```

The tests still pass, but it seems there is some problem.

## Goal

No error message.

## Proposed solution

1. Investigation (check on jest side)
2. Fix (could be done in a separate issue if big)

luckysori:
_Waiting for comit-network/comit-js-sdk#176 to be merged and for those changes to be included in a release of `comit-js-sdk` before we can update the last commit in this PR to use that release instead of a local dependency created via [`yalc`](https://github.com/whitecolor/yalc) (which seems to work **much** better than `yarn link`)._

Fixes #2211.

1. Turns out Jest will hang until every timeout is cleared, so racing against a timeout that is only cleared after it resolves was not good.
2. The proper teardown of the e2e test environment is fully achieved with 3698f3a, after a lot of debugging. Unfortunately, Jest recommends this

```
Consider running Jest with `--detectOpenHandles` to troubleshoot this issue
```

when tests hang after passing, but it just [doesn't seem to do anything](https://github.com/facebook/jest/issues/9473). I tried using [`leaked-handles`](https://github.com/Raynos/leaked-handles) without success, but finally found [`wtfnode`](https://github.com/myndzi/wtfnode), which identified `bcoin` as the culprit.

---

My OCD was definitely triggered with this one, as a different error got me to investigate how to get rid of `smack-my-jasmine-up`. Turns out removing it did not fix that other error, but I still see it as a win, since depending on such an obscure library is bad.

---

And another bug squashed! No more

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGHUP listeners added to [process]. Use emitter.setMaxListeners() to increase limit
```

errors, by combining cde8e0c and 127a4a2:
 - shutting down the logger during teardown.
 - closing log files after 5 seconds of inactivity, to prevent us from reaching the limit of listeners.

I believe log files can be reopened if something else needs to be logged.


bonomat:
`smack-my-jasmine-up` always annoyed me a bit aswell. 
So what's your strategy here now as your PR does not seem to build?

luckysori:
> `smack-my-jasmine-up` always annoyed me a bit aswell.
> So what's your strategy here now as your PR does not seem to build?

Silly me, forgot to do the same for e2e tests (not just dry)!

api_tests/src/utils.ts
```diff
@@ -47,16 +47,20 @@ export async function sleep(time: number) {
...
-        const id = setTimeout(() => {
-            clearTimeout(id);
+        id = setTimeout(() => {
             reject(new Error(`timed out after ${ms}ms`));
         }, ms);
     });
 
     // Returns a race between our timeout and the passed in promise
-    return Promise.race([promise, timeout]);
+    return Promise.race([promise, timeout]).then((result) => {
```
thomaseizinger: Should we maybe use .finally here, just to be sure?

api_tests/src/configure_jasmine.ts
```diff
@@ -0,0 +1,5 @@
+jasmine.getEnv().addReporter({
+    specStarted: (result) =>
+        // @ts-ignore
```
thomaseizinger: Can we get typings for this instead of using `ts-ignore`?

api_tests/src/e2e_test_environment.ts
```diff
@@ -251,26 +252,33 @@ export default class E2ETestEnvironment extends NodeEnvironment {
...
+
         this.logger.info("Tearing down complete");
     }
 
     async cleanupAll() {
         const tasks = [];
 
         if (this.bitcoinLedger) {
-            tasks.push(this.bitcoinLedger.stop);
+            tasks.push(this.bitcoinLedger.stop());
```
thomaseizinger: Lol, stupid me :sweat_smile:

api_tests/src/utils.ts
```diff
@@ -47,16 +47,19 @@ export async function sleep(time: number) {
 }
 
 export async function timeout<T>(ms: number, promise: Promise<T>): Promise<T> {
```
thomaseizinger: Should we replace our timeout with this?

https://www.npmjs.com/package/p-timeout

Just checked and basically does the same thing as we do here and closes the timeout properly.

thomaseizinger:
So many good fixes in this one :)

api_tests/src/utils.ts
luckysori: Do you reckon we should do the same for the SDK then? I'm all for not having to maintain a timeout function haha.

luckysori: > Do you reckon we should do the same for the SDK then?

We can leave that for later I guess, to not introduce more complexity into these changes, but I'm happy to do it in `comit-rs` now.

api_tests/src/configure_jasmine.ts
luckysori: Maybe you can help me figure this one out. 

I now realise that I'm using the `jasmine` namespace to set and read a value which wasn't there in the first place i.e. there is no type because I'm creating that "field".

Would it make more sense to save this to the global context of the e2e test suite?

luckysori: > Would it make more sense to save this to the global context of the e2e test suite?

I guess not because this global context can be accessed by two tests running concurrently.

api_tests/package.json
```diff
@@ -42,16 +43,16 @@
         "chai-json-schema": "^1.5.0",
         "chai-string": "^1.5.0",
         "chai-subset": "^1.6.0",
-        "comit-sdk": "^0.15.0",
+        "comit-sdk": "file:.yalc/comit-sdk",
```
bonomat: I guess that wasn't on purpose :D 
```suggestion
        "comit-sdk": "file:.yalc/comit-sdk",
```

luckysori: Oh yeah, fixing this as soon as I tackle Thomas' suggestions!

api_tests/src/configure_jasmine.ts
luckysori: I'm just gonna leave this one as an optional follow-up PR. Had a look at merging namespaces to extend the one provided by `jest` with the type we need here, but there's probably a much easier approach that I'm missing!

api_tests/src/utils.ts
thomaseizinger: > Do you reckon we should do the same for the SDK then? I'm all for not having to maintain a timeout function haha.

Exactly this, the code duplication makes it obvious that we shouldn't be maintaining this I think.

api_tests/src/configure_jasmine.ts
thomaseizinger: > I guess not because this global context can be accessed by two tests running concurrently.

The test environment provides a unique instance of global to every test so you can safely set things there!

thomaseizinger: But yes, we can do this as a follow-up. I don't have time at the moment so you could write a short issue please? :)

Issue #2139, opened 2020-03-03
Remove chai from end2end tests


bonomat:
This is a follow-up ticket from https://github.com/comit-network/comit-rs/pull/2106

# Problem

After moving to jest we should only use jest matchers and no chai matchers

# Goal

Don't have any chai matchers and remove chai from package.json

# Recommendation

Remove https://github.com/comit-network/comit-rs/blob/a9aa5720fedc297cb7005b9461d727f97911b7ba/api_tests/package.json#L59-L66 (including the types a few lines up) and fix the e2e tests.

Note: straightforward replace, don't refactor everything.

D4nte:
Note: json schema validation with jest looks hard.

I tried using https://github.com/epoberezkin/ajv/releases/tag/5.0.0

But our schema is `draft-04` and the steps in the migration section of the release above threw an error.

D4nte:
@thomaseizinger is it fine if we update our json schema to draft-06?

thomaseizinger:
> @thomaseizinger is it fine if we update our json schema to draft-06?

Yes,

Issue #2013, opened 2020-02-11
Clean-up dry tests


bonomat:
# Problem

The PR https://github.com/comit-network/comit-rs/pull/1978 introduced code which asks for optimization as the goal of this PR was on moving to the new e2e test framework so that we can run the tests in parallel.

Generally the idea is to make those "old", moved dry tests similar to how we write e2e tests.

Concrete issues that were recorded:

**Assertions**
Instead of having the assert functions in the defined in the test, add them to the `Actor` as it is done in other e2e tests. This should make the code more readable.

https://github.com/comit-network/comit-rs/blob/5c2f4e06f24c84c751af76f906808b1d5310b8d0/api_tests/tests/peers_using_ip.ts#L10-L26
**Properly leverage the SDK for communication with cnd*

**Leverage the Cnd instance in Actor**

Instead of using the `actor.cndHttpApiUrl().get(...)`, we should use the `Cnd` instance that is provided in the `Actor` and use the `fetch()` method; inspire yourself from other tests. Consider using `ComitClient` if possible.

**Remove globally added test data**

Remove the globally added test data from `utils.ts`; consider moving all tests that rely on that data into one file and move the test data there.
Creating global test data leads to reusing it in different tests, which is bad for maintanability. 

https://github.com/comit-network/comit-rs/blob/b3a5012409787e86431bbb0817e04ba243eb64c3/api_tests/src/utils.ts#L52-L82

bonomat:
Hi @thomaseizinger : this ticket has been on to-refine for a while now. What's the status here?  Might this even become obsolete soon once we have a new API?

thomaseizinger:
> Hi @thomaseizinger : this ticket has been on to-refine for a while now. What's the status here? Might this even become obsolete soon once we have a new API?

The comments I made on the PR are still valid. I put it in to-refine because the issue you created doesn't actually say what should be done and the title is very generic.

Unfortunately, I never got around to actually do the refinement then. Since you opened it originally, do you mind if you do the refinement?
Otherwise I can do it when I get the time for it.

I labelled it as technical-debt because we took a shortcut in the PR to move faster.

da-kami:
@thomaseizinger I gave it a try and updated this, hope I got all the things right :)

da-kami:
Note: Some chai `expect` statements were already removed, might be interesting to take a look at when tackling this issues: https://github.com/comit-network/comit-rs/pull/2346

thomaseizinger:
[...]

PR #2346, opened 2020-03-30
Remove chai expect from some tests

D4nte:
Helps with #2139

Not doing all of them to not fall in a sinkhole.


api_tests/src/actors/actor.ts
```diff
@@ -489,7 +488,7 @@ export class Actor {
                 defaultAssetDescription(asset, ledger)
             );
 
-            expect(currentWalletBalance).to.be.bignumber.gte(balanceInclFees);
+            expect(currentWalletBalance.gte(balanceInclFees)).toBeTruthy();
```
thomaseizinger: :mega: This is going to produce horribly unhelpful error messages (false is not equal to true).

api_tests/tests/dry/peers_using_ip.ts
```diff
@@ -14,19 +14,27 @@ import { Actor } from "../../src/actors/actor";
 async function assertNoPeersAvailable(actor: Actor, message: string) {
     const peersResponse = await request(actor.cndHttpApiUrl()).get("/peers");
 
-    expect(peersResponse.status).to.equal(200);
-    expect(peersResponse.body.peers, message).to.be.empty;
+    expect(peersResponse.status).toEqual(200);
+    try {
+        expect(peersResponse.body.peers).toBeArrayOfSize(0);
+    } catch (e) {
+        throw new Error(message);
```
thomaseizinger: This kind of evolution of code is why I often prefer to not push changes like the one requested in https://github.com/comit-network/comit-rs/pull/1978 too much down the line :(

cc @bonomat: https://github.com/comit-network/comit-rs/issues/2013

They just don't get addressed and things get progressively worse from there ..

D4nte:
Not needed for lnd mob as the code I was doing that for is being moved away from actor.ts by @thomaseizinger.

PR #2378, opened 2020-04-06
Improve config for LND

tcharding:
Improve the cnd configuration of LND by doing:

- Add the TLS certificate path and macaroon path to LND configuration.
- Don't use shared data structure for LND between `file.rs` and
`settings.rs`.
- Update unit tests.

Done as part of the LN squad.


cnd/src/config/settings.rs
```diff
@@ -489,4 +454,22 @@ mod tests {
                 lnd: Lnd::default(),
             })
     }
+
+    #[test]
+    fn error_on_http_url_for_lnd() {
```
luckysori, line 459: :thinking: I guess it would be just as valuable and more precise to test `fn check_url_lnd`.

thomaseizinger: I would argue that testing the public API that is actually in use is more robust because it allows us to refactor the internals and still use the tests to make sure we didn't break anything :)

PR #2405, opened 2020-04-06
Remove dependency on `primitive_types` for Ethereum types such as Address and TransactionHash

da-kami:
`H256` (Hash) and `H160` (Address) are converted to our own types.
Note that for deserialization we always have to pass the Hash/Address by reference, otherwise we don't reach the right code-path in serde hex. @thomaseizinger already [tackled that problem](https://github.com/fspmarshall/serde-hex/pull/8). 

Unfortunately, the `U256` [value in Ethereum is not fixed hex](https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding) (although it actually is of fixed size, namely `[u8, 32]`). 
This leads to the problem, that we would have to implement our own `Serializer` and `Deserializer` because serde-hex does not implement `Compact` for `[u8, 32]`.

Implementing a custom `Serializer` and `Deserializer` is, however, not trivial, because our hex library does not support compacted hex neither.

I started pulling in the [serialization and deserialization code from the parity codebase](https://github.com/paritytech/parity-common/blob/master/primitive-types/impls/serde/src/serialize.rs), but it is quite a lot of code to be pulled in, that is rather complicated.
Given, that this ticket was under the premise that we (most likely...) can just use serde-hex or our hex library for encoding/decoding the value, this is getting a rabbit hole. Since I already wasted quite some time on this ticket due to a tricky [serde-hex bug](https://github.com/fspmarshall/serde-hex/pull/8) I would put the implementation of `U256` aside now and keep the dependency on `primitive_types` for it. 

We can decide how/if we want to deal with `U256` later. I am not sure we want to own the custom serialization/deserialization code, or if it is not better to use `primitive_types::U256`.

### DIFF
--- a/cnd/src/btsieve/ethereum.rs
+++ b/cnd/src/btsieve/ethereum.rs
@@ -6,15 +6,12 @@ use crate::{
     btsieve::{
         find_relevant_blocks, BlockByHash, BlockHash, LatestBlock, Predates, PreviousBlockHash,
     },
-    ethereum::{Address, Bytes, Input, Log, Transaction, TransactionReceipt, H256, U256},
+    ethereum::{Address, Block, Bytes, Hash, Input, Log, Transaction, TransactionReceipt, U256},
 };
 use anyhow;
 use async_trait::async_trait;
 use chrono::NaiveDateTime;
 use genawaiter::{sync::Gen, GeneratorState};
-
-type Hash = H256;
-type Block = crate::ethereum::Block;
 
 #[async_trait]
 pub trait ReceiptByHash: Send + Sync + 'static {
@@ -24,7 +21,7 @@ pub trait ReceiptByHash: Send + Sync + 'static {
 impl BlockHash for Block {
     type BlockHash = Hash;
 
-    fn block_hash(&self) -> H256 {
+    fn block_hash(&self) -> Hash {
         self.hash
             .expect("Connector returned latest block with null hash")
     }
@@ -33,7 +30,7 @@ impl BlockHash for Block {
 impl PreviousBlockHash for Block {
     type BlockHash = Hash;
 
-    fn previous_block_hash(&self) -> H256 {
+    fn previous_block_hash(&self) -> Hash {
         self.parent_hash
     }
 }
@@ -174,7 +171,7 @@ where
                     topic.as_ref().map_or(true, |topic| {
                         block
                             .logs_bloom
-                            .contains_input(Input::Raw(topic.0.as_ref()))
+                            .contains_input(Input::Raw(&topic.0.as_bytes()))
                     })
                 });
                 if !maybe_contains_transaction {
@@ -227,7 +224,7 @@ impl Predates for Block {
 
 #[derive(Clone, Copy, Default, Eq, PartialEq, serde::Serialize, serdebug::SerDebug)]
 #[serde(transparent)]
-pub struct Topic(pub H256);
+pub struct Topic(pub Hash);
 
 /// Event works similar to web3 filters:
 /// https://web3js.readthedocs.io/en/1.0/web3-eth-subscribe.html?highlight=filter#subscribe-logs
--- a/cnd/src/ethereum.rs
+++ b/cnd/src/ethereum.rs
@@ -2,21 +2,141 @@
 #![forbid(unsafe_code)]
 
 pub use ethbloom::{Bloom as H2048, Input};
-pub use primitive_types::{H160, H256, U128, U256};
+use hex::FromHexError;
+pub use primitive_types::U256;
 use serde::{Deserialize, Serialize};
 use serde_hex::{CompactPfx, SerHex, SerHexSeq, StrictPfx};
-
-pub type Address = H160;
+use std::{
+    fmt,
+    fmt::{Display, Formatter, LowerHex},
+    str::FromStr,
+};
 
 #[derive(Debug, Default, Copy, Clone, PartialEq, Deserialize)]
 pub struct H64(#[serde(with = "SerHex::<StrictPfx>")] [u8; 8]);
+
+#[derive(
+    Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize,
+)]
+pub struct Address(#[serde(with = "SerHex::<StrictPfx>")] [u8; 20]);
+
+impl Address {
+    pub fn from_slice(src: &[u8]) -> Self {
+        let mut address = Address([0u8; 20]);
+        address.0.copy_from_slice(src);
+        address
+    }
+}
+
+impl From<[u8; 20]> for Address {
+    fn from(bytes: [u8; 20]) -> Self {
+        Address(bytes)
+    }
+}
+
+impl From<Address> for [u8; 20] {
+    fn from(s: Address) -> Self {
+        s.0
+    }
+}
+
+impl FromStr for Address {
+    type Err = FromHexError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        hex::decode(s).map(|v| Address::from_slice(v.as_slice()))
+    }
+}
+
+impl From<Address> for Hash {
+    fn from(address: Address) -> Self {
+        let mut h256 = Hash([0u8; 32]);
+        h256.0[(32 - 20)..32].copy_from_slice(&address.0);
+        h256
+    }
+}
+
+impl LowerHex for Address {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        if f.alternate() {
+            write!(f, "0x")?;
+        }
+        for i in &self.0[..] {
+            write!(f, "{:02x}", i)?;
+        }
+        Ok(())
+    }
+}
+
+#[derive(
+    Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize,
+)]
+pub struct Hash(#[serde(with = "SerHex::<StrictPfx>")] [u8; 32]);
+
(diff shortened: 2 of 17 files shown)


cnd/src/ethereum.rs
```diff
@@ -1,22 +1,140 @@
...
+
+impl FromStr for Address {
+    type Err = FromHexError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        hex::decode(s).map(|v| Address::from_slice(v.as_slice()))
+    }
+}
+
+impl From<Address> for Hash {
```
thomaseizinger, line 49: That looks weird, why do we need to convert between an Address and a Hash?

cnd/src/ethereum.rs
```diff
@@ -1,22 +1,140 @@
...
 
+#[derive(
+    Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize,
+)]
+pub struct Address(#[serde(with = "SerHex::<StrictPfx>")] [u8; 20]);
+
+impl Address {
+    pub fn from_slice(src: &[u8]) -> Self {
+        let mut address = Address([0u8; 20]);
+        address.0.copy_from_slice(src);
```
thomaseizinger, line 24: I think this can panic if the slice is not long enough, i.e. < 20 bytes. Can you please create a follow-up ticket for auditing the usages of this function with the goal of removing it?

cnd/src/ethereum.rs
thomaseizinger: If we currently do this in production code and removing that is non-trivial, can you please record a follow-up issue?

cnd/src/ethereum.rs
```diff
@@ -1,22 +1,140 @@
...
+impl From<Hash> for [u8; 32] {
+    fn from(s: Hash) -> Self {
+        s.0
+    }
+}
+
+impl Hash {
+    pub fn from_slice(src: &[u8]) -> Self {
+        let mut h256 = Hash([0u8; 32]);
+        h256.0.copy_from_slice(src);
```
thomaseizinger, line 89: Same as above, this can panic and we should avoid using this function.

cnd/src/ethereum.rs
```diff
@@ -86,7 +204,70 @@ impl<T: Into<Vec<u8>>> From<T> for Bytes {
...
+
+    #[test]
+    fn deserialise_address() {
+        let json =
+            serde_json::Value::String("0xc5549e335b2786520f4c5d706c76c9ee69d0a028".to_owned());
+        let _: Address = Address::deserialize(&json).unwrap();
+    }
+
+    #[test]
+    #[should_panic]
```
thomaseizinger, line 219: Note: Instead of `should_panic`, you could also test that `serde_json::from_value` returns an `Err` :)

cnd/src/http_api/problem.rs
```diff
@@ -72,7 +72,7 @@ pub fn from_anyhow(e: anyhow::Error) -> HttpApiProblem {
     }
 
     if e.is::<serde_json::Error>() {
-        tracing::error!("deserialization error: {}", e);
+        tracing::error!("deserialization error: {:?}", e);
```
thomaseizinger, line 75: I think this should be reverted :)

cnd/src/network.rs
```diff
@@ -1012,9 +1012,10 @@ impl SendRequest for Swarm {
 
                 match decision {
                     Some(Decision::Accepted) => {
-                        match serde_json::from_value::<rfc003::messages::AcceptResponseBody<AI, BI>>(
-                            response.body().clone(),
```
thomaseizinger, line 1016: Nice, we are avoiding to calls to `clone` :)

cnd/src/ethereum.rs
```diff
@@ -1,22 +1,140 @@
...
+            write!(f, "0x")?;
+        }
+        for i in &self.0[..] {
+            write!(f, "{:02x}", i)?;
+        }
+        Ok(())
+    }
+}
+
+impl Display for Hash {
```
thomaseizinger, line 118: Eventually, I would like this `Display` implementation to not print `...` but do what the implementation of `LowerHex` does.
We can then remove the implementation of `LowerHex`.

This will likely require a few changes to log statements in our code where we currently use `{:x}` to print the full transaction hash.

Can be done as a follow-up ticket :)

cnd/src/ethereum.rs
```diff
@@ -1,22 +1,140 @@
...
+
+impl From<Address> for Hash {
+    fn from(address: Address) -> Self {
+        let mut h256 = Hash([0u8; 32]);
+        h256.0[(32 - 20)..32].copy_from_slice(&address.0);
+        h256
+    }
+}
+
+impl fmt::LowerHex for Address {
```
thomaseizinger, line 57: Instead of `LowerHex`, I would like us to use `Display` for printing the full address. That is much more natural in log statements.

cnd/src/ethereum.rs
```diff
@@ -86,7 +204,70 @@ impl<T: Into<Vec<u8>>> From<T> for Bytes {
 #[cfg(test)]
 mod tests {
     use super::Log;
-    use crate::ethereum::TransactionReceipt;
+    use crate::ethereum::{Address, Hash, TransactionReceipt};
```
thomaseizinger, line 207: Nit: If you do `use super::*` in `tests` modules, you automatically get all types from the parent module which are very likely the ones you want to test :)

cnd/src/ethereum.rs
```diff
@@ -60,10 +178,10 @@ pub struct Log {
 #[derive(Debug, Default, Clone, PartialEq, Deserialize)]
 pub struct Block {
     /// Hash of the block
-    pub hash: Option<H256>,
+    pub hash: Option<Hash>,
```
thomaseizinger, line 181: Can you record a follow-up issue for making this not an `Option`? It is so annoying :grimacing:

cnd/src/ethereum.rs
da-kami: Recorded as follow up: https://github.com/comit-network/comit-rs/issues/2411

The problem is that the [`Topic`](https://github.com/comit-network/comit-rs/blob/1ceaa33ee93ac80d465a90001d21a0c96079ce6a/cnd/src/btsieve/ethereum.rs#L230) used to search in `Event`s just accepts a `Hash` specifically, but we also use it for searching for `Address`es - we should change this to be generic.

cnd/src/ethereum.rs
da-kami: follow up: https://github.com/comit-network/comit-rs/issues/2412

cnd/src/ethereum.rs
da-kami: follow up: #2412

cnd/src/ethereum.rs
thomaseizinger: Okay interesting, I would put this functionality on the specific topic then and not as a generall conversion between addresses and hashes.

cnd/src/ethereum.rs
```diff
@@ -1,22 +1,140 @@
 #![warn(unused_extern_crates, missing_debug_implementations, rust_2018_idioms)]
 #![forbid(unsafe_code)]
 
+use bitcoin::hashes::core::{fmt::Formatter, str::FromStr};
```
tcharding, line 4: Why are we importing these via `bitcoin`?  Can they not come from `std`?

cnd/src/ethereum.rs
```diff
@@ -1,22 +1,140 @@
...
+use core::fmt;
```
tcharding, line 5: This one is confusing also.  There exists `std::fmt::LowerHex`, is that not the same as the usage below of `fmt::LowerHex`?

cnd/src/ethereum.rs
da-kami: recorded: https://github.com/comit-network/comit-rs/issues/2413

cnd/src/ethereum.rs
da-kami: recorded: https://github.com/comit-network/comit-rs/issues/2413

cnd/src/ethereum.rs
da-kami: recorded: https://github.com/comit-network/comit-rs/issues/2414

cnd/src/ethereum.rs
da-kami: done

cnd/src/btsieve/ethereum.rs
```diff
@@ -6,14 +6,14 @@ use crate::{
...
+    ethereum::{Address, Bytes, Input, Log, Transaction, TransactionReceipt, U256},
 };
 use anyhow;
 use async_trait::async_trait;
 use chrono::NaiveDateTime;
 use genawaiter::{sync::Gen, GeneratorState};
 
-type Hash = H256;
+type Hash = crate::ethereum::Hash;
 type Block = crate::ethereum::Block;
```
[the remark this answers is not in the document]
da-kami: True, I left it because we basically still do the same for the Block. Originally it was to properly map the types so it is easier to read. Can clean this up for both and just import.

cnd/src/ethereum.rs
da-kami: oh, good catch, fixing :)

cnd/src/ethereum.rs
thomaseizinger: Likely, @da-kami just went with importing stuff through CLion and not writing those himself. Good catch :)

It is unfortunate that it does that :(

luckysori: Auto-importing with rust-analyzer in emacs can also cause similar issues :(

PR #2409, opened 2020-04-07
Check for redeem/refund spent outpoint using identity

### LINKED ISSUE #2270: Make the API for watching of spent outpoints on Bitcoin misuse resistant
## Problem

The current API for watching spent outpoints looks like this:

https://github.com/comit-network/comit-rs/blob/0cbf97d91edf318707b17c62c8b9c7fe7fa739c1/cnd/src/btsieve/bitcoin/mod.rs#L39-L44

Notice the `unlock_script` parameter?

This allows you to match on any part of the witness stack to determine, how a specific outpoint is spent. We use this to differentiate between redeemed and refunded HTLCs.

This is problematic:

1. With the planned move to miniscript in https://github.com/comit-network/blockchain-contracts/issues/23, we can no longer pass `1` or `0` here because the script generated by miniscript does not use a single byte on the witness stack to differentiate between redeem and refund. Instead if tries to match the signature directly and checks, whether it was succesful or not.

2. The API is depending on implementation details of our HTLCs which are defined in a completely different crate. To use the API correctly, you need to know, whether the redeem path is triggered with a 1 or 0. This creates an extremely high risk to be misused. Potentially, this is even a security problem if our script is malleable and an attacker can somehow sneak a 1 or a 0 in there. The bottom line is: Only relying on a `0` or a `1` is very bad.

This API is like that only for historical reasons: It used to be part of the btsieve binary, which didn't know about HTLCs. So we had to match on something that we certainly knew was going to be part of the stack.

## Goal

With btsieve being fully-integrated into cnd, we can now design the API in exactly the way we need it to be and make it a lot safer to use.

The goal is to make this API misuse resistant, that means: whatever we pass in there, it should be unambiguous about whether we want to match the refund or the redeem path of the script.

## Recommendation

Keep in mind that with miniscript, we don't have control over specific details of the script.

Here are some ideas how we could go about this:

1. Pass in a public key that we want to appear on the stack.

It should be safe to assume that public keys only appear on the stack if we want to verify a signature with them. Hence, if the refunder's public key appears on the stack, it is very likely that the HTLC was refunded.

rishflab:
Decide whether refund or redeem transaction using identity
rather than checking for a 1 or 0 on the witness stack.

Closes #2270

### DIFF
--- a/cnd/src/btsieve/bitcoin.rs
+++ b/cnd/src/btsieve/bitcoin.rs
@@ -5,11 +5,13 @@ pub use self::{
     bitcoind_connector::{BitcoindConnector, ChainInfo},
     cache::Cache,
 };
-use crate::btsieve::{
-    find_relevant_blocks, BlockByHash, BlockHash, LatestBlock, Predates, PreviousBlockHash,
+use crate::{
+    btsieve::{
+        find_relevant_blocks, BlockByHash, BlockHash, LatestBlock, Predates, PreviousBlockHash,
+    },
+    identity,
 };
 use bitcoin::{
-    blockdata::script::Instruction,
     consensus::{encode::deserialize, Decodable},
     BitcoinHash, OutPoint,
 };
@@ -40,7 +42,7 @@ pub async fn watch_for_spent_outpoint<C>(
     blockchain_connector: &C,
     start_of_swap: NaiveDateTime,
     from_outpoint: OutPoint,
-    unlock_script: Vec<Vec<u8>>,
+    identity: identity::Bitcoin,
 ) -> anyhow::Result<(bitcoin::Transaction, bitcoin::TxIn)>
 where
     C: LatestBlock<Block = Block> + BlockByHash<Block = Block, BlockHash = Hash>,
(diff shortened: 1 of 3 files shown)


cnd/src/swap_protocols/rfc003/bitcoin/htlc_events.rs
```diff
@@ -92,12 +92,14 @@ where
...
+        let (transaction, _) = watch_for_spent_outpoint(
+            self,
+            start_of_swap,
+            htlc_deployment.location,
+            htlc_params.redeem_identity,
+        )
+        .instrument(tracing::info_span!("htlc_redeemed"))
+        .await?;
 
         let secret = extract_secret(&transaction, &htlc_params.secret_hash)
```
rishflab, line 104: I should be able to remove this line as well

thomaseizinger: That would be good but I think that is out of scope of this ticket :)

If you have an idea, feel free to record it as an issue!
Thinking about it, I would say the best place for this function would actually be in `blockchain-contracts` because that crate knows, that `SHA256` is used an can therefore search for preimages in the witness stack that, when hashed, result in the given secret hash.

cnd/src/btsieve/bitcoin.rs
```diff
@@ -40,7 +42,7 @@ pub async fn watch_for_spent_outpoint<C>(
     blockchain_connector: &C,
     start_of_swap: NaiveDateTime,
     from_outpoint: OutPoint,
-    unlock_script: Vec<Vec<u8>>,
+    identity: identity::Bitcoin,
```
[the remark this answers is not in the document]
rishflab: yep sure

cnd/src/swap_protocols/rfc003/bitcoin/htlc_events.rs
rishflab: I don't think it is causing any problems so I don't think it is worth creating an issue to remove it. I just thought I could simply remote because the watch_outpoint call is already matching the spenders public key to decide whether it is a redeem or refund. The extract_secret check is redundant.

Issue #2290, opened 2020-03-20
Cnd did not crash due to address in use


bonomat:
# Problem

Cnd did try to bind the http server to an address which was already in use. It threw an error but did not crash.

# Goal 

If cnd can't start it's API due to `address in use`, it should not start

<details>
<summary>Old notes:</summary>

-----

# Old info

Yet again our e2e tests randomly failed: 
https://app.circleci.com/pipelines/github/comit-network/comit-rs/5779/workflows/af3bc470-a9cc-4c6a-8c16-61fff2a0f4cb/jobs/4878

```
$ cargo build --bin cnd && tsc
    Finished dev [unoptimized + debuginfo] target(s) in 0.22s
$ yarn dry && yarn e2e
$ jest --config jest.config-dry.js --maxWorkers=4
 FAIL  tests/dry/multiple_peers.ts (22.394s)
  ● Multiple peers tests › alice-sends-swap-request-to-bob-and-charlie

    connect ECONNREFUSED 0.0.0.0:41263



 PASS  tests/dry/peers_using_ip.ts (27.241s)
(node:16642) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGHUP listeners added to [process]. Use emitter.setMaxListeners() to increase limit
 PASS  tests/dry/lightning_routes.ts (5.876s)
 PASS  tests/dry/rfc003_schema.ts (29.642s)
 PASS  tests/dry/sanity.ts (31.475s)

Test Suites: 1 failed, 4 passed, 5 total
Tests:       1 failed, 21 passed, 22 total
Snapshots:   0 total
Time:        33.324s
Ran all test suites.
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Makefile:86: recipe for target 'e2e' failed
make: *** [e2e] Error 1

Exited with code exit status 2
CircleCI received exit code 2
```

# Goal

Time-box. Investigate. Fix.
If you have trouble finding an issues then stop after the time-box and keep it open for more reports (till stale-bot closes it).

</details>

bonomat:
Note: 

the log showed that cnd did not start for Bob: 

```
[...]

PR #2420, opened 2020-04-08
Exit cnd if HTTP port is in use

tcharding:
When starting cnd, if the HTTP port is in use, we would like to exit.  We can achieve this by checking if the port is available before trying to spawn the HTTP server.

This patch introduces a race condition between the call to `port_is_available()` and the actual spawn of the wrap server.  If we hit this race condition it is non-critical, it just means cnd will start without an HTTP server, the logs will still contain the bind error.

Fixes: #2290

### DIFF
--- a/cnd/src/main.rs
+++ b/cnd/src/main.rs
@@ -37,7 +37,7 @@ use cnd::{
 use rand::rngs::OsRng;
 use std::{process, sync::Arc};
 use structopt::StructOpt;
-use tokio::runtime;
+use tokio::{net::TcpListener, runtime};
 mod cli;
 mod trace;
 
@@ -165,10 +165,10 @@ fn main() -> anyhow::Result<()> {
     };
 
     runtime.block_on(load_swaps::load_swaps_from_database(deps.clone()))?;
-    runtime.spawn(spawn_warp_instance(settings, deps, facade2));
 
-    // Block the current thread.
-    ::std::thread::park();
+    let http_server_future = spawn_warp_instance(settings, deps, facade2);
+    runtime.block_on(http_server_future)?;
+
     Ok(())
 }
 
@@ -183,7 +183,11 @@ fn version() {
     println!("{} {} ({})", name, version, short);
 }
 
-async fn spawn_warp_instance(settings: Settings, dependencies: Facade, facade2: Facade2) {
+async fn spawn_warp_instance(
+    settings: Settings,
+    dependencies: Facade,
+    facade2: Facade2,
+) -> anyhow::Result<()> {
     let routes = route_factory::create(
         dependencies,
         facade2,
@@ -191,10 +195,12 @@ async fn spawn_warp_instance(settings: Settings, dependencies: Facade, facade2: 
     );
 
     let listen_addr = settings.http_api.socket;
+    let listener = TcpListener::bind(listen_addr).await?;
 
-    tracing::info!("Starting HTTP server on {}", listen_addr);
+    tracing::info!("Starting HTTP server on {} ...", listen_addr);
+    warp::serve(routes).serve_incoming(listener).await;
 
-    warp::serve(routes).bind(listen_addr).await
+    Ok(())
 }
 
 #[allow(clippy::print_stdout)] // We cannot use `log` before we have the config file
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -34,6 +34,7 @@ use crate::{
     },
     transaction,
 };
+use anyhow::Context;
 use async_trait::async_trait;
 use futures::{
     channel::oneshot::{self, Sender},
@@ -58,7 +59,7 @@ use std::{
     io,
     pin::Pin,
     sync::Arc,
-    task::{Context, Poll},
+    task::{self, Poll},
 };
 use tokio::{
     runtime::{Handle, Runtime},
@@ -114,8 +115,8 @@ impl Swarm {
             .build();
 
         for addr in settings.network.listen.clone() {
-            libp2p::Swarm::listen_on(&mut swarm, addr)
-                .expect("Could not listen on specified address");
+            libp2p::Swarm::listen_on(&mut swarm, addr.clone())
+                .with_context(|| format!("Address is not supported: {:?}", addr))?;
         }
 
         let swarm = Arc::new(Mutex::new(swarm));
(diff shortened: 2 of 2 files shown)


cnd/src/main.rs
```diff
@@ -142,6 +142,13 @@ fn main() -> anyhow::Result<()> {
...
 
     runtime.block_on(load_swaps::load_swaps_from_database(deps.clone()))?;
+
+    let listen_addr = settings.http_api.socket;
+    // This is racey but gives us some protection, if the race condition is hit it
+    // just means cnd will start even though the HTTP server did not come up.
+    if !port_is_available(listen_addr.port()) {
+        return Err(anyhow::anyhow!("HTTP port is unavailable: {}", listen_addr));
+    }
     runtime.spawn(spawn_warp_instance(settings, deps));
```
D4nte, line 152:
```suggestion
    runtime.block_on(spawn_warp_instance(settings, deps));
```
This one liner change is all you need. (and delete `::std::thread::park();`

tcharding: Cool, cheers.  Implemented as suggested.

cnd/src/main.rs
```diff
@@ -142,10 +142,15 @@ fn main() -> anyhow::Result<()> {
...
-    runtime.spawn(spawn_warp_instance(settings, deps));
 
-    // Block the current thread.
-    ::std::thread::park();
+    let listen_addr = settings.http_api.socket;
+    // This is racey but gives us some protection, if the race condition is hit it
+    // just means cnd will start even though the HTTP server did not come up.
+    if !port_is_available(listen_addr.port()) {
+        return Err(anyhow::anyhow!("HTTP port is unavailable: {}", listen_addr));
+    }
```
D4nte, line 151:
```suggestion
```

I don't think this is needed, the panic from warp is clear enough: `thread 'main' panicked at 'error binding to 0.0.0.0:8000: error creating server listener: Address already in use (os error 48)', <::std::macros::panic macros>:5:6
`

cnd/src/main.rs
```diff
@@ -160,14 +165,19 @@ fn version() {
...
 async fn spawn_warp_instance(settings: Settings, dependencies: Facade) {
     let routes = route_factory::create(dependencies, &settings.http_api.cors.allowed_origins);
-
     let listen_addr = settings.http_api.socket;
 
-    tracing::info!("Starting HTTP server on {}", listen_addr);
-
-    warp::serve(routes).bind(listen_addr).await
+    tracing::info!("Starting HTTP server on {} ...", listen_addr);
+    warp::serve(routes).try_bind(listen_addr).await
```
D4nte, line 180:
```suggestion
    warp::serve(routes).bind(listen_addr).await
```
`try_bind` returns a `Result`. I think we should keep `bind` so it panics.

cnd/src/main.rs
```diff
@@ -160,14 +165,19 @@ fn version() {
     println!("{} {} ({})", name, version, short);
 }
 
+fn port_is_available(port: u16) -> bool {
+    match TcpListener::bind(("127.0.0.1", port)) {
```
D4nte, line 169: I don't think this is entirely correct. A machine can have several interfaces (and support several ip protocols). In the config file, you can specify a full socket. If you were to have to check if a port is available you'd need to use the socket instead of assume it tries to bind to the loopback ipv4 address.

cnd/src/main.rs
thomaseizinger: We can return a result from `main` and avoid the panic by just shutting down the application like in any other error.
That would give us a nice error backtrace through anyhow instead of panicing.

I am in favor of `try_bind`.

thomaseizinger: Just make sure to handle the result then an bubble the error up in case it fails.

cnd/src/main.rs
tcharding: That is the whole point of this ticket, the panic from warp only panics the warp thread not the main thread so cnd continues to run.

cnd/src/main.rs
tcharding: `try_bind` does not return a result.

cnd/src/main.rs
tcharding: Good catch, you are indeed correct.

cnd/src/main.rs
tcharding:
``` warp server.rs

    /// Bind to a socket address, returning a `Future` that can be
    /// executed on any runtime.
    ///
    /// In case we are unable to bind to the specified address, resolves to an
    /// error and logs the reason.
    pub async fn try_bind(self, addr: impl Into<SocketAddr> + 'static) {
        let addr = addr.into();
        let srv = match try_bind!(self, &addr) {
            Ok((_, srv)) => srv,
            Err(err) => {
                log::error!("error binding to {}: {}", addr, err);
                return;
            }
        };

        srv.map(|result| {
            if let Err(err) = result {
                log::error!("server error: {}", err)
            }
        })
        .await;
    }
```

thomaseizinger:
I think going for https://docs.rs/warp/0.2.2/warp/struct.Server.html#method.serve_incoming is the best approach. You can pass the result of `tokio::net::TcpListener::bind` to it :)

This will allow us to bind to the socket ourselves, and we can fail if we cannot do that because it is taken.
Plus we get to keep the `runtime.spawn` aspect of just spawning away the warp future.

cnd/src/main.rs
D4nte: If you use `block_on` then the main thread panics too.

cnd/src/main.rs
thomaseizinger: Lol, why would it not return an error :roll_eyes:

tcharding: Stupid eh, especially confusing given the name.

tcharding:
> I think going for [docs.rs/warp/0.2.2/warp/struct.Server.html#method.serve_incoming](https://docs.rs/warp/0.2.2/warp/struct.Server.html#method.serve_incoming) is the best approach. You can pass the result of `tokio::net::TcpListener::bind` to it :)
> 
> This will allow us to bind to the socket ourselves, and we can fail if we cannot do that because it is taken.
> Plus we get to keep the `runtime.spawn` aspect of just spawning away the warp future.

Cool, I'll work this in if I can this afternoon.

bonomat:
Rebased for @tcharding as he will be on leave next week.

PR #2427, opened 2020-04-08
add matrix-os strategy to e2e job to run on mac-os

da-kami:
Added the e2e test for both OS similar to the build using matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,11 +39,6 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest]
-        include:
-          - os: ubuntu-latest
-            binary-suffix: ""
-          - os: macos-latest
-            binary-suffix: ""
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout sources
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -80,16 +80,13 @@ jobs:
         run: make test
 
       - name: Upload cnd ${{ matrix.os }} binary
-        if: matrix.os == 'ubuntu-latest'
         uses: actions/upload-artifact@v1
         with:
           name: cnd
           path: target/debug/cnd
```
bonomat, line 86: You will need to add an identifier here. 
Github action will upload a binary called `cnd` for both now, macOs and Linux. This means it can't work because the slower one will overwrite the faster one. 

You will also need to download the correct one after wards.

.github/workflows/ci.yml
```diff
@@ -80,16 +80,13 @@ jobs:
         run: make test
 
       - name: Upload cnd ${{ matrix.os }} binary
-        if: matrix.os == 'ubuntu-latest'
         uses: actions/upload-artifact@v1
         with:
           name: cnd
```
bonomat, line 85:
```suggestion
          name: cnd-${{ matrix.os }}
```

.github/workflows/ci.yml
da-kami: thanks! missed that :)

da-kami:
Weirdly circleci run just failed because cnd could not be compiled. 
Don't think that can be related to the changes I made, I re-triggered circleci.

.github/workflows/ci.yml
```diff
@@ -79,25 +79,33 @@ jobs:
         if: matrix.os != 'macos-latest'
         run: make test
 
+      # Store a cnd ${{ matrix.os }} archive that contains the cnd binary
```
bonomat, line 82: Do we need this comment? 
It literally says the same thing one line below 😅

da-kami: I added it because I confused myself earlier because I through that the artefact would be the cnd binary, but the artefact is a zip that contains the cnd binary (actually being named `cnd`). If you feel that is clear anyway then I can remove it.

.github/workflows/ci.yml
```diff
@@ -79,25 +79,33 @@ jobs:
...
 
-
   e2e_test:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest]
+        include:
+          - os: ubuntu-latest
+            binary-suffix: ""
```
thomaseizinger, line 95: I don't think we are using this binary-suffix for anything. That seems to be a copy-pasta artifact from create-comit-app.
We also don't need the `include` part because the matrix above describes all the combinations already.

thomaseizinger: I think the same applies for some other job further above.

.github/workflows/ci.yml
thomaseizinger: You could just update the step name to include the fact that it is an archive :)

da-kami: done

.github/workflows/ci.yml
da-kami: done

Issue #2177, opened 2020-03-06
Define the DB Schema for `finalize`d LN Swaps


D4nte:
## Problem

We learn new information about the other party during the communication phase and before the execution phase, concretely, identities and in the case of Bob, also the secret hash.

This information needs to be persisted in the database.

## Goal

Design a DB schema for the information we learn during the communication phase.
Aim for the design to be reusable across several swaps.

## Out of scope

The DB schema for `created`d swap should already be done in https://github.com/comit-network/comit-rs/issues/2173
Saving and retrieving from the new tables is not part of this issue.

## Recommendation

KISS.

Keep in mind that this issue was unblocked from https://github.com/comit-network/comit-rs/issues/2173 with the idea that we use separate tables for this.

tcharding:
Why is this blocked on https://app.zenhub.com/workspace/o/comit-network/comit-rs/issues/2170?

D4nte:
@luckysori to create ticket that tracks the "swap spawn" on the core for HALight. Such ticket will block this ticket as it will define the data that we need for HALight (and hence, this is the data to be saved in the DB).

Please ping @tcharding once such ticket is created so he can update his dependencies/tickets.

thomaseizinger:
Removing the blocker on https://github.com/comit-network/comit-rs/issues/2173 because we can design these tables independently from the one where the created swaps are stored.

tcharding:
At some stage the `start_of_swap` will most likely need to be created by the database i.e., with a table row: `at DATETIME DEFAULT CURRENT_TIMESTAMP` in either of the tables for 'created' or 'finalized' swaps.  Adding comment here to keep track of it.

`start_of_swap` is mentioned here: https://github.com/comit-network/comit-rs/issues/2408

And this should probably be either done in this ticket or in the [created](https://github.com/comit-network/comit-rs/issues/2173) side.

thomaseizinger:
Obsoleted by https://github.com/comit-network/comit-rs/issues/2485.

PR #2431, opened 2020-04-09
Define the DB schema for finalized swaps

tcharding:
Add a table to hold the data we learn from the communication protocols:

```
table! {
    finalized_swaps {
        id -> Integer,
        swap_id -> Text,
        alpha_identity -> Text,
        beta_identity -> Text,
        secret_hash -> Text,
    }
}
```

Create new migrations that include this table.

Choose to include the `secret_hash` even though for Alice this is known before the communication protocols start.  This keeps things symmetrical for Alice and Bob.  `secret_hash` may well appear in the 'created' swap table also (yet to be designed).  Having an extra row with the same data in two tables seems ok in favour of the added simplicity this enables.   Anyone think differently?

Resolves: #2177

### DIFF
--- a/cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
+++ b/cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
@@ -6,5 +6,6 @@ CREATE TABLE finalized_swaps
     swap_id UNIQUE      NOT NULL,
     alpha_identity      NOT NULL,
     beta_identity,      NOT NULL,
-    secret_hash         NOT NULL
+    secret_hash         NOT NULL,
+    at DATETIME DEFAULT CURRENT_TIMESTAMP
 )
--- a/cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
+++ b/cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
@@ -1,0 +1,10 @@
+-- Your SQL goes here
+
+CREATE TABLE finalized_swaps
+(
+    id INTEGER          NOT NULL PRIMARY KEY,
+    swap_id UNIQUE      NOT NULL,
+    aplpha_identity     NOT NULL,
+    beta_identity,      NOT NULL,
+    secret_hash         NOT NULL
+)
--- a/cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
+++ b/cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
@@ -4,7 +4,7 @@ CREATE TABLE finalized_swaps
 (
     id INTEGER          NOT NULL PRIMARY KEY,
     swap_id UNIQUE      NOT NULL,
-    aplpha_identity     NOT NULL,
+    alpha_identity      NOT NULL,
     beta_identity,      NOT NULL,
     secret_hash         NOT NULL
 )
--- a/cnd/src/db/schema.rs
+++ b/cnd/src/db/schema.rs
@@ -112,5 +112,6 @@ table! {
         alpha_identity -> Text,
         beta_identity -> Text,
         secret_hash -> Text,
+       at -> Timestamp,
     }
 }
--- a/cnd/src/db/schema.rs
+++ b/cnd/src/db/schema.rs
@@ -104,3 +104,13 @@ table! {
        counterparty -> Text,
    }
 }
+
+table! {
+    finalized_swaps {
+        id -> Integer,
+        swap_id -> Text,
+        alpha_identity -> Text,
+        beta_identity -> Text,
+        secret_hash -> Text,
+    }
+}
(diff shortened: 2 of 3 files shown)


D4nte:
> Add a table to hold the data we learn from the communication protocols:
> 
> ```
> table! {
>     finalized_swaps {
>         id -> Integer,
>         swap_id -> Text,
>         alpha_identity -> Text,
>         beta_identity -> Text,
>         secret_hash -> Text,
>     }
> }
> ```
> 
> Create new migrations that include this table.
> 
> Choose to include the `secret_hash` even though for Alice this is known before the communication protocols start. This keeps things symmetrical for Alice and Bob. `secret_hash` may well appear in the 'created' swap table also (yet to be designed). Having an extra row with the same data in two tables seems ok in favour of the added simplicity this enables. Anyone think differently?
> 
> Resolves: #2177

I don't think secret hash will appear in the created table so I think it's fine.

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
```diff
@@ -0,0 +1,110 @@
+-- Your SQL goes here
+
+
+CREATE TABLE rfc003_bitcoin_ethereum_bitcoin_ether_request_messages
```
D4nte, line 4: Why is this migration creating tables already created in a previous migration? I would expect the new migration to only create tables introduced with this patch.

Same for `down.sql`. The migration should only apply new changes, not re-create the whole DB from scratch.

If we apply this new migration to an existing DB, it'd fail as it'd try to create tables that already exist.

If we fallback this new migration to an existing DB, all data, even the one in the old table would be lost.

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
```diff
@@ -0,0 +1,110 @@
...
+    swap_id UNIQUE 	NOT NULL,
+    role 		NOT NULL,
+    counterparty 	NOT NULL
+);
+
+CREATE TABLE finalized_swaps
+(
+    id INTEGER          NOT NULL PRIMARY KEY,
+    swap_id UNIQUE      NOT NULL,
+    aplpha_identity     NOT NULL,
```
D4nte, line 107:
```suggestion
    alpha_identity     NOT NULL,
```

D4nte: The identities are those from the counterpart. I think it would be beneficial if they were tagged as so.

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
```diff
@@ -0,0 +1,110 @@
...
+    id INTEGER     	NOT NULL PRIMARY KEY,
+    swap_id UNIQUE 	NOT NULL,
+    role 		NOT NULL,
+    counterparty 	NOT NULL
+);
+
+CREATE TABLE finalized_swaps
+(
+    id INTEGER          NOT NULL PRIMARY KEY,
+    swap_id UNIQUE      NOT NULL,
```
D4nte, line 106: Which one is this swap id? The local or shared one? Either way, I think both should be stored.

cnd/src/db/schema.rs
```diff
@@ -104,3 +104,13 @@ table! {
        counterparty -> Text,
    }
 }
+
+table! {
+    finalized_swaps {
+        id -> Integer,
```
D4nte, line 110: Do we need a `finalized_at` timestamp? 🤔

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
tcharding: Oh thanks, I clearly do not understand how migrations work :)  Will fix

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
tcharding: I'm working on the 'created' side as well.  Please review that one when its ready and lets discuss this over there because I already add a comment dealing with this to `schema.rs`.  Otherwise the conversation is going to be broken/spilt over the two PRs.  Thanks

cnd/src/db/schema.rs
tcharding: Yes, timestamp needs to be done _somewhere_.  I think you may be right, finalized is the place for it not created.  Will add.

tcharding:
Thanks for the review @D4nte, I accidentally pushed a commit from another PR (create side) onto this so had to force push to remove it.  Patches to implement your suggestions are still separate on top of the original work.

tcharding:
Closed in favour of: https://github.com/comit-network/comit-rs/pull/2434

tcharding:
Re-opening after discussion with @D4nte

Issue #2173, opened 2020-03-05
Define the DB Schema for `create`d LN Swaps


D4nte:
## Problem

It is not possible to replicate the existing DB schema model used for rfc003. The main reason is that the messaging, and hence what is stores, differs for HALight. There are no more `swap request`, `accept message` or `decline message`. Instead, the 2 states to save in the DB for a swap are after `create` and `finalize` steps. See https://github.com/comit-network/comit-rs/issues/2134 for details.

This blocked by the REST API because it we want to keep some uniformity across the data types to make our life easier.
To design a data type, we have restrictions from each API.
I believe that the REST API would bring in the most restrictions, hence it would be smoother to drive the design from the REST API to then cascade it to the DB.

Coming with various design in parallel is likely to create unnecessary discrepancies.

## Goal

Design new DB schemas for `create`d HALight swaps. The DB schema must take in account:
- The [new data types](https://github.com/comit-network/comit-rs/issues/2158) that will need to be stored in the DB
- The fact that we want to be able to retrieve swaps which are `create`d but not yet `finalize`d
- Get consensus a name for a "created" swap

The DB Schemas should be defined in diesel and tables created when starting cnd, however, the saving will be done at a later stage (step 3 of https://github.com/comit-network/comit-rs/issues/2134#issuecomment-594308787)

## Out of scope

The DB schema for `finalize`d swap will be done in https://github.com/comit-network/comit-rs/issues/2177.

## Recommendation

KISS.

We want to support han(ether)-halight and halight-herc20 swaps at first. Make sure the database schema can support that. Consider a table design that is split into protocols similar to the body of the API request. This would/should allow composing the tables together for different combinations of swaps.

thomaseizinger:
Please also take into account that approach taken to define those schemas should be expandable to our current swaps as we will migrate the old stuff at some point.

thomaseizinger:
Removing the dependency on https://github.com/comit-network/comit-rs/issues/2158 because the first endpoint was implemented with the walking skeleton and can therefore act as an example.

tcharding:
At some stage the `start_of_swap` will most likely need to be created by the data base i.e., `at DATETIME DEFAULT CURRENT_TIMESTAMP`.  Adding comment here to keep track of it.

`start_of_swap` is mentioned here: https://github.com/comit-network/comit-rs/issues/2408

And this should probably be either done in this ticket or in the [finalized](https://github.com/comit-network/comit-rs/issues/2177) side.

[...]

PR #2433, opened 2020-04-09
Add 'create swaps' tables for the protocol pairs 

tcharding:
We need a way to persist data for created swaps.  We would like to be able to eventually restart swaps so we need everything that was handed to us by the user via the REST API.  Also, we will at some stage need to be able to link the data in 'created' swaps with the data for 'finalized' swaps, here we do this by storing the `local_swap_id`.

Add 'create swaps' tables for each of the protocol pairs we support.  Also add migrations.

This [closed PR](https://github.com/comit-network/comit-rs/pull/2432/files) shows another way of solving this same ticket.  We are currently unsure which method is best.

Fixes: #2173

### DIFF
--- a/cnd/migrations/2020-04-09-030124_create-message-tables/down.sql
+++ b/cnd/migrations/2020-04-09-030124_create-message-tables/down.sql
@@ -1,3 +1,8 @@
 -- This file should undo anything in `up.sql`
 
+DROP TABLE han_ethereum_created_swaps;
+DROP TABLE han_bitcoin_created_swaps;
+DROP TABLE herc20_created_swaps;
+DROP TABLE halight_created_swaps;
+DROP TABLE finalized_swaps;
 DROP TABLE finalized_swaps;
--- a/cnd/migrations/2020-04-09-030124_create-message-tables/down.sql
+++ b/cnd/migrations/2020-04-09-030124_create-message-tables/down.sql
@@ -1,0 +1,3 @@
+-- This file should undo anything in `up.sql`
+
+DROP TABLE finalized_swaps;
--- a/cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
+++ b/cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
@@ -1,0 +1,11 @@
+-- Your SQL goes here
+
+CREATE TABLE finalized_swaps
+(
+    id INTEGER          NOT NULL PRIMARY KEY,
+    swap_id UNIQUE      NOT NULL,
+    alpha_identity      NOT NULL,
+    beta_identity,      NOT NULL,
+    secret_hash         NOT NULL,
+    at DATETIME DEFAULT CURRENT_TIMESTAMP
+)
--- a/cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
+++ b/cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
@@ -1,11 +1,64 @@
 -- Your SQL goes here
 
+CREATE TABLE dial_infos
+(
+        id             INTEGER NOT NULL PRIMARY KEY,
+        local_swap_id  UNIQUE NOT NULL,
+        peer_id        NOT NULL
+);
+
+CREATE TABLE han_ethereum_created_swaps
+(
+        id             INTEGER NOT NULL PRIMARY KEY,
+        local_swap_id  UNIQUE NOT NULL,
+        role           NOT NULL,
+        chain_id       NOT NULL,
+        my_identity    NOT NULL,
+        expiry         NOT NULL,
+        amount         NOT NULL
+);
+
+CREATE TABLE han_bitcoin_created_swaps
+(
+        id             INTEGER NOT NULL PRIMARY KEY,
+        local_swap_id  UNIQUE NOT NULL,
+        role           NOT NULL,
+        network        NOT NULL,
+        my_identity    NOT NULL,
+        expiry         NOT NULL,
+        amount         NOT NULL
+);
+
+
+CREATE TABLE herc20_created_swaps
+(
+        id             INTEGER NOT NULL PRIMARY KEY,
+        local_swap_id  UNIQUE NOT NULL,
+        role           NOT NULL,
+        chain_id       NOT NULL,
+        my_identity    NOT NULL,
+        expiry         NOT NULL,
+        amount         NOT NULL,
+        token_contract NOT NULL
+);
+
+CREATE TABLE halight_created_swaps
+(
+        id             INTEGER NOT NULL PRIMARY KEY,
+        local_swap_id  UNIQUE NOT NULL,
+        role           NOT NULL,
+        network        NOT NULL,
+        my_identity    NOT NULL,
+        expiry         NOT NULL,
+        amount         NOT NULL
+);
+
 CREATE TABLE finalized_swaps
 (
-    id INTEGER          NOT NULL PRIMARY KEY,
-    swap_id UNIQUE      NOT NULL,
+    id                  INTEGER NOT NULL PRIMARY KEY,
+    swap_id             UNIQUE NOT NULL,
     alpha_identity      NOT NULL,
-    beta_identity,      NOT NULL,
+    beta_identity       NOT NULL,
     secret_hash         NOT NULL,
-    at DATETIME DEFAULT CURRENT_TIMESTAMP
-)
+    finalized_at        DATETIME DEFAULT CURRENT_TIMESTAMP
+);
--- a/cnd/src/db/schema.rs
+++ b/cnd/src/db/schema.rs
@@ -1,106 +1,181 @@
 table! {
-   rfc003_bitcoin_ethereum_bitcoin_ether_request_messages {
-       id -> Integer,
-       swap_id -> Text,
-       bitcoin_network -> Text,
-       ethereum_chain_id -> BigInt,
-       bitcoin_amount -> Text,
-       ether_amount -> Text,
-       hash_function -> Text,
-       bitcoin_refund_identity -> Text,
-       ethereum_redeem_identity -> Text,
-       bitcoin_expiry -> BigInt,
-       ethereum_expiry -> BigInt,
-       secret_hash -> Text,
-   }
+    rfc003_bitcoin_ethereum_bitcoin_ether_request_messages {
+        id -> Integer,
+        swap_id -> Text,
+        bitcoin_network -> Text,
+        ethereum_chain_id -> BigInt,
+        bitcoin_amount -> Text,
+        ether_amount -> Text,
+        hash_function -> Text,
+        bitcoin_refund_identity -> Text,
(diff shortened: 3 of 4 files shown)


tcharding:
Feel free to ignore this, I just read the 'recomendation' ...

cnd/src/db/schema.rs
```diff
@@ -104,3 +104,30 @@ table! {
        counterparty -> Text,
    }
 }
+
+// We need a way to tie the 'created' swaps with the 'finalized' swaps, possible
+// methods include:
+//
+// - Add a `local_id` row to the finalized swaps table then use an SQL join to
```
D4nte, line 111: I think `local_swap_id` should be in both tables and used for a join.

cnd/src/db/schema.rs
```diff
@@ -104,3 +104,30 @@ table! {
...
+table! {
+    han_ethereum_ether_halight_lightning_bitcoin_created_swaps {
+        id -> Integer,
+        role -> Text,
+        dial_information -> Text,
+        ethereum_chain_id -> BigInt,
+        bitcoin_network -> Text,
+        ether_amount -> Text,
+        bitcoin_amount -> Text,
+        alpha_identity -> Text,
```
D4nte, line 126: I find it strange to use `alpha` here when everything else is `bitcoin` or `ethereum`.

cnd/src/db/schema.rs
```diff
@@ -104,3 +104,30 @@ table! {
...
+        beta_identity -> Text,
+        secret_hash -> Text,
```
D4nte, line 128: Bob does not know the secret hash just yet

cnd/src/db/schema.rs
```diff
@@ -104,3 +104,30 @@ table! {
...
+        ethereum_expiry -> BigInt,
+        bitcoin_expiry -> BigInt, // FIXME: Is this correct?
```
D4nte, line 130: I think `lightning_cltv_expiry` or `beta_cltv_expiry` would be more precise.

cnd/src/db/schema.rs
```diff
@@ -104,3 +104,30 @@ table! {
...
+        local_id -> Text,
```
D4nte, line 131: Is that the local swap id? if so, please specify it because in this domain `id` is an integer so it is strange to have two fields with same noun in the name but different types.

tcharding:
Closed in favour of: https://github.com/comit-network/comit-rs/pull/2434

tcharding:
Re-opening after discussion with @D4nte.

cnd/src/db/schema.rs
tcharding: Done as suggested.

cnd/migrations/2020-04-09-052607_create-message-tables/up.sql
```diff
@@ -0,0 +1,133 @@
+-- Your SQL goes here
+
+CREATE TABLE han_ethereum_ether_han_bitcoin_bitcoin_created_swaps
+(
+    id INTEGER               NOT NULL PRIMARY KEY,
+    local_swap_id            UNIQUE NOT NULL,
+    role                     NOT NULL,
+    dial_information         NOT NULL,
```
thomaseizinger, line 8: Is the peer id or the address-hint? I cannot be both.

tcharding: Yes, sloppy work by me.  Any preference, we can have both as optional fields?

thomaseizinger: I think we don't need to save the address_hint at all :)

We should save the peer ID because it tells us who we are swapping with.

tcharding: Cool, worked this in.

thomaseizinger:
I feel like it is actually crucial to properly decide this **now** before we commit to a schema, so I would like us to invest some more time and write up, what the pros and cons of the different approaches are.

cnd/src/db/schema.rs
```diff
@@ -1,106 +1,179 @@
...
+        counterparty -> Text,
+    }
+}
+
+// The new split protocol tables i.e., Han, HErc20, HALight.
+//
+
+// This is used by Alice to save dial info for created swaps.
+table! {
+    dial_infos {
```
thomaseizinger, line 113: Not sure if we had this before like that but we could name this one swaps and include the role as-well. Then the individual tables would only contain ledger specific stuff.

cnd/src/db/schema.rs
```diff
@@ -1,106 +1,179 @@
...
+        expiry -> BigInt,
+        amount -> Text,
+    }
+}
+
+table! {
+    finalized_swaps {
+        id -> Integer,
+        local_swap_id -> Text,
+        shared_swap_id -> Text,
```
thomaseizinger, line 173: I like the local/shared notion!

tcharding: Good huh, found it by accident.

cnd/migrations/2020-04-09-030124_create-message-tables/down.sql
```diff
@@ -0,0 +1,8 @@
+-- This file should undo anything in `up.sql`
+
+DROP TABLE han_ethereum_created_swaps;
+DROP TABLE han_bitcoin_created_swaps;
+DROP TABLE herc20_created_swaps;
+DROP TABLE halight_created_swaps;
+DROP TABLE finalized_swaps;
+DROP TABLE finalized_swaps;
```
thomaseizinger, line 8: Duplicate.
```suggestion
```

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
```diff
@@ -0,0 +1,64 @@
+-- Your SQL goes here
+
+CREATE TABLE dial_infos
```
thomaseizinger, line 3: I would name this one `swaps` and move the role to this table.

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+(
+        id             INTEGER NOT NULL PRIMARY KEY,
+        local_swap_id  UNIQUE NOT NULL,
+        peer_id        NOT NULL
+);
+
+CREATE TABLE han_ethereum_created_swaps
+(
+        id             INTEGER NOT NULL PRIMARY KEY,
+        local_swap_id  UNIQUE NOT NULL,
```
thomaseizinger, line 13: Instead of doing this, we could use a foreign key to the `swaps` table.

cnd/src/db/schema.rs
tcharding: Nice, done.

tcharding:
> I like this but I think it is missing one crucial piece of information:
> 
> We need to track, which ledger is alpha and which is beta :)

Poo

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
```diff
@@ -0,0 +1,64 @@
+-- Your SQL goes here
+
+CREATE TABLE dial_infos
+(
+        id             INTEGER NOT NULL PRIMARY KEY,
+        local_swap_id  UNIQUE NOT NULL,
+        peer_id        NOT NULL
+);
+
+CREATE TABLE han_ethereum_created_swaps
```
thomaseizinger, line 10: We need to somehow track, whether the information in these tables is alpha or beta ledger.

We could:

- have a field in each of these tables that says alpha/beta
- have a field in the `swaps` table proposed above that says the protocol name for alpha/beta
- more options?

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+        amount         NOT NULL,
+        token_contract NOT NULL
+);
+
+CREATE TABLE halight_created_swaps
+(
+        id             INTEGER NOT NULL PRIMARY KEY,
+        local_swap_id  UNIQUE NOT NULL,
+        role           NOT NULL,
+        network        NOT NULL,
```
thomaseizinger, line 50: Theoretically, `halight` has network and chain. Should we track this here accordingly?

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+);
+
+CREATE TABLE finalized_swaps
+(
+    id                  INTEGER NOT NULL PRIMARY KEY,
+    swap_id             UNIQUE NOT NULL,
+    alpha_identity      NOT NULL,
+    beta_identity       NOT NULL,
+    secret_hash         NOT NULL,
+    finalized_at        DATETIME DEFAULT CURRENT_TIMESTAMP
```
thomaseizinger, line 63: I know I suggested to do this at some point but I think it would be better if we would generate the timestamp in the code. Once we extract COMIT as a library, the timestamp should be created in there and not in the database I feel.

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+        network        NOT NULL,
+        my_identity    NOT NULL,
+        expiry         NOT NULL,
+        amount         NOT NULL
+);
+
+CREATE TABLE finalized_swaps
+(
+    id                  INTEGER NOT NULL PRIMARY KEY,
+    swap_id             UNIQUE NOT NULL,
```
thomaseizinger, line 59:
```suggestion
    shared_swap_id             UNIQUE NOT NULL,
```

cnd/migrations/2020-04-09-030124_create-message-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+        id             INTEGER NOT NULL PRIMARY KEY,
+        local_swap_id  UNIQUE NOT NULL,
+        role           NOT NULL,
+        network        NOT NULL,
+        my_identity    NOT NULL,
+        expiry         NOT NULL,
+        amount         NOT NULL
+);
+
+CREATE TABLE finalized_swaps
```
thomaseizinger, line 56: This table should either contain `local_swap_id` or have a FK to the `swaps` table.

cnd/src/db/schema.rs
```diff
@@ -1,106 +1,179 @@
...
+        role -> Text,
+        counterparty -> Text,
+    }
+}
+
+// The new split protocol tables i.e., Han, HErc20, HALight.
+//
+
+// This is used by Alice to save dial info for created swaps.
+table! {
```
thomaseizinger, line 112: I think these are outdated when it comes to the migrations above ..

tcharding:
Made redundant by https://github.com/comit-network/comit-rs/issues/2485

Issue #2389, opened 2020-04-06
Bob to ensure that an incoming swap comes from the expected counterparty


D4nte:
## Problem

When creating a swap on Bob, we pass the dial information (peer id and address hint) of Alice in the http request.
However, at no point do we check whether the announcement for the given swap indeed comes from the given peer id. Only the digests are used to match, there is no verification of who is sending the announcement.
Them the one-shot messages are sent to the peer id that did the announcement, and not to the peer id present in the create swap http request.

## Goal

Reject the swap if it does not come from the expected party.
While I do not believe this is a day one feature, it is security feature in case Bob has KYC'd Alice or provided a preferential rate, etc.

Make sure to add an e2e test for this!

## Note

- Remove the inclusion of the address hint on Bob side when creating the swap.
- Consider returning an error if a swap is created for Bob with an address hint.

PR #2434, opened 2020-04-09
Add SQL tables for split protocols - 10%

tcharding:
New Draft PR to combine both the 'create' and 'finalize' tickets - this is going to have a lot of comments and will be difficult to follow having two PRs.  Thanks for your reviews so far @D4nte I believe I have included your comments in this work, but lets see.

This PR uses a totally different approach, adds a table for each protocol (Han, HErc20, HALight).  Also adds other tables for data that is known to Alice at a different stage than to Bob i.e., Alice knows the secret hash after the swap is created and before it is finalized whereas Bob only knows it after the swap is finalized. 

Resolves: #2177
Resolves: #2173

### DIFF
--- a/cnd/src/db/schema.rs
+++ b/cnd/src/db/schema.rs
@@ -1,106 +1,181 @@
 table! {
-   rfc003_bitcoin_ethereum_bitcoin_ether_request_messages {
-       id -> Integer,
-       swap_id -> Text,
-       bitcoin_network -> Text,
-       ethereum_chain_id -> BigInt,
-       bitcoin_amount -> Text,
-       ether_amount -> Text,
-       hash_function -> Text,
-       bitcoin_refund_identity -> Text,
-       ethereum_redeem_identity -> Text,
-       bitcoin_expiry -> BigInt,
-       ethereum_expiry -> BigInt,
-       secret_hash -> Text,
-   }
+    rfc003_bitcoin_ethereum_bitcoin_ether_request_messages {
+        id -> Integer,
+        swap_id -> Text,
+        bitcoin_network -> Text,
+        ethereum_chain_id -> BigInt,
+        bitcoin_amount -> Text,
+        ether_amount -> Text,
+        hash_function -> Text,
+        bitcoin_refund_identity -> Text,
+        ethereum_redeem_identity -> Text,
+        bitcoin_expiry -> BigInt,
+        ethereum_expiry -> BigInt,
+        secret_hash -> Text,
+    }
 }
 
 table! {
-   rfc003_ethereum_bitcoin_ether_bitcoin_request_messages {
-       id -> Integer,
-       swap_id -> Text,
-       ethereum_chain_id -> BigInt,
-       bitcoin_network -> Text,
-       ether_amount -> Text,
-       bitcoin_amount -> Text,
-       hash_function -> Text,
-       ethereum_refund_identity -> Text,
-       bitcoin_redeem_identity -> Text,
-       ethereum_expiry -> BigInt,
-       bitcoin_expiry -> BigInt,
-       secret_hash -> Text,
-   }
+    rfc003_ethereum_bitcoin_ether_bitcoin_request_messages {
+        id -> Integer,
+        swap_id -> Text,
+        ethereum_chain_id -> BigInt,
+        bitcoin_network -> Text,
+        ether_amount -> Text,
+        bitcoin_amount -> Text,
+        hash_function -> Text,
+        ethereum_refund_identity -> Text,
+        bitcoin_redeem_identity -> Text,
+        ethereum_expiry -> BigInt,
+        bitcoin_expiry -> BigInt,
+        secret_hash -> Text,
+    }
 }
 
 table! {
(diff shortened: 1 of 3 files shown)


cnd/src/db/schema.rs
```diff
@@ -1,106 +1,181 @@
...
+        expiry -> BigInt,
+        amount -> Text,
+    }
+}
+
+table! {
+    herc20_created_swaps {
+        id -> Integer,
+        local_swap_id -> Text,
+        network -> Text,            // i.e., 'regtest' for Bitcoin or '17' for Ethereum.  FIXME: Is this ok?
```
thomaseizinger, line 152: There is no `herc20` on Bitcoin?!

cnd/src/db/schema.rs
```diff
@@ -1,106 +1,181 @@
...
+        amount -> Text,
+        token_contract -> Text,
+    }
+}
+
+table! {
+    halight_created_swaps {
+        id -> Integer,
+        local_swap_id -> Text,
+        network -> Text,            // i.e., 'regtest' for Bitcoin or '17' for Ethereum.  FIXME: Is this ok?
```
thomaseizinger, line 164: HAlight doesn't work on Ethereum.

cnd/src/db/schema.rs
```diff
@@ -1,106 +1,181 @@
...
+table! {
+    secret_hashes {
+        id -> Integer,
+        local_swap_id -> Text,
+        secret_hash -> Text,
+    }
+}
+
+table! {
+    han_created_swaps {
```
thomaseizinger, line 138: If we add the role to this then the `created_swaps` table is obsolete, no?

cnd/src/db/schema.rs
```diff
@@ -1,106 +1,181 @@
...
+
+table! {
+    created_swaps {
+        id -> Integer,
+        local_swap_id -> Text,
+        role -> Text,
+    }
+}
+
+// This is used by Alice to save dial info for created swaps.
```
thomaseizinger, line 119: Bob also has dialinformation on the API because he is going to need it to verify that the swap came from Alice. See https://github.com/comit-network/comit-rs/issues/2389.

cnd/src/db/schema.rs
```diff
@@ -1,106 +1,181 @@
...
+table! {
+    dial_infos {
+        id -> Integer,
+        local_swap_id -> Text,
+        dial_information -> Text,
+    }
+}
+// Alice can write this soon as she creates the swap, Bob can write it
```
thomaseizinger, line 127: Alice doesn't ever need to write this. She can always derive the secret from the swap id.

We can still store it for symmetric reasons but I am not sure that is a good idea :man_shrugging:

cnd/src/db/schema.rs
```diff
@@ -1,106 +1,181 @@
...
+        token_contract -> Text,
+    }
+}
+
+table! {
+    halight_created_swaps {
+        id -> Integer,
+        local_swap_id -> Text,
+        network -> Text,            // i.e., 'regtest' for Bitcoin or '17' for Ethereum.  FIXME: Is this ok?
+        identity -> Text,
```
thomaseizinger, line 165: Should we make this explicit?

```suggestion
        my_identity -> Text,
```

cnd/src/db/schema.rs
```diff
@@ -1,106 +1,181 @@
...
+        amount -> Text,
+    }
+}
+
+table! {
+    finalized_swaps {
+        id -> Integer,
+        local_id -> Text,
+        shared_swap_id -> Text,
+        alpha_redeem_identity -> Text,
```
thomaseizinger, line 176: :mega: This notion only makes sense for Alice. Bob is learning the `alpha_refund_identity` during the communication phase.

cnd/src/db/schema.rs
thomaseizinger, line 138: I think having han_ethereum and han_bitcoin tables would be better.

tcharding:
Thanks for your review @thomaseizinger, I just had a chat wit @D4nte and decided to go back to _not_ having han/herc20 split tables.  Closing this and re-opening the other.

cnd/src/db/schema.rs
tcharding: Yep, nice.

cnd/src/db/schema.rs
tcharding: Removed it in favour of having it in finalized_swaps table.

cnd/src/db/schema.rs
tcharding: Yes, nice.  When done with @D4nte's suggestion of using `counterparty_alpha_identity` in finalized_swaps table, this works well.

cnd/src/db/schema.rs
tcharding: I think this is out of date, all instances of 'redeem' and 'refund' removed.

cnd/src/db/schema.rs
tcharding: Added, this clears up the one remaining comment on `network` too, great.

PR #2439, opened 2020-04-10
BitcoindWallet and upgrade bitcoind to 0.19.1

bonomat:
This is a potential fix for this bug in the sdk : comit-network/comit-js-sdk#177.
In order to make this work I had to upgrade bitcoind to 0.19.1

If we are happy with this wallet I propose to: 

1) move the `BitcoindWallet` into the sdk, i.e. fix the original issue comit-network/comit-js-sdk#177
2) release the sdk and 
  a) use the latest sdk in the e2e tests (i.e. remove the `BitcoindWallet` and use the one from the sdk)
  b) use the latest sdk version in the examples. For that, we will need to upgrade bitcoind in create-comit-app

### DIFF
--- a/api_tests/src/bitcoin_miner.ts
+++ b/api_tests/src/bitcoin_miner.ts
@@ -18,14 +18,15 @@ async function run(configFile: string) {
         port: config.rpcPort,
         username: config.username,
         password: config.password,
+        wallet: "miner",
     });
 
     // only coins after the first 101 are spendable
-    await client.generateToAddress(101, await client.getNewAddress());
+    const address = await client.getNewAddress();
+    await client.generateToAddress(101, address);
 
     while (true) {
-        await client.generateToAddress(1, await client.getNewAddress());
-
+        await client.generateToAddress(1, address);
         await sleep(1000);
     }
 }
--- a/api_tests/src/ledgers/bitcoind_instance.ts
+++ b/api_tests/src/ledgers/bitcoind_instance.ts
@@ -42,7 +42,7 @@ export class BitcoindInstance implements LedgerInstance {
     ) {}
 
     public async start() {
-        const bin = await this.findBinary("0.17.0");
+        const bin = await this.findBinary("0.19.1");
 
         this.logger.info("Using binary", bin);
 
--- a/api_tests/src/ledgers/bitcoind_instance.ts
+++ b/api_tests/src/ledgers/bitcoind_instance.ts
@@ -156,7 +156,7 @@ export class BitcoindInstance implements LedgerInstance {
 server=1
 printtoconsole=1
 rpcallowip=0.0.0.0/0
-nodebug=1
+debug=http, db, rpc
 rest=1
 acceptnonstdtxn=0
 zmqpubrawblock=tcp://127.0.0.1:${this.zmqPubRawBlockPort}
@@ -165,6 +165,7 @@ zmqpubrawtx=tcp://127.0.0.1:${this.zmqPubRawTxPort}
 [regtest]
 bind=0.0.0.0:${this.p2pPort}
 rpcbind=0.0.0.0:${this.rpcPort}
+wallet=miner
 `;
         const config = path.join(dataDir, "bitcoin.conf");
         await writeFileAsync(config, output);
--- a/api_tests/src/wallets/bitcoind_wallet.ts
+++ b/api_tests/src/wallets/bitcoind_wallet.ts
@@ -1,0 +1,88 @@
+import { BitcoinWallet } from "comit-sdk";
+import { BitcoinNodeConfig } from "../ledgers";
+import BitcoinRpcClient from "bitcoin-core";
+
+export class BitcoindWallet implements BitcoinWallet {
+    constructor(private readonly client: BitcoinRpcClient) {}
+
+    public static async newInstance(
+        network: string,
+        hdKey: string,
+        config: BitcoinNodeConfig
+    ): Promise<BitcoindWallet> {
+        const bitcoinRpcClient = new BitcoinRpcClient({
+            network,
+            host: "localhost",
+            port: config.rpcPort,
+            username: config.username,
+            password: config.password,
+            wallet: "miner",
+        });
+
+        // everything before is the same for every xprv
+        const walletName = hdKey.substr(16, 20);
+        const createWalletResponse = await bitcoinRpcClient.createWallet(
+            walletName
+        );
+
+        const client = new BitcoinRpcClient({
+            network,
+            host: "localhost",
+            port: config.rpcPort,
+            username: config.username,
+            password: config.password,
+            wallet: createWalletResponse.name,
+        });
+
+        let descriptor = `wpkh(${hdKey}/0h/0h/*h)`;
+
+        const descriptorInfo = await client.getDescriptorInfo(descriptor);
+        descriptor = `${descriptor}#${descriptorInfo.checksum}`;
+
+        const request = {
+            desc: descriptor,
+            timestamp: 0,
+            range: 0,
+        };
+        // no need to rescan as we mint only after the initialization
+        const options = { rescan: false };
+        const importMultiResponse = await client.importMulti(
+            [request],
+            options
+        );
+
+        if (!importMultiResponse[0].success) {
+            return Promise.reject("Could not import xprv key");
+        }
+
+        return new BitcoindWallet(client);
+    }
+
+    public async broadcastTransaction(
+        transactionHex: string,
+        _network: string
+    ): Promise<string> {
+        return this.client.sendRawTransaction(transactionHex);
+    }
+
+    public async getAddress(): Promise<string> {
+        return this.client.getNewAddress();
+    }
+
+    public async getBalance(): Promise<number> {
+        return this.client.getBalance();
+    }
+
+    public getFee(): string {
+        return "150";
+    }
+
+    public async sendToAddress(
+        address: string,
+        satoshis: number,
+        _network: string
+    ): Promise<string> {
+        const bitcoin = satoshis / 100_000_000;
+        return this.client.sendToAddress(address, bitcoin);
+    }
+}
--- a/api_tests/src/actor_test.ts
+++ b/api_tests/src/actor_test.ts
@@ -33,7 +33,7 @@ function nActorTest(
         } finally {
             for (const actorName of actorNames) {
                 const actor = actors.getActorByName(actorName);
-                await Promise.all([actor.stop(), actor.wallets.close()]);
+                await actor.stop();
             }
         }
         done();
--- a/api_tests/src/ledgers/bitcoin_miner_instance.ts
+++ b/api_tests/src/ledgers/bitcoin_miner_instance.ts
@@ -34,6 +34,10 @@ export default class BitcoinMinerInstance {
 
         miner.unref();
 
+        miner.on("error", (error) => {
+            logger.error("bitcoin miner threw an error ", error);
+        });
+
         miner.on("exit", (code) => {
             logger.warn("bitcoin miner exited with code ", code);
         });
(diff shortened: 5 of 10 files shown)


thomaseizinger:
If only the tests would pass 😂

D4nte:
Tests fail :) Please re-request once it works ;)

bonomat:
The problem is obviously bcoin :S 
So I provide a fix for the bcoin-bug from https://github.com/comit-network/comit-js-sdk/issues/177 in this branch: https://github.com/comit-network/comit-rs/pull/2445

api_tests/src/bitcoin_miner.ts
```diff
@@ -18,14 +18,15 @@ async function run(configFile: string) {
         port: config.rpcPort,
         username: config.username,
         password: config.password,
+        wallet: "miner",
     });
 
     // only coins after the first 101 are spendable
-    await client.generateToAddress(101, await client.getNewAddress());
+    const address = await client.getNewAddress();
```
D4nte, line 25: Who is using this address?

D4nte: I don't understand because you are generating a new address in this scope. You are then sending money to this address but the address is a local variable.

How do you then use the money sent to this address?

Or is it because this address is the one of a bitcoind wallet meaning you can then use the money by using `sendtoaddress` rpc call?

api_tests/src/wallets/bitcoind_wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+        const bitcoinRpcClient = new BitcoinRpcClient({
+            network,
+            host: "localhost",
+            port: config.rpcPort,
+            username: config.username,
+            password: config.password,
+            wallet: "miner",
+        });
+
+        // everything before is the same for every xprv
```
D4nte, line 22: everything before what?

api_tests/src/wallets/bitcoind_wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+        const walletName = hdKey.substr(16, 20);
```
D4nte, line 23: Can you clarify why you are doing this?

api_tests/src/bitcoin_miner.ts
bonomat: That's the address of the miner. No need to know it and we can safely reuse. When minting Bitcoin the money will be sent from this address.

api_tests/src/wallets/bitcoind_wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+        if (!importMultiResponse[0].success) {
+            return Promise.reject("Could not import xprv key");
+        }
+
+        return new BitcoindWallet(client);
+    }
+
+    public async broadcastTransaction(
+        transactionHex: string,
+        _network: string
```
D4nte, line 63: I am surprise this passes the check. This is a Rust practice, not TypeScript

api_tests/src/bitcoin_miner.ts
```diff
@@ -18,14 +18,15 @@ async function run(configFile: string) {
         port: config.rpcPort,
         username: config.username,
         password: config.password,
+        wallet: "miner",
     });
 
     // only coins after the first 101 are spendable
```
thomaseizinger, line 24: This comment should move one line down now :)

api_tests/src/bitcoin_miner.ts
thomaseizinger: Any danger in not reusing it?

To make it more descriptive, we could pass a "label" to this function call like "miner blockreward": https://bitcoin.org/en/developer-reference#getnewaddress

api_tests/src/ledgers/bitcoind_instance.ts
```diff
@@ -165,6 +165,7 @@ zmqpubrawtx=tcp://127.0.0.1:${this.zmqPubRawTxPort}
 [regtest]
 bind=0.0.0.0:${this.p2pPort}
 rpcbind=0.0.0.0:${this.rpcPort}
+wallet=miner
```
thomaseizinger, line 168: Why are we setting wallet related functionality here? I would prefer to keep this part of the code unaware of wallets. Can we create a `miner` wallet whereever we need it and make sure there is a dataflow one can trace?

Like:

- Initialize the bitcoin miner with a specific wallet name
- Use the same variable to initialize the `BitcoindWallet` to make sure it can spend the money

api_tests/src/wallets/bitcoind_wallet.ts
thomaseizinger: I think he is referring to the first 16 characters of the hdkey.

api_tests/src/wallets/bitcoind_wallet.ts
thomaseizinger: I would suggest to either:

- use a library for parsing this that makes it more readable
- include a code comment of the structure of an hdkey that describes what you are doing (maybe including an example what an hdkey looks like?)

api_tests/src/wallets/bitcoind_wallet.ts
thomaseizinger: We should still be making this check and not ignore the parameter!

api_tests/src/wallets/bitcoind_wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+    public getFee(): string {
+        return "150";
+    }
+
+    public async sendToAddress(
+        address: string,
+        satoshis: number,
+        _network: string
+    ): Promise<string> {
+        const bitcoin = satoshis / 100_000_000;
```
thomaseizinger, line 85: Are you sure that JavaScript is not going to do any precision errors here ever? :D

api_tests/src/wallets/bitcoind_wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+        hdKey: string,
+        config: BitcoinNodeConfig
+    ): Promise<BitcoindWallet> {
+        const bitcoinRpcClient = new BitcoinRpcClient({
+            network,
+            host: "localhost",
+            port: config.rpcPort,
+            username: config.username,
+            password: config.password,
+            wallet: "miner",
```
thomaseizinger, line 19: Is it required that we are using the `miner`s wallet here? From what I can see, we are not using this wallet here right?

bonomat:
@thomaseizinger (and other interested reader :) )

Bitcoind is a bit complicated in regards of wallets.
If wallets are enabled, bitcoind has a default wallet `""` (or left empty) and it allows for creating additional wallets with `createwallet <>`. 

Once we create an additional wallet the default wallet is not accessible anymore and one will always need to specify a wallet name.
Note, it is now REQUIRED to specify a wallet name for any rpc command. That's what makes is so awkward to have two RPC clients. First, the "default" client which allows us to create a new wallet, and second, for creating the wallet we want :)

In order to overcome this issue, I introduced a `miner` wallet (see bitcoin config) which serves as the default wallet for the miner :D 
The wallet is created during startup which is much faster (i.e. it replaces the default wallet).

api_tests/src/bitcoin_miner.ts
bonomat: I don't think this is necessary here. 
I opted for only 1 address because it's a speed improvement (1 rpc call vs 2 every second).

api_tests/src/ledgers/bitcoind_instance.ts
bonomat: See response on the PR.

api_tests/src/wallets/bitcoind_wallet.ts
bonomat: What do you want to use a library for? 

Maybe it's better to just use Math.random() for the name as proposed by @luckysori :D

thomaseizinger:
> @thomaseizinger (and other interested reader :) )
> 
> Bitcoind is a bit complicated in regards of wallets.
> If wallets are enabled, bitcoind has a default wallet `""` (or left empty) and it allows for creating additional wallets with `createwallet <>`.
> 
> Once we create an additional wallet the default wallet is not accessible anymore and one will always need to specify a wallet name.
> Note, it is now REQUIRED to specify a wallet name for any rpc command. That's what makes is so awkward to have two RPC clients. First, the "default" client which allows us to create a new wallet, and second, for creating the wallet we want :)
> 
> In order to overcome this issue, I introduced a `miner` wallet (see bitcoin config) which serves as the default wallet for the miner :D
> The wallet is created during startup which is much faster (i.e. it replaces the default wallet).
> 
> Additionally, I experienced concurrency issues when creating the wallet from within the miner, i.e. first you would need to instantiate a RPCClient without wallet (or the default) and then reinstantiate one with the miner.
> While the miner is starting we already kick-off our tests and might instantiate another RPCClient somewhere else, if we would be using `miner` wallet there, we might get a random error because this wallet does not exist. If we don't, then we have an error the otherway around :)

Interesting and annoying :D

Can we have a dedicated `default` wallet then that we can use for commands where we don't need a specific wallet? I would still be very much in favor if our usage of the `miner` wallet would be explicit and not also the one that we use for making other RPC calls. Just to avoid the mental question of "why do we need to wear the `miner` hat here if we talk to bitcoind". Having it say `default` in some places and `miner` in others would clear that up.

Do you think that is technically feasible?

In order to avoid the concurrency issues, I would recommend to create the wallet in the e2e test environment, after bitcoind is started up here: https://github.com/comit-network/comit-rs/blob/33374ba0717736eb3725f82f5d78740d4fc67190/api_tests/src/test_environment.ts#L173.

This callback is only invoked once if the ledger actually needs to be started: https://github.com/comit-network/comit-rs/blob/f3f6fa046e7de19a1e6721f81c7b67fdbaec1c5d/api_tests/src/test_environment.ts#L358

Semantically, I would say creating the miner's wallet is part of creating the config that we store globally for the Bitcoin ledger, see the startup of the Ethereum ledger as an example.
The config is passed into the `BitcoinWallet` here:

https://github.com/comit-network/comit-rs/blob/f3f6fa046e7de19a1e6721f81c7b67fdbaec1c5d/api_tests/src/wallets/index.ts#L66

Which allows you to access the name of the miner's wallet here:

https://github.com/comit-network/comit-rs/blob/f3f6fa046e7de19a1e6721f81c7b67fdbaec1c5d/api_tests/src/wallets/bitcoin.ts#L20

luckysori:
@thomaseizinger @bonomat I have incorporated most of the changes introduced here into the two PRs I opened yesterday:

1. [Replace bcoin with bitcoind in SDK](https://github.com/comit-network/comit-js-sdk/pull/183).
2. [Use the new wallet in comit-rs e2e tests](https://github.com/comit-network/comit-rs/pull/2446).

Sorry to make you look through similar changes again, but could you have a quick look at both PRs and see if your comments have been addressed, @thomaseizinger. I think it's all there except for maybe the `miner` wallet vs `defaultWallet` issue, which I handled differently.

luckysori:
> @thomaseizinger @bonomat I have incorporated most of the changes introduced here into the two PRs I opened yesterday:
> 
>     1. [Replace bcoin with bitcoind in SDK](https://github.com/comit-network/comit-js-sdk/pull/183).
> 
>     2. [Use the new wallet in comit-rs e2e tests](https://github.com/comit-network/comit-rs/pull/2446).
> 
> 
> Sorry to make you look through similar changes again, but could you have a quick look at both PRs and see if your comments have been addressed, @thomaseizinger. I think it's all there except for maybe the `miner` wallet vs `defaultWallet` issue, which I handled differently.

I would recommend closing this PR in favour of the other two, since it moves us closer to the end goal.

bonomat:
Closed in favor of: 

1. [Replace bcoin with bitcoind in SDK](https://github.com/comit-network/comit-js-sdk/pull/183).
2. [Use the new wallet in comit-rs e2e tests](https://github.com/comit-network/comit-rs/pull/2446).

PR #2440, opened 2020-04-10
Use swap factory for for all halight dry tests

### LINKED ISSUE #2384: e2e tests: Use swap factory
## Problem

`swap_factory.ts` is use to create the body to create a swap on the new REST API routes.
However, some of the functions to default values for the body swap are present in `default.ts` and used by `actor.ts`.

Note that `swap_factory.ts` already support han-ethereum-ether-halight-lightning-bitcoin pair. Similar logic needs to be written for the other pairs.

## Goal

- Consolidate all functions to generate the body for the new cnd REST API routes in `swap_factory.ts`.
- Ensure all LN tests (e.g. dry tests and lnd sanity) use `SwapFactory`.
- This means the halight dry tests need to be converted to 2 actor tests.
- The expiries (and other info) to use for a swap are decided within the swap factory for all actors (and not each actor generate the expiry) as both actor need to inject the exact same in the create swap request to cnd REST API.

D4nte:
Resolves #2384.

### DIFF
--- a/api_tests/src/actors/swap_factory.ts
+++ b/api_tests/src/actors/swap_factory.ts
@@ -19,18 +19,36 @@ import {
     Herc20EthereumErc20RequestParams,
     Peer,
 } from "comit-sdk";
+import { HarnessGlobal } from "../utils";
+
+declare var global: HarnessGlobal;
 
 export default class SwapFactory {
     public static async newSwap(
         alice: Actor,
-        bob: Actor
-    ): Promise<
-        [
-            HanEthereumEtherHalightLightningBitcoinRequestBody,
-            HanEthereumEtherHalightLightningBitcoinRequestBody
-        ]
-    > {
-        const ledgers: (keyof AllWallets)[] = ["ethereum", "lightning"];
+        bob: Actor,
+        dry?: boolean
+    ): Promise<{
+        hanEthereumEtherHalightLightningBitcoin: {
+            alice: HanEthereumEtherHalightLightningBitcoinRequestBody;
+            bob: HanEthereumEtherHalightLightningBitcoinRequestBody;
+        };
+        herc20EthereumErc20HalightLightningBitcoin: {
+            alice: Herc20EthereumErc20HalightLightningBitcoinRequestBody;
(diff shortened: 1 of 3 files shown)


api_tests/tests/ether_halight.ts
```diff
@@ -10,13 +10,14 @@ import { sleep } from "../src/utils";
 it(
     "han-ethereum-ether-halight-lightning-bitcoin-alice-redeems-bob-redeems",
     twoActorTest(async ({ alice, bob }) => {
-        const [aliceBody, bobBody] = await SwapFactory.newSwap(alice, bob);
+        const bodies = (await SwapFactory.newSwap(alice, bob))
+            .hanEthereumEtherHalightLightningBitcoin;
```
thomaseizinger, line 14: I've recently found that using a mix of `then` and await is a nice way of chaining things :)


```suggestion
        const bodies = await SwapFactory.newSwap(alice, bob).then(bodies => bodies.hanEthereumEtherHalightLightningBitcoin)
```

api_tests/src/actors/swap_factory.ts
```diff
@@ -19,18 +19,36 @@ import {
...
-        bob: Actor
-    ): Promise<
-        [
-            HanEthereumEtherHalightLightningBitcoinRequestBody,
-            HanEthereumEtherHalightLightningBitcoinRequestBody
-        ]
-    > {
-        const ledgers: (keyof AllWallets)[] = ["ethereum", "lightning"];
+        bob: Actor,
+        dry?: boolean
```
thomaseizinger, line 30: I am not seeing any use of the `dry` parameter?

api_tests/src/actors/swap_factory.ts
```diff
@@ -19,18 +19,36 @@ import {
...
+    ): Promise<{
```
thomaseizinger, line 31: Interesting idea to return all of them at once :)

api_tests/src/actors/swap_factory.ts
D4nte: in `lightning_routes.ts`

thomaseizinger:
> wow... those type names sweat_smile

We'll find something better eventually :)

PR #2446, opened 2020-04-13
Upgrade comit-sdk to 0.15.3

luckysori:
Blocked by comit-network/comit-js-sdk#183 and the release of said changes in 0.16.0.

Build fails because of the local dependency, but test pass locally.


api_tests/src/wallets/bitcoin.ts
```diff
@@ -1,31 +1,32 @@
...
+    BitcoindWallet,
 } from "comit-sdk";
 import { toBitcoin, toSatoshi } from "satoshi-bitcoin";
 import { pollUntilMinted, Wallet } from "./index";
 import { BitcoinNodeConfig } from "../ledgers";
+import { newBitcoinRpcClientInstance } from "../bitcoin_rpc_client";
 
 export class BitcoinWallet implements Wallet {
     public static async newInstance(config: BitcoinNodeConfig, logger: Logger) {
-        const hdKey = bcoin.HDPrivateKey.generate().xprivkey(config.network);
```
bonomat, line 15: Nice 👍

thomaseizinger:
> I'd be in favor of initializing the `miner` or `defaultWallet` wallet through the config file.
> This saves us 2 seconds for every run :)
> 
> from my logs. As you can see, we need to create 2 wallets instead of only 1. If we provide the `wallet=miner` in the config, we can get rid of the additional wallet creation --> speed up by 2 seconds :D

We only do that once for the whole test suite though :man_shrugging: 
I would also like our tests to be faster but I am sure there are other optimizations that are worthwhile pursuing that don't introduce hacks in the form of implicit dependencies between components :)

PR #2448, opened 2020-04-13
Deserialization for networks main/test/regtest as retrieved by bitcoind

### LINKED ISSUE #2438: Bitcoin::network does not support `test`/`main`-net
# Problem

We use `(rust_bitcoin)::constants::Network` for parsing the response we get from bitcoind.
However, the enum variants of `Network` do not support `test`. 

Implementation:
https://github.com/rust-bitcoin/rust-bitcoin/blob/master/src/network/constants.rs#L50-L56

RPC/Rest Documentation:
https://bitcoincore.org/en/doc/0.17.0/rpc/blockchain/getblockchaininfo/
```
...
"chain": "xxxx",              (string) current network name as defined in BIP70 (main, test, regtest)
...
```

This is a problem because we check `/rest/chaininfo.json` to get the latest blockheight. 
Deserializing the response fails for testnet (and mainnet) as the `Network` variant for testnet is `"testnet"` and for mainnet `main`.

# Goal
Correctly decode bitcoind's response (e.g. networks `main`, `test` and `regtest`). 

* https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki

bonomat:
Fixes #2438 

Manually testing against a bitcoind testnet node worked.

### DIFF
--- a/cnd/src/btsieve/bitcoin/bitcoind_connector.rs
+++ b/cnd/src/btsieve/bitcoin/bitcoind_connector.rs
@@ -5,11 +5,12 @@ use crate::{
 use async_trait::async_trait;
 use bitcoin::{BlockHash, Network};
 use reqwest::{Client, Url};
-use serde::Deserialize;
+use serde::{de, export::fmt, Deserialize, Deserializer};
 
 #[derive(Copy, Clone, Debug, Deserialize)]
 pub struct ChainInfo {
     bestblockhash: BlockHash,
+    #[serde(deserialize_with = "deserialize_bitcoind_values")]
     pub chain: Network,
 }
 
@@ -96,6 +97,35 @@ impl FetchNetworkId<Network> for BitcoindConnector {
     }
 }
 
+pub fn deserialize_bitcoind_values<'de, D>(deserializer: D) -> Result<bitcoin::Network, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    struct Visitor;
+
+    impl<'de> de::Visitor<'de> for Visitor {
+        type Value = bitcoin::Network;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+            formatter.write_str("a bitcoin network")
+        }
+
+        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            match v {
+                "main" => Ok(bitcoin::Network::Bitcoin),
+                "test" => Ok(bitcoin::Network::Testnet),
+                "regtest" => Ok(bitcoin::Network::Regtest),
+                unknown => Err(E::custom(format!("unknown bitcoin network {}", unknown))),
+            }
+        }
+    }
+
+    deserializer.deserialize_str(Visitor)
+}
+
 #[cfg(test)]
 mod tests {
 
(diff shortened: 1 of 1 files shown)


bonomat:
> ❓ Could this be solved by [annotating](https://serde.rs/field-attrs.html#rename) the enum variants instead?
> 
> ```rust
> enum Network {
>     #[serde(rename(deserialize = "main"))]
>     Bitcoin,
>     #[serde(rename(deserialize = "test"))]
>     Testnet,
>     #[serde(rename(deserialize = "regtest"))]
>     Regtest,
> }
> ```
@luckysori 

I'm not sure if I understand you correctly. We don't have access to this as the network is coming from rust-bitcoin. 
We could solve this by having our own Network enum.

luckysori:
> I'm not sure if I understand you correctly. We don't have access to this as the network is coming from rust-bitcoin.

Ahh, you're right. Forgot it wasn't our enum!

cnd/src/btsieve/bitcoin/serde_bitcoind_network.rs
```diff
@@ -0,0 +1,30 @@
...
+        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+            formatter.write_str("a bitcoin network")
+        }
+
+        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            match v {
+                "main" => Ok(bitcoin::Network::Bitcoin),
```
thomaseizinger, line 21: This means we are incompatible with older versions now, can we maintain backwards compatibility here?

cnd/src/btsieve/bitcoin/bitcoind_connector.rs
```diff
@@ -10,6 +10,7 @@ use serde::Deserialize;
 #[derive(Copy, Clone, Debug, Deserialize)]
 pub struct ChainInfo {
     bestblockhash: BlockHash,
+    #[serde(with = "crate::btsieve::bitcoin::serde_bitcoind_network")]
```
thomaseizinger, line 13: Given that we don't need this functionality anywhere else, can we just have a local function instead of whole new module?

We only need `deserialize` so you can use `deserialize_with` here: https://serde.rs/field-attrs.html#deserialize_with

thomaseizinger: Can be addressed in a follow-up if you don't want to do it now although I would appreciate if we don't have ugly identifiers like this :)

cnd/src/btsieve/bitcoin/serde_bitcoind_network.rs
bonomat: I checked with 0.17. and it was also `main`, `test` , and `regtest`.

cnd/src/btsieve/bitcoin/bitcoind_connector.rs
bonomat: It's not that I don't want to but rather that I didn't know how to do this :D

bonomat: I think I got it, rebase + push in a minute.

cnd/src/btsieve/bitcoin/serde_bitcoind_network.rs
thomaseizinger: Oh I didn't know that, fine with me then :)

cnd/src/btsieve/bitcoin/bitcoind_connector.rs
thomaseizinger: Good on you, you learned something about serde :)

PR #2449, opened 2020-04-14
Harmonize usage of async fn when deal with HTTP and networking code

thomaseizinger:
Previously, we used a mix of `Runtime::block_on` and `Runtime::spawn`
to kick off all the important background worker processes that make
cnd work.

This patch harmonizes this usage by:

- Always using `Runtime::spawn` to spawn background workers
- Always using `Runtime::block_on` to execute code that should
run synchronously as we start up cnd.

To achieve this, we separate binding to the socket for the HTTP API
(something that should happen synchronously during startup because
we want to terminate if we fail) from the actual worker future
that handles the incoming requests.

We also separate creating an instance of `Swarm` from actually
spawning the background worker to align it with the HTTP API
background worker.

This introduces the `std::thread::park` function call again to idle
the main process.

This patch also prepares for potential follow-up work where we want
to gracefully shutdown by listening on signals like SIGINT. Such code
would go in place of the current `std::thread::park`.

### DIFF
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -64,10 +64,7 @@ use std::{
     sync::Arc,
     task::{self, Poll},
 };
-use tokio::{
-    runtime::{Handle, Runtime},
-    sync::Mutex,
-};
+use tokio::{runtime::Handle, sync::Mutex};
 use tracing_futures::Instrument;
 
 #[derive(Clone, derivative::Derivative)]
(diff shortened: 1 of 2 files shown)


cnd/src/network.rs
```diff
@@ -178,16 +171,24 @@ impl libp2p::core::Executor for TokioExecutor {
     }
 }
 
-struct SwarmWorker {
-    swarm: Arc<Mutex<libp2p::Swarm<ComitNode>>>,
+/// The SwarmWorker, when spawned into a runtime, continuously polls the
+/// underlying swarm for events.
+///
+/// This is the main driver of the networking code.
+/// Note that the inner swarm is wrapped in an `Arc<Mutex>` any we only old the
```
luckysori, line 178: :wrench: 

```suggestion
/// Note that the inner swarm is wrapped in an `Arc<Mutex>` and we only hold the
```

PR #2450, opened 2020-04-14
Add herc20 protocol

### LINKED ISSUE #2422: Introduce the `herc20` protocol
## Problem

We want to support halight-herc20 swaps in cnd but the internal structure still uses rfc003.

## Goal

- Support halight-herc20 swaps with as little effort as possible.
- Prove it by creating a `new_herc20_swap` function similar to the existing `new_halight_swap` function (see _Usage_ link below)

## Out of scope

Using this new module from the old rfc003 routes/code paths.

## Recommendation

Duplicate the necessary code from `rfc003` in a `herc20` module for use in the new halight swaps.
Note that you can remove a lot of unnecessary type parameters because `herc20` is specific to Ethereum.
Try to use the new design of the `halight` protocol as a template: 

Definition: https://github.com/comit-network/comit-rs/blob/95cf45db1141c0ed2857bd7750731665ebc85cda/cnd/src/swap_protocols/halight.rs#L176-L221
Usage: https://github.com/comit-network/comit-rs/blob/95cf45db1141c0ed2857bd7750731665ebc85cda/cnd/src/network/comit_ln.rs#L678-L705

D4nte:
Resolves #2422

### DIFF
--- a/cnd/src/swap_protocols/herc20.rs
+++ b/cnd/src/swap_protocols/herc20.rs
@@ -1,0 +1,281 @@
+use crate::{
+    asset, htlc_location, identity,
+    swap_protocols::{
+        rfc003::{Secret, SecretHash},
+        state, SwapId,
+    },
+    transaction,
+};
+use chrono::NaiveDateTime;
+use futures::{
+    future::{self, Either},
+    Stream,
+};
+use genawaiter::sync::{Co, Gen};
+use std::collections::{hash_map::Entry, HashMap};
+use tokio::sync::Mutex;
+
+mod connector_impls;
+
+use crate::{ethereum::Bytes, timestamp::Timestamp};
+use blockchain_contracts::ethereum::rfc003::Erc20Htlc;
+pub use connector_impls::*;
+
+/// Resolves when said event has occured.
+#[async_trait::async_trait]
+pub trait WaitForDeployed {
+    async fn wait_for_deployed(&self, params: Params) -> anyhow::Result<Deployed>;
+}
+
+#[async_trait::async_trait]
+pub trait WaitForFunded {
+    async fn wait_for_funded(&self, params: Params, deployed: Deployed) -> anyhow::Result<Funded>;
+}
+
+#[async_trait::async_trait]
+pub trait WaitForRedeemed {
+    async fn wait_for_redeemed(
+        &self,
+        params: Params,
+        deployed: Deployed,
+    ) -> anyhow::Result<Redeemed>;
+}
+
+#[async_trait::async_trait]
+pub trait WaitForRefunded {
+    async fn wait_for_refunded(
+        &self,
+        params: Params,
+        deployed: Deployed,
+    ) -> anyhow::Result<Refunded>;
+}
+
+/// Represents states that an ERC20 HTLC can be in.
+#[derive(Debug, Clone)]
+pub enum State {
+    None,
+    Deployed(Deployed),
+    Funded(Funded),
+    Redeemed(Redeemed),
+    Refunded(Refunded),
+}
+
+/// Represents the events in the herc20 protocol.
+#[derive(Debug, Clone, PartialEq, strum_macros::Display)]
+pub enum Event {
+    /// The protocol was started.
+    Started,
+
+    /// The HTLC was deployed and is pending funding.
+    Deployed(Deployed),
+
+    /// The HTLC has been funded with ERC20 tokens.
+    Funded(Funded),
+
+    /// The HTLC has been destroyed via the redeem path, token have been sent to
+    /// the redeemer.
+    Redeemed(Redeemed),
+
+    /// The HTLC has been destroyed via the refund path, token has been sent
+    /// back to funder.
+    Refunded(Refunded),
+}
+
+/// Represents the data available at said state.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Deployed {
+    pub transaction: transaction::Ethereum,
+    pub location: htlc_location::Ethereum,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum Funded {
+    Correctly {
+        transaction: transaction::Ethereum,
+        asset: asset::Erc20,
+    },
+    Incorrectly {
+        transaction: transaction::Ethereum,
+        asset: asset::Erc20,
+    },
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Redeemed {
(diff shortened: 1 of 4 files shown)


cnd/src/swap_protocols/herc20.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// the redeemer.
+    Redeemed(data::Redeemed),
+
+    /// The HTLC has been destroyed via the refund path, token has been sent
+    /// back to funder.
+    Refunded(data::Refunded),
+}
+
+/// Represents the data available at said state.
+pub mod data {
```
thomaseizinger, line 87: The `data` module is bugging me for `halight` as-well. Can we get rid of this?

The naming conflict should urge us to think about a better name for one of the two concepts instead of moving one of the types into a module with a generic name.

To start with, I would say the names of the events are more naturally expressed as they currently are so I would probably change the names of the traits. What about `WaitForFunded` etc?
If we find a way for either not passing in `Deployed` separately, we could even to something like `WaitFor<Funded>` :exploding_head:

cnd/src/swap_protocols/herc20.rs
```diff
@@ -0,0 +1,263 @@
...
+            let (error, _other_future) = either.factor_first();
+            return Err(error);
+        }
+    }
+
+    Ok(())
+}
+
+#[derive(Clone, Debug)]
+pub struct Params {
```
thomaseizinger, line 258: These parameters are not orthogonal to each other.

thomaseizinger: The asset is inside the htlc and so is the secret hash. I would say just compute the HTLC from the remaining parameters like we used to.

cnd/src/swap_protocols/herc20.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    #[derive(Debug, Clone, PartialEq)]
+    pub struct Refunded {
+        pub transaction: transaction::Ethereum,
+    }
+}
+
+#[derive(Default, Debug)]
+pub struct StateStore {
```
thomaseizinger, line 123: As mentioned in the other PR, not really happy with this naming.

cnd/src/swap_protocols/herc20.rs
```diff
@@ -0,0 +1,263 @@
...
+    Redeemed(data::Redeemed),
+
+    /// The HTLC has been destroyed via the refund path, token has been sent
+    /// back to funder.
+    Refunded(data::Refunded),
+}
+
+/// Represents the data available at said state.
+pub mod data {
+    // These empty types are useful because they give us additional type safety.
```
thomaseizinger, line 88: Sounds like a copy pasta error to me.

cnd/src/swap_protocols/herc20.rs
```diff
@@ -0,0 +1,263 @@
...
+use chrono::NaiveDateTime;
+use futures::{
+    future::{self, Either},
+    Stream,
+};
+use genawaiter::sync::{Co, Gen};
+use std::collections::{hash_map::Entry, HashMap};
+use tokio::sync::Mutex;
+
+mod connector;
```
thomaseizinger, line 18: This module doesn't really define the connector though, maybe `connector_impls` is a better name?

cnd/src/swap_protocols/herc20.rs
D4nte: The comment was valid at first :)

cnd/src/swap_protocols/herc20.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+/// Represents states that an ERC20 HTLC can be in.
+#[derive(Debug, Clone)]
+pub enum State {
+    None,
+    Deployed(data::Deployed),
+    Funded(data::Funded),
+    Redeemed(data::Redeemed),
+    Refunded(data::Refunded),
```
bonomat, line 62: What about over-/underfunded or IncorrectlyFunded as it was called before?

bonomat: Ignore, found it a few lines below.

cnd/src/swap_protocols/herc20.rs
D4nte: Ah yes indeed, apologies on this one, I intended to clean it up once it compiled but did not give myself a todo.

cnd/src/swap_protocols/herc20.rs
D4nte: > If we find a way for either not passing in `Deployed` separately, we could even to something like `WaitFor<Funded>` 🤯

There are ways but it gives less type safety so for now I'll leave it with `WaitFor[..]` traits

cnd/src/swap_protocols/herc20/connector.rs
```diff
@@ -0,0 +1,121 @@
...
+    },
+};
+use std::cmp::Ordering;
+use tracing_futures::Instrument;
+
+lazy_static::lazy_static! {
+    static ref REDEEM_LOG_MSG: Hash = blockchain_contracts::ethereum::rfc003::REDEEMED_LOG_MSG.parse().expect("to be valid hex");
+    static ref REFUND_LOG_MSG: Hash = blockchain_contracts::ethereum::rfc003::REFUNDED_LOG_MSG.parse().expect("to be valid hex");
+    /// keccak('Transfer(address,address,uint256)')
+    static ref TRANSFER_LOG_MSG: Hash = "ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef".parse().expect("to be valid hex");
```
bonomat: Should we update `blockchain_contracts::ethereum::rfc003:` and export it from there? Feels weird to have it in here.

D4nte: Currently compiling so I can't change branch but not sure. Do we have this information in blockchain-contracts? Or are you saying we should add it to the blockchain-contracts crate?

bonomat: Sorry for not being clear: I meant that *if* we don't have it in blockchain-contracts then we should add it there. 
I didn't verify if this is not the case.

D4nte: Yes I see. It's not really something introduced by our protocols contrary to the other values but it is something that belongs to the blockchain contract domain so yes, I think it's a good idea.

D4nte:
bors cancel

PR #2451, opened 2020-04-14
Halight cleanup

D4nte:
After [introducing herc20](https://github.com/comit-network/comit-rs/pull/2450) I'd thought I'd align few things.

While I was there, I also fixed a couple of open issues.


cnd/src/swap_protocols/halight.rs
```diff
@@ -91,34 +94,34 @@ pub mod data {
 }
 
 #[derive(Default, Debug)]
-pub struct InvoiceStates {
+pub struct StateStore {
```
thomaseizinger, line 97: We deliberately went away from this naming, did you have to introduce it again? :grimacing:

thomaseizinger: What about naming this `States` instead and refer to it as `halight::States` everywhere?

D4nte: Best to review in chronological order with https://github.com/comit-network/comit-rs/pull/2450 first :)

D4nte: Can you please describe in more details why we went away from this naming?

thomaseizinger: I find it to be a very awkward name because it is super generic. It also has quite an object-oriented touch which I'd consider to be an anti-pattern in Rust. Maybe @luckysori can also say something, I know he dislikes it aswell :)

D4nte:
> I have no hard feelings about the naming :)
> The tests fail though, did you give them a try locally?

Yes I actually did, will double check when I align this PR with #2450.

.circleci/config.yml
```diff
@@ -130,7 +130,7 @@ commands:
             - cargo
       - restore_cache:
           keys:
-            - target-folder-{{ checksum "Cargo.lock" }}
+            - target-folder-v1-{{ checksum "Cargo.lock" }}
```
thomaseizinger, line 133: Can you explain this change please?

cnd/src/main.rs
```diff
@@ -23,21 +23,20 @@ use cnd::{
...
-    load_swaps,
+    jsonrpc, load_swaps,
     network::Swarm,
     seed::RootSeed,
     swap_protocols::{
-        halight::InvoiceStates, Facade, Facade2, LedgerStates, SwapCommunicationStates,
-        SwapErrorStates,
+        halight::States, Facade, Facade2, LedgerStates, SwapCommunicationStates, SwapErrorStates,
     },
 };
```
thomaseizinger, line 33: This line should not be here, the formatter didn't pick up the below use statements :)

cnd/src/swap_protocols/facade2.rs
```diff
@@ -95,8 +71,8 @@ pub struct Facade2 {
     pub swarm: Swarm,
     pub alpha_ledger_state: Arc<LedgerStates>, /* We currently only support Han-HALight, this is
                                                 * Ethereum. */
-    pub beta_ledger_state: Arc<InvoiceStates>, /* We currently only support Han-HALight, this is
-                                                * Lightning. */
+    pub beta_ledger_state: Arc<States>, /* We currently only support Han-HALight, this is
```
thomaseizinger, line 74:
```suggestion
    pub beta_ledger_state: Arc<halight::States>, /* We currently only support Han-HALight, this is
```

.circleci/config.yml
D4nte: Yes, I keep getting an error during the build that is strange and only appears in the CI. I wonder if it's due to the change around the proc macro. I am renewing the target folder cache to start from fresh to see if it helps.

cnd/src/main.rs
D4nte: I have a fix locally and I'll push once I see how things are going.

PR #2454, opened 2020-04-14
Libp2p integration tests for `ComitLn` network behaviour

da-kami:
outline the solution in comments for 10% review

Please just comment on this draft PR, no approvals. This is purely for discussing the solution.

The idea I discussed with @rishflab is to pull the code that actually starts the execution (by triggering the blockchain watchers) out of the `NetworkBehaviour` (`ComitLn`).

`finalize` would just return the information necessary to start the execution rather than starting it itself. We use the [`next_event()`](https://docs.rs/libp2p/0.18.0/libp2p/swarm/struct.ExpandedSwarm.html#method.next_event) function to listen for `finalize`. `finalize` returns all the necessary parameters to to actually start the execution.

Test for announce protocol that uses this pattern: https://github.com/comit-network/comit-rs/blob/announce-tests/cnd/src/network/protocols/announce/behaviour.rs#L213-L269

### DIFF
--- a/cnd/src/main.rs
+++ b/cnd/src/main.rs
@@ -152,6 +152,9 @@ fn main() -> anyhow::Result<()> {
         beta_ledger_state: Arc::clone(&invoice_states),
     };
 
+    // await finalized on the swarm here (using [swarm].next() and condition inside loop - is there a more elegant way...?)
+    // trigger the execution from parameters returned by finalized
+
     let deps = Facade {
         bitcoin_connector,
         ethereum_connector,
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -6,6 +6,7 @@ pub mod transport;
 
 pub use transport::ComitTransport;
 
+use crate::network::protocols::finalized;
 use crate::{
     asset::AssetKind,
     btsieve::{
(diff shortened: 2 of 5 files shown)


da-kami:
> Not sure if moving to `main.rs` is so good, it is not really a concern of the application so I would say this is too far up. But I just had a different idea:
> 
> You could be making use of custom poll function and a custom `OutEvent` of the `NetworkBehaviour` derive:
> 
> [docs.rs/libp2p/0.18.0/libp2p/swarm/trait.NetworkBehaviour.html#deriving-networkbehaviour](https://docs.rs/libp2p/0.18.0/libp2p/swarm/trait.NetworkBehaviour.html#deriving-networkbehaviour)
> 
> This would allow you to implement a custom function on `ComitLn` that checks for any swaps that been finalized and yields an event to the "parent" network behaviour, in this case `ComitNode`. You would then receive this event here:
> 
> https://github.com/comit-network/comit-rs/blob/1859b4b4e1ad5c4f181bebeb9414edc5b4cc33fd/cnd/src/network.rs#L1131-L1133
> 
> But instead of `()`, the type would be the `OutEvent` you specify on the `derive` of `NetworkBehaviour` of `ComitLn`.
> 
> I think that would make it really easy to test because you can just create a small swarm that uses just the `ComitLn` behaviour and check whether it emits the correct event.
> 
> In the real event handler, you then do the necessary stuff for spawning swaps etc.

That's what we had in mind for testing. Just spin up swarm for Alice and Bob and then see if the correct out event is emitted. Would call. it `finalized`)

Question concerning the poll function: Is that really needed?
Wouldn't it be enough to do something like this?

```
impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalized::Message>> for ComitNode {}
```

Would that not work?

da-kami:
> > Not sure if moving to `main.rs` is so good, it is not really a concern of the application so I would say this is too far up. But I just had a different idea:
> > You could be making use of custom poll function and a custom `OutEvent` of the `NetworkBehaviour` derive:
> > [docs.rs/libp2p/0.18.0/libp2p/swarm/trait.NetworkBehaviour.html#deriving-networkbehaviour](https://docs.rs/libp2p/0.18.0/libp2p/swarm/trait.NetworkBehaviour.html#deriving-networkbehaviour)
> > This would allow you to implement a custom function on `ComitLn` that checks for any swaps that been finalized and yields an event to the "parent" network behaviour, in this case `ComitNode`. You would then receive this event here:
> > https://github.com/comit-network/comit-rs/blob/1859b4b4e1ad5c4f181bebeb9414edc5b4cc33fd/cnd/src/network.rs#L1131-L1133
> > 
> > But instead of `()`, the type would be the `OutEvent` you specify on the `derive` of `NetworkBehaviour` of `ComitLn`.
> > I think that would make it really easy to test because you can just create a small swarm that uses just the `ComitLn` behaviour and check whether it emits the correct event.
> > In the real event handler, you then do the necessary stuff for spawning swaps etc.
> 
> That's what we had in mind for testing. Just spin up swarm for Alice and Bob and then see if the correct out event is emitted. Would call. it `finalized`)
> 
> Question concerning the poll function for the implementation: Is that really needed?
> Wouldn't it be enough to do something like this?
> 
> ```
> impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalized::Message>> for ComitNode {}
> ```
> 
> Would that not work?

@rishflab just pointed out that the `poll()` function would have to be implemented on `ComitLn` in order to emit the event from the `ComitLn` network behaviour.

da-kami:
closing for now, working on the branch

PR #2459, opened 2020-04-15
Upload e2e test logs

bonomat:
Note: lnd are ignored for now. Not sure if there are useful logs in there. @D4nte : any thoughts? 

I opted against uploading the whole `api/log` folder due to its size: `api/log` contains `api/log/parity` and `api/log/bitcoind` which can get rather big. These folders contain the whole blockchain data. 

Uploaded files are auto-deleted after [90 days](https://github.community/t5/GitHub-Actions/Managing-Actions-storage-space/m-p/41424/highlight/true#M4618). 

I'm using a preview feature of the upload archive action. Afaik, eventually they will offer multi-file upload and we can merge the multiple steps into one. 

This plugin is not yet _stable_ hence I recommend to only use it for the logs for now and not for other actions such as binary or releases. If these uploads fail it's way more annoying than if a log file upload fails :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,3 +135,31 @@ jobs:
           cd api_tests
           yarn install
           yarn ci
+
+      - name: Upload bitcoind log
+        if: failure()
+        uses: actions/upload-artifact@v2-preview
+        with:
+          name: e2e-logs-bitcoind.log
+          path: api_tests/log/bitcoind/regtest/debug.log
+
+      - name: Upload parity log
+        if: failure()
+        uses: actions/upload-artifact@v2-preview
+        with:
+          name: e2e-logs-parity.log
+          path: api_tests/log/parity/parity.log
+
+      - name: Upload lnd logs
+        if: failure()
+        uses: actions/upload-artifact@v2-preview
+        with:
+          name: e2e-logs-lnd
+          path: api_tests/log/lnd-*/logs/bitcoin/regtest/lnd.log
+
+      - name: Upload e2e logs
+        if: failure()
+        uses: actions/upload-artifact@v2-preview
+        with:
+          name: e2e-test-logs
+          path: api_tests/log/tests/


D4nte:
> Note: lnd are ignored for now. Not sure if there are useful logs in there. @D4nte : any thoughts?

lnd logs are actually quite useful if there is an issue with lightning tests.

> Btw: we need to delete these files manually!

Oh boy, how do we do that?

.github/workflows/ci.yml
```diff
@@ -122,3 +122,24 @@ jobs:
...
+        with:
+          name: e2e-logs-parity.log
+          path: api_tests/log/parity/parity.log
+
+      - name: Upload e2e logs
+        if: failure()
+        uses: actions/upload-artifact@v2-preview
+        with:
+          name: e2e-test-logs
+          path: api_tests/log/tests/
```
D4nte, line 145: Did you consider just uploading the `api_test/log` folder? And if so, why deciding not to upload it entirely?

bonomat: I did consider it and decided against it due to it's size: `api/log` contains `api/log/parity` and `api/log/bitcoind` which can get rather big. These folder contain the whole blockchain data. 

--> update the PR description.

D4nte:
Should we be creating an action to delete the log files regularly?

bonomat:
@D4nte : 
> lnd logs are actually quite useful if there is an issue with lightning tests.

Ok, I'll update the script.

> Should we be creating an action to delete the log files regularly?

That's a good idea. I'll create a ticket for that.

bonomat:
Lol, perfectly on time we had a random test failure 😅

bonomat:
@D4nte : 
Until we run out of storage we don't have to delete the files actively. 
Archives are removed after 90 days according to [this thread](https://github.community/t5/GitHub-Actions/Managing-Actions-storage-space/m-p/41424/highlight/true#M4618).

Alternatively there would be [an action](https://github.com/c-hive/gha-remove-artifacts) which would automate the deletion on a monthly basis. But space is not yet limited so I just leave this for now :)

PR #2466, opened 2020-04-16
Remove Circle CI config

D4nte:
Also ensure that GitHub Action CI run the docs.

FYI @comit-network/comit-devs

I'll update the repo settings to block on Linux GitHub actions and not on Circle CI to merge once this is approved.


bonomat:
Bye bye circle-ci...

D4nte:
Need to find a way to not cancel the ubuntu GitHub actions when mac os fails.

bonomat:
Approved too early :

Can you fix the badge as well please?

https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

bonomat:
@D4nte : 
> Need to find a way to not cancel the ubuntu GitHub actions when mac os fails.

I think what you are looking for is `fail-fast`: 

https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

If you want, I can integrate it into this PR where I speed up the build:
https://github.com/comit-network/comit-rs/pull/2469

D4nte:
> @D4nte :
> 
> > Need to find a way to not cancel the ubuntu GitHub actions when mac os fails.
> 
> I think what you are looking for is `fail-fast`:
> 
> [help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)
> 
> If you want, I can integrate it into this PR where I speed up the build:
> #2469

I opted for `continue-on-error` on the e2e tests instead.

.mergify.yml
```diff
@@ -2,7 +2,10 @@ pull_request_rules:
   - name: comment on broken dependabot PRs, asking for help
     conditions:
       - "author=dependabot-preview[bot]"
-      - "status-failure=ci/circleci: debug-build-test"
+      - "status-failure=static_analysis"
```
thomaseizinger, line 5: This whole rule doesn't really make sense because it never worked the way it was intended (mergify cannot assign for some reason).

I think these conditions are combined with a logical AND so now that it is more than one, I think it makes no more sense. It is unlikely a dependency upgrade breaks all of those.

D4nte: Ooooh. Cool, I can update to something that makes sense.

PR #2467, opened 2020-04-16
pull execution code from `ComitLn` up to `ComitNode` 

da-kami:
This is in preparation for the libp2p tests, but it can be merged separately.
Pulls the "execution" (spawn watchers) code out of `ComitLn` into `ComitNode` and send an event after `finalize` from `ComitLn` to `ComitNode`. 

@D4nte @thomaseizinger decided to bring this in separate to avoid more brain-intense rebase on changes in `ComitLn` and `ComitNode`.

Note with `.danger_accept_invalid_certs(true)` on the reqwest client active the e2e tests pass locally for me. 

Note that the tests (to be added later) are planned like this in `ComitLn`:

```
#[cfg(test)]
mod tests {

    #[test]
    fn swap_communication_yields_correct_execution_parameters_for_alice() {

        // -- Basic outline of the test for Alice --

        // GIVEN a swap as Alice
        // use CreateSwapParams to create a swap

        // WHEN passed to the network behavior
        //      init swarm for Alice and Bob
        //      use [swarm].next_event() pattern to loop through events

        // THEN (the necessary communications happens and) we receive the
        // event with the identities and hash of secret set correctly.
    }

    #[test]
    fn swap_communication_yields_correct_execution_parameters_for_bob() {

        // same for Bob
        // The output should include hash_secret in addition to identities
    }
}
```

### DIFF
--- a/cnd/src/main.rs
+++ b/cnd/src/main.rs
@@ -127,7 +127,7 @@ fn main() -> anyhow::Result<()> {
     let swap_communication_states = Arc::new(SwapCommunicationStates::default());
 
     // HALight
-    let invoice_states = Arc::new(States::default());
+    let halight_state = Arc::new(States::default());
 
     let swap_error_states = Arc::new(SwapErrorStates::default());
 
@@ -141,14 +141,14 @@ fn main() -> anyhow::Result<()> {
         Arc::clone(&swap_communication_states),
         Arc::clone(&alpha_ledger_state),
         Arc::clone(&beta_ledger_state),
-        Arc::clone(&invoice_states),
+        Arc::clone(&halight_state),
         &database,
     )?;
 
     let facade2 = Facade2 {
         swarm: swarm.clone(),
         alpha_ledger_state: Arc::clone(&alpha_ledger_state),
-        beta_ledger_state: Arc::clone(&invoice_states),
+        beta_ledger_state: Arc::clone(&halight_state),
     };
 
     let deps = Facade {
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -7,27 +7,30 @@ pub mod transport;
 pub use transport::ComitTransport;
 
 use crate::{
+    asset,
     asset::AssetKind,
     btsieve::{
         bitcoin::{self, BitcoindConnector},
-        ethereum::{self, Web3Connector},
+        ethereum::{self, Cache, Web3Connector},
     },
     comit_api::LedgerKind,
     config::Settings,
     db::{Save, Sqlite, Swap},
-    htlc_location,
+    htlc_location, identity,
     libp2p_comit_ext::{FromHeader, ToHeader},
     network::comit_ln::ComitLN,
     seed::RootSeed,
     swap_protocols::{
-        halight::{LndConnectorParams, States},
-        ledger,
+        halight,
+        halight::{LndConnectorAsReceiver, LndConnectorAsSender, LndConnectorParams, States},
+        han, herc20, ledger,
         rfc003::{
             self,
+            create_swap::HtlcParams,
             messages::{Decision, DeclineResponseBody, Request, RequestBody, SwapDeclineReason},
-            LedgerState, SwapCommunication,
+            LedgerState, SecretHash, SwapCommunication,
         },
-        state::Insert,
+        state::{Insert, Update},
         HanEtherereumHalightBitcoinCreateSwapParams, HashFunction, LedgerStates, NodeLocalSwapId,
         Role, SwapCommunicationStates, SwapId, SwapProtocol,
     },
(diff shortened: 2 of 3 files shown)


cnd/src/network.rs
```diff
@@ -1131,6 +1137,185 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<()> for ComitNode {
...
+                            .expect("Failure reading tls certificate")
+                            .read_macaroon()
+                            .expect("Failure reading macaroon");
+
+                        new_halight_swap(local_swap_id, secret_hash, invoice_states, lnd_connector)
+                            .instrument(
+                                tracing::error_span!("beta_ledger", swap_id = %local_swap_id, role = %role),
+                            )
+                    });
+                } else {
```
D4nte, line 1168: This is not related to your code but I already noticed it in the past.
Could you replace the `if..else` with a `match` block on `role` as to avoid the "// This is Bob` comment?

cnd/src/network.rs
```diff
@@ -1131,6 +1137,185 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<()> for ComitNode {
...
+        }
+    }
+}
+
+/// Creates a new instance of the halight protocol.
+///
+/// This function delegates to the `halight` module for the actual protocol
+/// implementation. Its main purpose is to annotate the protocol instance with
+/// logging information and store the events yielded by the protocol in
+/// `InvoiceStates`.
```
D4nte, line 1249:
```suggestion
/// `halight::States`.
```
This is my bad :)

cnd/src/network/comit_ln.rs
```diff
@@ -1,49 +1,63 @@
...
+};
+
+/// Event emitted  by the `ComitLn` behaviour.
+#[derive(Debug)]
+pub enum BehaviourOutEvent {
+    SwapIsFinalizedCanStartExecution {
+        local_swap_id: NodeLocalSwapId,
+        swap_params: CreateSwapParams,
+        secret_hash: SecretHash,
+        invoice_states: Arc<States>,
```
thomaseizinger, line 43: I don't think these should be part of the event, they are global dependencies so they should go into `ComitNode` as fields.

cnd/src/network/comit_ln.rs
```diff
@@ -1,49 +1,63 @@
...
+        ethereum_identity: identity::Ethereum,
+        ethereum_ledger_state: Arc<LedgerStates>,
```
thomaseizinger, line 45: Same with these, please make them fields on `ComitNode`.

cnd/src/network/comit_ln.rs
```diff
@@ -1,49 +1,63 @@
...
+use std::{
+    collections::{HashMap, VecDeque},
+    sync::Arc,
+    task::{Context, Poll},
+};
+
+/// Event emitted  by the `ComitLn` behaviour.
+#[derive(Debug)]
+pub enum BehaviourOutEvent {
+    SwapIsFinalizedCanStartExecution {
```
thomaseizinger, line 39:
```suggestion
    SwapFinalized {
```

cnd/src/network/comit_ln.rs
da-kami: oh, true, I missed that those are actually just used for execution. Will more them.

cnd/src/network/comit_ln.rs
da-kami: Will see to merge this one with alpha/beta ledger states depending on what ledger Ethereum is.

cnd/src/network.rs
da-kami: note: I did a bit more cleanup to make the naming more consistent

da-kami:
bors retry

cnd/src/main.rs
```diff
@@ -127,7 +127,7 @@ fn main() -> anyhow::Result<()> {
     let swap_communication_states = Arc::new(SwapCommunicationStates::default());
 
     // HALight
-    let invoice_states = Arc::new(States::default());
+    let halight_state = Arc::new(States::default());
```
D4nte, line 130:
```suggestion
    let halight_states = Arc::new(States::default());
```

da-kami: in accordance with all the other state names I made it singular.
(alpha_ledger_state, beta_ledger_state)

is this semantically different in lightning, so plural would be justified?

thomaseizinger: It is not different but I guess it is more the other states that should be renamed :)

D4nte: In Rust, the convention is to have the name of the variable matching the type of the variable as closely as possible.

da-kami: OK, then I rename all of them 😄

da-kami: I'll do this rename in a follow-up PR

PR #2477, opened 2020-04-17
Use lnd binary instead of compiling it in the e2e tests

### LINKED ISSUE #2369: Download `lnd` binary from GitHub releases instead of building them in CI
## Problem

Currently, we build the `lnd` binary during CI and locally we assume it is in the path.
This:

a) increases CI time unnecessarily
b) makes it harder to run the tests locally

## Goal

- Our e2e test suite should not have implicit dependencies.
- Provide a fast way for developer to get ready in the readme by mentioning the binary availability.

## Recommendation

The binaries on the `lnd` release page are built with the `invoicesrpc` tag. (It says there and I also confirmed by downloading one of the binaries and checking for the hold-invoice functionality.)
Hence, we can just download the desired release from their instead of building it locally.

I would recommend to wait until https://github.com/comit-network/comit-rs/pull/2307 is merged, as it makes it a breeze to do the same thing for lnd that I've done for bitcoind and parity there.

Use the latest beta version of lnd.

D4nte:
Resolves #2369

### DIFF
--- a/api_tests/src/actors/actor.ts
+++ b/api_tests/src/actors/actor.ts
@@ -664,12 +664,10 @@ export class Actor {
     }
 
     public async assertAlphaNotDeployed() {
-        await sleep(3000); // It is meaningless to assert before cnd processes a new block
         await this.assertLedgerState("alpha_ledger", "NOT_DEPLOYED");
     }
 
     public async assertBetaNotDeployed() {
-        await sleep(3000); // It is meaningless to assert before cnd processes a new block
         await this.assertLedgerState("beta_ledger", "NOT_DEPLOYED");
     }
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,22 +117,9 @@ jobs:
           path: api_tests/node_modules
           key: ${{ matrix.os }}-node-modules-directory-${{ hashFiles('api_tests/package.json') }}
 
-      - name: Install Go 1.13.3.
-        uses: actions/setup-go@v1
-        with:
-          go-version: '1.13.3'
-
-      - name: Install LND v0.9.1-beta
-        run: |
-          go get -d github.com/lightningnetwork/lnd
-          cd ~/go/src/github.com/lightningnetwork/lnd
-          git checkout v0.9.1-beta
-          make tags=invoicesrpc
-          make tags=invoicesrpc install
-
       - name: Run e2e tests
         run: |
-          export PATH=$HOME/.cargo/bin:$HOME/.local/bin:$HOME/go/bin:$PATH
+          export PATH=$HOME/.cargo/bin:$HOME/.local/bin:$PATH
           cd api_tests
           yarn install
           yarn ci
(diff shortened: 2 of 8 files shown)


da-kami:
> Resolves #2369

❓

There is no specific build with the tags needed any more? 
The provided binary is build like that?

```
make tags=invoicesrpc	
make tags=invoicesrpc install
```

D4nte:
> > Resolves #2369
> 
> ❓
> 
> There is no specific build with the tags needed any more?
> The provided binary is build like that?
> 
> ```
> make tags=invoicesrpc	
> make tags=invoicesrpc install
> ```

Read #2369 in more details ;)

D4nte:
Hum, running locally see if I can reproduce the issue.

api_tests/src/actors/actor.ts
```diff
@@ -664,12 +664,10 @@ export class Actor {
     }
 
     public async assertAlphaNotDeployed() {
-        await sleep(3000); // It is meaningless to assert before cnd processes a new block
```
bonomat, line 667: Can you explain why this is not valid anymore please?

bonomat:
The error you've received is because the log-files were busy. 
`    spawn ETXTBSY

      76 | 
      77 |         this.logger.debug(`Using binary ${bin}`);
    > 78 |         this.process = spawn(bin, ["--lnddir", this.dataDir], {
         |                        ^
      79 |             stdio: ["ignore", "ignore", "ignore"], // stdin, stdout, stderr.  These are all logged already.
      80 |         });
      81 | 
`

I had that one before and googled it. No idea why this happened though.

api_tests/src/actors/actor.ts
D4nte: Please see commit message: https://github.com/comit-network/comit-rs/pull/2477/commits/943af8ba3469e5607308761308e6156cd3ba6032

> Stabilize e2e by reducing sleep time
> 
> The sleep was done up to 4*3s in one test: Each time a
`assert*NotDeployed` is called.
> 
> This causes failure as it was toying with the timeout line.

D4nte:
bors retry

D4nte:
> The error you've received is because the log-files were busy.
> ` spawn ETXTBSY
> 
> ```
>   76 | 
>   77 |         this.logger.debug(`Using binary ${bin}`);
> > 78 |         this.process = spawn(bin, ["--lnddir", this.dataDir], {
>      |                        ^
>   79 |             stdio: ["ignore", "ignore", "ignore"], // stdin, stdout, stderr.  These are all logged already.
>   80 |         });
>   81 | 
> ```
> 
> `
> 
> I had that one before and googled it. No idea why this happened though.

Can you clarify what log files you are referring too. This command does not use log files. Or are you saying that lnd is the one that is exiting with an error because it cannot read the lnd dir?

api_tests/src/actors/actor.ts
bonomat: Thanks for pointing this out. Nevertheles, the comment said: 
> It is meaningless to assert before cnd processes a new block

Does this not apply anymore or has this been resolved otherwise?

D4nte:
bors retry

"Post Cache target directory" keeps failing for some reason.

PR #2478, opened 2020-04-17
Fix and enable windows builds

### LINKED ISSUE #2337: GitHub Action CI: Fix Windows Build
# Problem

With https://github.com/comit-network/comit-rs/pull/2321 we build comit-rs using GitHub Actions. 
Unfortunately the windows build is broken due to an unimplemented function - see here: https://github.com/comit-network/comit-rs/blob/93f58a1762c7c7a8da038a1ec2f22dccea1a21cb/cnd/src/lib.rs#L108-L118

There might be further unknown pitfalls.

# Goal

- Fix windows build by implementing said function,
- Enable windows build as part of our GitHub CI.

At this stage we only want the build to pass (cargo tests and e2e test are out of scope).

Note: If there are many other changes to be done, better to halt and discuss.

D4nte:
Resolves #2337

### DIFF
--- a/cnd/src/lib.rs
+++ b/cnd/src/lib.rs
@@ -108,18 +108,18 @@ pub fn data_dir() -> Option<PathBuf> {
         .map(|proj_dirs| proj_dirs.data_dir().to_path_buf())
 }
 
-/// Returns `/Users/[username]/Library/Application Support/Lnd/`.
-/// exists.
-#[cfg(target_os = "macos")]
+/// Returns `/Users/[username]/Library/Application Support/Lnd/` for macos.
+/// Returns `%LOCALAPPDATA%/Lnd for windows.
+/// Returns `~/.lnd` if $HOME exists for linux.
 pub fn lnd_default_dir() -> Option<PathBuf> {
-    directories::ProjectDirs::from("", "", "Lnd")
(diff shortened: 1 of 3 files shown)


cnd/src/lib.rs
```diff
@@ -108,18 +108,18 @@ pub fn data_dir() -> Option<PathBuf> {
...
-#[cfg(target_os = "linux")]
-pub fn lnd_default_dir() -> Option<PathBuf> {
-    directories::UserDirs::new().map(|d| d.home_dir().to_path_buf().join(".lnd"))
+    if cfg!(target_os = "macos") || cfg!(target_os = "windows") {
+        directories::ProjectDirs::from("", "", "Lnd")
+            .map(|proj_dirs| proj_dirs.data_dir().to_path_buf())
+    } else if cfg!(target_os = "linux") {
+        directories::UserDirs::new().map(|d| d.home_dir().to_path_buf().join(".lnd"))
+    } else {
+        panic!("platform not supported.");
```
thomaseizinger, line 121: Should we log and return `None` instead?

D4nte: Indeed, I didn't event notice it was an option 😊Should have returned `None` for windows in the first place!

PR #2479, opened 2020-04-17
ComitLN Integration Test

### LINKED ISSUE #2303: Integration tests for libp2p code
## Problem
We currently don't have any high level/integration tests for libp2p code. We cannot confirm whether everything works how we think it does. These tests will help us test future p2p functionality such as the decentralised orderbook. 

Currently the `finalize` handler in the `NetworkBehaviour` directly calls the `create_watcher` functions for both ledger, passing identity (and selecting the `create_watcher`) base on the current role (Alice/Bob).

As highlighted in #2388 , this is not a _clean_ design but also makes the isolated testing of the `NetworkBehaviour` impossible.

## Goal
Add integration tests that allow us to check whether our libp2p code (protocols) perform as we expect.

## Recommendation
Look at the network behaviour tests of the libp2p chat/identity examples. Implement tests similar to these. Spawning of a new process is not required.

Do not get in the rabbit hole of testing the `create_watcher` function behaviour. The output of the communication should be tested (transmission and selection of the correct identities).

rishflab:
Add test to check announced ethereum lightning swap is finalized

Added a happy path integration test that checks if Alice and
Bob have agreed upon an Ethereum-Lightning swap to execute.
This test is used to check that the libp2p network behaviour
abstraction generates a SwapFinalized event as expected.

Closes #2303

I can't assert_eq on most of the fields because partial_eq is not implemented for the types

### DIFF
--- a/cnd/Cargo.toml
+++ b/cnd/Cargo.toml
@@ -67,6 +67,7 @@ warp = { version = "0.2", default-features = false }
 
 [dev-dependencies]
 base64 = "0.12"
+bitcoin = { version = "0.23", features = ["rand"] }
 bitcoincore-rpc = "0.10.0"
 quickcheck = "0.9.2"
 regex = "1.3"
@@ -74,3 +75,4 @@ serde_urlencoded = "0.6"
 spectral = { version = "0.6", default-features = false }
 tempfile = "3.1.0"
 testcontainers = "0.9"
+
--- a/cnd/src/network/comit_ln.rs
+++ b/cnd/src/network/comit_ln.rs
@@ -578,3 +578,122 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        asset::{ethereum::FromWei, Ether},
+        lightning,
+        network::{test_swarm, DialInformation},
+        swap_protocols::EthereumIdentity,
+    };
+    use digest::Digest;
+    use futures::future;
+    use libp2p::{multiaddr::Multiaddr, PeerId};
+    use rand::thread_rng;
+
+    fn make_alice_swap_params(
+        bob_peer_id: PeerId,
+        bob_addr: Multiaddr,
+        ether: asset::Ether,
+        lnbtc: asset::Lightning,
+        ethereum_absolute_expiry: Timestamp,
+        lightning_cltv_expiry: Timestamp,
+    ) -> HanEtherereumHalightBitcoinCreateSwapParams {
+        HanEtherereumHalightBitcoinCreateSwapParams {
+            role: Role::Alice,
+            peer: DialInformation {
+                peer_id: bob_peer_id,
+                address_hint: Some(bob_addr),
+            },
+            ethereum_identity: EthereumIdentity::from(identity::Ethereum::random()),
+            ethereum_absolute_expiry,
+            ethereum_amount: ether,
+            lightning_identity: lightning::PublicKey::random(),
+            lightning_cltv_expiry,
+            lightning_amount: lnbtc,
+        }
+    }
+
+    fn make_bob_swap_params(
+        alice_peer_id: PeerId,
+        ether: asset::Ether,
+        lnbtc: asset::Lightning,
+        ethereum_absolute_expiry: Timestamp,
+        lightning_cltv_expiry: Timestamp,
+    ) -> HanEtherereumHalightBitcoinCreateSwapParams {
+        HanEtherereumHalightBitcoinCreateSwapParams {
+            role: Role::Bob,
+            peer: DialInformation {
+                peer_id: alice_peer_id,
+                address_hint: None,
+            },
+            ethereum_identity: EthereumIdentity::from(identity::Ethereum::random()),
+            ethereum_absolute_expiry,
+            ethereum_amount: ether,
+            lightning_identity: lightning::PublicKey::random(),
+            lightning_cltv_expiry,
+            lightning_amount: lnbtc,
+        }
+    }
+
+    #[tokio::test]
+    async fn finalize_lightning_ethereum_swap_success() {
+        // arrange
+        let (mut alice_swarm, _, alice_peer_id) =
+            test_swarm::new(ComitLN::new(RootSeed::new_random(thread_rng()).unwrap()));
+        let (mut bob_swarm, bob_addr, bob_peer_id) =
+            test_swarm::new(ComitLN::new(RootSeed::new_random(thread_rng()).unwrap()));
+
+        let ether = Ether::from_wei(9_001_000_000_000_000_000_000u128);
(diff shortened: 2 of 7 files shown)


da-kami:
> Add test to check announced ethereum lightning swap is finalized
> 
> Added a happy path integration test that checks if Alice and
> Bob have agreed upon an Ethereum-Lightning swap to execute.
> This test is used to check that the libp2p network behaviour
> abstraction generates a SwapFinalized event as expected.
> 
> Closes #2303
> 
> I can't assert_eq on most of the fields because partial_eq is not implemented for the types

I think it's OK to derive `Partial_Eq` to the types.

cnd/src/network/comit_ln.rs
```diff
@@ -577,3 +577,182 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+                    swap_params: bob_swap_params,
+                    secret_hash: bob_secret_hash,
+                    ethereum_identity: _bob_ethereum_id,
+                },
+            ) => {
+                assert_eq!(alice_swap_params.peer.peer_id, bob_peer_id);
+                assert_eq!(bob_swap_params.peer.peer_id, alice_peer_id);
+                assert_eq!(alice_local_swap_id, alice_node_id);
+                assert_eq!(bob_local_swap_id, bob_node_id);
+                assert_eq!(alice_secret_hash, bob_secret_hash);
```
da-kami, line 754: We should check on the identities as well, if it is necessary to derive `Partial_Eq` for that I would say it's OK. @thomaseizinger @tcharding what do you think?

tcharding: I don't think its worth checking the identities, we already check that this all works in the unit tests.  This test is just checking we can glue everything together correctly.

rishflab:
> I pushed 4 minor changes to the branch. In general good work @rishflab and @da-kami, the code works and does test that we have everything glued together correctly, which is no small feat. I do have some reservations though; firstly the code needs cleaning up, its a long single function for example. But more importantly, I feel, and please take this with all due respect, that this test is going to be a maintenance burden considering there is a good chance that the way we use network behaviours is going to change considerably. I not sure how much value the test adds in comparison to the maintenance burden. I'm happy to ack it if the test was super cleanly written so the next guy that changes the network behaviour finds it quick and easy to update this test. I wouldn't mind also seeing some comments on the test stating what it aims to test, since typically cfg code inside a module is a unit test. With the five asserts it is not easy to work out what the aim of this test is.

I think the main value a test like this adds is it allows you to change ComitLN and and check if all the libp2p internal wiring has been done correctly without running an e2e test. These sort of test will be really helpful when we start testing more complicated p2p functionality like the propagating a trade through a network of peers. It will be a lot simpler to run and maintain than e2e tests with separate processes.

tcharding:
> > I pushed 4 minor changes to the branch. In general good work @rishflab and @da-kami, the code works and does test that we have everything glued together correctly, which is no small feat. I do have some reservations though; firstly the code needs cleaning up, its a long single function for example. But more importantly, I feel, and please take this with all due respect, that this test is going to be a maintenance burden considering there is a good chance that the way we use network behaviours is going to change considerably. I not sure how much value the test adds in comparison to the maintenance burden. I'm happy to ack it if the test was super cleanly written so the next guy that changes the network behaviour finds it quick and easy to update this test. I wouldn't mind also seeing some comments on the test stating what it aims to test, since typically cfg code inside a module is a unit test. With the five asserts it is not easy to work out what the aim of this test is.
> 
> I think the main value a test like this adds is it allows you to change ComitLN and and check if all the libp2p internal wiring has been done correctly without running an e2e test. These sort of test will be really helpful when we start testing more complicated p2p functionality like the propagating a trade through a network of peers. It will be a lot simpler to run and maintain than e2e tests with separate processes.

That's a fair point, and probably a start in the right direction actually to get us away from relying so much on the e2e tests.

> I should be able to move a lot of the setup code into a reusable function which will make the test easier to understand.

Lets merge it as is and perhaps you could put it on your todo list to clean it up some time when you have a spare hour?  Unless someone else dis-approves of such a move?

tcharding:
Hey @rishflab, if you use the 'Connect with an issue' feature then ZenHub will move this PR to the correct pipeline for you.  If you don't see it perhaps you need the Zenhub extension for your browser.

thomaseizinger:
@rishflab I pushed some commits about how I would go about this test. I hope you find them useful :)

rishflab:
> @rishflab I pushed some commits about how I would go about this test. I hope you find them useful :)

yep awesome thanks! I think the memory transport fixed an issue I was having with tcp port allocation. OS was assigning the same tcp port to each swarm because of the single process test.

thomaseizinger:
> > @rishflab I pushed some commits about how I would go about this test. I hope you find them useful :)
> 
> yep awesome thanks! I think the memory transport fixed an issue I was having with tcp port allocation. OS was assigning the same tcp port to each swarm because of the single process test.

Just squash it all together once the build is finally passing :)

cnd/Cargo.toml
```diff
@@ -11,7 +11,7 @@ anyhow = "1"
 async-trait = "0.1"
 base64 = "0.12.0"
 bigdecimal = "0.1.0"
-bitcoin = { version = "0.23", features = ["use-serde"] }
+bitcoin = { version = "0.23", features = ["use-serde", "rand"] }
```
thomaseizinger, line 14: We only use the `rand` feature in tests so you could also just add a `bitcoin` dependency with the `rand` feature in `dev-dependencies`.

cnd/src/network/test_swarm.rs
```diff
@@ -0,0 +1,43 @@
...
+    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
+        let _ = tokio::spawn(future);
+    }
+}
+
+pub fn new<B: NetworkBehaviour>(behaviour: B) -> (libp2p::Swarm<B>, Multiaddr, PeerId) where <<<B as NetworkBehaviour>::ProtocolsHandler as IntoProtocolsHandler>::Handler as ProtocolsHandler>::InEvent: Clone{
+    let keypair = libp2p::identity::Keypair::generate_ed25519();
+    let peer_id = PeerId::from(keypair.public());
+
+    let transport = libp2p::core::transport::memory::MemoryTransport::default()
```
da-kami, line 23: sweet, memory transport 👍

cnd/src/network/comit_ln.rs
```diff
@@ -578,3 +578,122 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+                ethereum_expiry,
+                lightning_expiry,
+            ),
+        );
+
+        // act
+        let (alice_event, bob_event) = future::join(alice_swarm.next(), bob_swarm.next()).await;
+
+        // assert
+        match (alice_event, bob_event) {
```
da-kami, line 680: really neat now, great work!

PR #2492, opened 2020-04-20
Comitln it joined futures

da-kami:
Still hangs once we spin it off 😐

@thomaseizinger you mentioned, that we would have to `poll()` the swarm in the test. But I think `next_event()` should be enough. No? 
Too late to think now, will have another look tomorrow.

Note: I only worked on the `ComitLN` test. 

Packed the initialisation of the communication into an async function, because I felt it would be necessary, otherwise it would not be properly called within the right context? (am I mistaken?)

There are 4 futures:
1. Future to kick-off the communication for Alice
2. Future to kick-off the communication for Bob
3. Future to listen on the finalized event for Alice
4. Future to listen on the finalized event for Bob

### DIFF
--- a/cnd/src/network/comit_ln.rs
+++ b/cnd/src/network/comit_ln.rs
@@ -577,3 +577,166 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        ethereum::Address,
+        network::{derive_key_pair, oneshot_behaviour, transport, DialInformation, TokioExecutor},
+        swap_protocols::{EthereumIdentity, SwapCommunicationStates, SwapErrorStates},
+    };
+    use anyhow::Context;
+    use bitcoin::secp256k1;
+    use futures::{pin_mut, prelude::*};
+    use libp2p::{
+        core::{muxing::StreamMuxer, upgrade},
+        identity,
+        mplex::MplexConfig,
+        multihash::{Multihash, Sha3_256},
+        secio::SecioConfig,
+        swarm::{Swarm, SwarmBuilder, SwarmEvent},
+        tcp::TcpConfig,
+        Multiaddr, PeerId, Transport,
+    };
+    use rand::{random, thread_rng};
+    use spectral::prelude::*;
+    use std::{str::FromStr, sync::Arc};
+    use tokio::runtime;
+    use uuid::Uuid;
+
+    fn random_swap_digest() -> SwapDigest {
+        SwapDigest::new(Sha3_256::digest(b"hello world"))
+    }
+
+    #[test]
+    fn lightning_to_ethereum_integration_test() {
+        let (alice_key_pair, alice_peer_id, alice_seed) = {
+            let seed = RootSeed::new_random(thread_rng()).unwrap();
+            let key_pair = derive_key_pair(&seed);
+            let peer_id = PeerId::from(key_pair.clone().public());
+            (key_pair, peer_id, seed)
+        };
+
+        let (bob_key_pair, bob_peer_id, bob_seed) = {
+            let seed = RootSeed::new_random(thread_rng()).unwrap();
+            let key_pair = derive_key_pair(&seed);
+            let peer_id = PeerId::from(key_pair.clone().public());
+            (key_pair, peer_id, seed)
+        };
+
+        let mut alice_runtime = runtime::Builder::new()
+            .enable_all()
+            .threaded_scheduler()
+            .thread_stack_size(1024 * 1024 * 8) // the default is 2MB but that causes a segfault for some reason
+            .build()
+            .unwrap();
+
+        let mut alice_swarm = SwarmBuilder::new(
+            transport::build_comit_transport(alice_key_pair).unwrap(),
+            ComitLN::new(alice_seed),
+            alice_peer_id.clone(),
+        )
+        .executor(Box::new(TokioExecutor {
+            handle: alice_runtime.handle().clone(),
+        }))
+        .build();
+
+        // let mut alice_swarm = {
+        //     let transport= transport::build_comit_transport(alice_key_pair).unwrap();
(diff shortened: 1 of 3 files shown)


da-kami:
Note, that I ran into the problem that the swarms are passed to 2 different futures, that's why I wrapped the swarm into `Arc<Mutex<..>>`. The implementation changed several times on the way, and I am quite sure the wrapping is still needed, but did not try to remove it again.

da-kami:
Added another commit, noticed that somehow the initialisation is never triggered.
I figure this might be because the mutex is always locked.

But if that is the case we have to somehow trigger the initialisation from within the loop for listening...?

cnd/src/network/comit_ln.rs
```diff
@@ -577,3 +577,281 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+    }
+
+    async fn start_communication(
+        alice_swarm: Arc<Mutex<Swarm<ComitLN>>>,
+        bob_swarm: Arc<Mutex<Swarm<ComitLN>>>,
+        alice_params: HanEtherereumHalightBitcoinCreateSwapParams,
+        bob_params: HanEtherereumHalightBitcoinCreateSwapParams,
+        alice_node_id: NodeLocalSwapId,
+        bob_node_id: NodeLocalSwapId,
+    ) {
```
tcharding, line 621: Perhaps
```
    struct Arg {
        swarm: Arc<Mutex<Swarm<ComitLN>>>,
        params: HanEtherereumHalightBitcoinCreateSwapParams,
        local_swap_id: NodeLocalSwapId,
    }

    async fn start_communication(alice: Arg, bob: Arg) {
```

cnd/src/network/comit_ln.rs
```diff
@@ -577,3 +577,281 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+        println!("entered init...");
+
+        async {
+            // init bob first, because his swap has to be available for alice' announce
+            // message
+            sleep(time::Duration::from_secs(1));
```
tcharding, line 627: I do not think this will work as you expect, we should probably be using tokio `sleep` not `std::thread::sleep`.  ref: https://blog.hwc.io/posts/rust-futures-threadsleep-and-blocking-calls-inside-async-fn/.  Although from the code I cannot see why the sleeps are here in the first place :)

cnd/src/network/comit_ln.rs
```diff
@@ -577,3 +577,281 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+            let mut alice_guard = alice_swarm.lock().await;
+            alice_guard.initiate_communication(alice_node_id, alice_params);
+            println!("Alice initialised");
+        }
+        .await;
+    }
+
+    #[test]
+    fn lightning_to_ethereum_integration_test() {
```
tcharding, line 642: nit: We currently only support Han/ether to HALight/bitcoin so this name seems a little misleading.

cnd/src/network/comit_ln.rs
```diff
@@ -577,3 +577,281 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+        // future to poll for the finalized event of alice behaviour
+        let alice_future = async {
+            println!("Start listening for events for alice");
+            loop {
+                // sleep(time::Duration::from_millis(300));
+
+                let mut guard = alice_swarm.lock().await;
+
+                let next = guard.next_event().await;
+
```
tcharding, line 755: Are you sure its ok to do this?  I don't think one can hold a lock while calling `await`?

cnd/src/network/comit_ln.rs
```diff
@@ -577,3 +577,281 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+                        continue;
+                    }
+                }
+            }
+        };
+
+        // join all futures
+        let joined = join3(alice_future, bob_future, future_init);
+
+        runtime.block_on(joined);
```
tcharding, line 808: Why do you join the init future into this?  Why not just run it first then join the other two?

cnd/src/network/oneshot_behaviour.rs
```diff
@@ -142,3 +142,152 @@ where
...
+                match bob_swarm_fut.await {
+                    SwarmEvent::NewListenAddr(addr) => return addr,
+                    _ => {}
+                }
+            }
+        });
+
+        alice_swarm.register_addresses(bob_peer_id.clone(), vec![bob_addr]);
+
+        alice_swarm.send(bob_peer_id, Message);
```
tcharding, line 256: This is not going to work because polling for alice_swarm is never initiated.

cnd/src/network/comit_ln.rs
```diff
@@ -577,3 +577,281 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+            ethereum_amount: asset::Ether::zero(),
+            lightning_identity: lightning_id,
+            lightning_cltv_expiry: Timestamp::from(10),
+            lightning_amount: asset::Lightning::from_sat(0),
+        };
+
+        let swap_params_bob = HanEtherereumHalightBitcoinCreateSwapParams {
+            role: Role::Bob,
+            peer: dial_info, /* not relevant for Bob, should be optional or somehow related
+                              * to role */
```
tcharding, line 718: TIL; This comment is not true, the dial info of alice is passed in by Bob since he learned it during the negotiation phase.

da-kami: Yeah, eventually Bob could check if an announce connection opened by Alice was actually opened by Alice' peer-id that he learned during negotiation. 

Thanks @tcharding

da-kami:
closing this in favour of https://github.com/comit-network/comit-rs/tree/comitln-it
Packing `initiate_communication` is an unnecessary overkill.

Issue #2398, opened 2020-04-06
Remove conversion between `NodeLocalSwapId` and `SwapId`


tcharding:
## Problem

We currently have two identifiers for swaps:
1. A local swap identifier: generated by each node when a swap is posted to the REST API and used in the REST API route.
2. Synced/shared swap identifier, created by Bob and shared with Alice.

This leads to potentially buggy code if/when we mix these up.  We currently have two types but the usage is clunky.  `NodeLocalSwapId` and `SwapId`.

## Goal

Make good use of the type system to be safe.

Have a nice clean abstraction and data types that correctly model our desired usage.  On top of this, make it hard to write buggy code.

We should remove any possible conversion between both swaps (and remove `pub` on inner field if present).
We should only have a mapping between a given `NodeLocalSwapId` and a given `SwapId` (from memory/db).

## Recommendation

Don't changes types across the whole codebase, duplicate some code from rfc003 if necessary.
Consider making the inner field of `SwapId` private to avoid further misusage.

D4nte:
After giving more thought on this issue after the grooming, I'd like to invite the dev (and possibly @thomaseizinger) who grabs this issue to consider the following:

The REST API uses the `local swap id`, meaning the controler needs to be able to find a given swap using the local id.

The COMIT API uses the `shared swap id`, meaning the network behaviour/swap needs to be able to find a given swap using the shared id.

We currently use the same `Uuid` to get the shared id from the local id: 

https://github.com/comit-network/comit-rs/blob/70e9010198c7f6d0ee75a1d31d8c093ffc8c1836/cnd/src/http_api/routes.rs#L48

This can only be incorrect as the local id for Alice is totally unrelated to the shared id (generated by Bob).

The relation between local and shared ids needs to be saved in the DB as, again, they are not correlated for Alice. Hence, this should be part of the DB design spike (@thomaseizinger what is the tracking issue for the spike). The fact that it is store in the DB *might* be considered as part of this ticket.

The way I see it is that in memory, there must be 2 hashmaps:
- `id1 -> id2`
- `id2 -> swap`

id1 and id2 are shared id and local id, either way is fine, I would favor id2 to be local id as I expect more call on the HTTP API than on the network, giving a very slight performance advantage.

In the DB, there are various ways to store this information so I'll let the dev doing the spike reason about it.

tcharding:
Putting this back in 'groomed' because not confident I can get this done and merged before sprint closes tomorrow.

PR #2503, opened 2020-04-22
Local and shared swap identifiers

tcharding:
During the LN squad we got a little loose with how we handled the `SwapId` type.  Specifically, we wanted to [ab]use the `LedgerStates` and its associated traits (get, insert, update) which use the `SwapId` as a parameter.  In order to do so we munged the swap id and a newly added `NodeLocalSwapId`.  This led to debugging issues during the LN squad but was left till after merge to clean up.  This PR is the clean up of those hacks.

Adds two new types:
- `LocalSwapId` (this is just a rename of `NodeLocalSwapId`): Used as the swap identifier throughout the REST API, created by each node (Alice and Bob) when a swap is created by the user via REST endpoint.
- `ShardeSwapId`: Created by Bob and shared with Alice.  Used throughout the network layer to identify swaps.

There are a bunch of patches in this PR, each has a decent commit log for reviewers if interested.  The overall strategy is to duplicate the state traits, one (unchanged) into `swap_protocols/rfc003` and the new one stays in `swap_protocols` having all the uses of `SwapId` replaced with `LocalSwapId`.

The new spilt protocol network layer code is then cleaned up to use `SharedSwapId`.

Please note: `SwapId` is left in place (albeit moved to the `rfc003` module), this PR makes a good push towards deprecating `rfc003`.

Resolves: #2398


cnd/src/swap_protocols/swap_id.rs
```diff
@@ -38,3 +38,39 @@ impl fmt::Display for LocalSwapId {
...
+}
+
+impl FromStr for SharedSwapId {
+    type Err = uuid::Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Uuid::from_str(s).map(SharedSwapId)
+    }
+}
+
+impl From<Uuid> for SharedSwapId {
```
D4nte, line 73: I don't think this trait is needed. `Uuid` should just be an internal implementation detail.

cnd/src/swap_protocols/swap_id.rs
```diff
@@ -38,3 +38,39 @@ impl fmt::Display for LocalSwapId {
...
+    fn from(uuid: Uuid) -> Self {
+        SharedSwapId(uuid)
+    }
+}
+
+impl From<SharedSwapId> for Uuid {
```
D4nte, line 79: I don't think this trait is needed. `Uuid` should just be an internal implementation detail.

cnd/src/main.rs
```diff
@@ -121,10 +122,14 @@ fn main() -> anyhow::Result<()> {
         macaroon_path: settings.lightning.lnd.readonly_macaroon_path.clone(),
     };
 
-    // Han protocol
+    // RCF003 protocol
+    let rfc003_alpha_ledger_states = Arc::new(rfc003::LedgerStates::default());
+    let rfc003_beta_ledger_states = Arc::new(rfc003::LedgerStates::default());
+    let swap_communication_states = Arc::new(SwapCommunicationStates::default());
+
+    // Han/HErc20 protocols (A.K.A split protocols)
```
D4nte, line 130:
```suggestion
    // Han/Herc20 protocols (A.K.A split protocols)
```

cnd/src/main.rs
```diff
@@ -138,31 +143,35 @@ fn main() -> anyhow::Result<()> {
...
+        beta_ledger_states: Arc::clone(&&rfc003_beta_ledger_states),
         swap_communication_states,
         swap_error_states,
         seed,
         db: database,
         swarm: swarm.clone(),
     };
 
+    // split protocols
+    let facade2 = Facade2 {
```
D4nte, line 169:
```suggestion
    let split_protocols_facade = SplitProtocolsFacade {
```

cnd/src/network.rs
```diff
@@ -215,33 +218,46 @@ fn derive_key_pair(seed: &RootSeed) -> Keypair {
 #[derive(NetworkBehaviour)]
 #[allow(missing_debug_implementations)]
 pub struct ComitNode {
+    /// The rfc003 network behaviour.
     comit: Comit,
+    /// The new spilt protocols network behaviour.
```
D4nte, line 223:
```suggestion
    /// The new split protocols network behaviour.
```

cnd/src/swap_protocols/swap_id.rs
```diff
@@ -50,6 +50,13 @@ impl fmt::Display for LocalSwapId {
 #[derive(Serialize, Deserialize, Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct SharedSwapId(Uuid);
 
+impl SharedSwapId {
+    /// Construct an all zero swap id (for testing).
+    pub fn nil() -> Self {
+        SharedSwapId(Uuid::nil())
+    }
+}
```
D4nte, line 58:
```suggestion
#[cfg(test)]
impl SharedSwapId {
    /// Construct an all zero swap id (for testing).
    pub fn nil() -> Self {
        SharedSwapId(Uuid::nil())
    }
}
```

Don't push to prod code if it is only for test.

cnd/src/main.rs
D4nte: This is actually a good name, Or even better: rename `Facade` to `Rfc003Facade` if you feel keen.

cnd/src/swap_protocols/swap_id.rs
tcharding: Removed, thanks man.

cnd/src/swap_protocols/swap_id.rs
tcharding: Gone!

cnd/src/main.rs
tcharding:
```
tobin@ares|~/build/github.com/comit-network/spikes on NAT-traversal
± gg -i herc20
0021-protocol-naming.adoc:15:- `HErc20` - (atomic swaps using ) **H**TLCs for the **Erc20** asset
```

tcharding: Damn fancy acronyms, who did that?

cnd/src/main.rs
tcharding: Cool, I like it.  Will use `Rfc003Facade` in line with the other renames.  thanks

cnd/src/network.rs
tcharding: catch!

cnd/src/swap_protocols/swap_id.rs
tcharding: Yes, thanks good idea - schooled me too.  Cheers man.

tcharding:
Thanks for the review @D4nte, top review.  

> Also, there are a number of typos in the commit message of 08e27e3 (s/spilt/split, s/hand/han, etc).

We will make a kernel programmer out of you yet :)

Cargo.lock
```diff
@@ -45,9 +45,9 @@ dependencies = [
```
luckysori, line 45: :question: Did a `cargo update` sneak in by mistake? :stuck_out_tongue:

cnd/src/main.rs
luckysori, line 130: :wrench: :clown_face: 
```suggestion
    // HAN/HErc20 protocols (A.K.A split protocols)
```

cnd/src/network.rs
```diff
@@ -215,33 +218,46 @@ fn derive_key_pair(seed: &RootSeed) -> Keypair {
 #[derive(NetworkBehaviour)]
 #[allow(missing_debug_implementations)]
 pub struct ComitNode {
```
luckysori, line 220: :thinking: A struct with these many fields should be illegal.

cnd/src/network.rs
```diff
@@ -215,33 +218,46 @@ fn derive_key_pair(seed: &RootSeed) -> Keypair {
...
+    /// The rfc003 network behaviour.
     comit: Comit,
+    /// The new spilt protocols network behaviour.
     comit_ln: ComitLN,
```
luckysori, line 224: :question: Would it make sense to call this one `Comit` and the old one `ComitRfc003`?

cnd/src/network/halight.rs
```diff
@@ -0,0 +1,35 @@
+use crate::swap_protocols::{
```
luckysori, line 1: :question: Maybe a dumb question, but do all these files that orchestrate a swap need to live in the `network` module? None of them seem to care about being called by the network later.

cnd/src/network.rs
tcharding: Try this one from the Linux kernel
```
/**
 * struct device - The basic device structure
 * @parent:	The device's "parent" device, the device to which it is attached.
 * 		In most cases, a parent device is some sort of bus or host
 * 		controller. If parent is NULL, the device, is a top-level device,
 * 		which is not usually what you want.
 * @p:		Holds the private data of the driver core portions of the device.
 * 		See the comment of the struct device_private for detail.
 * @kobj:	A top-level, abstract class from which other classes are derived.
 * @init_name:	Initial name of the device.
 * @type:	The type of device.
 * 		This identifies the device type and carries type-specific
 * 		information.
 * @mutex:	Mutex to synchronize calls to its driver.
 * @lockdep_mutex: An optional debug lock that a subsystem can use as a
 * 		peer lock to gain localized lockdep coverage of the device_lock.
 * @bus:	Type of bus device is on.
 * @driver:	Which driver has allocated this
 * @platform_data: Platform data specific to the device.
 * 		Example: For devices on custom boards, as typical of embedded
 * 		and SOC based hardware, Linux often uses platform_data to point
 * 		to board-specific structures describing devices and how they
 * 		are wired.  That can include what ports are available, chip
 * 		variants, which GPIO pins act in what additional roles, and so
 * 		on.  This shrinks the "Board Support Packages" (BSPs) and
 * 		minimizes board-specific #ifdefs in drivers.
 * @driver_data: Private pointer for driver specific info.
 * @links:	Links to suppliers and consumers of this device.
 * @power:	For device power management.
 *		See Documentation/driver-api/pm/devices.rst for details.
 * @pm_domain:	Provide callbacks that are executed during system suspend,
 * 		hibernation, system resume and during runtime PM transitions
 * 		along with subsystem-level and driver-level callbacks.
 * @pins:	For device pin management.
 *		See Documentation/driver-api/pinctl.rst for details.
 * @msi_list:	Hosts MSI descriptors
 * @msi_domain: The generic MSI domain this device is using.
 * @numa_node:	NUMA node this device is close to.
 * @dma_ops:    DMA mapping operations for this device.
 * @dma_mask:	Dma mask (if dma'ble device).
 * @coherent_dma_mask: Like dma_mask, but for alloc_coherent mapping as not all
 * 		hardware supports 64-bit addresses for consistent allocations
 * 		such descriptors.
 * @bus_dma_limit: Limit of an upstream bridge or bus which imposes a smaller
 *		DMA limit than the device itself supports.
 * @dma_pfn_offset: offset of DMA memory range relatively of RAM
 * @dma_parms:	A low level driver may set these to teach IOMMU code about
 * 		segment limitations.
 * @dma_pools:	Dma pools (if dma'ble device).
 * @dma_mem:	Internal for coherent mem override.
 * @cma_area:	Contiguous memory area for dma allocations
 * @archdata:	For arch-specific additions.
 * @of_node:	Associated device tree node.
 * @fwnode:	Associated device node supplied by platform firmware.
 * @devt:	For creating the sysfs "dev".
 * @id:		device instance
 * @devres_lock: Spinlock to protect the resource of the device.
 * @devres_head: The resources list of the device.
 * @knode_class: The node used to add the device to the class list.
 * @class:	The class of the device.
 * @groups:	Optional attribute groups.
 * @release:	Callback to free the device after all references have
 * 		gone away. This should be set by the allocator of the
 * 		device (i.e. the bus driver that discovered the device).
 * @iommu_group: IOMMU group the device belongs to.
 * @iommu_fwspec: IOMMU-specific properties supplied by firmware.
 * @iommu_param: Per device generic IOMMU runtime data
 *
 * @offline_disabled: If set, the device is permanently online.
 * @offline:	Set after successful invocation of bus type's .offline().
 * @of_node_reused: Set if the device-tree node is shared with an ancestor
 *              device.
 * @state_synced: The hardware state of this device has been synced to match
 *		  the software state of this device by calling the driver/bus
 *		  sync_state() callback.
 * @dma_coherent: this particular device is dma coherent, even if the
 *		architecture supports non-coherent devices.
 *
 * At the lowest level, every device in a Linux system is represented by an
 * instance of struct device. The device structure contains the information
 * that the device model core needs to model the system. Most subsystems,
 * however, track additional information about the devices they host. As a
 * result, it is rare for devices to be represented by bare device structures;
 * instead, that structure, like kobject structures, is usually embedded within
 * a higher-level representation of the device.
 */
struct device {
	struct kobject kobj;
	struct device		*parent;

	struct device_private	*p;

	const char		*init_name; /* initial name of the device */
	const struct device_type *type;

	struct bus_type	*bus;		/* type of bus device is on */
	struct device_driver *driver;	/* which driver has allocated this
					   device */
	void		*platform_data;	/* Platform specific data, device
					   core doesn't touch it */
	void		*driver_data;	/* Driver data, set and get with
					   dev_set_drvdata/dev_get_drvdata */
#ifdef CONFIG_PROVE_LOCKING
	struct mutex		lockdep_mutex;
#endif
	struct mutex		mutex;	/* mutex to synchronize calls to
					 * its driver.
					 */

	struct dev_links_info	links;
	struct dev_pm_info	power;
	struct dev_pm_domain	*pm_domain;

#ifdef CONFIG_GENERIC_MSI_IRQ_DOMAIN
	struct irq_domain	*msi_domain;
#endif
#ifdef CONFIG_PINCTRL
	struct dev_pin_info	*pins;
#endif
#ifdef CONFIG_GENERIC_MSI_IRQ
	struct list_head	msi_list;
#endif

	const struct dma_map_ops *dma_ops;
	u64		*dma_mask;	/* dma mask (if dma'able device) */
	u64		coherent_dma_mask;/* Like dma_mask, but for
					     alloc_coherent mappings as
					     not all hardware supports
					     64 bit addresses for consistent
					     allocations such descriptors. */
	u64		bus_dma_limit;	/* upstream dma constraint */
	unsigned long	dma_pfn_offset;

	struct device_dma_parameters *dma_parms;

	struct list_head	dma_pools;	/* dma pools (if dma'ble) */

#ifdef CONFIG_DMA_DECLARE_COHERENT
	struct dma_coherent_mem	*dma_mem; /* internal for coherent mem
					     override */
#endif
#ifdef CONFIG_DMA_CMA
	struct cma *cma_area;		/* contiguous memory area for dma
					   allocations */
#endif
	/* arch specific additions */
	struct dev_archdata	archdata;

	struct device_node	*of_node; /* associated device tree node */
	struct fwnode_handle	*fwnode; /* firmware device node */

#ifdef CONFIG_NUMA
	int		numa_node;	/* NUMA node this device is close to */
#endif
	dev_t			devt;	/* dev_t, creates the sysfs "dev" */
	u32			id;	/* device instance */

	spinlock_t		devres_lock;
	struct list_head	devres_head;

	struct class		*class;
	const struct attribute_group **groups;	/* optional groups */

	void	(*release)(struct device *dev);
	struct iommu_group	*iommu_group;
	struct iommu_fwspec	*iommu_fwspec;
	struct iommu_param	*iommu_param;

	bool			offline_disabled:1;
	bool			offline:1;
	bool			of_node_reused:1;
	bool			state_synced:1;
#if defined(CONFIG_ARCH_HAS_SYNC_DMA_FOR_DEVICE) || \
    defined(CONFIG_ARCH_HAS_SYNC_DMA_FOR_CPU) || \
    defined(CONFIG_ARCH_HAS_SYNC_DMA_FOR_CPU_ALL)
	bool			dma_coherent:1;
#endif
};
```

cnd/src/network.rs
tcharding: Yes, will add this to my todo list, thanks.

cnd/src/network/halight.rs
tcharding: Nice observation, these are literally just a wrapper around `create_watcher` that adds tracing.  I didn't even notice that while doing this.  Adding that to my list also, I'll do a follow up PR.  thanks

PR #2504, opened 2020-04-22
Do not panic during halight communication if no lnd

### LINKED ISSUE #2489: Instantiating a new LndConnector should be safe and easy
## Problem

Currently, if you just instantiate an `LndConnector` you have to call `read_certificate` and `read_macaroon` on the instance **otherwise it will panic at runtime**:

https://github.com/comit-network/comit-rs/blob/5a7b9af1e304dab3b0c54b6404a769237fed8dd6/cnd/src/network.rs#L1166-L1171

This API is error prone and should be re-worked.
The reason why this is tricky is because the user might not have `lnd` running and have no intentions of doing a swap that involves `halight`, hence we lazy-load the certificate and the macaroon only when creating a `halight` swap.

## Goal

- Do not panic if lnd files are not present during execution while still allowing the user to start cnd without those fails.
- Have a nice, typesafe API for creating instances of `LndConnector`
- Allow users to start cnd without requiring them to have those files available as they might not have `lnd` on their system.

D4nte:
In the main, only the path to lnd certificate and macaroon
were passed. The files were read only during the finalize
step of the communication, leading to a panic if not
present or unreadable.

The logic has been modified to provide the following
functionality:
- An attempt is made to read the files when starting cnd,
the result is an `Option` passed down to the swarm.
- When a halight swap is created on the REST API,
the `Option` is checked and an error is returned to the
caller if it's `None`
- During the finalize phase, we assume that the check was
previously done so we should not be able to finalize a
halight swap if the `Option` is `None`. However, if there
is a bug, we log an error a do not finalize (nor panic).

Resolves #2489

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ## Fixed
--   Fix windows build
+-   Fix windows build.
+-   Return an error when creating a halight swap if lnd certificate or macaroon are unavailable instead of failing silently.
 
 ## [0.7.3] - 2020-04-14
 
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -1188,12 +1188,11 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
         match event {
             comit_ln::BehaviourOutEvent::SwapFinalized {
                 local_swap_id,
-                swap_params,
+                swap_params: create_swap_params,
                 secret_hash,
                 ethereum_identity,
             } => {
-                let role = swap_params.role;
-                let create_swap_params = swap_params;
+                let role = create_swap_params.role;
 
                 match self.lnd_connector_params {
                     None => { tracing::error!("Internal Failure: lnd connectors are not initialised, no action has been taken. This should be unreachable.") }
@@ -1216,7 +1215,6 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
                                     let asset = create_swap_params.ethereum_amount.clone();
                                     let ledger = ledger::Ethereum::default();
                                     let expiry = create_swap_params.ethereum_absolute_expiry;
-                                    let secret_hash = secret_hash;
 
                                     han::new_han_ethereum_ether_swap(
                                         local_swap_id,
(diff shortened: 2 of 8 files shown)


cnd/src/network.rs
```diff
@@ -1188,90 +1197,82 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
                 let role = swap_params.role;
                 let create_swap_params = swap_params;
```
tcharding, line 1198: I know you didn't touch this but I don't see the reason for it, the prefix 'create_' adds no extra information IMO.

cnd/src/network.rs
```diff
@@ -1188,90 +1197,82 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
...
+
+                                tokio::task::spawn({
+                                    let connector = self.ethereum_connector.clone();
+                                    let alice_ethereum_identity = create_swap_params.ethereum_identity;
+                                    let bob_ethereum_identity = ethereum_identity;
+
+                                    let asset = create_swap_params.ethereum_amount.clone();
+                                    let ledger = ledger::Ethereum::default();
+                                    let expiry = create_swap_params.ethereum_absolute_expiry;
+                                    let secret_hash = secret_hash;
```
tcharding, line 1221: Cut'n'pasta / sed residue?

D4nte: This was here before :)

tcharding: Oh, we should definitely remove it then, its probably just a relic of hacking fast during LN squad.

CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ## Fixed
--   Fix windows build
+-   Fix windows build.
+-   Do not panic if lnd is not available during a halight swap execution.
```
thomaseizinger, line 12: Can we rephrase this so it makes sense for the user? I think we should be talking about certificate and macaroon files here and what it "panic" means if you want to use that word. For example, only the tokio worker thread would have panicked which is something the user doesn't even see.

The only observable behaviour is that the swap doesn't do anything.

cnd/src/swap_protocols/facade.rs
```diff
@@ -81,8 +81,8 @@ impl Facade {
         &self,
         id: LocalSwapId,
         swap_params: HanEtherereumHalightBitcoinCreateSwapParams,
-    ) {
-        self.swarm.initiate_communication(id, swap_params).await;
+    ) -> anyhow::Result<()> {
+        self.swarm.initiate_communication(id, swap_params).await
```
thomaseizinger, line 85: It is a question of style but I prefer to never return `Result`s directly because I think it is less readable of what is going on.
```suggestion
        self.swarm.initiate_communication(id, swap_params).await?;
        
        Ok(())
```

tcharding: I agree with you @thomaseizinger for a longer function but on a one line function like this I think that the added `?` and `Ok(())` are just noise.

PR #2505, opened 2020-04-22
Follow up clean up work of network module

tcharding:
Clean up the network layer as suggested by @luckysori during review of #2085 

Two patches:

- Move wrapper functions out of the `network` module and into the `swap_protocols` module with the functions they wrap.

- Re-name `Comit` network behaviour to `Rfc003Comit` in line with the other types #2085 renamed.

Please Note: I elected to _not_ rename `ComitLN` to `Comit`, although this will probably happen later, at the moment this behaviour is not complete so leaving the `LN` on it seems fitting.  It keeps the name change from breaking anything also (renaming one network behaviour to the same name as another at the same time could prevent the compiler from catching all the call sites that need fixing).


tcharding:
> Awesome, thanks for doing this

My pleasure, you keep the good reviews coming and I'll keep the follow up patches coming - team WIN.

PR #2506, opened 2020-04-22
Add peer-id check for Bob

### LINKED ISSUE #2389: Bob to ensure that an incoming swap comes from the expected counterparty

da-kami:
resolves #2389

Note that I packed the rejection upon Bob providing an `address_hint` into a separate PR https://github.com/comit-network/comit-rs/pull/2507
I don't think we need this at the moment.

### DIFF
--- a/api_tests/src/actors/actor.ts
+++ b/api_tests/src/actors/actor.ts
@@ -354,42 +354,32 @@ export class Actor {
         }
     }
 
-    public async init() {
+    public async init(config?: {
+        maxTimeoutSecs: number;
+        tryIntervalSecs: number;
+    }) {
         if (!this.swap) {
             throw new Error("Cannot init nonexistent swap");
         }
 
         const response = await this.swap.tryExecuteSirenAction<LedgerAction>(
             "init",
-            {
-                maxTimeoutSecs: 30,
-                tryIntervalSecs: 1,
-            }
+            config ? config : Actor.defaultActionConfig
         );
         await this.swap.doLedgerAction(response.data);
     }
 
-    public async waitForAction(
-        action: string,
-        maxTimeoutSecs?: number,
-        tryIntervalSecs?: number
-    ) {
-        if (!this.swap) {
-            throw new Error("Cannot init nonexistent swap");
-        }
-
-        return this.swap.tryExecuteSirenAction<LedgerAction>(action, {
-            maxTimeoutSecs: maxTimeoutSecs ? maxTimeoutSecs : 3,
-            tryIntervalSecs: tryIntervalSecs ? tryIntervalSecs : 1,
-        });
-    }
-
-    public async fund() {
+    public async fund(config?: {
+        maxTimeoutSecs: number;
+        tryIntervalSecs: number;
+    }) {
         if (!this.swap) {
             throw new Error("Cannot fund nonexistent swap");
         }
 
-        const txid = await this.swap.fund(Actor.defaultActionConfig);
+        const txid = await this.swap.fund(
+            config ? config : Actor.defaultActionConfig
+        );
 
         if (txid instanceof Transaction) {
             await txid.status(1);
--- a/api_tests/src/actors/actor.ts
+++ b/api_tests/src/actors/actor.ts
@@ -369,6 +369,21 @@ export class Actor {
         await this.swap.doLedgerAction(response.data);
     }
 
+    public async waitForAction(
+        action: string,
+        maxTimeoutSecs?: number,
+        tryIntervalSecs?: number
+    ) {
+        if (!this.swap) {
+            throw new Error("Cannot init nonexistent swap");
+        }
+
+        return this.swap.tryExecuteSirenAction<LedgerAction>(action, {
+            maxTimeoutSecs: maxTimeoutSecs ? maxTimeoutSecs : 3,
+            tryIntervalSecs: tryIntervalSecs ? tryIntervalSecs : 1,
+        });
+    }
+
     public async fund() {
         if (!this.swap) {
             throw new Error("Cannot fund nonexistent swap");
--- a/api_tests/tests/ether_halight.ts
+++ b/api_tests/tests/ether_halight.ts
@@ -3,9 +3,9 @@
  * @ledger lightning
  */
 
-import { twoActorTest } from "../src/actor_test";
 import SwapFactory from "../src/actors/swap_factory";
 import { sleep } from "../src/utils";
+import { twoActorTest } from "../src/actor_test";
 
 it(
     "han-ethereum-ether-halight-lightning-bitcoin-alice-redeems-bob-redeems",
@@ -34,5 +34,37 @@ it(
 
         await alice.assertBalances();
         await bob.assertBalances();
+    })
+);
+
+it(
+    "han-ethereum-ether-halight-lightning-bitcoin-alice-announces-with-wrong-peer-id",
+    twoActorTest(async ({ alice, bob }) => {
+        const bodies = (await SwapFactory.newSwap(alice, bob))
+            .hanEthereumEtherHalightLightningBitcoin;
+
+        // Simulate that Bob is awaiting a swap from a different peer-id than Alice node's peer-id.
+        bodies.bob.peer.peer_id =
+            "QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N";
+        await bob.createSwap(bodies.bob);
+        await sleep(500);
+
+        await alice.createSwap(bodies.alice);
+
+        // Due to the asynchronous nature of the announce protocol (and the fact that there is not error handling (yet)),
+        // this test only tests that the "init" action does not become available.
+        //
+        // Scenario details:
+        // 0) Alice and Bob agree on the swap params (negotiation); Alice gives Bob a peer-id that is not her node's peer-id.
+        // 1) Bob posts swap, retrieves local swap-ID
+        // 2) Alice posts swap, retrieves local swap-ID
+        //      2.1) Internally the announce protocol runs into an Error because Alice' peer-id does not match the one she gave Bob.
+        //              This Error, is however not visible to the outside.
+        //              The application is responsible for removing swaps that do not move forward after a given time.
+        const aliceResponsePromise = alice.waitForAction("init");
+        return expect(aliceResponsePromise).rejects.toThrowError();
+        // Same for Bob's side
+        const bobResponsePromise = bob.waitForAction("fund");
+        return expect(bobResponsePromise).rejects.toThrowError();
     })
 );
--- a/api_tests/tests/ether_halight.ts
+++ b/api_tests/tests/ether_halight.ts
@@ -51,6 +51,11 @@ it(
 
         await alice.createSwap(bodies.alice);
 
+        const config = {
+            maxTimeoutSecs: 1,
+            tryIntervalSecs: 0.3,
+        };
+
         // Due to the asynchronous nature of the announce protocol (and the fact that there is not error handling (yet)),
         // this test only tests that the "init" action does not become available.
         //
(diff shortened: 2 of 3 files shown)


cnd/src/network/comit_ln.rs
```diff
@@ -306,18 +306,47 @@ impl NetworkBehaviourEventProcess<announce::behaviour::BehaviourOutEvent> for Co
     fn inject_event(&mut self, event: announce::behaviour::BehaviourOutEvent) {
         match event {
             announce::behaviour::BehaviourOutEvent::ReceivedAnnouncement { peer, mut io } => {
+                if let Some(local_swap_id) =
```
da-kami, line 309: Note that initially I munched the error inside the 
```
if let Some(local_swap_id) =
                    self.swaps_waiting_for_announcement.remove(&io.swap_digest)	                    self.swaps_waiting_for_announcement.remove(&io.swap_digest)
```

block. 

But that has the disadvantage that we remove the event for a few milliseconds (and then add it again).

I feel the current solution is the cleanest, as it first checks if there are any error without modifying the HashMap and only starts modifying if there are no errors.

tcharding: Can we have some comments in the code please about this.  I was stumped as to why the error handling was all at the top of the function until I read this comment here, it did not occur to me that we would have to add the swap back into the hash map while reading the code.

nit: We don't need the second `if let Some(local_swap_id)` construct, we could use `local_swap_id = self.swaps_waiting_for_announcement.remove(&io.swap_digest).unwrap()` with a comment saying `// unwrap is ok since we check this above` or some such thing.

cnd/src/network/comit_ln.rs
```diff
@@ -306,18 +306,47 @@ impl NetworkBehaviourEventProcess<announce::behaviour::BehaviourOutEvent> for Co
...
+                    self.swaps_waiting_for_announcement.get(&io.swap_digest)
+                {
+                    let create_swap_params = self.swaps.get(&local_swap_id).unwrap();
```
tcharding, line 312: Since you are touching this code can we clean up these `unwraps` while we are at it?  Either with and explicit check on the return value and a trace for error OR use an expect with a message saying why this should never be hit.

da-kami: I know that this is possible, but I would keep the if-let construct here. If we ever change the code we might overlook the `unwrap`. 
I assume that the if-let "makes you think" (because there is an if without an else close), but I don't see this as a big problem.

cnd/src/network/comit_ln.rs
da-kami: I added comments.
Additionally changed the logic to `match` instead of `if let` because I think it makes it more readable. (then the `else` case is clearly `None`)

api_tests/src/actors/actor.ts
```diff
@@ -369,6 +369,30 @@ export class Actor {
         await this.swap.doLedgerAction(response.data);
     }
 
+    public async init_does_not_become_available() {
```
thomaseizinger, line 372: TypeScript uses camelCase for function names :)

Also, I am not quite sure, whether this is an assertion or not? I don't see an assertion but the name kind of suggests that there should maybe be one?

api_tests/tests/ether_halight.ts
```diff
@@ -3,36 +3,70 @@
  * @ledger lightning
  */
 
-import { twoActorTest } from "../src/actor_test";
 import SwapFactory from "../src/actors/swap_factory";
 import { sleep } from "../src/utils";
+import { twoActorTest } from "../src/actor_test";
+
+describe("E2E: Ethereum/ether - Lightning/bitcoin", () => {
```
thomaseizinger, line 10: Why did you introduce `describe` blocks again? 

I feel like they just add a level of indentation to the file and don't provide any value outside of that.

api_tests/tests/ether_halight.ts
```diff
@@ -3,36 +3,70 @@
...
+            // this test only tests that the "init" action does not become available.
+            //
+            // Scenario details:
+            // 0) Alice and Bob agree on the swap params (negotiation); Alice gives Bob a peer-id that is not her node's peer-id.
+            // 1) Bob posts swap, retrieves local swap-ID
+            // 2) Alice posts swap, retrieves local swap-ID
+            //      2.1) Internally the announce protocol runs into an Error because Alice' peer-id does not match the one she gave Bob.
+            //              This Error, is however not visible to the outside.
+            //              The application is responsible for removing swaps that do not move forward after a given time.
+            const aliceResponsePromise = alice.init_does_not_become_available();
```
thomaseizinger, line 65: I think just calling the regular `init` function and _not_ awaiting here would be sufficient.

api_tests/tests/ether_halight.ts
da-kami: Oh, I thought that was still our best practice. Will remove

thomaseizinger:
Side note: Please keep our commit-message guidelines (https://chris.beams.io/posts/git-commit/) in mind: Commit messages should start with an uppercase letter :)

Also, a good commit message doesn't focus _what_ you did in terms of code changes (anyone who can read code can see that, like adding a peer-id check) but rather and what the overall feature is. The original issue title/description should give you a good hint on what that is.

api_tests/tests/ether_halight.ts
da-kami: I don't agree with this, but maybe I have a knowledge lack here.

To my understanding the announce protocol exchanges several messages async between Alice and Bob, so after "announcing" we have to wait to be sure that init "did not become available". 
In the Rust code we only await that the announce was triggered correctly before returning the local-swap-id, but we don't await all the async messages underneath. 
I would keep this async.

api_tests/src/actors/actor.ts
da-kami: Changed.

D4nte:
See https://github.com/comit-network/comit-rs/pull/2507#issuecomment-618165618

api_tests/tests/ether_halight.ts
thomaseizinger: Differently asked: What is the value that `init_does_not_become_available` adds over the regular `init`?

da-kami:
> See [#2507 (comment)](https://github.com/comit-network/comit-rs/pull/2507#issuecomment-618165618)

I would still merge this @D4nte - it is basically independent of #2507 - would you approve?

da-kami:
Going for merge in one hour unless there are objects, we can still check decide on the follow up PR afterwards.

api_tests/tests/ether_halight.ts
```diff
@@ -36,3 +36,35 @@ it(
...
+        // this test only tests that the "init" action does not become available.
+        //
+        // Scenario details:
+        // 0) Alice and Bob agree on the swap params (negotiation); Alice gives Bob a peer-id that is not her node's peer-id.
+        // 1) Bob posts swap, retrieves local swap-ID
+        // 2) Alice posts swap, retrieves local swap-ID
+        //      2.1) Internally the announce protocol runs into an Error because Alice' peer-id does not match the one she gave Bob.
+        //              This Error, is however not visible to the outside.
+        //              The application is responsible for removing swaps that do not move forward after a given time.
+        const aliceResponsePromise = alice.waitForAction("init");
```
D4nte, line 64: Excellent, can you please trash the `assert*NotDeployed` functions and replace them with this?

thomaseizinger, line 64: How is this different from:

```ts
const aliceResponsePromise = alice.init();
```

I don't see the difference. The first thing that happens inside `init` is waiting for the action to be available.

da-kami: Not sure how that is related; I am not asserting on a `ledger_state` here. Could you please elaborate what you mean?

da-kami: That is a good point, I wonder why I did not see that before. 
Adapted so `init` and `fund` take an optional `config` (tryparams) that would be preferred over the default configuration.

api_tests/tests/ether_halight.ts
```diff
@@ -36,3 +36,40 @@ it(
...
+        // 0) Alice and Bob agree on the swap params (negotiation); Alice gives Bob a peer-id that is not her node's peer-id.
+        // 1) Bob posts swap, retrieves local swap-ID
+        // 2) Alice posts swap, retrieves local swap-ID
+        //      2.1) Internally the announce protocol runs into an Error because Alice' peer-id does not match the one she gave Bob.
+        //              This Error, is however not visible to the outside.
+        //              The application is responsible for removing swaps that do not move forward after a given time.
+        const aliceResponsePromise = alice.init(config);
+        return expect(aliceResponsePromise).rejects.toThrowError();
+        // Same for Bob's side
+        const bobResponsePromise = bob.fund(config);
```
da-kami, line 72: Note: Reflected about this over night - on Bob's side that actually does not totally make sense.
Yes, we would run into a timeout, but given that we only call `init` on Alice side this is incomplete anyway, because Alice did not `fund`.

I think it's OK to leave the test like this, because usually those two actors are in different programs, but it is a bit weird to look at it compared to the other test.

api_tests/tests/ether_halight.ts
```diff
@@ -36,3 +36,40 @@ it(
...
+        //
+        // Scenario details:
+        // 0) Alice and Bob agree on the swap params (negotiation); Alice gives Bob a peer-id that is not her node's peer-id.
+        // 1) Bob posts swap, retrieves local swap-ID
+        // 2) Alice posts swap, retrieves local swap-ID
+        //      2.1) Internally the announce protocol runs into an Error because Alice' peer-id does not match the one she gave Bob.
+        //              This Error, is however not visible to the outside.
+        //              The application is responsible for removing swaps that do not move forward after a given time.
+        const aliceResponsePromise = alice.init(config);
+        return expect(aliceResponsePromise).rejects.toThrowError();
```
thomaseizinger, line 70: You can only return once from a function :)

```suggestion
        await expect(aliceResponsePromise).rejects.toThrowError();
```

api_tests/tests/ether_halight.ts
```diff
@@ -36,3 +36,40 @@ it(
...
+        // this test only tests that the "init" action does not become available.
+        //
+        // Scenario details:
+        // 0) Alice and Bob agree on the swap params (negotiation); Alice gives Bob a peer-id that is not her node's peer-id.
+        // 1) Bob posts swap, retrieves local swap-ID
+        // 2) Alice posts swap, retrieves local swap-ID
+        //      2.1) Internally the announce protocol runs into an Error because Alice' peer-id does not match the one she gave Bob.
+        //              This Error, is however not visible to the outside.
+        //              The application is responsible for removing swaps that do not move forward after a given time.
+        const aliceResponsePromise = alice.init(config);
```
thomaseizinger, line 69:
```suggestion
        const initActionPromise = alice.init(config);
```

api_tests/tests/ether_halight.ts
thomaseizinger: I think it is fine to assert on both sides that whatever their first action would be, they never get to it!

cnd/src/network/comit_ln.rs
```diff
@@ -306,18 +306,55 @@ impl NetworkBehaviourEventProcess<announce::behaviour::BehaviourOutEvent> for Co
     fn inject_event(&mut self, event: announce::behaviour::BehaviourOutEvent) {
         match event {
             announce::behaviour::BehaviourOutEvent::ReceivedAnnouncement { peer, mut io } => {
+                // Check if there are any errors before modifying the hash-map.
+                match self.swaps_waiting_for_announcement.get(&io.swap_digest) {
```
thomaseizinger, line 310: Once https://github.com/comit-network/comit-rs/pull/2479 is merged, it should be easy to have an integration test for this :)

PR #2507, opened 2020-04-22
Reject swap request if Bob specifies an `address_hint`

da-kami:
I don't think we really need this but #2389 states that it should be considered. 
In order to implement this we should think about this further and restrict it role-specific in the `Body` model rather than the `Error` that I throw. This, however, is not trivial. I would just not implement this for now.

This is a follow-up for https://github.com/comit-network/comit-rs/pull/2506

Note: If I misunderstood something or this can actually be handled in a more trivial way let me know :)

### DIFF
--- a/api_tests/src/actors/swap_factory.ts
+++ b/api_tests/src/actors/swap_factory.ts
@@ -94,7 +94,7 @@ export default class SwapFactory {
                     bobIdentities.lightning
                 ),
                 role: "Bob" as "Alice" | "Bob",
-                peer: await makePeer(alice),
+                peer: await makePeerIdOnly(alice),
             },
         };
 
@@ -123,7 +123,7 @@ export default class SwapFactory {
                     bobIdentities.lightning
                 ),
                 role: "Bob" as "Alice" | "Bob",
-                peer: await makePeer(alice),
+                peer: await makePeerIdOnly(alice),
             },
         };
 
@@ -152,7 +152,7 @@ export default class SwapFactory {
                 ),
 
                 role: "Bob" as "Alice" | "Bob",
-                peer: await makePeer(alice),
+                peer: await makePeerIdOnly(alice),
             },
         };
 
@@ -183,7 +183,7 @@ export default class SwapFactory {
                 ),
 
                 role: "Bob" as "Alice" | "Bob",
-                peer: await makePeer(alice),
+                peer: await makePeerIdOnly(alice),
             },
         };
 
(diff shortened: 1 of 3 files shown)


api_tests/src/actors/swap_factory.ts
```diff
@@ -234,6 +234,12 @@ async function makePeer(actor: Actor): Promise<Peer> {
     };
 }
 
+async function makePeerIdOnly(actor: Actor): Promise<Peer> {
+    return {
+        peer_id: await actor.cnd.getPeerId(),
```
thomaseizinger, line 239: I think the deserializer also supports not using an object here but just the peer id as value.

cnd/src/http_api/routes/index.rs
```diff
@@ -89,7 +89,19 @@ pub async fn post_han_ethereum_halight_bitcoin(
 
     facade.save(id, ()).await;
 
-    facade.initiate_communication(id, body.into()).await;
+    // This is kinda dirty, we should properly restrict the DialInformation by role
```
thomaseizinger, line 92: A better way of solving this would be to move the role to the URL, based on that, you can have dedicated structs of each side and let serde handle this error :)

thomaseizinger: That is a bigger refactoring though because it needs changes to the SDK, so I wouldn't do it right now!

cc @bonomat @D4nte : Some nice validation here that we should iterate on our stuff in the e2e tests before pushing it to the SDK!

da-kami:
@thomaseizinger it seems we go forward with merging this, so I will update the comment accordingly. 
I am still not 100% this check is really important but am happy to merge it after the other PR went in.

api_tests/src/actors/swap_factory.ts
da-kami: I understand what you write, but have trouble achieving this in code because of the `await`. 
I could make a variable that awaits the peer-id and then return that variable, but I don't really see much value in that. 
What exactly did you have in mind?

D4nte:
After reading the comments here I feel like we should drop this in favour of moving the role in the create swap url.

thomaseizinger:
> After reading the comments here I feel like we should drop this in favour of moving the role in the create swap url.

Happy to drop it :)
I would suggest to defer the change a bit longer until we are confident that it is a good idea.

da-kami:
> After reading the comments here I feel like we should drop this in favour of moving the role in the create swap url.

I am in favour of that solution - I close the PR and create a follow-up ticket.
@bonomat @thomaseizinger any objections?

bonomat:
> > After reading the comments here I feel like we should drop this in favour of moving the role in the create swap url.
> 
> I am in favour of that solution - I close the PR and create a follow-up ticket.
> @bonomat @thomaseizinger any objections?

No objection your honor

thomaseizinger:
> > After reading the comments here I feel like we should drop this in favour of moving the role in the create swap url.
> 
> I am in favour of that solution - I close the PR and create a follow-up ticket.
> @bonomat @thomaseizinger any objections?

Go for it!

da-kami:
closed in favour of follow up ticket: https://github.com/comit-network/comit-rs/issues/2520

Issue #2158, opened 2020-03-04
Define the interface between REST-API controllers and the database component for LN swaps


thomaseizinger:
## Problem

Once an endpoint is being triggered for creating a new LN swap, the controller has to take the payload deserialized from the HTTP body and convert it to a model that can be passed to the database component.

The payload defined in https://github.com/comit-network/comit-rs/issues/2154 cannot just be straight passed to the database because it is lacking the information that is defined in the route. Hence, we need to create a different struct that fully represents the data and pass it to the database.

## Goals

- Define data types that combine the information from the payload with the information from the model.
- Avoid code duplication up to a reasonable point. The payloads contain similar chunks of information in different combinations (erc20-ln, ln-erc20, ln-ether, etc ...), make use of this fact.

## Recommendation

Note that in order to refer back to the swap later on, we need to create an identifier (a "local" swap id if you want). It is recommended that the controller defines this identifier and passes it along to the database.

Implement the `Save` trait on the database for these types and keep it `unimplemented!()`.

PR #2513, opened 2020-04-23
Define the interface between the controller and the database

tcharding:
Define the interface between controller and database

Users can create an ether/lightning swap via the REST API.  We would like a way to save the data that the user provides into the database.  As a first step in doing so define an interface between the controller and the database.

This patch introduces a new naming convention for the split protocols.  Instead of fully describing them as we currently do, this patch proposes the following convention

HanEthereumEther -> HNETH # Htlc (asset) Native ETH 
HalightLightningBitcoin -> HLNBTC # Htlc Lightning BTC

While not included in this patch this implies:

HanBitcoinBitcoin -> HNBTC # Htlc (asset) Native BTC

And assumes Herc20 will be shortened to:

Herc20EthereumEther -> HERC20 # Htlc ERC-20

This naming convention allows modules to be placed under `swap_protocols` for each concrete protocol i.e., `hnbtc`, `hneth`, `hlnbtc`, and `herc20`.  This allows for the shortening of the super long identifiers we are actively adding e.g., `new_han_ethereum_ether_swap` could become `hneth::new_swap`.

Fixes: #2158


cnd/src/db.rs
```diff
@@ -132,6 +133,16 @@ pub enum Error {
     SwapNotFound,
 }
 
+// Initial attempt at a created swap, this is only for a single protocol pair.
+#[derive(Debug, Clone)]
+pub struct CreatedSwap<A, B> {
+    pub swap_id: LocalSwapId,
+    pub alpha: A,
+    pub beta: B,
+    pub peer: DialInformation,
```
thomaseizinger, line 142: I think we only care about the peer id, the address hint is unlikely to be valid in the long run.

cnd/src/db/save.rs
```diff
@@ -505,3 +505,28 @@ impl Save<Decline> for Sqlite {
         Ok(())
     }
 }
+
+#[async_trait]
+impl<A, B> Save<CreatedSwap<A, B>> for Sqlite
+where
+    A: Save<A>,
```
thomaseizinger, line 512: This is an interesting approach, I like it.

I do think though that this is optimizing too early. For now, I would just write only a `Save` impl where `A` and `B` are filled in.
Doing that will ultimately lead to quite a bit of code duplication between those `Save` impls. I suspect that we should be able to create another trait that goes on these bounds that allows us to write a single `Save` impl where `A` and `B` fulfill this trait. I am pretty certain that that one would be internal to the database though and not the `Save` trait but more something like a conversion to some internal DB model and then calling to diesel.

cnd/src/swap_protocols.rs
```diff
@@ -84,3 +85,28 @@ pub trait RefundAction {
 
     fn refund_action(&self) -> Option<Self::Output>;
 }
+
+#[derive(Clone, Debug)]
+pub struct HanEthereumEther {
```
thomaseizinger, line 90: We already have modules for protocols, can we move these types into there and shorten their names by importing them with more paths?

cnd/src/db.rs
D4nte: I guess the question is what recovery behaviour if the node crash between swap creation and before finalizing.
If we have no hope of recovering, the address hint is indeed useless.
If a node that starts with a swap created but not finalised would try to automatically recover by re-initiating the announcement communication then the address hint will be needed.

cnd/src/http_api/routes/index.rs
```diff
@@ -83,16 +84,24 @@ pub async fn post_han_ethereum_halight_bitcoin(
...
 
-    let reply = warp::reply::reply();
+    let swap_params: HanEtherereumHalightBitcoinCreateSwapParams = body.clone().into();
 
-    let id = LocalSwapId::default();
+    let swap_id = LocalSwapId::default();
+    let reply = warp::reply::reply();
 
-    facade.save(id, ()).await;
+    let swap: CreatedSwap<HanEthereumEther, HalightLightningBitcoin> = CreatedSwap {
```
D4nte, line 92:
```suggestion
    let swap = CreatedSwap::<HanEthereumEther, HalightLightningBitcoin> { {
```

cnd/src/http_api/routes/index.rs
```diff
@@ -83,16 +84,24 @@ pub async fn post_han_ethereum_halight_bitcoin(
...
+        swap_id,
+        alpha: body.alpha,
+        beta: body.beta,
+        peer: body.peer,
+        role: body.role.0,
+    };
 
-    facade.initiate_communication(id, body.into()).await;
+    facade.save(swap).await;
```
D4nte, line 100:
```suggestion
    facade.save(swap).await?;
```

`save` should return a result.
I think that we should proceed if we cannot save the swap in the DB as it means that if cnd is stopped later, we will not be able to resume the swap and hence the user risks to lose his assets.

D4nte: `save` actually return a result so let's use it (I guess clippy would say the same?).

cnd/src/db.rs
thomaseizinger: Good question. I would still separate the information and maybe have an on-disk cache of peer-id <-> addresses somewhere that this would go in!

thomaseizinger: Sounds like a feature to me though so I would not save the information for now to keep the scope small.

cnd/src/swap_protocols.rs
tcharding: Good idea, will do.

cnd/src/http_api/routes/index.rs
tcharding: In the name of learning, is there any reason why you favour the turbo fish syntax over the explicit type syntax @D4nte?

cnd/src/http_api/routes/index.rs
tcharding: Yeah, I back and forth'ed on this point during development.  I totally agree with you but in the end I decided that adding the error return was out of scope for this ticket, since the facade's `save` implementation is empty changing the signature of an empty function seemed like a dirty thing to do.  I elected to leave the correction of the error handling for when the facade method is done (and I assume this will be within the scope of the ticket that does the save implementations in the database (introduced as `unimplemented` in this PR).  Thoughts?

cnd/src/http_api/routes/index.rs
thomaseizinger: I usually prefer turbo-fish because it some robust.

Annotating the final type breaks if you refactor the code by chaining more things onto the expression for example.
Turbo-fish syntax doesn't suffer from that because it is localized to the actual type that needs hints.

tcharding:
@da-kami, this PR proposes a new naming convention built on top of the one you and @thomaseizinger designed with knowledge gained over the last month using it.

cnd/src/swap_protocols/lnbtc.rs
```diff
@@ -0,0 +1,12 @@
+use crate::swap_protocols::{asset, identity};
+
+/// Data required to create a swap that involves a bitcoin on the lightning
+/// network.
+#[derive(Clone, Debug)]
+pub struct CreatedSwap {
+    // FIXME: Should this be `asset::bitcoin`?
```
thomaseizinger, line 7: I think so yes.

cnd/src/swap_protocols.rs
```diff
@@ -1,10 +1,11 @@
 pub mod actions;
 mod facade;
 pub mod halight;
-pub mod han;
 pub mod herc20;
+pub mod hneth;
 pub mod ledger;
 pub mod ledger_states;
+pub mod lnbtc;
```
thomaseizinger, line 8: I like the new naming, good step forward :)

I am sort of missing the `H` in `LNBTC` though. To signal compatibility with other HTLC based swaps, I think this one should say it as-well.
Also, if LN ever gets scriptless script support, then we will need a different protocol but it would still be Bitcoin on Lightning, just a different version of lnd probably.

tcharding: `hlnbtc`?

tcharding:
Heavy rebasing down @thomaseizinger, all suggestions by yourself and @D4nte should be worked in at this stage.  I struggled to find a clean scope limit and in the end settled on changing all identifiers and filenames _without_ touching the REST API (`http_api/routes/index.rs` got a little tickle though).

cnd/src/swap_protocols/lnbtc.rs
tcharding: Went a step further and removed the `asset::Lightning` module all together, I should never have added it - it was an overdone abstraction.

cnd/src/swap_protocols.rs
tcharding: Changed to use `hlnbtc`.  These 'words' are a little hard to parse in code (.e.g., `HnethHlnbtcFoo`), not sure if our eyes will get used to parsing them of if we want to mess with the format a bit e.g., `HnEth` but I'm hesitant to do this because it violates idiomatic rust (how to capitalise acronyms).  Any thoughts?

cnd/src/http_api/routes/index.rs
D4nte: At best, you nnow ask the next developer to go through the same process than we just did to decide whether or not the failure of save should stop the action.
At worst, this will not be caught up by the next dev/reviewers and user will lose money.

> since the facade's save implementation is empty changing the signature of an empty function seemed like a dirty thing to do

I would disagree, on the contrary. Creating a skeleton with functions that have an unimplemented body is the perfect timing to update the interfaces through and through. After all, if you create a function but do not implement its logic, isn't it because you want to do just that: refine and prepare the interfaces so that the follow-up work will be about implementing the logics and not revising the interfaces again?

cnd/src/swap_protocols.rs
D4nte: I'd suggest `haneth` and not `hneth`. The removal of one vowel just makes it worse.

cnd/src/http_api/routes/index.rs
D4nte: It's simply shorter @tcharding. Your code repeat the `CreateSwap` symbol twice, the information for the reader is redundant.

D4nte: Please note if you mark a comment as "resolved" it does not entice the author to reply to the comment :)

cnd/src/db.rs
```diff
@@ -132,6 +133,20 @@ pub enum Error {
     SwapNotFound,
 }
 
+/// Date required to create a swap.
```
D4nte, line 136:
```suggestion
/// Data required to create a swap.
```

cnd/src/http_api/routes.rs
```diff
@@ -136,15 +136,15 @@ fn make_siren_action(swap_id: LocalSwapId, action_name: &str) -> siren::Action {
 pub struct AliceHanEthereumHalightBitcoinState {
     pub alpha_ledger_state:
         LedgerState<asset::Ether, htlc_location::Ethereum, transaction::Ethereum>,
-    pub beta_ledger_state: halight::State,
+    pub beta_ledger_state: hlnbtc::State,
```
D4nte, line 139:
```suggestion
    pub beta_ledger_state: halight_btc::State,
```
I do not find `hltnbtc` readable.

cnd/src/http_api/routes/index.rs
```diff
@@ -182,32 +190,66 @@ impl From<Body<HanEthereumEther, HalightLightningBitcoin>>
...
             lightning_cltv_expiry: body.beta.cltv_expiry.into(),
-            lightning_amount: body.beta.amount.0,
+            bitcoin_amount: body.beta.amount.0,
         }
     }
 }
 
 #[derive(serde::Deserialize, Clone, Debug)]
-pub struct HanEthereumEther {
+struct Hneth {
```
D4nte, line 199:
```suggestion
struct HanEth {
```
This is my proposal re readability.

cnd/src/http_api/routes/index.rs
```diff
@@ -182,32 +190,66 @@ impl From<Body<HanEthereumEther, HalightLightningBitcoin>>
...
+            chain_id: p.chain_id,
+            absolute_expiry: p.absolute_expiry,
+        }
+    }
+}
+
 #[derive(serde::Deserialize, Clone, Debug)]
-pub struct HalightLightningBitcoin {
-    pub amount: Http<asset::Lightning>,
+struct Hlnbtc {
```
D4nte, line 218:
```suggestion
struct HlnBtc {
```

cnd/src/network.rs
```diff
@@ -350,9 +348,9 @@ impl ComitNode {
     pub fn initiate_communication(
         &mut self,
         id: LocalSwapId,
-        swap_params: HanEtherereumHalightBitcoinCreateSwapParams,
+        swap_params: HnethHlnbtcCreateSwapParams,
```
D4nte, line 351:
```suggestion
        swap_params: HanEthHlnBtcCreateSwapParams,
```

D4nte:
```suggestion
        swap_params: HanEthHalightBtcCreateSwapParams,
```

cnd/src/http_api/routes.rs
da-kami: I agree with @D4nte - we chose `halight` because it is nicely readable and spell-able. @thomaseizinger ?

cnd/src/http_api/routes/index.rs
da-kami: Again agree with @D4nte here. 
Removing the ledger is a good idea though, because `HTLC for Asset on Native ledger` can imply the ledger I suppose. (because there is only one Native asset per ledger for that protocol if I am not mistaken.)

cnd/src/http_api/routes/index.rs
da-kami: In accordance with @D4nte's change above I would actually vote for `HalightBtc`.

da-kami:
> > I struggled to find a clean scope limit and in the end settled on changing all identifiers and filenames without touching the REST API (http_api/routes/index.rs got a little tickle though).
> 
> I have some ideas about these names but I am afraid that we never merge this PR if we start / continue to discuss these here.
> 
> Do you think we could merge everything apart from the naming changes and discuss the names in a different place?
> 
> If it is too difficult to tear apart we can also merge as is and continue the discussion somewhere else :)

I actually think it is too early to iterate on the names. I started going over this, but it is a discussion rabbit hole...

cnd/src/http_api/routes/index.rs
tcharding: > At worst, this will not be caught up by the next dev/reviewers and user will lose money.

The facade save method has to delegate to the db save trait which returns an error - I don't see how it would be possible to forget to deal with this error.

cnd/src/swap_protocols.rs
tcharding: > I'd suggest haneth and not hneth. The removal of one vowel just makes it worse.

Not a bad suggestion, this makes `hlnbtc` and `hanbtc` have the same number of characters, which is nicer.  Also makes the possibility of writing identifiers as `HanBtc` more palatable.  Are there any arguments _against_ `haneth`/`hanbtc`?

tcharding:
> > HalightLightningBitcoin -> HLNBTC # Htlc Lightning BTC
> 
> After starting the review, I'd be in favour of removing the `n` of this abbrev.

So, you prefer `hnbtc` for lightning but `hanbtc` for Han BTC?  As commented above, I feel there may be some merit in keeping the number of characters in these terms the same.

> `HalightLightningBitcoin` -> `HalightBitcoin`: There is not Halight protocol without lightning
> `HalightBitcoin` -> `HalightBtc`: This is short enough.
> 
> Edit: I'd also favour capitalising the crypto/ledger symbol: `HanEth` `HanBtc`.

What do you say to these @D4nte ?

- `HanBtc`
- `HanEth`
- `HlnBtc`
- `Herc20`

cnd/src/http_api/routes.rs
tcharding: Fair points, `hlnbtc` is not readable and `halight` is nicer.

cnd/src/http_api/routes/index.rs
tcharding: > because there is only one Native asset per ledger for that protocol if I am not mistaken.

That is my understanding.

tcharding:
Thanks for everyone's input @da-kami, @D4nte, @thomaseizinger.  I'm closing this PR as is, keeping the patches and discussion as a record for when we revisit the current naming scheme.  This PR was a success in my opinion, it was only by working on this that the naming improvement suggestions came out.  To summarise, for the record, consensus in the above thread seems to be

- We do need to modify the current naming scheme, the redundancy in 'protocol/ledger/asset' is to much of a burden in code identifiers.
- Separating the protocol/ledger/assets into there own _single_ construct is probably more useful than explicitly composing out of the three (most combinations of protocol/ledger/asset are not valid so this level of generalisation is not needed).
- The acronyms should be easily readable and easy to write in code, consideration of Rust identifier capitalisation needs to be considered.

Best suggestions currently are:
- 'haneth' (written as `HanEth` when possible)
- 'hanbtc' (written as `HanBtc` when possible)
- 'herc20' (written as `Herc20` when possible)

Lightning (bitcoin) naming has not reached consensus, we currently have: (Please add vote on one of these names if you feel inclined to do so @thomaseizinger, @D4nte, @da-kami, @luckysori, @rishflab, @bonomat)  
- [+0] 'hlnbtc' (`HlnBtc`)
- [+1] 'hlbtc' (`HlBtc`)
- [+0] 'halightbtc' (`HalightBtc`)

thomaseizinger:
> Thanks for everyone's input @da-kami, @D4nte, @thomaseizinger. I'm closing this PR as is, keeping the patches and discussion as a record for when we revisit the current naming scheme. (Code changes excluding re-naming will be extracted into new PRs.) This PR was a success in my opinion, it was only by working on this that the naming improvement suggestions came out. To summarise, for the record, consensus in the above thread seems to be
> 
>     * We do need to modify the current naming scheme, the redundancy in 'protocol/ledger/asset' is to much of a burden in code identifiers.
> 
>     * Separating the protocol/ledger/assets into there own _single_ construct is probably more useful than explicitly composing out of the three (most combinations of protocol/ledger/asset are not valid so this level of generalisation is not needed).
> 
>     * The acronyms should be easily readable and easy to write in code, consideration of Rust identifier capitalisation needs to be considered.
> 
> 
> Best suggestions currently are:
> 
>     * 'haneth' (written as `HanEth` when possible)
> 
>     * 'hanbtc' (written as `HanBtc` when possible)
> 
>     * 'herc20' (written as `Herc20` when possible)
> 
> 
> Lightning (bitcoin) naming has not reached consensus, we currently have: (Please add vote on one of these names if you feel inclined to do so @thomaseizinger, @D4nte, @da-kami, @luckysori, @rishflab, @bonomat)
> 
>     * [+0] 'hlnbtc' (`HlnBtc`)
> 
>     * [+1] 'hlbtc' (`HlBtc`)
> 
>     * [+1] 'halightbtc' (`HalightBtc`)

Thanks for writing this up @tcharding !
Can I ask you to open a follow-up (spike?-)issue where this current state is recorded and the goal is define a better naming scheme?
I think the last time we did this, it worked pretty well and actually didn't take long at all. We just kept pushing it away for way too long :D

I would like to add to / rephrase this part:

> Separating the protocol/ledger/assets into there own _single_ construct is probably more useful than explicitly composing out of the three (most combinations of protocol/ledger/asset are not valid so this level of generalisation is not needed).

Conceptually, we are doing "inheritance" here (`han` as an abstract protocol with two concrete instances). A future naming scheme should avoid that IMO.

> Best suggestions currently are:
> 
>     * 'haneth' (written as `HanEth` when possible)
> 
>     * 'hanbtc' (written as `HanBtc` when possible)
> 
>     * 'herc20' (written as `Herc20` when possible)

I prefer that over the current scheme already although I think we could do _even better_ if the naming didn't even suggest an abstract concept / similarity between `haneth` and `hanbtc`.

tcharding:
Does Liquid have HTLCs?  Have we discussed supporting swaps in/out of Liquid with COMIT?  If so `hlbtc` would likely be needed for Liquid implying `hlnbtc` for Lightning.

bonomat:
> Does Liquid have HTLCs? Have we discussed supporting swaps in/out of Liquid with COMIT? If so `hlbtc` would likely be needed for Liquid implying `hlnbtc` for Lightning.

yes, and they look exactly the same as on Bitcoin as liquid(d)? is a bitcoind fork. 

We haven't _discussed_ it but afaik we can easily support it and probably are going to in the near future.

Issue #2498, opened 2020-04-09
The requested combination of ledgers and assets is not supported.


yosriady:
## Problem

On the rfc003 routes, errors returned for deserialisation issues are unhelpful

--- 

This is an odd one, but **for some inputs** the cnd HTTP API returns an error when you `sendSwap`:
```
HTTP_URL_CND_1 http://127.0.0.1:51378/swaps/rfc003

{"status":400,"title":"Swap not supported.","detail":"The requested combination of ledgers and assets is not supported."}
```

Here is the swap that triggers the error (this is the request payload passed to the `/swaps/rfc003`endpoint):

```
{
	"alpha_ledger": {
		"name": "ethereum",
		"chain_id": 17
	},
	"beta_ledger": {
		"name": "bitcoin",
		"network": "regtest"
	},
	"alpha_asset": {
		"name": "erc20",
		"token_contract": "0x62d69f6867a0a084c6d313943dc22023bc263691",
		"quantity": "27000000000000000000"
	},
	"beta_asset": {
		"name": "bitcoin",
		"quantity": "376433.99999999994"
	},
	"alpha_ledger_refund_identity": "0xf8960e4BB96108aBD63910A3aC7BCdCd3d53696F",
	"alpha_expiry": 1586409200,
	"beta_expiry": 1586405600,
	"peer": {
		"peer_id": "QmcWiij22RbVB1goi1DR6fz61xRRu1FiXC4Ue6dbiK6E6V",
		"address_hint": "/ip4/127.0.0.1/tcp/9939"
	}
}
```

The error message isn't saying anything about what went wrong with the payload.

## Goal

As we are migrated to new routes, we should prioritise those new routes and ensure that the handling of deserialisation is more verbose by returning the actual deser error (what field cannot be deserialized).

**Do not touch the old routes.**

Ensure (or make it) that the errors send back if there is a deserialisation issue are verbose enough to allow the dev to know what field is malformed (.e.g float amount in satoshi or wei field).
Please note more broad error tests are already present `create-*-returns-invalid-body`.

D4nte:
Should this go back to grooming then?

tcharding:
There is only one new endpoint implemented so far.  The error handling for this endpoint _does_ return the serde error.  I believe this issue is resolved already. (initial issue description was for the rfc003 endpoints and during grooming we decided to only do the new endpoints, we missed the fact that these were correctly handled already.)

[...]

PR #2514, opened 2020-04-23
Move get_swaps() into rfc003 module

tcharding:
Done while investigating (and closing) #2498.

The get swaps endpoint is tied to `rfc003` by the fact that it uses the function `build_rfc003_siren_entity()`.  This function uses the `with_swap_types` macro, this is never going to work with the new split protocols. `get_swaps()` and its handler should be over in the `rfc003` module along with the other `rfc003` endpoint handlers to make this explicit.


cnd/src/http_api/problem.rs
```diff
@@ -32,7 +32,7 @@ pub struct UnexpectedQueryParameters {
     pub parameters: &'static [&'static str],
 }
 
-// tracing trippers clippy warning, issue reported: https://github.com/tokio-rs/tracing/issues/553
+// tracing triggers clippy warning, issue reported: https://github.com/tokio-rs/tracing/issues/553
```
D4nte, line 35: 🤣

PR #2523, opened 2020-04-25
Remove `Option` for `Hash` in Ethereum `Block` struct #2414

### LINKED ISSUE #2414: Remove `Option` for `Hash` in Ethereum `Block` struct
## Problem

The Ethereum `Block` currently has an `Option<Hash>` field, this should just be `Hash` - the `Web3Connector` should deal with this. 

https://github.com/comit-network/comit-rs/blob/b93cd01c57620579ab9703555c98622df83d3b96/cnd/src/ethereum.rs#L181-L183

Most of those `Option`-wrappers were already removed, let's also remove that one. 

## Goal

```
pub struct Block {
    /// Hash of the block
    pub hash: Hash,
...
}
```

dhaileytaha:
Removed the `Option` as recommended :)

Resolves #2414.


thomaseizinger:
I've the result of why the CI is not running and raised a PR here: https://github.com/comit-network/comit-rs/pull/2524

You will need to wait until that one is merged and rebase against latest `dev` before the CI is being triggered.

We'll ping you here once you can go ahead and rebase :)

dhaileytaha:
@thomaseizinger 
Thank you!
Sure, will wait for your notification.

dhaileytaha:
> Hi @dhaileytaha,
> 
> thanks for the contribution!
> For some reason the CI doesn't seem to kick in, which is strange 🤔
> 
> But regardless of that, I am surprised that just removing the Option actually builds because that would mean we are never using this field! 😂
> 
> If that is the case, do you mind changing your patch to just remove the field? :)
> 
> Also, if you could squash all your commits into a single one and remove the merge commits, that would be great!

Its great to find myself coming up with strange and surprising things 😅

thomaseizinger:
Hi @dhaileytaha,

https://github.com/comit-network/comit-rs/pull/2524 is now merged. Please rebase your PR against **origin/dev** (we are using a Git-Flow based branching model) and squash everything into a single commit.

CI should be triggered once you force-push your branch.

PR #2521, opened 2020-04-24
Implement GET /swaps/:id (lightning)

tcharding:
Currently we do not return any data when a user hits the newly created `/swaps/:id` endpoints.

The schema was designed in https://github.com/comit-network/comit-js-sdk/pull/185

Return the correct data as defined in the js-sdk for the ethereum-ether/lightning endpoint.

Please note: Does not include tests as described in the issue.

Resolves: #2387

PR #2524, opened 2020-04-26
Run CI on pull-requests on selected push actions

thomaseizinger:
Currently, we are only running the CI on branch pushes which prevents external contributors from triggering the CI for a PR they raised.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,12 @@ name: CI
 
 on:
   push:
-    branches-ignore:
-      - 'staging.tmp'
-      - 'trying.tmp'
+    branches:
+      - 'staging'
+      - 'trying'
+      - 'master'
+      - 'dev'
+  pull_request:
 
 jobs:
   static_analysis:


.github/workflows/ci.yml
```diff
@@ -2,9 +2,12 @@ name: CI
 
 on:
   push:
-    branches-ignore:
-      - 'staging.tmp'
-      - 'trying.tmp'
+    branches:
+      - 'staging'
+      - 'trying'
```
D4nte, line 7: These both are for bors, right?

thomaseizinger: Yes!

bonomat:
@thomaseizinger 

I'm afraid that this did not turn out as planned as the PR below is now getting built multiple times in parallel: 

https://github.com/comit-network/comit-rs/pull/2521

image

image

thomaseizinger:
@bonomat hmm, that is not what I intended, do we need to exclude other branches somehow maybe?

D4nte:
> @bonomat hmm, that is not what I intended, do we need to exclude other branches somehow maybe?

The PR that was cited was not yet rebased on dev so it is possible that the "push" events in the PR were coming for the PR's branch own GitHub Action [config](https://github.com/comit-network/comit-rs/blob/2387-get-swaps/.github/workflows/ci.yml) and the "pull_request" event was coming from the `dev` branch github action config.

Let's observe PRs that have Thomas' commit in their tree.

PR #2539, opened 2020-04-28

dhaileytaha:
PR against origin/dev


thomaseizinger:
Your build is failing because you need to adapt all the existing call sites that at the moment dealt with the `Option` wrapper. In most places, that should simplify things because you can just use the existing field without calling `unwrap` or `expect` on it.

cnd/src/btsieve/ethereum.rs
```diff
@@ -202,8 +201,7 @@ where
         match block_generator.async_resume().await {
             GeneratorState::Yielded(block) => {
                 let block_hash = block
-                    .hash
-                    .ok_or_else(|| anyhow::anyhow!("block without hash"))?;
+                    .hash;   // .ok_or_else(|| anyhow::anyhow!("block without hash"))?;
```
thomaseizinger, line 204: Please remove the comment and inline the variable.

cnd/src/btsieve/ethereum.rs
```diff
@@ -146,8 +145,8 @@ where
         match block_generator.async_resume().await {
             GeneratorState::Yielded(block) => {
                 let block_hash = block
```
thomaseizinger, line 147: This variable is no longer needed now I guess, you can just inline that :)

cnd/src/btsieve/ethereum/cache.rs
```diff
@@ -51,7 +51,7 @@ where
     async fn latest_block(&self) -> anyhow::Result<Self::Block> {
         let block = self.connector.latest_block().await?;
 
-        let block_hash = block.hash.expect("no blocks without hash");
+        let block_hash = block.hash;
```
thomaseizinger, line 54: Same here, no need for this variable, it can be inlined.

cnd/src/btsieve/ethereum.rs
dhaileytaha: done

dhaileytaha:
closed this one and routed to https://github.com/comit-network/comit-rs/pull/2580

PR #2541, opened 2020-04-28
Remove dead code

thomaseizinger:


D4nte:
How come clippy does not catch this kind of thing?

Issue #2490, opened 2020-04-20
Handle multiple, concurrent swaps with the same peer properly


thomaseizinger:
## Problem

Currently, the announce protocol emits the following event once a swap was confirmed:

https://github.com/comit-network/comit-rs/blob/5a7b9af1e304dab3b0c54b6404a769237fed8dd6/cnd/src/network/protocols/announce/behaviour.rs#L285-L292

Now imagine a party announcing the same swap again before the first one is finalized. We cannot distinguish which of those swaps get confirmed and which ones get declined because the event only returns peer-id, swap-id and the digest. Announcing the same swap to the same peer means peer and digest are equal and hence you cannot distinguish.

## Goal

Handle this situation better so that this ambiguity doesn't happen.

## Recommendation

Remove the concurrency aspect of the current behaviour.

D4nte:
Please mark as `added-mid-sprint` and add the milestone.

tcharding:
> Please mark as `added-mid-sprint` and add the milestone.

Done, thanks for the reminder.

thomaseizinger:
We seem to have lost this functionality as discovered in #2718.

PR #2543, opened 2020-04-28
 Do not initiate a swap for identical parameters

tcharding:
If a swap has already been initiated we do not want to initiate a second time otherwise there will be ambiguity in the system later during the communication protocols since the swap digest is based off of the created swap parameters.

There is more than one solution to this problem but for now just iterate over the current swaps and if one exists with these parameters already abort initiation.

Now with testing, patch 2 adds a failing e2e test, based on @thomaseizinger observation that patch 1 returns a 500 error.  Patch 3 fixes the tests.  PR is 'untested' because I cannot run e2e tests on my machine currently ...

Fixes: #2490

### DIFF
--- a/api_tests/tests/ether_halight.ts
+++ b/api_tests/tests/ether_halight.ts
@@ -74,3 +74,22 @@ it(
         return expect(bobResponsePromise).rejects.toThrowError();
     })
 );
+
+// This test could be anywhere, it is just here because this endpoint was the first of the spilt protocols implemented.
+it(
+    "alice-cannot-create-two-identical-swaps",
+    twoActorTest(async ({ alice, bob }) => {
+        // arrange
+        const bodies = (await SwapFactory.newSwap(alice, bob))
+            .hanEthereumEtherHalightLightningBitcoin;
+
+        await alice.createSwap(bodies.alice);
+        await sleep(200);
+
+        // act
+        const response = alice.createSwap(bodies.alice);
+
+        // assert
+        await expect(response).rejects.toThrow("Swap already exists.");
+    })
+);
--- a/cnd/src/http_api/problem.rs
+++ b/cnd/src/http_api/problem.rs
@@ -4,6 +4,7 @@ use crate::{
         rfc003::handlers::{post_swap::UnsupportedSwap, InvalidAction, InvalidActionInvocation},
         LndActionError,
     },
+    network::comit_ln::SwapExists,
 };
 use http_api_problem::HttpApiProblem;
 use warp::{
@@ -47,7 +48,13 @@ pub fn from_anyhow(e: anyhow::Error) -> HttpApiProblem {
     };
 
     if let Some(db::Error::SwapNotFound) = e.downcast_ref::<db::Error>() {
+        tracing::error!("swap was not found");
         return HttpApiProblem::new("Swap not found.").set_status(StatusCode::NOT_FOUND);
+    }
+
+    if e.downcast_ref::<SwapExists>().is_some() {
+        tracing::error!("swap already exists, returning 400");
+        return HttpApiProblem::new("Swap already exists.").set_status(StatusCode::BAD_REQUEST);
     }
 
     if let Some(e) = e.downcast_ref::<UnexpectedQueryParameters>() {
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -335,9 +335,7 @@ impl ComitNode {
         swap_params: HanEtherereumHalightBitcoinCreateSwapParams,
     ) -> anyhow::Result<()> {
         self.supports_halight()?;
-
-        self.comit_ln.initiate_communication(id, swap_params);
-        Ok(())
+        self.comit_ln.initiate_communication(id, swap_params)
     }
 
     pub fn get_finalized_swap(&mut self, id: LocalSwapId) -> Option<comit_ln::FinalizedSwap> {
--- a/cnd/src/network/comit_ln.rs
+++ b/cnd/src/network/comit_ln.rs
@@ -27,6 +27,7 @@ use libp2p::{
 };
 use std::{
     collections::{HashMap, VecDeque},
+    fmt,
     task::{Context, Poll},
 };
 
@@ -105,30 +106,27 @@ impl ComitLN {
         &mut self,
         id: LocalSwapId,
         create_swap_params: HanEtherereumHalightBitcoinCreateSwapParams,
-    ) {
+    ) -> anyhow::Result<()> {
         let digest = create_swap_params.clone().digest();
 
+        if self.swaps_waiting_for_announcement.contains_key(&digest) {
+            anyhow::bail!(SwapExists)
+        }
         self.swaps.insert(id, create_swap_params.clone());
+        self.swaps_waiting_for_announcement
+            .insert(digest.clone(), id);
 
         match create_swap_params.role {
             Role::Alice => {
-                if self.swaps_waiting_for_announcement.contains_key(&digest) {
-                    // To fix this panic, we should either pass the local swap id to the
-                    // announce behaviour or get a unique token from the behaviour that
-                    // we can use to track the progress of the announcement
-                    panic!("cannot send two swaps with the same digest at the same time!")
-                }
-
                 self.announce
-                    .start_announce_protocol(digest.clone(), create_swap_params.peer);
-
-                self.swaps_waiting_for_announcement.insert(digest, id);
+                    .start_announce_protocol(digest, create_swap_params.peer);
             }
             Role::Bob => {
                 tracing::info!("Swap waiting for announcement: {}", digest);
-                self.swaps_waiting_for_announcement.insert(digest, id);
             }
         }
+
+        Ok(())
     }
 
     pub fn get_finalized_swap(&self, swap_id: LocalSwapId) -> Option<FinalizedSwap> {
@@ -208,6 +206,17 @@ impl ComitLN {
 
         // We trust in libp2p to poll us.
         Poll::Pending
+    }
+}
+
+#[derive(thiserror::Error, Clone, Copy, Debug)]
+pub struct SwapExists;
+
+impl fmt::Display for SwapExists {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // This impl is required to build but want to use a static string for
+        // this when returning it via the REST API.
+        write!(f, "")
     }
 }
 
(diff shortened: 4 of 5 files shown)


cnd/src/network.rs
```diff
@@ -133,26 +133,6 @@ impl Swarm {
         })
     }
 
-    /// This is the API for Alice to call in order to execute the appropriate
```
thomaseizinger, line 136: Lol, I literally just open a PR to remove the same thing XD

cnd/src/network/comit_ln.rs
```diff
@@ -106,10 +106,18 @@ impl ComitLN {
         id: LocalSwapId,
         create_swap_params: HanEtherereumHalightBitcoinCreateSwapParams,
     ) {
-        let digest = create_swap_params.clone().digest();
-
+        for (_, params) in self.swaps.iter() {
+            if *params == create_swap_params {
```
thomaseizinger, line 110: Can we compare based on `digest` please?

cnd/src/network/comit_ln.rs
```diff
@@ -106,10 +106,18 @@ impl ComitLN {
...
+                tracing::warn!(
+                    "swap with same parameters already exists, aborting swap initiation"
+                );
+                return;
```
thomaseizinger, line 114: I would like to have error handling for this. Up till here, we are still in the context of the HTTP request, so we can actually return an error to the user right away.

tcharding: Cool, will do.

cnd/src/network/comit_ln.rs
tcharding: Sure thing!

cnd/src/network.rs
tcharding: I'll drop this from this one then.

cnd/src/network/comit_ln.rs
```diff
@@ -105,30 +105,31 @@ impl ComitLN {
...
         let digest = create_swap_params.clone().digest();
 
+        if self.swaps_waiting_for_announcement.contains_key(&digest) {
+            return Err(anyhow::anyhow!(
+                "Swap with same digest already exists: {}",
+                digest
+            ));
+        }
         self.swaps.insert(id, create_swap_params.clone());
+        // No race condition because we are called within a mutex guard.
```
thomaseizinger, line 118: This is guaranteed by the mutable reference at compile-time, not the Mutex guard!

Long live Rust!

cnd/src/network/comit_ln.rs
```diff
@@ -197,6 +198,26 @@ impl ComitLN {
         })
     }
 
+    // Once a swap has been finalized this removes it from the ComitLN cache of
+    // active swaps i.e., all the hash maps that we use to track state during
+    // the communication protocols.
+    fn remove_finalized_swap(&mut self, shared_id: SharedSwapId) {
```
thomaseizinger, line 204: Wouldn't it be enough to remove the entry in `swaps_waiting_for_announcement`?

cnd/src/network/comit_ln.rs
```diff
@@ -105,30 +105,31 @@ impl ComitLN {
         &mut self,
         id: LocalSwapId,
         create_swap_params: HanEtherereumHalightBitcoinCreateSwapParams,
-    ) {
+    ) -> anyhow::Result<()> {
         let digest = create_swap_params.clone().digest();
 
+        if self.swaps_waiting_for_announcement.contains_key(&digest) {
```
thomaseizinger, line 111: This is going to result in a 500 for the user because we cannot identify this particular case.

Please create a proper error and send a meaningful error (400) back to the user. An e2e test would also be nice for this case.

I would suggest a single struct (not enum) that derives `thiserror::Error`. You can then do:

```suggestion
        if self.swaps_waiting_for_announcement.contains_key(&digest) {
            anyhow::bail!(TheErrorYouCreated)
        }
```

(I don't think we need to share the digest with the user as it is an internal thing, so I omitted it from the struct ctor)

cnd/src/network/comit_ln.rs
tcharding: Long live Rust

cnd/src/network/comit_ln.rs
tcharding: Yes I think it most likely would, but for future proof-ness at the cost of some run time efficiency I figure why not clean the whole thing up.

cnd/src/network/comit_ln.rs
tcharding: Oh man, that macro is just what I have been looking for.  I have typed `return Err(anyhow!(Error::Foo))` so many times this last week.

thomaseizinger: RTFM man :)

tcharding: > This is going to result in a 500 for the user because we cannot identify this particular case.

Thanks, I did not even begin to think all the way up to the user, good reviewing.

D4nte:
> Patch one and two are clean up of stale items that were added during LN mob but not used, part of this PR because they both relate to the communication initiation.

I only see one commit in this PR.

tcharding:
> > Patch one and two are clean up of stale items that were added during LN mob but not used, part of this PR because they both relate to the communication initiation.
> 
> I only see one commit in this PR.

My bad, they were dropped - I have updated the comment.  Thanks.

api_tests/tests/ether_halight.ts
```diff
@@ -73,3 +73,20 @@ it(
         return expect(bobResponsePromise).rejects.toThrowError();
     })
 );
+
+// This test could be anywhere, it is here because this is the first test file that uses the new comunication protocols.
```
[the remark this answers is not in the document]
tcharding: Sure thing.

cnd/src/network/comit_ln.rs
```diff
@@ -211,6 +228,10 @@ impl ComitLN {
     }
 }
 
+#[derive(thiserror::Error, Clone, Copy, Debug)]
+#[error("swap already exists")]
+pub struct SwapExistsError {}
```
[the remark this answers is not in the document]
tcharding: Done, thanks.  You mean nitpick :)  It was a neat pick however, naming is hard.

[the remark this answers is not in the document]
D4nte: haha lol yes indeed. Thanks.

cnd/src/http_api/problem.rs
```diff
@@ -50,6 +51,10 @@ pub fn from_anyhow(e: anyhow::Error) -> HttpApiProblem {
         return HttpApiProblem::new("Swap not found.").set_status(StatusCode::NOT_FOUND);
     }
 
+    if let Some(e) = e.downcast_ref::<SwapExists>() {
+        return HttpApiProblem::new(format!("{}", e)).set_status(StatusCode::BAD_REQUEST);
```
thomaseizinger, line 55: I would prefer user-facing errors to be static strings. It is not a breaking change per-se but to me it would be very surprising that changing the error message of an error defined 3 abstraction levels deep has a user-facing impact.

Also, the error message here is not consistent with the others as it is:

- not capitalized
- doesn't end with a dot

thomaseizinger: In the other handlers, we log the error :)

cnd/src/network/comit_ln.rs
```diff
@@ -211,6 +228,10 @@ impl ComitLN {
     }
 }
 
+#[derive(thiserror::Error, Clone, Copy, Debug)]
+#[error("swap already exists")]
+pub struct SwapExists {}
```
thomaseizinger, line 233: Nit: If you decide to not include an data, you can do this (saves you 3 characters when instantiating it because you can just say `anyhow::bail!(SwapExists)`:
```suggestion
pub struct SwapExists;
```

If we decide to log this error instead of sending it to the user, it might be helpful to include the digest in the log message so it shows up in the log file.

cnd/src/http_api/problem.rs
thomaseizinger: It might also be worth to tell the user a little more of what is going here and they just need to wait until the first swap finalizes before they can make a new one. Esp. the temporary nature of this error is something that should be pointed out IMO.

api_tests/tests/lightning_routes.ts
```diff
@@ -131,4 +132,21 @@ describe("Lightning routes tests", () => {
...
+            const bodies = (await SwapFactory.newSwap(alice, bob))
+                .hanEthereumEtherHalightLightningBitcoin;
+
+            await alice.createSwap(bodies.alice);
+            await sleep(200);
+            try {
+                await alice.createSwap(bodies.bob);
+            } catch (err) {
+                expect(err.status).toBe(400);
+            }
```
thomaseizinger, line 149: I personally find `try-catch` to be quite ugly so I usually do this:

```suggestion
			await expect(alice.createSwap(bodies.bob)).toThrow("Swap already exists.");
```

How does this work?

`createSwap` returns a Promise, which is equivalent to a Future from Rust.
`await` does the same thing except that it throws an exception if the Promise _rejects_. You can think of Promises as Futures that always return a Result.

So instead of `await`ing the Promise right away and catching the exception, we pass the returned Promise directly to our test framework and say:

Please assert that this Promise will eventually reject with the following error.
We still need to `await` the whole thing because it is an async operation. (Not lazy, like in Rust but we want the test to wait until it is done).

Note that I also changed your assertion to actually check if we return the correct error by asserting the error message.

thomaseizinger: A test purist would probably argue that you should even separate the assertion from the acting:

```ts
	// arrange
	const bodies = (await SwapFactory.newSwap(alice, bob))

	await alice.createSwap(bodies.alice);
	await sleep(200);

	// act
	let response = alice.createSwap(bodies.bob);

	// assert
	await expect(response).toThrow("Swap already exists.");
```

tcharding: Nice knowledge sharing, thanks for making the effort.  Will process and update the PR where it seems fit.

cnd/src/http_api/problem.rs
tcharding: Cool, I didn't understand this on first reading but after looking at the code again I've got you.  Cheers man, will fix up this PR as suggested.

cnd/src/network/comit_ln.rs
tcharding: Cool thanks.

api_tests/tests/ether_halight.ts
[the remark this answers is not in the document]
tcharding: @D4nte I moved this back into `ether_halight.ts`.  I cannot get the test to run when its in `lightning_routes.ts` and it runs just fine in `ether_haliht.ts`. To be blunt I do not care to learn why (TS, Jest) right now, this PR has been open for forever - I'm bored with it already.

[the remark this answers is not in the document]
D4nte: I have approved the PR already.

cnd/src/network/comit_ln.rs
tcharding: ok, is in the end I did as you suggest.  Because removing the swap from all the hash maps makes other code break. I could not get to the bottom of why, this makes me think that the code is currently brittle.  We have to re-think the whole hash map thing any way so I don't think its a big issue.

PR #2553, opened 2020-04-29
Upgrade comit-sdk to 0.16.0

D4nte:
Uses the change done in https://github.com/comit-network/comit-js-sdk/pull/229

The `sendPayment` API now resolves when the payment
is `in flight`, allowing us to await it.

We can then later await the returned function to wait
for the payment to be settled.

Related to comit-network/comit-sdk#180


D4nte:
bors retry

PR #2554, opened 2020-04-29
Actually assert the full error response

thomaseizinger:
I realized while taking a look at our current e2e tests that we are not always asserting error responses in the same way and also not all aspects of it.

I've played around with Jest a bit and I believe I've found a concise way of covering the whole error response.

Putting this PR up for people to take inspiration from.
Please use this format from now on when you assert error responses from cnd and migrate other tests to test format if you come across some as part of other work.

Thanks :)

Asking everyone for review so everyone knows!

### DIFF
--- a/api_tests/tests/lightning_routes.ts
+++ b/api_tests/tests/lightning_routes.ts
@@ -18,18 +18,22 @@ describe("Lightning routes tests", () => {
             const bodies = (await SwapFactory.newSwap(alice, bob, true))
                 .hanEthereumEtherHalightLightningBitcoin;
 
-            await expect(
-                alice.cnd.createHanEthereumEtherHalightLightningBitcoin(
-                    bodies.alice
-                )
-            ).rejects.toThrow("lightning is not configured.");
-            try {
-                await bob.cnd.createHanEthereumEtherHalightLightningBitcoin(
-                    bodies.bob
-                );
-            } catch (err) {
-                expect(err.status).toBe(400);
-            }
+            const aliceResponse = alice.cnd.createHanEthereumEtherHalightLightningBitcoin(
+                bodies.alice
+            );
+            const bobResponse = bob.cnd.createHanEthereumEtherHalightLightningBitcoin(
+                bodies.bob
+            );
+
+            const expectedProblem = {
+                status: 400,
+                title: "lightning is not configured.",
+                detail:
+                    "lightning ledger is not properly configured, swap involving this ledger are not available.",
+            };
+
+            await expect(aliceResponse).rejects.toMatchObject(expectedProblem);
+            await expect(bobResponse).rejects.toMatchObject(expectedProblem);
         })
     );
 


api_tests/tests/lightning_routes.ts
```diff
@@ -18,18 +19,29 @@ describe("Lightning routes tests", () => {
...
+            const bobResponse = bob.cnd.createHanEthereumEtherHalightLightningBitcoin(
+                bodies.bob
+            );
+
+            await expect(aliceResponse).rejects.toMatchObject(
+                new Problem({
+                    status: 400,
+                    title: "lightning is not configured.",
+                    detail:
+                        "lightning ledger is not properly configured, swap involving this ledger are not available.",
```
tcharding, line 34: Does this mean this test now breaks if someone edits this string over in `cnd/src/http_api/problem.rs`?  If so, whats the failure going to look like?  Is it going to be obvious that its just caused by the string changing so we don't go chasing shadows?

tcharding: After reading your comments https://github.com/comit-network/comit-rs/pull/2543#discussion_r417147453 I can answer this myself I believe.  If we use static strings for the error and have them at the 'top level' i.e., not returned from deeply nested errors, then this works nicely.

thomaseizinger: The error could be nicer but I think it is acceptable:

image

thomaseizinger: You have to do a manual comparison unfortunately.

thomaseizinger: Actually, now you got me to writing a custom matcher :joy:

thomaseizinger: I tried for a bit but couldn't get anything useful out of it. Will leave as is!

api_tests/tests/lightning_routes.ts
```diff
@@ -18,18 +19,29 @@ describe("Lightning routes tests", () => {
...
-                expect(err.status).toBe(400);
-            }
+            const aliceResponse = alice.cnd.createHanEthereumEtherHalightLightningBitcoin(
+                bodies.alice
+            );
+            const bobResponse = bob.cnd.createHanEthereumEtherHalightLightningBitcoin(
+                bodies.bob
+            );
+
+            await expect(aliceResponse).rejects.toMatchObject(
```
D4nte, line 29: 📣 This only looks at the properties of the object. You need to use `toStrictEqual` instead.

thomaseizinger: Good call, thank you :)

thomaseizinger: That doesn't seem to work though?

image

thomaseizinger: It should fail because of wrong status AND wrong detail. It seems to compare just the message of the error ...

thomaseizinger: Not instantiating the problem seems to do the trick combined with `toMatchObject`:

image

D4nte: Ok, weird, the jest doc is basically misleading.

D4nte: Ok I read the doc again and now I understand, all good :)

PR #2557, opened 2020-04-30
Wait for a transaction being included before releasing the lock

tcharding:
Our Geth test node mines a block every second.  When sending a
transaction we set the nonce to be the value that we first read from the
blockchain, if we try to do this twice within a block it fails (Geth
thinks we are trying to do a transaction replacement).

Solution; after sending a transaction we wait until it has 1 confirmation. Only then we continue and release the lock.

Fixes: Your pain.

### DIFF
--- a/api_tests/src/wallets/ethereum.ts
+++ b/api_tests/src/wallets/ethereum.ts
@@ -76,18 +76,8 @@ export class EthereumWallet implements Wallet {
             value: "0x0",
             data,
         };
-        const transactionResponse = await this.sendTransaction(tx);
-        const transactionReceipt = await this.jsonRpcProvider.waitForTransaction(
-            transactionResponse.hash,
-            1
-        );
-        // const transactionReceipt = await transactionResponse.wait(1);
 
-        if (!transactionReceipt.status) {
-            throw new Error(
-                `Minting ${asset.quantity} tokens to address ${toAddress} failed`
-            );
-        }
+        await this.sendTransaction(tx);
 
         this.logger.info(
             "Minted",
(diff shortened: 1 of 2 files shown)


api_tests/src/wallets/ethereum.ts
```diff
@@ -119,7 +109,32 @@ export class EthereumWallet implements Wallet {
...
+                " sent, waiting to be confirmed."
+            );
+
+            const transactionReceipt = await this.jsonRpcProvider.waitForTransaction(
+                transactionResponse.hash,
+                1
+            );
+            if (!transactionReceipt.status) {
+                throw new Error(
+                    `Minting ${tx.value} tokens to address ${tx.to} failed`
```
thomaseizinger, line 128: This error message is wrong now because we are not just calling this for minting right?

bonomat: That is correct. I rephrased and rebased :)

api_tests/src/wallets/ethereum.ts
```diff
@@ -119,42 +109,51 @@ export class EthereumWallet implements Wallet {
...
+                " sent, waiting to be confirmed."
+            );
+
+            const transactionReceipt = await this.jsonRpcProvider.waitForTransaction(
+                transactionResponse.hash,
+                1
+            );
+            if (!transactionReceipt.status) {
+                throw new Error(
+                    `Sending amount ${tx.value} to address ${tx.to} failed.`
```
thomaseizinger, line 128: Also not entirely true to be honest :see_no_evil: We are using this function to send any kind of TX, also contract deployments which don't send any money. Here is what I would say:

```suggestion
                    `Transaction ${tx.transactionHash} failed with status 1`
```

tcharding:
Nice, thanks @bonomat.  I don't see the error @thomaseizinger is referring to. 
 Can I leave this to you to squash and merge?

PR #2565, opened 2020-05-01
Remove chai schema

### LINKED ISSUE #2139: Remove chai from end2end tests

D4nte:
Resolves #2139


thomaseizinger:
NACK https://github.com/comit-network/comit-rs/pull/2565/commits/915d2c8d969d94269ec27f564860e31c503a2d35.

I don't think we should change to absolute URIs, it doesn't make sense as we will unlikely return an action that points to something outside of cnd. The additional configuration options is just an unnecessary burden.
I would suggest to just patch the JSON schema and maybe raise a PR upstream, even though given the number of unmerged PRs, that is unlikely to be merged.

api_tests/tests/rfc003_schema.ts
```diff
@@ -1,80 +1,24 @@
...
-import {
-    Action,
-    EmbeddedRepresentationSubEntity,
-    Entity,
-    Link,
-} from "comit-sdk";
+import { Action, EmbeddedRepresentationSubEntity, Link } from "comit-sdk";
+import { extendSchemaMatcher } from "../src/schema_matcher";
+
+extendSchemaMatcher();
```
thomaseizinger, line 9: Is there a more global place where we could do this?

api_tests/tests/rfc003_schema.ts
```diff
@@ -179,59 +118,43 @@ describe("Rfc003 schema swap reject tests", () => {
...
-                {
-                    name: "decline",
-                },
-            ]);
+            expect(bobSwapDetails.properties).toMatchSchema(
+                swapPropertiesJsonSchema
+            );
+            expect(bobSwapDetails.actions.map((action) => action.name)).toEqual(
+                expect.arrayContaining(["accept", "decline"])
+            );
```
thomaseizinger, line 126: :+1:

api_tests/tests/rfc003_schema.ts
D4nte: Let's try :)

Issue #2430, opened 2020-04-09
Only deserialize fields from `Lnd`'s response that we actually use


thomaseizinger:
## Problem

Currently, we deserialize an arbitrary set of fields for invoices and payments in the responses from lnd:

Invoice: https://github.com/comit-network/comit-rs/blob/8b960306b3bfa06e12fbc7478b0e08bb8c4ef0be/cnd/src/lnd.rs#L40-L50

Payment: https://github.com/comit-network/comit-rs/blob/8b960306b3bfa06e12fbc7478b0e08bb8c4ef0be/cnd/src/lnd.rs#L57-L63

## Goal

Only deserialize what we need, too make it less likely that we suffer from backwards-incompatible changes in lnd's API.

thomaseizinger:
This is blocked by https://github.com/comit-network/comit-rs/issues/2395 because we are going to need some of the fields for the validation.

thomaseizinger:
I think we did this in #2573.

PR #2573, opened 2020-05-04
Validate lnd invoice and payment against swap params

### LINKED ISSUE #2395: Validate lnd's Invoices/Payments against swap params
## Problem

Currently, we don't validate whether the invoice added or the payment made by the user matches what we told them to do.

## Goal

Fail early if the user doesn't do what we told them to in the action.
- Fail if the invoice (with the given hash) added has incorrect details
- Fail if the payment has incorrect details

## Ideas

Extend `halight::Params` with whatever you need to do appropriate validation inside the traits implemented on the lnd connectors. For example for an accepted payment (sender side): https://github.com/comit-network/comit-rs/blob/4fcb8b36b327ae8cc7225b80fa55a64fe6709970/cnd/src/lnd.rs#L221

rishflab:
Check whether the parameters specified in the invoice and payment
returned by lnd match the parameters specified in the swap
parameters. Remove fields from the invoice that are no being used.

Resolves #2395, #2430

### DIFF
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -1187,6 +1187,12 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
                 ethereum_identity,
             } => {
                 let role = create_swap_params.role;
+                let params = halight::Params {
+                    identity: create_swap_params.lightning_identity,
+                    cltv_expiry: create_swap_params.lightning_cltv_expiry,
+                    amount: create_swap_params.lightning_amount,
+                    secret_hash,
+                };
 
                 match self.lnd_connector_params {
                     None => { tracing::error!("Internal Failure: lnd connectors are not initialised, no action has been taken. This should be unreachable.") }
@@ -1195,7 +1201,7 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
                             Role::Alice => {
                                 tokio::task::spawn({
                                     let lnd_connector: LndConnectorAsReceiver = (**lnd_connector_params).clone().into();
-                                    halight::new_halight_swap(local_swap_id, secret_hash, create_swap_params.clone().into(), self.halight_states.clone(), lnd_connector)
+                                    halight::new_halight_swap(local_swap_id, params, self.halight_states.clone(), lnd_connector)
                                         .instrument(
                                             tracing::error_span!("beta_ledger", swap_id = %local_swap_id, role = %role),
                                         )
@@ -1232,7 +1238,7 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
                             Role::Bob => {
                                 tokio::task::spawn({
                                     let lnd_connector: LndConnectorAsSender = (**lnd_connector_params).clone().into();
-                                    self::halight::new_halight_swap(local_swap_id, secret_hash,create_swap_params.clone().into(), self.halight_states.clone(), lnd_connector)
+                                    self::halight::new_halight_swap(local_swap_id, params, self.halight_states.clone(), lnd_connector)
                                         .instrument(
                                             tracing::error_span!("beta_ledger", swap_id = %local_swap_id, role = %role),
                                         )
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -1195,7 +1195,7 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
                             Role::Alice => {
                                 tokio::task::spawn({
                                     let lnd_connector: LndConnectorAsReceiver = (**lnd_connector_params).clone().into();
-                                    halight::new_halight_swap(local_swap_id, secret_hash, self.halight_states.clone(), lnd_connector)
+                                    halight::new_halight_swap(local_swap_id, secret_hash, create_swap_params.clone().into(), self.halight_states.clone(), lnd_connector)
                                         .instrument(
                                             tracing::error_span!("beta_ledger", swap_id = %local_swap_id, role = %role),
                                         )
@@ -1206,7 +1206,7 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
                                     let alice_ethereum_identity = create_swap_params.ethereum_identity;
                                     let bob_ethereum_identity = ethereum_identity;
 
-                                    let asset = create_swap_params.ethereum_amount.clone();
+                                    let asset = create_swap_params.ethereum_amount;
                                     let ledger = ledger::Ethereum::default();
                                     let expiry = create_swap_params.ethereum_absolute_expiry;
                                     let token_contract = create_swap_params.token_contract.into();
@@ -1232,7 +1232,7 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
                             Role::Bob => {
                                 tokio::task::spawn({
                                     let lnd_connector: LndConnectorAsSender = (**lnd_connector_params).clone().into();
-                                    self::halight::new_halight_swap(local_swap_id, secret_hash, self.halight_states.clone(), lnd_connector)
+                                    self::halight::new_halight_swap(local_swap_id, secret_hash,create_swap_params.clone().into(), self.halight_states.clone(), lnd_connector)
                                         .instrument(
                                             tracing::error_span!("beta_ledger", swap_id = %local_swap_id, role = %role),
                                         )
--- a/cnd/src/swap_protocols/halight.rs
+++ b/cnd/src/swap_protocols/halight.rs
@@ -5,11 +5,10 @@ use crate::{
         rfc003::{Secret, SecretHash},
         state,
         state::Update,
-        HanEtherereumHalightBitcoinCreateSwapParams, Ledger, LocalSwapId,
+        Ledger, LocalSwapId,
     },
     timestamp::Timestamp,
 };
-
 use futures::{
     future::{self, Either},
     Stream, TryFutureExt, TryStreamExt,
@@ -23,6 +22,7 @@ use tokio::sync::Mutex;
 
 mod connector;
 
+use crate::timestamp::RelativeTime;
 pub use connector::*;
 
 /// Htlc Lightning Bitcoin atomic swap protocol.
@@ -54,14 +54,13 @@ pub struct InProgressSwap {
 /// logging information and store the events yielded by the protocol.
 pub async fn new_halight_swap<C>(
     id: LocalSwapId,
-    secret_hash: SecretHash,
-    swap_params: Params,
+    params: Params,
     state_store: Arc<States>,
     connector: C,
 ) where
     C: WaitForOpened + WaitForAccepted + WaitForSettled + WaitForCancelled,
 {
-    let mut events = new(&connector, secret_hash, swap_params)
+    let mut events = new(&connector, params)
         .inspect_ok(|event| tracing::info!("yielded event {}", event))
         .inspect_err(|error| tracing::error!("swap failed with {:?}", error));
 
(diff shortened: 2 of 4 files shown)


cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -33,39 +38,77 @@ enum InvoiceState {
...
+    pub cltv_expiry: Timestamp,
     pub state: InvoiceState,
+    pub fallback_addr: ethereum::Address,
     #[serde(deserialize_with = "deserialize_r_preimage")]
     pub r_preimage: Option<[u8; 32]>,
 }
 
+impl ValidateParams for Invoice {
+    fn validate(self, params: halight::Params) -> Result<(), ValidationError> {
+        if params.lightning_cltv_expiry == self.cltv_expiry
```
rishflab, line 76: I am still not sure about these checks particularly the fallback_addr and lightning_amount

cnd/src/network.rs
```diff
@@ -1181,7 +1183,7 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
                             Role::Alice => {
                                 tokio::task::spawn({
                                     let lnd_connector: LndConnectorAsReceiver = (**lnd_connector_params).clone().into();
-                                    halight::new_halight_swap(local_swap_id, secret_hash, self.halight_states.clone(), lnd_connector)
+                                    halight::new_halight_swap(local_swap_id, secret_hash, Params::from(create_swap_params.clone()), self.halight_states.clone(), lnd_connector)
```
D4nte, line 1186:
```suggestion
                                    halight::new_halight_swap(local_swap_id, secret_hash, create_swap_params.clone().into(), self.halight_states.clone(), lnd_connector)
```
Easier to read.

cnd/src/network.rs
```diff
@@ -1216,7 +1218,7 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit_ln::BehaviourOutEvent> fo
                             Role::Bob => {
                                 tokio::task::spawn({
                                     let lnd_connector: LndConnectorAsSender = (**lnd_connector_params).clone().into();
-                                    self::halight::new_halight_swap(local_swap_id, secret_hash, self.halight_states.clone(), lnd_connector)
+                                    self::halight::new_halight_swap(local_swap_id, secret_hash, Params::from(create_swap_params.clone()), self.halight_states.clone(), lnd_connector)
```
D4nte, line 1221:
```suggestion
                                    self::halight::new_halight_swap(local_swap_id, secret_hash, create_swap_params.clone().into(), self.halight_states.clone(), lnd_connector)
```

cnd/src/swap_protocols/halight.rs
```diff
@@ -243,7 +263,25 @@ where
...
+    pub lightning_cltv_expiry: Timestamp,
+    pub lightning_amount: asset::Bitcoin,
+}
+
+impl Params {
+    pub fn from(params: HanEtherereumHalightBitcoinCreateSwapParams) -> Self {
+        Params {
+            ethereum_identity: params.ethereum_identity,
+            ethereum_absolute_expiry: params.ethereum_absolute_expiry,
+            // ethereum_amount: params.ethereum_amount,
```
D4nte, line 281:
```suggestion
```

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -33,39 +38,77 @@ enum InvoiceState {
...
+            && params.ethereum_identity == EthereumIdentity::from(self.fallback_addr)
```
D4nte, line 77:
```suggestion
```
We do not use the fallback address.

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -33,39 +38,77 @@ enum InvoiceState {
...
+#[derive(Clone, Copy, Debug, Deserialize)]
+pub struct Invoice {
+    #[serde(deserialize_with = "deserialize_amount")]
+    pub value: asset::Bitcoin,
+    #[serde(deserialize_with = "deserialize_timestamp")]
+    pub expiry: Timestamp,
+    #[serde(deserialize_with = "deserialize_timestamp")]
+    pub cltv_expiry: Timestamp,
     pub state: InvoiceState,
+    pub fallback_addr: ethereum::Address,
```
D4nte, line 69: It is not possible to have an Ethereum Address on lnd.
```suggestion
```

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -33,39 +38,77 @@ enum InvoiceState {
...
     #[serde(deserialize_with = "deserialize_r_preimage")]
     pub r_preimage: Option<[u8; 32]>,
 }
 
+impl ValidateParams for Invoice {
+    fn validate(self, params: halight::Params) -> Result<(), ValidationError> {
+        if params.lightning_cltv_expiry == self.cltv_expiry
+            && params.ethereum_identity == EthereumIdentity::from(self.fallback_addr)
+            && params.ethereum_absolute_expiry == self.expiry
```
D4nte, line 78:
```suggestion
```
`self.expiry` is the invoice's expiry. It is not (directly) related to the expiry on Ethereum.

cnd/src/swap_protocols/halight.rs
```diff
@@ -243,7 +263,25 @@ where
...
-    pub secret_hash: SecretHash,
+    pub ethereum_identity: EthereumIdentity,
+    pub ethereum_absolute_expiry: Timestamp,
+    // pub ethereum_amount: asset::Ether,
+    pub lightning_identity: identity::Lightning,
+    pub lightning_cltv_expiry: Timestamp,
+    pub lightning_amount: asset::Bitcoin,
+}
+
+impl Params {
```
D4nte, line 276:
```suggestion
impl From<HanEtherereumHalightBitcoinCreateSwapParams> for Params {
```
Implement the `From` trait to have `Into` for free.

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -94,7 +137,7 @@ impl LndConnectorParams {
 
 fn read_file<T>(path: PathBuf) -> anyhow::Result<T>
 where
-    T: TryFrom<Vec<u8>, Error = Error>,
+    T: TryFrom<Vec<u8>, Error = anyhow::Error>,
```
D4nte, line 140:
```suggestion
    T: TryFrom<Vec<u8>, Error = Error>,
```
`anyhow::Error` is already imported in this file.

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -432,6 +513,56 @@ fn client(certificate: &Certificate, macaroon: &Macaroon) -> Result<reqwest::Cli
...
+where
+    D: Deserializer<'de>,
+{
+    struct Visitor;
+
+    impl<'de> de::Visitor<'de> for Visitor {
+        type Value = asset::Bitcoin;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+            formatter.write_str("a lightning r_preimage which is a base64 of an 32 byte array")
```
D4nte, line 526: This does not seem to be correct as we are deserialising a bitcoin asset here.

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -432,6 +513,56 @@ fn client(certificate: &Certificate, macaroon: &Macaroon) -> Result<reqwest::Cli
     Ok(client)
 }
 
+pub fn deserialize_amount<'de, D>(deserializer: D) -> Result<asset::Bitcoin, D::Error>
```
D4nte, line 516: We already implemented this deserialisations somewhere else, can we try to re-use the code instead?

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -432,6 +513,56 @@ fn client(certificate: &Certificate, macaroon: &Macaroon) -> Result<reqwest::Cli
...
+where
+    D: Deserializer<'de>,
+{
+    struct Visitor;
+
+    impl<'de> de::Visitor<'de> for Visitor {
+        type Value = Timestamp;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+            formatter.write_str("a lightning r_preimage which is a base64 of an 32 byte array")
```
D4nte, line 551: Copy-pasta

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -490,6 +619,7 @@ mod tests {
 
         let invoice_json = r#"{
       "r_preimage": "FyNMCPE5nm96/QZUNXmFNzzDYYEaBtpXNcNe07bC+f8=",
+      "fallback_addr": "3MXqbMwf457U4Jaw35WnJdnL99mq7Q8oQQ",
```
D4nte, line 622: We don't use the fallback address

cnd/src/swap_protocols/halight.rs
```diff
@@ -243,7 +263,23 @@ where
     })
 }
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Copy, Clone, Debug)]
 pub struct Params {
-    pub secret_hash: SecretHash,
+    pub ethereum_identity: EthereumIdentity,
+    pub ethereum_absolute_expiry: Timestamp,
```
thomaseizinger, line 269: :mega: How are these two concerned with the halight protocol?

cnd/src/swap_protocols/halight.rs
```diff
@@ -243,7 +263,23 @@ where
...
+    pub lightning_identity: identity::Lightning,
+    pub lightning_cltv_expiry: Timestamp,
+    pub lightning_amount: asset::Bitcoin,
```
thomaseizinger, line 272: :mega: We are within the `halight` scope here, the `lightning_` prefix is just noise in this module.

cnd/src/swap_protocols/halight.rs
```diff
@@ -243,7 +263,23 @@ where
...
+}
+
+impl From<HanEtherereumHalightBitcoinCreateSwapParams> for Params {
```
thomaseizinger, line 275: :mega: If you want to do an impl like this, it shouldn't live in this module because you are creating a circular dependency here: The code that uses defines `HanEtherereumHalightBitcoinCreateSwapParams` also uses `Params` so both modules import from each other. Also `HanEtherereumHalightBitcoinCreateSwapParams` is not halight-specific but knows "more" about the codebase than this particular module.

cnd/src/swap_protocols/halight/connector.rs
thomaseizinger: We haven't interacted with LN before so I wouldn't want to re-use API-specific code that just happens to be the same.
There is no super harmful duplication here as we just delegate to `from_sat` for the crucial part.
It is annoying that we have to do this, but if we want to avoid it then the solution shouldn't be: "just use the HTTP API serialization code".

cnd/src/swap_protocols/halight.rs
```diff
@@ -243,7 +263,23 @@ where
     })
 }
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Copy, Clone, Debug)]
 pub struct Params {
```
D4nte, line 267: Looking at the usage of `halight::Params` I think it would make sense to add the secret hash to it as they are always used together.

```suggestion
pub struct Params {
	pub secret_hash: SecretHash,
```

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -432,6 +508,56 @@ fn client(certificate: &Certificate, macaroon: &Macaroon) -> Result<reqwest::Cli
...
+where
+    D: Deserializer<'de>,
+{
+    struct Visitor;
+
+    impl<'de> de::Visitor<'de> for Visitor {
+        type Value = Timestamp;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+            formatter.write_str("a blocknumber as a string")
```
thomaseizinger, line 546: Is it actually a block number? I think it is seconds :thinking:

cnd/src/swap_protocols/halight/connector.rs
thomaseizinger: I would actually prefer if we never import it because it makes it hard to know, which `Error` it is without jumping to the definition. `anyhow::Error` is still short and avoids any kind of ambiguity.

tcharding:
Looks like the boys have got you covered on review for this one @rishflab, ping me if I'm needed please, otherwise I'll leave this one to them.

cnd/src/swap_protocols/halight.rs
rishflab: I had a look and only facade is importing halight. I cant see any circular dependency?

thomaseizinger: Okay, forget the circular dependency talk :D
This `impl` deals with more than just halight (the current module) and should be defined in a different place IMO :)
Probably wherever `HanEtherereumHalightBitcoinCreateSwapParams` is defined makes the most sense!

You can see it as: conceptually, the halight module doesn't (and shouldn't) "know" in which swaps it is going to be involved. This is a concern of code that composes those modules together.

cnd/src/swap_protocols/halight.rs
D4nte: @rishflab not keen on this one?

thomaseizinger: I agree with Franck here, for other connectors the secret hash is part of the params :)

thomaseizinger:
@rishflab I fixed up the merge conflicts and pushed two follow-up commits that address some of the comments that were still open. Hope you find it useful :)

Issue #2242, opened 2020-03-13
Test docker image build as part of the CI


D4nte:
## Problem

We build a docker image for cnd as part of the release process. This is then used in create-comit-app.

The image build process is failing because we never test it and only do it at release time.

## Goal

- Use CI to ensure that the docker image build is never broken.
- Verification should include that cnd runs: Start container and do a curl `GET /`

## Recommendation

Build the docker image as part of the regular PR runs in GitHub actions.

Note: Preferably do this as part of the release script. we should build it for the release, start it up, start cnd and only commence with releasing if cnd starts up.

bonomat:
@D4nte : please refine this issue.

PR #2574, opened 2020-05-05
Fix changelog format to avoid release failure

da-kami:
`Fixed` is a 2nd level header but in changelogs, it should be 3rd level.

I had trouble test-releasing when working on https://github.com/comit-network/comit-rs/issues/2242
@thomaseizinger pointed me to the problem.


thomaseizinger:
> @thomaseizinger are you aware of a lib that could check that in the CI?

No, we could built it ourselves if we want to but I don't think it is worth the effort atm :)

da-kami:
bors retry

PR #2575, opened 2020-05-05
Replace `parity` config section with `web3_url`

### LINKED ISSUE #2552: Use of the identifier `Parity` in config module
## Problem

We switched from Parity to Geth for our e2e tests with https://github.com/comit-network/comit-rs/pull/2499.
The `config` module however still uses the name `Parity` as we thought we only support `Parity` since we never ran our tests against a Geth node. 
 
This changed now and we know that cnd is able to deal with a Parity node as well as a Geth node. 

Hence, the config module is not correct anymore which might be confusing to users to have to configure "parity" if they are not using Parity as their Ethereum node.

## Goal

Create software that is easy to understand and configure.

D4nte:
Resolves #2552

### DIFF
--- a/cnd/src/config.rs
+++ b/cnd/src/config.rs
@@ -61,14 +61,14 @@ impl From<Bitcoin> for file::Bitcoin {
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct Ethereum {
     pub chain_id: ethereum::ChainId,
-    pub parity: Parity,
+    pub geth: Geth,
 }
 
 impl From<Ethereum> for file::Ethereum {
     fn from(ethereum: Ethereum) -> Self {
         file::Ethereum {
             chain_id: ethereum.chain_id,
-            parity: Some(ethereum.parity),
+            geth: Some(ethereum.geth),
         }
     }
 }
@@ -77,7 +77,7 @@ impl Default for Ethereum {
     fn default() -> Self {
         Self {
             chain_id: ethereum::ChainId::regtest(),
-            parity: Parity {
+            geth: Geth {
                 node_url: Url::parse("http://localhost:8545")
                     .expect("static string to be a valid url"),
             },
@@ -86,7 +86,7 @@ impl Default for Ethereum {
 }
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
-pub struct Parity {
+pub struct Geth {
     pub node_url: Url,
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Fix windows build.
 -   Return an error when creating a halight swap if lnd certificate or macaroon are unavailable instead of failing silently.
 
+## Changed
+
+-   **Breaking Change**: Rename `parity` to `geth` in the configuration file as we are only testing the software with Geth as part of the CI.
+
 ## [0.7.3] - 2020-04-14
 
 ### Fixed
--- a/api_tests/src/config.ts
+++ b/api_tests/src/config.ts
@@ -112,13 +112,13 @@ interface LedgerConnectors {
     lightning?: LightningConnector;
 }
 
-interface Parity {
+interface Geth {
     node_url: string;
 }
 
 interface EthereumConnector {
     chain_id: number;
-    parity: Parity;
+    geth: Geth;
 }
 
 interface Bitcoind {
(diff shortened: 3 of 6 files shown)


cnd/src/config.rs
```diff
@@ -61,14 +61,14 @@ impl From<Bitcoin> for file::Bitcoin {
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct Ethereum {
     pub chain_id: ethereum::ChainId,
-    pub parity: Parity,
+    pub web3_url: Url,
```
thomaseizinger, line 64: That is kinda inconsistent with the bitcoin side of things now.

Why not leave the section there but name it `web3` and the field can remain `node_url`?

D4nte: For Bitcoin we do specify that we support `bitcoind`. Hence, it is sufficient to have `node_url` terminology as the context tells us already that the given node must be `bitcoind`.

However, in the case of Ethereum, we support a generic node and only named the section `ethereum`.
By naming the url `web3_url` we do specify that we are expecting the `web3` api.

thomaseizinger: What I was pointing out is that we are missing a layer now for Ethereum. From what you are saying, we would have to change the `node_url` for Bitcoin to `bitcoind_node_url` to be consistent again.

The point of the additional "layer" in the configuration was to make it trivial to extend with other connectors even if we just support one at the moment.

I don't mind removing the "connector" layer but it should be consistent for all blockchains we support. At the moment we have:

```
[bitcoin]
network = "regtest"

[bitcoin.bitcoind]
node_url = "http://localhost:18443/"

[ethereum]
chain_id = 1337

[ethereum.parity]
```

but with this change, we have:

```
[bitcoin]
network = "regtest"

[bitcoin.bitcoind]
node_url = "http://localhost:18443/"

[ethereum]
chain_id = 1337
web3_url = "http://localhost:8545/"
```

tcharding: I'd like to see, what ever way is chosen, that its uniform across ledgers also please.

D4nte: Thanks for clarifying. I don't think it makes to align with Bitcoin just for the sake of uniformity.

In this case, I would suggest:

```
[bitcoin]
network = "regtest"

[bitcoin.bitcoind]
node_url = "http://localhost:18443/"

[ethereum]
chain_id = 1337

[ethereum.web3]
url = "http://localhost:8545/"
```

Note that a web3 endpoint is not necessarily a `node`. Infura is probably a cloud base service and most likely request does not target a specific `node`. In any case, what matters is that we expect a web3 api and hence the `node` terminology is unnecessary at best.

I am happy to also change `bitcoin.bitcoind.node_url` to `url` for the sake of _uniformity_. But again, I do not see any point in doing so.

bonomat: The reason why we had `parity` in the first place was that we explicitly said we only support Parity as we never tested against Geth. Now it's the other way around: we will keep testing against Geth but don't know about Parity for sure. 

I'm not proposing a solution but just wanted to mention this.

D4nte: > The reason why we had `parity` in the first place was that we explicitly said we only support Parity as we never tested against Geth. Now it's the other way around: we will keep testing against Geth but don't know about Parity for sure.
> 
> I'm not proposing a solution but just wanted to mention this.

Thanks for the input. Is that something we still want to uphold? If someone come to us and they use infura and we need to do a code change to support infura, would we reject and say "use geth" or we will attempt to support any reasonable Ethereum software?

thomaseizinger: To avoid ambiguity, what about just changing `parity` to `geth` and leaving the rest as is?

Infura uses API-keys to my knowledge, so just providing a URL won't work for that even though it is a web3 compatible API. (see here: https://blog.infura.io/getting-started-with-infura-28e41844cc89/).

If we say `ethereum.geth.node_url`, it is unambigous what we officially support and the extension path for more connectors is trivial as-well.

Maybe `ethereum.geth.web3_url` is even better to make it clear that we don't want a websocket endpoint or anything like that!

If over time, we find a more abstract model for representing the same configuration, we can always refactor towards that :)

bonomat: > Thanks for the input. Is that something we still want to uphold? 

Yes, I think it would make sense to explicitly state what we support, e.g. Geth instead of Parity. 

> If someone come to us and they use infura and we need to do a code change to support infura, would we reject and say "use geth" or we will attempt to support any reasonable web3 provider?

Good question. I don't think there is an answer for this right now and I would wait until this happens and decide then.

D4nte: > Yes, I think it would make sense to explicitly state what we support, e.g. Geth instead of Parity.

Ok, I did not think this would be the answer. I'll proceed with the change.

D4nte:
> It's still a breaking change right?
> 
> I've added a no-mergify label just in case.

Well spotted.

D4nte:
bors retry

PR #2576, opened 2020-05-05
Test that cnd is reachable in created docker image before releasing 

da-kami:
resolves: #2242 

Note that so far the commands were only tested locally on my machine, I did not run the action on github-actions yet.

### DIFF
--- a/.github/workflows/publish-new-release.yml
+++ b/.github/workflows/publish-new-release.yml
@@ -94,11 +94,23 @@ jobs:
 
           echo "::set-output name=version::$VERSION"
 
+      - name: Build docker image as ${{ steps.extract-version.outputs.version }} and latest
+        run: |
+          VERSION="${{ steps.extract-version.outputs.version }}"
+          docker build . -t comitnetwork/cnd:$VERSION -t comitnetwork/cnd:latest
+
+      - name: Test docker image by starting it and curling the cnd API
+        run: |
+          VERSION="${{ steps.extract-version.outputs.version }}"
+          echo "::starting docker-container comitnetwork/cnd:$VERSION"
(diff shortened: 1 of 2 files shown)


da-kami:
I'm still running some tests against my fork.
I had trouble properly creating the docker image locally after a release build. 
When starting up the docker image I get the message 

```
/usr/local/bin/cnd: 1: /usr/local/bin/cnd: Syntax error: word unexpected (expecting ")")
```

It is hard to figure out if this is related to how I do things on my machine or if the current release actually causes trouble. Since I don't want to waste time setting up more things that are actually different than what we do in the actual jobs I decided to:

- reduce the release jobs to the new functionality to be tested on my fork (removing all the actual release parts in the end)
- run a happy-path release on the fork (and see that a message is printed in the actual release part - but nothing more)
- run an error-path where the URL that is triggered points to a wrong port. 

I expect that the functionality will not have to be changed, but will wait for those 2 scenarios to pass and then remove the label.

bonomat:
> I'm still running some tests against my fork.
I had trouble properly creating the docker image locally after a release build.
When starting up the docker image I get the message

You can't build the docker image on your local machine in the same way as we do it in the action. You are running MacOS and the docker container Ubuntu. The build file will be incompatible.

da-kami:
> > I'm still running some tests against my fork.
> > I had trouble properly creating the docker image locally after a release build.
> > When starting up the docker image I get the message
> 
> You can't build the docker image on your local machine in the same way as we do it in the action. You are running MacOS and the docker container Ubuntu. The build file will be incompatible.
> 
> // Edit // I ran the action on [my fork](https://github.com/bonomat/comit-rs/pull/2) and it worked also in the error case. Feel free merge at your convenience ;)

I love you @bonomat ! - Merging then.

Issue #2546, opened 2020-04-28
Make the DB the source of truth for fetching swaps


thomaseizinger:
## Problem

Currently, we use the `ComitLN` network behaviour to fetch finalized swaps: https://github.com/comit-network/comit-rs/blob/d3821a5db2ba823f738aed60e9512ae69ccc5346/cnd/src/network/comit_ln.rs#L134
This was a temporary solution because we didn't want to deal with the database during the initial implementation of halight-swaps.

## Goal

Fetch finalized swaps from the database instead.

We are still dealing with only one ledger combination here.
Note: when starting cnd we do not load anything from the db for halight swaps (and we keep that way).
Note: DB API should already be in place thanks to https://github.com/comit-network/comit-rs/issues/2545

Issue #2545, opened 2020-04-28
Implement the new database design for herc20 <-> halight swaps


thomaseizinger:
## Problem

We are currently not persisting any data to disk for swaps involving the halight protocol.

## Goal

- Persist data for the herc20 <-> halight swap combination after the swap has been created through the API and after it has been finalized. Due to the modular design of the schema, one implementation should already allow **both directions** for this protocol pair (herc20 <-> halight).
- Test the design proposed in https://github.com/comit-network/spikes/pull/62.

Note that, even though the DB design is split across many tables, this is an implementation detail of the DB layer and should not leak to the outside of it.

The Database needs to provide you an API that allows you to fetch and save a whole swap by local id and that needs to be unit tested. It will be integrated in prod code with #2399 and #2546.
No code changes outside of the database module are expected.

## Definition of Done

- [ ] You implemented two tests that save and retrieve a swap by ID (one per protocol combination).

thomaseizinger:
@da-kami @tcharding Please note the updated description and a slight change in scope, in particular, this ticket now includes two combinations instead of just one. Due to the modular design of the database schema, this should be almost for free though.

PR #2577, opened 2020-05-05
Add finalized swap api

tcharding:
Add the rest of the database API.  We now have three entry points to the database, save created swap, save finalized swap, and retrieve live swap.

This PR introduces a new naming scheme for swaps, please comment if you wish to

```
    Swaps can exist at various stages of their lifecycle, we have tentative
    names for two of these stages so far

    - 'Created' swap - this is a swap that has been created by a POST call to
    the REST API.

    - 'Finalized' swap - this is swap after we have shared the data required
      to/from the counterparty node.

    The data base module controls these types.  This patch introduces a
    third term 'live' swap. The term 'finalized' is reduced to mean just the
    additional data shared during the communication protocols.

    - 'Live' swap - a swap that has been created, has been 'finalized' (has
      the additional data shared during communication protocols) and is now
      'live' in the system.

    RFC: Request For Comments on these names please.
```

Notes: should `GetLiveSwap` use 'retrieve' or 'load' instead of 'get'?


cnd/src/network/comit_ln.rs
```diff
@@ -129,6 +129,9 @@ impl ComitLN {
         Ok(())
     }
 
+    // TODO: This function needs to get its data from the database _not_ from
```
thomaseizinger, line 132: There is a ticket for this: https://github.com/comit-network/comit-rs/issues/2546

cnd/src/db.rs
```diff
@@ -152,6 +159,53 @@ pub struct CreatedSwap<A, B> {
...
 
+// This only supports han-eth/halight-btc swap.
+/// Data learned during the communication protocols that enables a created swap
+/// to be finalized.
+#[derive(Debug, Clone, Copy)]
+pub struct FinalizedSwap {
+    /// The shared swap id, generated by Bob.
+    pub swap_id: SharedSwapId,
+    /// The secret hash, generated by Alice.
+    pub secret_hash: SecretHash,
```
thomaseizinger, line 170: I think we decided on Alice not storing the secret hash at all.

cnd/src/db.rs
```diff
@@ -152,6 +159,53 @@ pub struct CreatedSwap<A, B> {
...
+    pub beta_asset: asset::Bitcoin,
+    pub alpha_ledger_refund_identity: identity::Ethereum,
+    pub alpha_ledger_redeem_identity: identity::Ethereum,
+    pub beta_ledger_refund_identity: identity::Lightning,
+    pub beta_ledger_redeem_identity: identity::Lightning,
+    pub alpha_expiry: Timestamp,
+    pub beta_expiry: Timestamp,
+    pub swap_id: LocalSwapId,
+    pub secret_hash: SecretHash,
+    pub secret: Option<Secret>,
```
thomaseizinger, line 199: This is weird. Do we actually need all of this data in one place at some point? It is probably easier to name things if we don't create "god" structs that know everything but approach it more on a usecase basis.

For example the API never needs to know the secret hash as far as I know and neither does it need the identities.

cnd/src/db.rs
da-kami: We only discussed not storing the secret. But we could extend it to the hash for Alice. Since she computes the secret, she can also compute the hash from it. @bonomat do you have an opinion on that?

cnd/src/db.rs
da-kami: What separates a `LiveSwap` from a `FinalizedSwap`? Isn't it considered `live` once it is finalized? We would spawn the watchers once the party received and sent `finalized` (and only then a swap is `finalized` in my head). What am I missing?

cnd/src/network/comit_ln.rs
da-kami: yeah, I'd remove that comment.

da-kami:
> Add the rest of the database API. We now have three entry points to the database, save created swap, save finalized swap, and retrieve live swap.
> 
> This PR introduces a new naming scheme for swaps, please comment if you wish to
> 
> ```
>     Swaps can exist at various stages of their lifecycle, we have tentative
>     names for two of these stages so far
> 
>     - 'Created' swap - this is a swap that has been created by a POST call to
>     the REST API.
> 
>     - 'Finalized' swap - this is swap after we have shared the data required
>       to/from the counterparty node.
> 
>     The data base module controls these types.  This patch introduces a
>     third term 'live' swap. The term 'finalized' is reduced to mean just the
>     additional data shared during the communication protocols.
> 
>     - 'Live' swap - a swap that has been created, has been 'finalized' (has
>       the additional data shared during communication protocols) and is now
>       'live' in the system.
> 
>     RFC: Request For Comments on these names please.
> ```
> 
> Notes: should `GetLiveSwap` use 'retrieve' or 'load' instead of 'get'?

Note that I read this, but I still don't understand the need to separate `finalized` from `live` swaps.

cnd/src/network/comit_ln.rs
tcharding: Woops this was added before i decided to do this as a separate PR :) will remove all the TODOs.

cnd/src/db.rs
tcharding: If Alice doesn't store it then its not symmetrical for Bob and Alice, what benefit does this bring?  I'd prefer to see symmetry here.

cnd/src/db.rs
tcharding: The idea was to have two types used for storing the data in the database (created and finalized) and then a type for retrieving the data (live).  This way _how_ the database stores the data is kept hidden within the database module.  The `LiveSwap` type added in this PR is just the data structure that was `FinalizedSwap` in the network code (created by lookups in the hash maps).  Since this is the functionality that we are trying to replace in #2545 it seems likely that we will need to expose exactly the same data from the db module as we do currently from the network module.  Does that make sense?  If this is not clear as a stand alone PR we (@da-kami and I) can just base our work on top of this and do a single large PR.

tcharding:
On second thoughts, I can't be bothered debating this here.  The point of this PR was just to help us move faster, it hasn't achieved that so I'm closing it and building a bigger PR with these as the first two commits.

PR #2580, opened 2020-05-06

dhaileytaha:
removed Option<Hash> from eth block struct


thomaseizinger:
There also seem to be still some compile errors in the tests: https://github.com/comit-network/comit-rs/pull/2580/checks?check_run_id=648007278#step:7:665

You can run `make clippy` to trigger what the CI is doing here.

dhaileytaha:
Complete in https://github.com/comit-network/comit-rs/pull/2582

PR #2583, opened 2020-05-06
Bob does not reject announcement if the swap is not yet created

### LINKED ISSUE #2488: Bob shouldn't drop incoming announcements for swaps he doesn't know about (yet)
## Problem

Currently, if Bob receives an announcement with a swap digest that he doesn't know about, he will simply reject the announcement by closing the substream:

https://github.com/comit-network/comit-rs/blob/5a7b9af1e304dab3b0c54b6404a769237fed8dd6/cnd/src/network/comit_ln.rs#L347-L356

This is a problem because it might very well happen that Bob's user was just slightly slower than Alice's with adding the swap to cnd.

## Goal

Don't drop incoming announcements immediately but instead wait for some amount of time before dropping it.
Non-configurable time can be decided at implementation time.
No changes around the DB are expected with this issue.
e2e test is expected.

D4nte:
Bob now stores swaps announced by Alice even if they are not yet created.

When Bob creates a swap (via REST API), then it checks such swaps and proceed with confirmation and communication if present.

Please note I amended the current test to check this behaviour to avoid adding one more full swap e2e test.

Helps with #2488.

### DIFF
--- a/api_tests/tests/ether_halight.ts
+++ b/api_tests/tests/ether_halight.ts
@@ -13,11 +13,9 @@ it(
         const bodies = (await SwapFactory.newSwap(alice, bob))
             .hanEthereumEtherHalightLightningBitcoin;
 
-        // Bob needs to know about the swap first because he is not buffering incoming announcements about swaps he doesn't know about
-        await bob.createSwap(bodies.bob);
-        await sleep(500);
-
         await alice.createSwap(bodies.alice);
+        await sleep(500);
+        await bob.createSwap(bodies.bob);
 
         await alice.init();
 
--- a/api_tests/tests/ether_halight.ts
+++ b/api_tests/tests/ether_halight.ts
@@ -6,6 +6,8 @@
 import SwapFactory from "../src/actors/swap_factory";
 import { sleep } from "../src/utils";
 import { twoActorTest } from "../src/actor_test";
+import { Actor } from "../src/actors/actor";
+import { SwapStatus } from "../src/swap_response";
 
 it(
     "han-ethereum-ether-halight-lightning-bitcoin-alice-redeems-bob-redeems",
@@ -45,31 +47,17 @@ it(
         // Simulate that Bob is awaiting a swap from a different peer-id than Alice node's peer-id.
         bodies.bob.peer.peer_id =
             "QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N";
-        await bob.createSwap(bodies.bob);
-        await sleep(500);
-
         await alice.createSwap(bodies.alice);
+        await bob.createSwap(bodies.bob);
 
-        const config = {
-            maxTimeoutSecs: 1,
-            tryIntervalSecs: 0.3,
-        };
+        await assertSwapCreated(alice);
+        await assertSwapCreated(bob);
 
-        // Due to the asynchronous nature of the announce protocol (and the fact that there is not error handling (yet)),
-        // this test only tests that the "init" action does not become available.
-        //
-        // Scenario details:
-        // 0) Alice and Bob agree on the swap params (negotiation); Alice gives Bob a peer-id that is not her node's peer-id.
-        // 1) Bob posts swap, retrieves local swap-ID
-        // 2) Alice posts swap, retrieves local swap-ID
-        //      2.1) Internally the announce protocol runs into an Error because Alice' peer-id does not match the one she gave Bob.
-        //              This Error, is however not visible to the outside.
-        //              The application is responsible for removing swaps that do not move forward after a given time.
-        const aliceResponsePromise = alice.init(config);
-        return expect(aliceResponsePromise).rejects.toThrowError();
-        // Same for Bob's side
-        const bobResponsePromise = bob.fund(config);
-        return expect(bobResponsePromise).rejects.toThrowError();
+        await sleep(1000);
+
+        // Assert that the swaps are still not finalized
+        await assertSwapCreated(alice);
+        await assertSwapCreated(bob);
     })
 );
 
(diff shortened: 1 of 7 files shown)


cnd/src/network/comit_ln.rs
```diff
@@ -54,8 +55,17 @@ pub struct ComitLN {
     #[behaviour(ignore)]
     events: VecDeque<BehaviourOutEvent>,
 
+    /// Swaps As Alice
```
da-kami, line 58:
```suggestion
    /// Swaps as Alice
```

D4nte:
> Still an interesting question to me, if "cleanup" should be done by `cnd` or the application on top.
Clean up is done at restart + We can worry about it once it becomes an issue as this goes in the memory performance bucket.

cnd/src/network/comit_ln.rs
```diff
@@ -109,20 +121,29 @@ impl ComitLN {
     ) -> anyhow::Result<()> {
         let digest = create_swap_params.clone().digest();
 
-        if self.swaps_waiting_for_announcement.contains_key(&digest) {
+        if self.swaps_waiting_for_announcement.contains_key(&digest)
+            || self.swaps_waiting_for_confirmation.contains_key(&digest)
+        {
             anyhow::bail!(SwapExists)
         }
+        tracing::info!("Swap created: {}", digest);
```
thomaseizinger, line 129: I think this log message is pretty misleading, we are not "creating" anything here.

api_tests/tests/ether_halight.ts
```diff
@@ -13,11 +13,9 @@ it(
         const bodies = (await SwapFactory.newSwap(alice, bob))
             .hanEthereumEtherHalightLightningBitcoin;
 
-        // Bob needs to know about the swap first because he is not buffering incoming announcements about swaps he doesn't know about
-        await bob.createSwap(bodies.bob);
-        await sleep(500);
-
         await alice.createSwap(bodies.alice);
+        await sleep(500);
+        await bob.createSwap(bodies.bob);
```
luckysori, line 18: :thinking: Instead of replacing this test, can we just add a dry test that ensures that creating a swap in the opposite order succeeds?

cnd/src/network/comit_ln.rs
```diff
@@ -315,6 +318,7 @@ impl NetworkBehaviourEventProcess<announce::behaviour::BehaviourOutEvent> for Co
     fn inject_event(&mut self, event: announce::behaviour::BehaviourOutEvent) {
         match event {
             announce::behaviour::BehaviourOutEvent::ReceivedAnnouncement { peer, mut io } => {
+                tracing::info!("Peer {} announced a swap ({})", peer, io.swap_digest);
```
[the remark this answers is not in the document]
thomaseizinger: Interesting idea! Yes, more tracing spans please <3

D4nte:
@thomaseizinger what change are you requesting?

thomaseizinger:
> @thomaseizinger what change are you requesting?

To cleanup of these pending announcements after some period of time.

The ticket says: "Don't drop incoming announcements immediately but instead wait for some amount of time before dropping it."

api_tests/tests/ether_halight.ts
D4nte: This is currently not possible with the current e2e test framework.

Doing https://github.com/comit-network/comit-rs/issues/2589 to make it possible.

D4nte:
This PR is getting bigger but not yet finished. I suggest reviewers to proceed with some pre-review to get familiar with code.
Only need to implement @thomaseizinger's comment now.

D4nte:
@thomaseizinger please let me do the clean up in a follow-up PR. I am already working on it but I am doing some refactoring and it will just explose the LOC change of this PR.

Please note I am removed the GitHub keywords so that #2488 does not get closed ;)

api_tests/tests/ether_halight.ts
```diff
@@ -47,31 +47,17 @@ it(
...
         await bob.createSwap(bodies.bob);
-        await sleep(500);
 
-        await alice.createSwap(bodies.alice);
+        await assertSwapCreated(alice);
+        await assertSwapCreated(bob);
+
+        await sleep(1000);
 
-        const config = {
```
thomaseizinger, line 55: This is a different test now.

We are no longer asserting the peer id check. We could do that as an internal test though and just get rid of this one?

cnd/src/network/comit_ln.rs
```diff
@@ -340,15 +459,11 @@ impl NetworkBehaviourEventProcess<announce::behaviour::BehaviourOutEvent> for Co
                         }
                     }
                     None => {
-                        tracing::warn!(
-                            "Peer {} announced a swap ({}) we don't know about",
-                            peer,
-                            io.swap_digest
+                        tracing::debug!(
+                            "Swap has not been created yet, waiting for REST API request."
```
thomaseizinger, line 463: Log messages should adhere to module boundaries in the same way that regular code does: Nothing in this module knows about the REST API, so IMO also log messages shouldn't.

D4nte: Ok, fixing in new branch.

D4nte: The main reason is that I want to be sure everyone understand the terminology of "created" but I guess being consistent with it will be the most helpful way to fix the meaning.

api_tests/tests/ether_halight.ts
D4nte: I don't agree, the test is equivalent.
But instead of checking for the availability of the actions (which seems quite an indirect way to check if the swap is finalized), we check the swap status.
If the swap status is still "created" then it means it has not been finalized.
When finalized, the swap status change to "In Progress".

I agree re internal check. I'll see if I get time to do it.

thomaseizinger: Ah ok I missed that one, sorry about that :)

PR #2590, opened 2020-05-07
Leverage cnd response to actually assert halight swaps

### LINKED ISSUE #2589: Use `GET /swap/:id` in e2e test
## Problem

There are hackery of various order in the halight tests because we were not able to check a swap status against the cnd REST API until #2521 was done.

Also, cnd provides a different interface for rfc003 and the new split protocol but currently it's all processed in the same Actor methods.

## Goal

- Use cnd's information to assert the swap status.
- Make code easier to manipulate by splitting code that uses rfc003 API vs split protocol API

D4nte:
Also separates rfc003 from split protocols code to make it easier to work on.

Resolves #2589.


api_tests/tests/lightning_routes.ts
```diff
@@ -18,22 +18,23 @@ describe("Lightning routes tests", () => {
             const bodies = (await SwapFactory.newSwap(alice, bob, true))
                 .hanEthereumEtherHalightLightningBitcoin;
 
-            const aliceResponse = alice.cnd.createHanEthereumEtherHalightLightningBitcoin(
```
thomaseizinger, line 21: The reason this was split was so the test follows the "arrange-act-assert" structure :)

D4nte: Yes but the test is unstable. For some reason it fails for the exact reason it should fail but jest does not catch the promise rejection. See commit comment.

thomaseizinger: Yeah I noticed that. Does this actually fix it? That is so weird ...

D4nte: I did ~5 full runs since then and the issue did not come back...

thomaseizinger: Hmm, well spotted then :)
I wonder if we could reproduce that somehow and report a bug!

PR #2594, opened 2020-05-08
Move all swap hashmaps to a dedicated struct

### LINKED ISSUE #2488: Bob shouldn't drop incoming announcements for swaps he doesn't know about (yet)

D4nte:
Having several different hashmap for each swap information
and status creates complexity in `ComitLn` and also creates
the risk of having bugs that lead to discrepancy between
the different hashmaps.

By moving all hashmap to a given struct that has a consistent
interface, it allows the focus on avoiding discrepancy
to be localized to this new `Swaps` struct.

It also later allows the migration of this `Swaps` struct
to a state machine if we see it as a gain.

Old pending swaps are also cleaned up. For now I put a hardcoded 5 minutes time out. We can decide to make it configurable at a later stage.

Resolves #2488

### DIFF
--- a/api_tests/src/swap_response.ts
+++ b/api_tests/src/swap_response.ts
@@ -53,23 +53,23 @@ export interface Properties {
  */
 export enum SwapStatus {
     /**
-     * The swap was created but the communication phase hasn't yet been finalized
+     * The swap was created via the REST API but the communication phase hasn't yet been finalized.
      */
     Created = "CREATED",
     /**
-     * The communication was finalized and blockchain actions are needed or happening
+     * The communication was finalized and blockchain actions are needed or happening.
      */
     InProgress = "IN_PROGRESS",
     /**
-     * The swap is finished and the assets were swapped
+     * The swap is finished and the assets were swapped.
      */
     Swapped = "SWAPPED",
     /**
-     * The swap is finished and the assets were not swapped
+     * The swap is finished and the assets were not swapped.
      */
     NotSwapped = "NOT_SWAPPED",
     /**
-     * An unexpected internal failure aborted the swap
+     * An unexpected internal failure aborted the swap.
      */
     InternalFailure = "INTERNAL_FAILURE",
 }
--- a/cnd/src/network/comit_ln.rs
+++ b/cnd/src/network/comit_ln.rs
@@ -4,7 +4,7 @@ use crate::{
         oneshot_behaviour,
         protocols::{
             announce,
-            announce::{behaviour::Announce, protocol::ReplySubstream, SwapDigest},
+            announce::{behaviour::Announce, protocol::ReplySubstream},
             ethereum_identity, finalize, lightning_identity, secret_hash,
         },
     },
@@ -31,6 +31,9 @@ use std::{
     fmt,
     task::{Context, Poll},
 };
+use swaps::Swaps;
+
+mod swaps;
 
 /// Event emitted  by the `ComitLn` behaviour.
 #[derive(Debug)]
@@ -54,27 +57,8 @@ pub struct ComitLN {
 
     #[behaviour(ignore)]
     events: VecDeque<BehaviourOutEvent>,
-
-    /// In role of Alice; swaps exist in here once a swap is created by Alice
-    /// (and up until an announce confirmation is received from Bob).
     #[behaviour(ignore)]
-    swaps_waiting_for_confirmation: HashMap<SwapDigest, LocalSwapId>,
-    /// In role of Bob; swaps exist in here if Bob creates the swap _before_ an
-    /// announce message is received from Alice (and up until the announce
-    /// message arrives).
-    #[behaviour(ignore)]
-    swaps_waiting_for_announcement: HashMap<SwapDigest, LocalSwapId>,
-    /// In role of Bob; swaps exist in here if Bob creates the swap _after_ an
-    /// announce message is received from Alice (and up until Bob creates the
-    /// swap).
-    #[behaviour(ignore)]
-    swaps_waiting_for_creation:
-        HashMap<SwapDigest, (libp2p::PeerId, ReplySubstream<NegotiatedSubstream>)>,
-
-    #[behaviour(ignore)]
-    swaps: HashMap<LocalSwapId, HanEtherereumHalightBitcoinCreateSwapParams>,
-    #[behaviour(ignore)]
-    swap_ids: HashMap<LocalSwapId, SharedSwapId>,
+    all_swaps: Swaps,
     #[behaviour(ignore)]
     ethereum_identities: HashMap<SharedSwapId, identity::Ethereum>,
     #[behaviour(ignore)]
@@ -106,11 +90,7 @@ impl ComitLN {
             lightning_identity: Default::default(),
             finalize: Default::default(),
             events: VecDeque::new(),
-            swaps_waiting_for_confirmation: Default::default(),
-            swaps_waiting_for_announcement: Default::default(),
-            swaps_waiting_for_creation: Default::default(),
-            swaps: Default::default(),
-            swap_ids: Default::default(),
+            all_swaps: Default::default(),
             ethereum_identities: Default::default(),
             lightning_identities: Default::default(),
             communication_state: Default::default(),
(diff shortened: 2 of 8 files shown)


D4nte:
@tcharding FYI wip

api_tests/src/swap_response.ts
```diff
@@ -53,23 +53,23 @@ export interface Properties {
  */
 export enum SwapStatus {
     /**
-     * The swap was created but the communication phase hasn't yet been finalized
+     * The swap was created via the REST API but the communication phase hasn't yet been finalized.
```
da-kami, line 56: Sorry, I missed that a new state `CREATED` was added.

I discussed this with @thomaseizinger recently and was under the impression, that there should not be a new `CREATED` state in the aggregated swap state. `IN_PROGRESS` means that `cnd` was made aware of the swap.

All the other states are final states (`SWAPPED`, `NOT_SWAPPED`, `INTERNAL_FAILURE`).

`CREATED` for me is within the communication protocol.

What was the motivation for adding this state? 😬

thomaseizinger: I don't think there is a point in discussing these states now :)
I would suggest to do whatever we think works currently. I see us adopting these names/states anyway to something that makes more sense once everything is in place.

cnd/src/network/comit_ln.rs
```diff
@@ -126,28 +109,27 @@ impl ComitLN {
...
         match create_swap_params.role {
             Role::Alice => {
                 tracing::info!("Starting announcement for swap: {}", digest);
                 self.announce
-                    .start_announce_protocol(digest.clone(), create_swap_params.peer);
-                self.swaps_waiting_for_confirmation.insert(digest, id);
+                    .start_announce_protocol(digest.clone(), (&create_swap_params.peer).clone());
+                self.swaps
+                    .create_as_pending_confirmation(digest, id, create_swap_params)
+                    .map_err(anyhow::Error::from)?;
```
thomaseizinger, line 120: No need for `map_err` with `?`.

cnd/src/network/comit_ln/swaps.rs
```diff
@@ -0,0 +1,602 @@
+use crate::{
+    network::protocols::announce::{protocol::ReplySubstream, SwapDigest},
+    swap_protocols::{HanEtherereumHalightBitcoinCreateSwapParams, LocalSwapId, SharedSwapId},
+    timestamp::Timestamp,
+};
+use digest::Digest;
+use libp2p::{swarm::NegotiatedSubstream, PeerId};
+use std::collections::HashMap;
+
+#[derive(Display, thiserror::Error, Clone, Copy, Debug)]
```
thomaseizinger, line 10: Deriving `Display` is not going to give you any helpful error. Please use `#[error]` instead.

You might also want to think about declaring single error structs and just use `anyhow::Error` for all APIs. Saves you from declaring this enum!

cnd/src/network/comit_ln/swaps.rs
```diff
@@ -0,0 +1,602 @@
...
+
+    /// Remove all pending (not finalized) swap older than `older_than`
+    pub fn clean_up_pending_swaps(&mut self, older_than: Timestamp) {
+        let digests: Vec<SwapDigest> = self
+            .timestamps
+            .iter()
+            .filter(|(_, timestamp)| **timestamp < older_than)
+            .map(|(digest, _)| digest)
+            .cloned()
+            .collect();
```
thomaseizinger, line 267: I always try to avoid type hints because they could change if you refactor the function. Giving type hints on the fn's that actually need them doesn't suffer from that :)

```suggestion
        let digests = self
            .timestamps
            .iter()
            .filter(|(_, timestamp)| **timestamp < older_than)
            .map(|(digest, _)| digest)
            .cloned()
            .collect::<Vec<_>>();
```

thomaseizinger: You could also avoid the allocation (i.e. using `collect`) by using the iterator with a for loop :)

```rust
for digest in digest_iter {
	self.pending_confirmation.remove(digest);
	// ...
}
```

cnd/src/network/comit_ln/swaps.rs
```diff
@@ -0,0 +1,602 @@
...
+            lightning_identity: identity::Lightning::random(),
+            lightning_cltv_expiry: 12345.into(),
+            lightning_amount: asset::Bitcoin::from_sat(1_000_000_000),
+        }
+    }
+
+    #[test]
+    fn created_swap_as_pending_confirmation_can_be_retrieved() {
+        let create_params = create_params();
+        let digest = create_params.clone().digest();
```
thomaseizinger, line 348: I didn't realize that digest consumes `self`, can we change that?

cnd/src/network/comit_ln/swaps.rs
```diff
@@ -0,0 +1,602 @@
...
+            create_params.clone(),
+        );
+
+        assert!(creation.is_ok());
+
+        let created_swap = swaps.get_created_swap(&local_swap_id);
+
+        assert!(created_swap.is_some());
+
+        assert_eq!(created_swap.unwrap(), create_params)
```
thomaseizinger, line 364: Nit: Avoid newlines to embrace arrange-act-assert :)

```suggestion
        assert!(creation.is_ok());
        let created_swap = swaps.get_created_swap(&local_swap_id);
        assert!(created_swap.is_some());
        assert_eq!(created_swap.unwrap(), create_params)
```

cnd/src/network/comit_ln/swaps.rs
```diff
@@ -0,0 +1,602 @@
...
+        swaps
+            .create_as_pending_confirmation(digest.clone(), local_swap_id, create_params.clone())
+            .unwrap();
+
+        let shared_swap_id = SharedSwapId::default();
+        let (_local_swap_id, _create_params) = swaps
+            .move_pending_confirmation_to_communicate(&digest, shared_swap_id)
+            .unwrap();
+
+        assert_eq!(local_swap_id, _local_swap_id);
```
thomaseizinger, line 439: Are these underscored because they are ignored? Should we give this a different name?

api_tests/src/swap_response.ts
D4nte: It is not new.

This was added as part of the design of the swap response on the cnd REST API many moons ago.

It is useful to test few things in the e2e test so I'd be keen to keep it.

da-kami:
Some e2e test failing locally, so I guess this needs more thinking... but now the weekend awaits 😅

D4nte:
> I think we should introduce a proper state-machine with enums indicating where we are here. The way `ComitLn.initiate_communication` and the functions in `Swap` change the state (based on the manipulation of multiple hashmaps) based on if statements is really hard to grasp.
> It would be great if this would be a more ordered state machine. 

Did you read the PR's description? 😁

da-kami:
> > I think we should introduce a proper state-machine with enums indicating where we are here. The way `ComitLn.initiate_communication` and the functions in `Swap` change the state (based on the manipulation of multiple hashmaps) based on if statements is really hard to grasp.
> > It would be great if this would be a more ordered state machine.
> 
> Did you read the PR's description? 😁

🙇‍♂️

da-kami:
> Some e2e test failing locally, so I guess this needs more thinking... but now the weekend awaits 😅

One more follow-up commit to re-add the peer-id check on Bob's side, that should fix the failing `communication` e2e test.

I have some trouble validating if all e2e tests are fixed now locally, because I temporarily switched to yarn 2 and that seems to have messed a few things up. 
Signing out now.

thomaseizinger:
I wouldn't want to invest too much time (i.e. state machine) into this before the spike regarding the redesign is done as it all might get invalidated :)

D4nte:
> I wouldn't want to invest too much time (i.e. state machine) into this before the spike regarding the redesign is done as it all might get invalidated :)

Oh yeah, I just want to make it work now.

cnd/src/network/comit_ln.rs
D4nte: 🤦‍♀️

cnd/src/network/comit_ln/swaps.rs
```diff
@@ -0,0 +1,749 @@
...
+            pending_creation: Default::default(),
+            swaps: Default::default(),
+            swap_ids: Default::default(),
+            timestamps: Default::default(),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
thomaseizinger, line 336: Thanks for adding all these tests!

cnd/src/network/comit_ln.rs
```diff
@@ -126,28 +109,28 @@ impl ComitLN {
...
-
-        self.swaps.insert(id, create_swap_params.clone());
 
         match create_swap_params.role {
             Role::Alice => {
                 tracing::info!("Starting announcement for swap: {}", digest);
                 self.announce
-                    .start_announce_protocol(digest.clone(), create_swap_params.peer);
-                self.swaps_waiting_for_confirmation.insert(digest, id);
+                    .start_announce_protocol(digest.clone(), (&create_swap_params.peer).clone());
```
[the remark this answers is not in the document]
D4nte: Nope :)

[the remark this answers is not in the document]
D4nte:
```
error[E0308]: mismatched types
   --> cnd/src/network/comit_ln.rs:117:62
    |
117 |                     .start_announce_protocol(digest.clone(), &create_swap_params.peer.clone());
    |                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                                                              |
    |                                                              expected struct `network::DialInformation`, found `&network::DialInformation`
    |                                                              help: consider removing the borrow: `create_swap_params.peer.clone()`
```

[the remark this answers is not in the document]
thomaseizinger: Ah okay, yeah the whole borrow is needless :)
I will push a follow-up!

Issue #2399, opened 2020-04-06
200 is returned for any swap id on `GET /swap/:id`


tcharding:
## Problem

Currently we always return 200 OK for any swap id that is passed to us when a user hits the HALight endpoint to get a particular swap.

## Goal

Do the right thing here: https://github.com/comit-network/comit-rs/blob/814467ec1a4c2a3319b59eb4566156f194122926/cnd/src/http_api/routes.rs#L55

We are manipulating `Option`s such as `finalized_swap`. It is fine for 200 to be returned if `finalized_swap` is not present for example.
Also, if the swap does not exist at all then 404 should be returned.
If an actual error is encountered, then it should actually be bubbled up, today we just have `None` from the option.

if the swap is in the database then return 200, else, return 404.

## Recommendation

Read from the database to verify the swap exists, assumes there is a database entry for HALight swaps.

thomaseizinger:
Removed `good-first-issue` because this is blocked and hence cannot be picked up.

PR #2595, opened 2020-05-08
Add new db model

tcharding:
In [spike 25](comit-network/spikes:0025-an-extensible-db-design.adoc@master) we designed a new database model. This patch is a first effort at implementing it. Includes code for herc20/halight swaps. Does unit testing of table loads/saves.

Ties in with the han/halight code already present. As such this is all a bit mixed up.  Some stuff is herc20/halight, some stuff is han/halgiht and some stuff is `unimplemented!`

Closes: #2545

### DIFF
--- a/api_tests/src/swap_response.ts
+++ b/api_tests/src/swap_response.ts
@@ -53,23 +53,23 @@ export interface Properties {
  */
 export enum SwapStatus {
     /**
-     * The swap was created but the communication phase hasn't yet been finalized
+     * The swap was created via the REST API but the communication phase hasn't yet been finalized.
      */
     Created = "CREATED",
     /**
-     * The communication was finalized and blockchain actions are needed or happening
+     * The communication was finalized and blockchain actions are needed or happening.
      */
     InProgress = "IN_PROGRESS",
     /**
-     * The swap is finished and the assets were swapped
+     * The swap is finished and the assets were swapped.
      */
     Swapped = "SWAPPED",
     /**
-     * The swap is finished and the assets were not swapped
+     * The swap is finished and the assets were not swapped.
      */
     NotSwapped = "NOT_SWAPPED",
     /**
-     * An unexpected internal failure aborted the swap
+     * An unexpected internal failure aborted the swap.
      */
     InternalFailure = "INTERNAL_FAILURE",
 }
--- a/api_tests/src/actors/actor.ts
+++ b/api_tests/src/actors/actor.ts
@@ -22,7 +22,12 @@ import { LedgerConfig, sleep } from "../utils";
 import { Actors } from "./index";
 import { Wallets } from "../wallets";
 import { defaultLedgerDescriptionForLedger } from "./defaults";
-import { EscrowStatus, LedgerState, SwapResponse } from "../swap_response";
+import {
+    EscrowStatus,
+    LedgerState,
+    SwapResponse,
+    SwapStatus,
+} from "../swap_response";
 
 export type ActorName = "alice" | "bob" | "charlie";
 
@@ -355,7 +360,7 @@ export class Actor {
         while (true) {
             await sleep(200);
             const entity = await this.swap.fetchDetails();
-            if (entity.properties.status === "SWAPPED") {
+            if (entity.properties.status === SwapStatus.Swapped) {
                 break;
             }
         }
--- a/api_tests/tests/communication.ts
+++ b/api_tests/tests/communication.ts
@@ -44,7 +44,42 @@ describe("communication", () => {
             await assertSwapFinalized(bob);
         })
     );
+
+    it(
+        "swap-announced-with-wrong-peer-id-does-not-finalize",
+        twoActorTest(async ({ alice, bob }) => {
+            const bodies = (await SwapFactory.newSwap(alice, bob))
+                .hanEthereumEtherHalightLightningBitcoin;
+
+            // Simulate that Bob is awaiting a swap from a different peer-id than Alice node's peer-id.
+            bodies.bob.peer.peer_id =
+                "QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N";
+            await alice.createSwap(bodies.alice);
+            await bob.createSwap(bodies.bob);
+
+            await assertSwapCreated(alice);
+            await assertSwapCreated(bob);
+
+            await sleep(1000);
+
+            // Assert that the swaps are still not finalized
+            await assertSwapCreated(alice);
+            await assertSwapCreated(bob);
+        })
+    );
 });
+
+/**
+ * Assert that the swap has been created via the REST API bu tthe communication phase has not been finalized.
+ *
+ * No point adding it to `Actor` as it is only use in this test.
+ */
+async function assertSwapCreated(actor: Actor) {
+    await actor.pollCndUntil(
+        actor.swap.self,
+        (swapResponse) => swapResponse.properties.status === SwapStatus.Created
+    );
+}
 
 /**
  * Assert that the communication phase has been finalized.
--- a/api_tests/tests/ether_halight.ts
+++ b/api_tests/tests/ether_halight.ts
@@ -6,8 +6,6 @@
 import SwapFactory from "../src/actors/swap_factory";
 import { sleep } from "../src/utils";
 import { twoActorTest } from "../src/actor_test";
-import { Actor } from "../src/actors/actor";
-import { SwapStatus } from "../src/swap_response";
 
 it(
     "han-ethereum-ether-halight-lightning-bitcoin-alice-redeems-bob-redeems",
@@ -37,57 +35,3 @@ it(
         await bob.assertBalances();
     })
 );
-
-it(
-    "han-ethereum-ether-halight-lightning-bitcoin-alice-announces-with-wrong-peer-id",
-    twoActorTest(async ({ alice, bob }) => {
-        const bodies = (await SwapFactory.newSwap(alice, bob))
-            .hanEthereumEtherHalightLightningBitcoin;
-
-        // Simulate that Bob is awaiting a swap from a different peer-id than Alice node's peer-id.
-        bodies.bob.peer.peer_id =
-            "QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N";
-        await alice.createSwap(bodies.alice);
-        await bob.createSwap(bodies.bob);
-
-        await assertSwapCreated(alice);
-        await assertSwapCreated(bob);
-
-        await sleep(1000);
-
-        // Assert that the swaps are still not finalized
-        await assertSwapCreated(alice);
-        await assertSwapCreated(bob);
-    })
-);
-
-// This test could be anywhere, it is just here because this endpoint was the first of the spilt protocols implemented.
-it(
-    "alice-cannot-create-two-identical-swaps",
-    twoActorTest(async ({ alice, bob }) => {
-        // arrange
-        const bodies = (await SwapFactory.newSwap(alice, bob))
-            .hanEthereumEtherHalightLightningBitcoin;
-
-        await alice.createSwap(bodies.alice);
-        await sleep(200);
-
-        // act
-        const response = alice.createSwap(bodies.alice);
-
-        // assert
-        await expect(response).rejects.toThrow("Swap already exists.");
-    })
-);
-
-/**
- * Assert that the swap has been created via the REST API bu tthe communication phase has not been finalized.
- *
- * No point adding it to `Actor` as it is only use in this test.
- */
-async function assertSwapCreated(actor: Actor) {
-    await actor.pollCndUntil(
-        actor.swap.self,
-        (swapResponse) => swapResponse.properties.status === SwapStatus.Created
-    );
-}
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -353,7 +353,7 @@ impl ComitNode {
         &mut self,
         id: LocalSwapId,
     ) -> Option<HanEtherereumHalightBitcoinCreateSwapParams> {
-        self.comit_ln.get_created_swap(id)
+        self.comit_ln.get_created_swap(&id)
     }
 
     fn supports_halight(&self) -> anyhow::Result<()> {
--- a/cnd/src/network/comit_ln.rs
+++ b/cnd/src/network/comit_ln.rs
@@ -4,7 +4,7 @@ use crate::{
         oneshot_behaviour,
         protocols::{
             announce,
-            announce::{behaviour::Announce, protocol::ReplySubstream, SwapDigest},
+            announce::{behaviour::Announce, protocol::ReplySubstream},
             ethereum_identity, finalize, lightning_identity, secret_hash,
         },
     },
@@ -31,6 +31,12 @@ use std::{
     fmt,
     task::{Context, Poll},
 };
+use swaps::Swaps;
+
+/// Setting it at 5 minutes
+const PENDING_SWAP_EXPIRY_SECS: u32 = 5 * 60;
+
+mod swaps;
 
 /// Event emitted  by the `ComitLn` behaviour.
 #[derive(Debug)]
@@ -54,27 +60,8 @@ pub struct ComitLN {
 
     #[behaviour(ignore)]
     events: VecDeque<BehaviourOutEvent>,
-
-    /// In role of Alice; swaps exist in here once a swap is created by Alice
-    /// (and up until an announce confirmation is received from Bob).
     #[behaviour(ignore)]
-    swaps_waiting_for_confirmation: HashMap<SwapDigest, LocalSwapId>,
-    /// In role of Bob; swaps exist in here if Bob creates the swap _before_ an
-    /// announce message is received from Alice (and up until the announce
-    /// message arrives).
-    #[behaviour(ignore)]
-    swaps_waiting_for_announcement: HashMap<SwapDigest, LocalSwapId>,
-    /// In role of Bob; swaps exist in here if Bob creates the swap _after_ an
-    /// announce message is received from Alice (and up until Bob creates the
-    /// swap).
-    #[behaviour(ignore)]
-    swaps_waiting_for_creation:
-        HashMap<SwapDigest, (libp2p::PeerId, ReplySubstream<NegotiatedSubstream>)>,
-
-    #[behaviour(ignore)]
-    swaps: HashMap<LocalSwapId, HanEtherereumHalightBitcoinCreateSwapParams>,
-    #[behaviour(ignore)]
-    swap_ids: HashMap<LocalSwapId, SharedSwapId>,
+    swaps: Swaps<ReplySubstream<NegotiatedSubstream>>,
     #[behaviour(ignore)]
     ethereum_identities: HashMap<SharedSwapId, identity::Ethereum>,
     #[behaviour(ignore)]
@@ -106,11 +93,7 @@ impl ComitLN {
             lightning_identity: Default::default(),
             finalize: Default::default(),
             events: VecDeque::new(),
-            swaps_waiting_for_confirmation: Default::default(),
-            swaps_waiting_for_announcement: Default::default(),
-            swaps_waiting_for_creation: Default::default(),
             swaps: Default::default(),
-            swap_ids: Default::default(),
             ethereum_identities: Default::default(),
             lightning_identities: Default::default(),
             communication_state: Default::default(),
@@ -126,28 +109,28 @@ impl ComitLN {
     ) -> anyhow::Result<()> {
         let digest = create_swap_params.clone().digest();
         tracing::trace!("Swap creation request received: {}", digest);
-        if self.swaps_waiting_for_announcement.contains_key(&digest)
-            || self.swaps_waiting_for_confirmation.contains_key(&digest)
-        {
-            anyhow::bail!(SwapExists)
-        }
-
-        self.swaps.insert(id, create_swap_params.clone());
 
         match create_swap_params.role {
             Role::Alice => {
                 tracing::info!("Starting announcement for swap: {}", digest);
                 self.announce
-                    .start_announce_protocol(digest.clone(), create_swap_params.peer);
-                self.swaps_waiting_for_confirmation.insert(digest, id);
+                    .start_announce_protocol(digest.clone(), (&create_swap_params.peer).clone());
+                self.swaps
+                    .create_as_pending_confirmation(digest, id, create_swap_params)?;
             }
             Role::Bob => {
-                if let Some((peer, io)) = self.swaps_waiting_for_creation.remove(&digest) {
+                if let Ok((shared_swap_id, peer, io)) = self
+                    .swaps
+                    .move_pending_creation_to_communicate(&digest, id, create_swap_params.clone())
+                {
                     tracing::info!("Confirm & communicate for swap: {}", digest);
-                    self.bob_communicate(peer, io, &id)
+                    self.bob_communicate(peer, io, shared_swap_id, create_swap_params)
                 } else {
-                    self.swaps_waiting_for_announcement
-                        .insert(digest.clone(), id);
+                    self.swaps.create_as_pending_announcement(
+                        digest.clone(),
+                        id,
+                        create_swap_params,
+                    )?;
                     tracing::debug!("Swap {} waiting for announcement", digest);
                 }
             }
@@ -158,25 +141,20 @@ impl ComitLN {
 
     pub fn get_created_swap(
         &self,
-        swap_id: LocalSwapId,
+        swap_id: &LocalSwapId,
     ) -> Option<HanEtherereumHalightBitcoinCreateSwapParams> {
-        self.swaps.get(&swap_id).cloned()
+        self.swaps.get_created_swap(swap_id)
     }
 
     pub fn get_finalized_swap(&self, swap_id: LocalSwapId) -> Option<FinalizedSwap> {
-        let create_swap_params = match self.swaps.get(&swap_id) {
-            Some(body) => body,
+        let (id, create_swap_params) = match self.swaps.get_announced_swap(&swap_id) {
+            Some(swap) => swap,
             None => return None,
         };
 
         let secret = match create_swap_params.role {
             Role::Alice => Some(self.seed.derive_swap_seed(swap_id).derive_secret()),
             Role::Bob => None,
-        };
-
-        let id = match self.swap_ids.get(&swap_id).copied() {
-            Some(id) => id,
-            None => return None,
         };
 
         let alpha_ledger_redeem_identity = match create_swap_params.role {
@@ -235,6 +213,7 @@ impl ComitLN {
         peer: PeerId,
         swap_id: SharedSwapId,
         local_swap_id: LocalSwapId,
+        create_swap_params: HanEtherereumHalightBitcoinCreateSwapParams,
     ) {
         let addresses = self.announce.addresses_of_peer(&peer);
         self.secret_hash
@@ -244,8 +223,6 @@ impl ComitLN {
         self.lightning_identity
             .register_addresses(peer.clone(), addresses.clone());
         self.finalize.register_addresses(peer.clone(), addresses);
-
-        let create_swap_params = self.swaps.get(&local_swap_id).unwrap();
 
         self.ethereum_identity.send(
             peer.clone(),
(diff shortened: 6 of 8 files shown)


cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
```diff
@@ -0,0 +1,64 @@
+-- Your SQL goes here
+
+CREATE TABLE swaps
```
thomaseizinger, line 3: Should be single according to the `drop` migration.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+    id INTEGER			NOT NULL PRIMARY KEY,
+    local_swap_id UNIQUE	NOT NULL,
+    role			NOT NULL,
+    counterparty_peer_id	NOT NULL
+);
+
+CREATE TABLE secret_hashes
+(
+    id INTEGER			NOT NULL PRIMARY KEY,
+    swap_id INTEGER		NOT NULL,
```
thomaseizinger, line 14: Shouldn't need this if we have the foreign key.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+    secret_hash		NOT NULL,
+    FOREIGN KEY (swap_id)	REFERENCES swaps(id)
+);
+
+CREATE TABLE shared_swap_ids
+(
+    id INTEGER			NOT NULL PRIMARY KEY,
+    swap_id INTEGER		NOT NULL,
```
thomaseizinger, line 22: Shouldn't need this if we have the foreign key.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+
+CREATE TABLE address_hints
+(
+    id INTEGER			NOT NULL PRIMARY KEY,
+    swap_id INTEGER		NOT NULL,
+    address_hint		NOT NULL,
+    FOREIGN KEY(swap_id)	REFERENCES swaps(id)
+);
+
+CREATE TABLE herc20_swap_details
```
thomaseizinger, line 35: Not a fan of the `swap_detail` terminology :)

```suggestion
CREATE TABLE herc20_protocol
```

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+(
+    id INTEGER			NOT NULL PRIMARY KEY,
+    swap_id INTEGER		NOT NULL,
```
thomaseizinger, line 38: Shouldn't need this if we have the foreign key.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+    amount			NOT NULL,
+    chain_id			NOT NULL,
+    expiry			NOT NULL,
+    hash_function		NOT NULL,
```
thomaseizinger, line 42: Don't think this should be configurable, would be a different protocol then.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+    token_contract		NOT NULL,
+    redeem_identity,
+    refund_identity,
+    ledger			NOT NULL CHECK (ledger IN ('alpha', 'beta')),
```
thomaseizinger, line 46: Should we rename this to `side` to avoid confusion with what we usually call ledger?

```suggestion
    side			NOT NULL CHECK (side IN ('alpha', 'beta')),
```

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+    FOREIGN KEY(swap_id)	REFERENCES swaps(id)
+);
+
+CREATE TABLE halight_swap_details
+(
+    id INTEGER			NOT NULL PRIMARY KEY,
+    swap_id INTEGER		NOT NULL,
```
thomaseizinger, line 53: Shouldn't need this if we have the foreign key.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+
+CREATE TABLE shared_swap_ids
+(
+    id INTEGER			NOT NULL PRIMARY KEY,
+    swap_id INTEGER		NOT NULL,
+    shared_swap_id		NOT NULL,
+    FOREIGN KEY (swap_id)	REFERENCES swaps(id)
+);
+
+CREATE TABLE address_hints
```
thomaseizinger, line 27: We could generalize this table to an `address_book` where we store addresses or peer IDs.

It is really only a "hint" for the initial request.
Anytime we receive a request, we could potentially load all addresses for this peer from this table and try all of them.

For now, we could just name it `address_book` but not anything besides saving entries.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
```diff
@@ -0,0 +1,64 @@
...
+(
+    id INTEGER			NOT NULL PRIMARY KEY,
+    swap_id INTEGER		NOT NULL,
+    address_hint		NOT NULL,
+    FOREIGN KEY(swap_id)	REFERENCES swaps(id)
```
thomaseizinger, line 32: I would remove the reference to swap and store the Peer ID instead.

cnd/src/db/load.rs
```diff
@@ -0,0 +1,33 @@
...
+    async fn load(&self, swap_id: LocalSwapId) -> anyhow::Result<Option<T>>;
+}
+
+#[async_trait]
+impl Load<LiveSwap<han::LiveSwap, halight::LiveSwap>> for Sqlite {
+    async fn load(
+        &self,
+        _: LocalSwapId,
+    ) -> anyhow::Result<Option<LiveSwap<han::LiveSwap, halight::LiveSwap>>> {
+        // NOTE: When implementing this be sure to return Ok(None) if the
```
thomaseizinger, line 19: Instead of returning an `Option` here, I would suggest the following:

- None if there is no swap with this ID
- Some if there is a swap.

The only information we learn during the communication phase is secret hash and identities, both are **not** returned from the API.
This struct can also have an enum that tells you the communication state based on what data is present in the database.
The point is, it should always return something if we have data, even if the communication is not yet finished. We are going to need this for https://github.com/comit-network/comit-rs/issues/2399

For the usecase of re-spawning watchers, where we need the additional information like secret-hash and identities, we probably want a completely different API: One that returns as all protocol watchers that need to be respawned (as a single list and not as pair of protocols).

cnd/src/db/save.rs
```diff
@@ -516,3 +518,84 @@ impl Save<CreatedSwap<han::CreatedSwap, halight::CreatedSwap>> for Sqlite {
...
+        let herc = IHerc20SwapDetail {
+            swap_id: 0, // FK, set during save.
+            amount: Text(created.alpha.amount.into()),
+            chain_id: U32(created.alpha.chain_id),
+            expiry: U32(created.alpha.absolute_expiry),
+            hash_function: Text(HashFunction::Sha256),
+            token_contract: Text(created.alpha.contract_address.into()),
+            redeem_identity,
+            refund_identity,
+            ledger: "alpha".to_string(),
```
thomaseizinger, line 558: Should be an enum I reckon :)

api_tests/src/swap_response.ts
```diff
@@ -59,7 +59,7 @@ export enum SwapStatus {
     /**
      * The communication was finalized and blockchain actions are needed or happening
      */
-    InProgress = "IN_PROGRESS",
+    InProgress = "IN_PROGRESS", // FIXME: Rename this to 'live' ... everywhere.
```
da-kami, line 62: I am not sure this is necessary. 
I am actually not sure `live` is less ambiguous than `in progress`.

cnd/src/db/save.rs
```diff
@@ -516,3 +518,84 @@ impl Save<CreatedSwap<han::CreatedSwap, halight::CreatedSwap>> for Sqlite {
...
+    async fn save(
+        &self,
+        created: CreatedSwap<herc20::CreatedSwap, halight::CreatedSwap>,
+    ) -> anyhow::Result<()> {
+        let local_swap_id = created.swap_id;
+        let role = created.role;
+
+        // Save the swap into the swaps table.
+        let swap = ISwap::new(created.swap_id, created.peer, created.role);
+        self.save_swap(&swap).await?;
```
thomaseizinger, line 533: I think all these things should happen within one DB transaction so they are atomic.

cnd/src/db/schema.rs
```diff
@@ -1,106 +1,66 @@
+// LocalSwapId and SharedSwapId are encoded as Text and named local_swap_id, and
+// shared_swap_id respectively.  swap_id (Integer) is always a foreign key link
+// to the `swaps` table.
 table! {
-   rfc003_bitcoin_ethereum_bitcoin_ether_request_messages {
```
thomaseizinger, line 2: These should not be deleted (yet) I guess :)

cnd/src/db/tables.rs
```diff
@@ -0,0 +1,733 @@
...
+    lightning,
+    swap_protocols::{self, rfc003, HashFunction, LocalSwapId, Role},
+};
+use diesel::{self, prelude::*, RunQueryDsl};
+use libp2p::{Multiaddr, PeerId};
+
+// TODO: Audit 'pub' scope on all structs.
+
+#[derive(Identifiable, Queryable, PartialEq, Debug)]
+#[table_name = "swaps"]
```
thomaseizinger, line 22: Sometimes singular, sometimes plural. I think the spike defined singular :)

cnd/src/db/save.rs
```diff
@@ -516,3 +518,84 @@ impl Save<CreatedSwap<han::CreatedSwap, halight::CreatedSwap>> for Sqlite {
...
+            Role::Alice => None,
+            Role::Bob => Some(Text(created.beta.identity.into())),
+        };
+        let network = BitcoinNetwork::from_str(&created.beta.network)?;
+        let halight = IHalightSwapDetail {
+            swap_id: 0, // FK, set during save.
+            amount: Text(created.beta.amount.into()),
+            network: Text(network),
+            chain: "bitcoin".to_string(), // FIXME: What should this be?
+            expiry: U32(0),               // FIXME: What should this be?
```
da-kami, line 578: I think we can remove this. We added `expiry` and `cltv_expiry` because they co-existed at the time when we wrote the spike. 
`expiry` was since removed as far as I could see from the code, so I'd remove this unless @thomaseizinger @bonomat @D4nte disagree.

(same for the `chain` parameter...?)

cnd/src/db/tables.rs
```diff
@@ -0,0 +1,733 @@
...
+            local_swap_id: swap.local_swap_id,
+            role: swap.role,
+            counterparty_peer_id: swap.counterparty_peer_id,
+        }
+    }
+}
+
+#[derive(Insertable, Debug, Clone)]
+#[table_name = "swaps"]
+pub struct ISwap {
```
thomaseizinger, line 42: I know it is longer but I was confused about the `I` because it means `Interface` in other languages.
```suggestion
pub struct InsertableSwap {
```

cnd/src/db/tables.rs
```diff
@@ -0,0 +1,733 @@
...
+}
+
+#[derive(Insertable, Debug, Clone, Copy)]
+#[table_name = "secret_hashes"]
+pub struct ISecretHash {
+    swap_id: i32,
+    secret_hash: Text<rfc003::SecretHash>,
+}
+
+// At the moment the address hint is just associated with the swap.
```
thomaseizinger, line 74: See my comment on the table definition. I think it should be related to the peer id.

cnd/src/db/tables.rs
```diff
@@ -0,0 +1,733 @@
...
+        address_hint: libp2p::Multiaddr,
+    ) -> anyhow::Result<()> {
+        self.do_in_transaction(|connection| {
+            let key = Text(swap_id);
+
+            let swap: Swap = swaps::table
+                .filter(swaps::local_swap_id.eq(key))
+                .first(connection)?;
+
+            // Needed because Multiaddr does not implement Copy
```
thomaseizinger, line 345: Clone it then? :)

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
da-kami: That is a valid solution, there would not be a relation to the swap then, though (we would just imply the relationship through the peer-id). 
I actually tried to model a relationship from `swap` to `address_hint` on the `counterparty_peer_id` but diesel did not let me do that.

If we are OK with just having a separate table `address_book` where we save `peer_id` `address_hint` relationships we can do so. 
We can also relate entries in there to a swap on the `swap_id`, but from an `address_book` point of view this relation does not make sense.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
da-kami: I was unable to create a proper FK relationship betweek `swap.counterparty_peer_id` and `address_hint.counterparty_peer_id` - if we don't need this relationship that is fine. But that would mean, that we would always have to try to dial all `address_hints` a `peer_id` yields in this table (that could be named `address_book`)

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
da-kami: For simplicity reasons I would actually keep it as currently modelled. Or is the `address_book` relevant for recovery?

da-kami:
> Looks good!
> 
> Made a couple of comments in regards to the schema!
> My biggest criticism would probably be that I think we should be doing all the inserts in one transaction and not in several.
> 
> Unfortunately, SQLite doesn't support `RETURNING` but this should help: [docs.rs/diesel/1.4.4/diesel/query_dsl/trait.RunQueryDsl.html#method.get_result](https://docs.rs/diesel/1.4.4/diesel/query_dsl/trait.RunQueryDsl.html#method.get_result)
> 
> Also, see this blogpost here: [diesel.rs/guides/all-about-inserts](https://diesel.rs/guides/all-about-inserts/)

After trying this and doing some research: `get_result` just adds `RETURNING` to the statement which is not supported for `SQLite`.

So, I think this is just not possible for sqlite.
Or did I misunderstand you @thomaseizinger ?

cnd/src/db/tables.rs
da-kami: that did not work, but I can try again.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
bonomat, line 32: I think this is a weird relation now. 
As discussed shortly on the call, I would either turn around the relationship or move the `address_hint` inline into `swaps` and make it nullable.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
thomaseizinger: I don't think there need to be any relations. We are just storing addresses of peers :)

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
thomaseizinger: As mentioned above, the address hint is completely unrelated to swaps!

It is just part of the request body so that the client doesn't have to do two requests when creating a swap!

There should be no relation what so ever with a swap IMO.

thomaseizinger:
> > Unfortunately, SQLite doesn't support `RETURNING` but this should help: [docs.rs/diesel/1.4.4/diesel/query_dsl/trait.RunQueryDsl.html#method.get_result](https://docs.rs/diesel/1.4.4/diesel/query_dsl/trait.RunQueryDsl.html#method.get_result)
> > Also, see this blogpost here: [diesel.rs/guides/all-about-inserts](https://diesel.rs/guides/all-about-inserts/)
> 
> After trying this and doing some research: `get_result` just adds `RETURNING *` to the statement which is not supported for `SQLite`.
> 
> So, I think this is just not possible for sqlite.
> It yields:
> 
> ```
> the trait `diesel::backend::SupportsReturningClause` is not implemented for `diesel::sqlite::Sqlite`
> ```
> 
> Or did I misunderstand you @thomaseizinger ?

Does this help?
https://github.com/diesel-rs/diesel/blob/master/examples/sqlite/all_about_inserts/src/lib.rs#L278

cnd/src/db/tables.rs
thomaseizinger: Where does this bound come from? Should be in the error message somewhere.

api_tests/src/swap_response.ts
tcharding: Good point about the states not being clearly defined, we should definitely define these clearly and have uniform terminology in the code base.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
tcharding: Oh, thanks.  The `DROP TABLE foo` statements are all wrong.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
tcharding: This _is_ the foreign key

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
tcharding: This should have been argued on the spike.  Happy to debate/improve it at the 99% review though :)

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
tcharding: Same goes for all comments below.  We use the `swap_id` in each table as the foreign key.  cnd swap ids are referred to as `local_swap_id` and `shared_swap_id` throughout the schema.

tcharding: This is inline with how diesel expects things to be named, the type (INTEGER) is also in line with this usage.

cnd/src/db/tables.rs
tcharding: @da-kami and I debated this, diesel works by default with plural so we went for plural.  (The drop statements where a left over of this debate that I missed.)

cnd/src/db/tables.rs
tcharding: I knew review would say this :(  I disagree but will change it at 99% review if requested.  For now I'm going to leave it like this because its me typing `Insertable` 400 times a day. Its a few simple perl one-liners to change later.

cnd/src/db/tables.rs
tcharding: All this database code is going to need iterating on, at the moment I'm optimising for the simplest thing that works. FK'ing on swap_id as is done for all the other tables makes the code the same so makes it easier to write quickly. Can we leave this to be improved upon later?

cnd/src/db/tables.rs
tcharding: Used 
```
            let insertable = IAddressHint {
                swap_id: swap.id,
                address_hint: Text(address_hint.clone()),
            };
```

api_tests/src/swap_response.ts
tcharding: According to the comments in the TS code this FIXME is correct
```
    /**
     * The swap was created but the communication phase hasn't yet been finalized
     */
    Created = "CREATED",
    /**
     * The communication was finalized and blockchain actions are needed or happening
     */
    InProgress = "IN_PROGRESS", // FIXME: Rename this to 'live' ... everywhere.
```

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
thomaseizinger: Ah okay, my mistake, I thought this was the UUID :)

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
thomaseizinger: > This should have been argued on the spike. Happy to debate/improve it at the 99% review though :)

I disagree, these kind of comments/discussions would have just delayed the spike from being merged.
Spikes shouldn't be exact instructions but just solve the hard problems.

As you said, names can always be changed so I've kept my feedback on this until we actually implement it.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
thomaseizinger: > But that would mean, that we would always have to try to dial all address_hints a peer_id yields in this table (that could be named address_book)

Yes that is why I would do.

cnd/src/db/tables.rs
thomaseizinger: Really, diesel cares about this? Wow :D

cnd/src/db/tables.rs
thomaseizinger: Can you please set up auto-completion in your editor so you don't have to type these things manually and we can stop these discussions?

For the record, if I type `ISwap` and hit tab, it will autocomplete all names that are `IwhatevertextishereSwap` because it recognizes capitals. Surely there is something for emacs that can do that.

cnd/src/db/tables.rs
thomaseizinger: The simplest thing IMO is not associating the address_hint with anything at all but just storing key-value pairs of peer-id to address, no?

api_tests/src/swap_response.ts
tcharding: Re-verted to using `InProgress` instead of `Live`.  Will add a ticket to rename this.

cnd/src/db/schema.rs
tcharding: They are moved to `rfc003_schema.rs`.

cnd/src/db/save.rs
bonomat: My understanding was that we need both: 
* `expiry` is for the invoice
* `cltv_expiry` is the actual payment

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
bonomat: Agreed, depending on what you store in `swaps` there should be either no relation or a  fk-relation from `swaps.peer_id` -> `address_hints.id`.

cnd/src/db/tables.rs
tcharding: Yes, that opens us up to clobbering address hints but that is probably not an issue.  Will update with this suggestion.

cnd/src/db/tables.rs
tcharding: I have auto-completion set up, I still don't like it :)

cnd/src/db/tables.rs
tcharding: I wouldn't go so far as saying it 'cares', there are ways to make it work but it is 'opinionated' and naming it like this makes the default marco behaviours work.

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
tcharding: Cool, wish I realized that before starting; I would have changed a bunch of names :) Will keep this in mind for next time I code off of a spike.

tcharding: How about just plain `herc20`, since this is the main table for her20 info does adding anything extra to the name add any meaning?

D4nte:
Removing myself from review as there are already 3 reviewers who made comments.

cnd/src/db/tables.rs
thomaseizinger: Good enough reason for me to go with diesel's opinion then, convention over configuration :)

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
thomaseizinger: Fine with me :)

api_tests/src/swap_response.ts
D4nte: Please note that IMHO, the state are clearly defined here.

I suggest that we indeed keep this up to date to harmonize the state terminology across the codebase and avoid using 3 different words to express the same state (live, in progress, finalized).

cnd/src/db/tables.rs
tcharding: FTR I tried this, we get naming clashes with rfc003 `Swap` and `InsertableSwap` - I'm going to leave it as `ISwap` until the rfc003 stuff is gone out of the db module.

cnd/src/db/tables.rs
tcharding: This is implemented now.

tcharding:
> My biggest criticism would probably be that I think we should be doing all the inserts in one transaction and not in several.

@thomaseizinger, I agree.  The inserts are not atomic in this PR.  However because this ticket is blocking the whole sprint I propose to do the simplest / braindead save/loads for now. Incremental improvements can then be done easily since they are confined to the db module and the sprint can go on.

thomaseizinger:
> > My biggest criticism would probably be that I think we should be doing all the inserts in one transaction and not in several.
> 
> @thomaseizinger, I agree. The inserts are not atomic in this PR. However because this ticket is blocking the whole sprint I propose to do the simplest / braindead save/loads for now. Incremental improvements can then be done easily since they are confined to the db module and the sprint can go on.

I am fine with this :)

thomaseizinger:
@tcharding I just found this: https://sqlite.org/foreignkeys.html#fk_enable

Can we have a test that ensures we are enabling foreign-key support?

tcharding:
> @tcharding I just found this: https://sqlite.org/foreignkeys.html#fk_enable
> 
> Can we have a test that ensures we are enabling foreign-key support?

Nice catch, had read that and forgotten about it.

cnd/src/swap_protocols/halight.rs
```diff
@@ -34,6 +36,16 @@ pub struct CreatedSwap {
...
 }
 
+/// Halight specific data for an in progress swap.
+#[derive(Debug, Clone, Copy)]
+pub struct InProgressSwap {
+    pub ledger: lightning::Regtest,
+    pub asset: asset::Bitcoin,
+    pub refund_identity: identity::Lightning,
+    pub redeem_identity: identity::Lightning,
+    pub expiry: Timestamp,
```
da-kami, line 46: Given the [comment that there should be 2 expiries (`cltv_exiry` and `expiry`)](https://github.com/comit-network/comit-rs/pull/2595/files#r422750934) I suppose we should add the `cltv_expiry` here.
If we only supply one expiry on the API level I would opt for setting this to the same expiry internally.

da-kami:
@tcharding I don't have anything else to add. Let's cleanup and get this in and create a few small follow-up tickets for improvements.
Feel free to ping me if you want help with e.g. creating tickets.

cnd/src/http_api/routes.rs
```diff
@@ -33,88 +34,105 @@ pub fn into_rejection(problem: HttpApiProblem) -> Rejection {
...
+        };
+
+    let swap = Load::<InProgressSwap<han::InProgressSwap, halight::InProgressSwap>>::load(
+        &facade, swap_id,
+    )
+    .await?;
+    match swap {
+        None => make_created_swap_entity(swap_id, HanEthereumHalightBitcoinCreatedState {
+            created_swap: created_swap.into(),
+        }),
```
thomaseizinger, line 99: Why do we have to go to the database twice here? Could we not just have an API that returns us exactly what we need (taking care of the difference between `InProgressSwap` and `CreatedSwap` inside of the datastructure by having some things be `Option`?

This looks to me like we are unnecessarily forcing this part of the codebase to work with the separation of `Created` and `InProgress` swaps and hence yielding sub-optimal code because we have to go to the database twice.

We don't need to find a model that works for every usecase in the codebase, we can have different models for different purposes :)
At this stage, we probably also want to re-consider, which fields the API actually needs and not load those (not for performance reasons, but to make the different needs more obvious and to avoid someone refactoring away "duplication" when there is none)

cnd/migrations/2020-05-06-105000_created-swap-tables/up.sql
bonomat: I thought I just responded? 🧐 Mhmm... message gone...

Nothing in particular. This property just was there somehow there. Feel free to remove it/don't add it. It does not add much value as it is always static.

thomaseizinger: Well, you edited my comment :joy: 

image

Thanks for clarifying :)

bonomat: haha... ups 😅

tcharding:
> @tcharding I just found this: [sqlite.org/foreignkeys.html#fk_enable](https://sqlite.org/foreignkeys.html#fk_enable)
> 
> Can we have a test that ensures we are enabling foreign-key support?

Added a new ticket for this.

tcharding:
I do not know what just happened but the commit off this PR disappeared?

PR #2601, opened 2020-05-11
Add context to ledger connector errors 

### LINKED ISSUE #2368: Trace network messages to blockchain nodes
## Problem

When getting a connection error from a blockchain (Ethereum, Bitcoin and Lightning) node, it is hard to know what is the issue as only the URL of the node is logged.

## Goal

Facilitate debugging by providing more information such as:
- headers
- http method
- body content

## Recommendation

Make good use of tracing spans to assist with context information.

Print information listed above for requests to Ethereum and Bitcoin node when logging is at trace level.

luckysori:
Fixes #2368.

Instead of using `tracing` like the original ticket suggests, we use `anyhow::Context`. It should be easier for a developer to get all the error information from the error stack, rather than having to cross-reference with cnd's logs.

Feel free to be nitpicky in terms of error messages, unnecessary errors, missing errors, etc.

### DIFF
--- a/cnd/src/btsieve/bitcoin/bitcoind_connector.rs
+++ b/cnd/src/btsieve/bitcoin/bitcoind_connector.rs
@@ -1,9 +1,11 @@
 use crate::{
-    btsieve::{bitcoin::bitcoin_http_request_for_hex_encoded_object, BlockByHash, LatestBlock},
+    btsieve::{BlockByHash, LatestBlock},
     config::validation::FetchNetworkId,
 };
+use anyhow::Context;
 use async_trait::async_trait;
-use bitcoin::{BlockHash, Network};
+use bitcoin::{consensus::deserialize, BlockHash, Network};
+use futures::TryFutureExt;
 use reqwest::{Client, Url};
 use serde::{de, export::fmt, Deserialize, Deserializer};
 
@@ -35,6 +37,23 @@ impl BitcoindConnector {
             .join(&format!("{}.hex", block_hash))
             .expect("building url should work")
     }
+
+    async fn chain_info(&self) -> anyhow::Result<ChainInfo> {
+        let url = &self.chaininfo_url;
+        let chain_info = self
+            .client
+            .get(url.clone())
+            .send()
+            .await
+            .with_context(|| GetRequestFailed(url.clone()))?
+            .json::<ChainInfo>()
+            .await
+            .context("failed to deserialize JSON response as chaininfo")?;
+
+        tracing::debug!("Fetched chain info: {:?} from bitcoind", chain_info);
+
+        Ok(chain_info)
+    }
 }
 
 #[async_trait]
@@ -42,16 +61,7 @@ impl LatestBlock for BitcoindConnector {
     type Block = bitcoin::Block;
 
     async fn latest_block(&self) -> anyhow::Result<Self::Block> {
-        let chaininfo_url = self.chaininfo_url.clone();
-
-        let chain_info = self
-            .client
-            .get(chaininfo_url)
-            .send()
-            .await?
-            .json::<ChainInfo>()
-            .await?;
-
+        let chain_info = self.chain_info().await?;
         let block = self.block_by_hash(chain_info.bestblockhash).await?;
 
         Ok(block)
@@ -65,8 +75,15 @@ impl BlockByHash for BitcoindConnector {
 
     async fn block_by_hash(&self, block_hash: Self::BlockHash) -> anyhow::Result<Self::Block> {
         let url = self.raw_block_by_hash_url(&block_hash);
-        let block =
-            bitcoin_http_request_for_hex_encoded_object::<Self::Block>(url, &self.client).await?;
+        let block = self
+            .client
+            .get(url.clone())
+            .send()
+            .await
+            .with_context(|| GetRequestFailed(url))?
+            .text()
+            .map_ok(decode_response)
+            .await??;
 
         tracing::debug!(
             "Fetched block {} with {} transactions from bitcoind",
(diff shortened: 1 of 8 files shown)


cnd/src/jsonrpc.rs
```diff
@@ -67,9 +64,20 @@ impl<T> Request<T> {
...
+#[derive(Debug, serde::Deserialize, thiserror::Error)]
+#[error("JSON-RPC request failed with code {code}: {message}")]
+pub struct JsonRpcError {
+    code: i64,
+    message: String,
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("connection error: {0}")]
+pub struct ConnectionFailed(#[from] reqwest::Error);
```
thomaseizinger, line 79: :mega: I don't think this is used anywhere now.

It used to be used implicitly through the `?` on line 30 but given that you are now using anyhow, the conversion to this no longer happens.
You can use `map_err` to restore this behaviour.

cnd/src/jsonrpc.rs
```diff
@@ -23,28 +16,32 @@ impl Client {
...
             .json::<Response<Res>>()
-            .await?;
+            .await
+            .context("failed to deserialize JSON response as JSON-RPC response")?;
 
         match response {
             Response::Success { result } => Ok(result),
-            Response::Error { code, message } => Err(Error::JsonRpc { code, message }),
+            Response::Error(error) => {
+                Err(error).with_context(|| format!("JSON-RPC request {:?} failed", request))
```
thomaseizinger, line 38: Is this `with_context` necessary? I think you are just duplicating the message that is on the `JsonRpcError` struct.

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -158,13 +158,14 @@ impl LndConnectorAsSender {
         &self,
         secret_hash: SecretHash,
         status: PaymentStatus,
-    ) -> Result<Option<Payment>, Error> {
+    ) -> anyhow::Result<Option<Payment>> {
         let response = client(&self.certificate, &self.macaroon)?
             .get(self.payment_url())
             .send()
             .await?
```
thomaseizinger, line 165: Do we want to add context here as-well?

cnd/src/swap_protocols/halight/connector.rs
```diff
@@ -305,7 +306,7 @@ impl LndConnectorAsReceiver {
                 // yes we can fail while we already encoundered an error ...
                 .with_context(|| format!("encountered {} while fetching invoice but couldn't deserialize error response 🙄", status_code))?;
```
thomaseizinger, line 307: Love it that my emoji is still in there :joy:

cnd/src/trace.rs
```diff
@@ -13,6 +13,7 @@ pub fn init_tracing(level: log::LevelFilter) -> anyhow::Result<()> {
 
     let subscriber = FmtSubscriber::builder()
         .with_max_level(level_from_level_filter(level))
+        .with_ansi(false)
```
thomaseizinger, line 16: Would prefer if we set this according to the output of `atty`: https://github.com/softprops/atty

cnd/src/btsieve/bitcoin/bitcoind_connector.rs
```diff
@@ -35,23 +37,35 @@ impl BitcoindConnector {
...
     }
+
+    async fn chain_info(&self) -> anyhow::Result<ChainInfo> {
+        let url = &self.chaininfo_url;
+        let chain_info = self
+            .client
+            .get(url.clone())
+            .send()
+            .await
+            .context(GetRequestFailed(url.clone()))?
```
thomaseizinger, line 48: We might want to use `with_context` here to avoid the allocation in case we don't actually fail.

cnd/src/jsonrpc.rs
luckysori: In fact the dry tests now fail because of this!

I tried, but I could not see the original implicit conversion :(

thomaseizinger: > I tried, but I could not see the original implicit conversion :(

Every `?` contains a `ReturnTypeError::from(inner_error)`.
This always works because of the blanket impl of `From<T>` for every `T` that just returns `self`.

luckysori: I'm still confused about one thing. Does `client.send` actually only fail with a connection error? The error enum used by `reqwest` seems to also include HTTP errors. Is this just another case of error enums being bad?

cnd/src/jsonrpc.rs
luckysori: My intention is to give the context of which _request_ resulted in a JSON-RPC _response_ error. Request contains method and params, whereas the response error only has code and message.

cnd/src/swap_protocols/halight/connector.rs
luckysori: My assumption here was that the payment URL will be part of the error returned by lnd. Do you think that's incorrect or would you add something else?

cnd/src/trace.rs
luckysori: Cool idea! Will do.

cnd/src/btsieve/bitcoin/bitcoind_connector.rs
luckysori: Yes!

cnd/src/jsonrpc.rs
thomaseizinger: > Is this just another case of error enums being bad?

I think so yes. "Bad" in the sense that they use just one error enum for all public APIs.

cnd/src/jsonrpc.rs
thomaseizinger: Aha! That makes sense :)

cnd/src/swap_protocols/halight/connector.rs
thomaseizinger: I don't think it is going to be part of it :)

PR #2602, opened 2020-05-11
Do not consume `self` on `digest`

D4nte:
Saves having a clone every time `digest` is called.

### DIFF
--- a/cnd/src/network/protocols/announce.rs
+++ b/cnd/src/network/protocols/announce.rs
@@ -2,7 +2,7 @@ pub mod behaviour;
 pub mod handler;
 pub mod protocol;
 
-use digest::IntoDigestInput;
+use digest::ToDigestInput;
 use libp2p::multihash::{self, Multihash};
 use serde::{de::Error, Deserialize, Deserializer, Serialize, Serializer};
 use std::fmt;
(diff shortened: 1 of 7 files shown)


cnd/src/network/protocols/announce.rs
```diff
@@ -17,8 +17,8 @@ impl SwapDigest {
 }
 
 impl IntoDigestInput for SwapDigest {
-    fn into_digest_input(self) -> Vec<u8> {
-        self.0.into_bytes()
+    fn into_digest_input(&self) -> Vec<u8> {
```
thomaseizinger, line 20: Should be called `to_digest_input` according to Rust conventions :)

digest/src/lib.rs
```diff
@@ -67,25 +67,25 @@ pub use digest_macro_derive::Digest;
 pub use hex;
 
 pub trait Digest {
-    type Hash: Hash + IntoDigestInput;
-    fn digest(self) -> Self::Hash;
+    type Hash: Hash + ToDigestInput;
```
thomaseizinger, line 70: I actually didn't notice before, why are we requiring this trait bound here?

D4nte: Can you read the doc in the `lib.rs` and let me know if the question still holds?

thomaseizinger: Makes sense now, thanks :)
I think I was mixing it up with the `Digest` trait somehow.

PR #2618, opened 2020-05-12
Lnd expiries

### LINKED ISSUE #2400: LND expiries are not fully thought through
## Problem

During the LN squad hacking we did not fully think through the expiries (both invoice and cltv) used by LND.  For example we hardcode 3600 for one of them and use the COMIT beta ledger expiry for another.

## Goal

Align lnd invoice expiry with what cnd does with other expiries.

Note: The invoice expiry (3600) is limited by lnd.
Note: Do not redo the whole mathematical model, fit the lnd expiries with the current default 12/24hrs model.

D4nte:
See 1a002196.
Resolves #2400.

### DIFF
--- a/cnd/src/http_api/routes.rs
+++ b/cnd/src/http_api/routes.rs
@@ -20,6 +20,7 @@ use crate::{
         DeployAction, Facade, FundAction, Herc20HalightBitcoinCreateSwapParams, InitAction,
         LocalSwapId, RedeemAction, RefundAction, Role,
     },
+    timestamp::RelativeTime,
     transaction,
 };
 use blockchain_contracts::ethereum::rfc003::{ether_htlc::EtherHtlc, Erc20Htlc};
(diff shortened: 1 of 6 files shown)


cnd/src/http_api/routes.rs
```diff
@@ -28,6 +28,8 @@ use serde::Serialize;
 use std::collections::HashMap;
 use warp::{http, http::StatusCode, Rejection, Reply};
 
+const INVOICE_EXPIRY_SECS: u32 = 3600;
```
thomaseizinger, line 31: If you make the constructor if `RelativeTime` a `const fn`, then you can initialize it directly as `RelativeTime` instead of `u32`.

cnd/src/timestamp.rs
```diff
@@ -3,7 +3,7 @@ use std::time::SystemTime;
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Deserialize, Serialize)]
 #[serde(transparent)]
-pub struct Timestamp(u32);
+pub struct Timestamp(u32); // seconds
```
thomaseizinger, line 6: Would prefer if we would add proper rustdocs instead of these inline comments.

cnd/src/http_api/routes.rs
D4nte: TIL I learn that we should have more `const fn` constructors.

thomaseizinger: It is a mild guess but I think you should be able to `const fn` the ctor :)

D4nte:
bors retry

PR #2619, opened 2020-05-12
Remove dead code

bonomat:
I think I found some dead code :)


thomaseizinger:
Did you run out of admin tasks @bonomat? :grimacing:

PR #2620, opened 2020-05-12
Switch ether to erc20 for halight swaps

da-kami:
In an attempt to get cnd into a state where we can bring the current db and the lightning swaps together this PR changes `HanEthereumEther-HalightLightningBitcoin` swaps to `Herc20EthereumErc20-HalightLightningBitcoin` swaps.

Note that I did not act upon the new `herc20` protocol code yet, but used the rfc003 watcher for monitoring Ethereum. 

Would get this merged asap and then finally connect use the new DB 😀

### DIFF
--- a/cnd/src/swap_protocols/facade.rs
+++ b/cnd/src/swap_protocols/facade.rs
@@ -13,7 +13,7 @@ use std::sync::Arc;
 /// before communication with the other node has started
 #[derive(Clone, Digest, Debug, PartialEq)]
 #[digest(hash = "SwapDigest")]
-pub struct HanEtherereumHalightBitcoinCreateSwapParams {
+pub struct Herc20HalightBitcoinCreateSwapParams {
     #[digest(ignore)]
     pub role: Role,
     #[digest(ignore)]
@@ -23,7 +23,9 @@ pub struct HanEtherereumHalightBitcoinCreateSwapParams {
     #[digest(prefix = "2001")]
     pub ethereum_absolute_expiry: Timestamp,
     #[digest(prefix = "2002")]
-    pub ethereum_amount: asset::Ether,
+    pub ethereum_amount: asset::Erc20Quantity,
+    #[digest(ignore)]
+    pub token_contract: EthereumIdentity,
     #[digest(ignore)]
     pub lightning_identity: identity::Lightning,
     #[digest(prefix = "3001")]
@@ -39,6 +41,12 @@ impl ToDigestInput for asset::Bitcoin {
 }
 
 impl ToDigestInput for asset::Ether {
+    fn to_digest_input(&self) -> Vec<u8> {
+        self.to_bytes()
+    }
+}
+
+impl ToDigestInput for asset::Erc20Quantity {
     fn to_digest_input(&self) -> Vec<u8> {
         self.to_bytes()
     }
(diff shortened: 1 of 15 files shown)


cnd/src/swap_protocols/facade.rs
```diff
@@ -23,7 +23,9 @@ pub struct HanEtherereumHalightBitcoinCreateSwapParams {
     #[digest(prefix = "2001")]
     pub ethereum_absolute_expiry: Timestamp,
     #[digest(prefix = "2002")]
-    pub ethereum_amount: asset::Ether,
+    pub ethereum_amount: asset::Erc20Quantity,
+    #[digest(ignore)]
+    pub token_contract: EthereumIdentity,
```
thomaseizinger, line 28: This should be part of the digest!
Two swaps on different ERC20 contracts are not the same :)

cnd/src/swap_protocols/herc20_rfc003_watcher.rs
```diff
@@ -0,0 +1,178 @@
...
+    role: Role,
+) {
+    create_watcher::<_, _, _, _, htlc_location::Ethereum, _, transaction::Ethereum>(
+        connector.as_ref(),
+        ethereum_ledger_state,
+        swap_id,
+        htlc_params,
+        Utc::now().naive_local(),
+    )
+    .instrument(tracing::error_span!("alpha_ledger", swap_id = %swap_id, role = %role))
```
thomaseizinger, line 64: This span is wrong now because the fn implies that it can be called for either side.

cnd/src/swap_protocols/herc20_rfc003_watcher.rs
```diff
@@ -0,0 +1,178 @@
...
+    A: Ord + Clone,
+    H: Clone,
+    I: Clone,
+    T: Clone,
+{
+    ledger_state
+        .insert(swap_id, LedgerState::<A, H, T>::NotDeployed)
+        .await;
+
+    // construct a generator that watches alpha and beta ledger concurrently
```
thomaseizinger, line 101: This comment does not longer make sense I guess :D

cnd/src/swap_protocols/facade.rs
da-kami: Thanks for pointing it out! had a feeling it should be - this is a great reminder.
Will record it as follow up!

cnd/src/swap_protocols/herc20_rfc003_watcher.rs
da-kami: hm... not sure I would invest much time in improving this. I see this complete file as a temporary solution that will go away once we properly use `herc20`

da-kami: happy to remove the log statement though

thomaseizinger: This is more of a reminder to not just blindly copy and paste code :)
You are welcome to copy code from other parts of the codebase (or from any other source if the license allows it) but once you copy it, it is YOUR code so you should take care of it the same way as if you would had written it yourself.

I am not asking to polish it, I am only asking to not introduce things that are wrong because the code is now used in a different setting and some assumptions just no longer apply.

Issue #2606, opened 2020-05-11
LN code is based on unsupported trading pair


tcharding:
## Problem

When we wrote the LN code we used heth/halight as the swap protocols. It turns out we do not want to support this combination.

## Goal

Have a useful first LN implementation to base future work off of.

DoD:
- An passing e2e test that tests the happy path for a herc20/halight swap.

PR #2622, opened 2020-05-12
LN clean up

tcharding:
While working on https://github.com/comit-network/comit-rs/issues/2606 clean up a few things.

In this PR do:
- Replace wrapper type `EthereumIdentity(identity::Ethereum)` with `identity::Ethereum`.
- Move `ToDigestInput` implementations into their respective modules (out of the facade).
- Rename facade field `beta_ledger_states` to be `halight_states` since it is of type `halight::States` _not_ beta ledger like we did in rfc003.

### DIFF
--- a/cnd/src/asset/bitcoin.rs
+++ b/cnd/src/asset/bitcoin.rs
@@ -1,4 +1,5 @@
 use bitcoin::{util::amount::Denomination, Amount};
+use digest::ToDigestInput;
 use std::fmt;
 
 #[derive(Debug, Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Hash)]
@@ -28,6 +29,12 @@ impl fmt::Display for Bitcoin {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
         let bitcoin = self.0.to_string_in(Denomination::Bitcoin);
         write!(f, "{} BTC", bitcoin)
+    }
+}
+
+impl ToDigestInput for Bitcoin {
+    fn to_digest_input(&self) -> Vec<u8> {
+        self.to_le_bytes().to_vec()
     }
 }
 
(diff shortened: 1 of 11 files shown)


thomaseizinger:
Thanks for cleaning this up :)

thomaseizinger:
@tcharding I fixed your merge conflicts.

thomaseizinger:
Just noticed but it is not super important:

> Move ToDigestInput implementations into their respective modules (out of the facade).

I think we might have a conceptual misalignment here.

I see it as one of Rust's strengths that trait impls and struct declarations don't have to be next too each other (like in Java/C# where you have to declare all interfaces implemented by a class right at the class definition).
This allows you to place trait impls next to the code that cares about them (i.e. for `ToDigestInput` that is the network layer) and not next to the actual struct definition.

The reason why I think this is important is because it decouples modules from each other: `ToDigestInput` comes from a our `digest` dependency. Ideally, the code that actually uses `.digest()` is the only module that needs to depend on this crate. If we spread the `ToDigestInput` trait impls all over the place, then a lot more modules have to depend on it.

tcharding:
> Just noticed but it is not super important:
> 
> > Move ToDigestInput implementations into their respective modules (out of the facade).
> 
> I think we might have a conceptual misalignment here.
> 
> I see it as one of Rust's strengths that trait impls and struct declarations don't have to be next too each other (like in Java/C# where you have to declare all interfaces implemented by a class right at the class definition).
> This allows you to place trait impls next to the code that cares about them (i.e. for `ToDigestInput` that is the network layer) and not next to the actual struct definition.
> 
> The reason why I think this is important is because it decouples modules from each other: `ToDigestInput` comes from a our `digest` dependency. Ideally, the code that actually uses `.digest()` is the only module that needs to depend on this crate. If we spread the `ToDigestInput` trait impls all over the place, then a lot more modules have to depend on it.

Thanks for the thorough explanation, I actually did a u-turn on this change over in the hbit module.  Had forgotten it was here also.  Will just close this PR since the hbit one has all these patches (excl. move of `ToDigestInput` impls).

PR #2623, opened 2020-05-12
Add hbit module

### LINKED ISSUE #2606: LN code is based on unsupported trading pair

tcharding:
We have been trying to implement the newly designed split protocols for ages now; this PR is a change of approach. Instead of pushing forward with the LN stuff hoping that a nice split protocols implementation will fall out this PR adds the herc20/hbit protocol pair.
 
Add an `hbit` module under `swap_protocols/` with the following things in mind:
- Do not use rfc003 types, duplicate rfc003 code into `hbit` if needed.
- Try not to break any current working code.

This PR adds two new end points for both swaps of the trading pair hbit/herc20. Save of the created swap is fully implemented.  No retrieval for this data from the data base, also the REST actions are not done (these are heavy users of data retrieval so will be interesting to implement on top of the new database design, this can be done after this merges without hampering our effort).
  
Adds `unimplemeneted!` for all entrypoints to the networking layer. 

Please note, during duplication of the generic rfc003 code the Deploy state was copied, this has not been removed - prefer doing so separately after merge.

### DIFF
--- a/cnd/src/bitcoin.rs
+++ b/cnd/src/bitcoin.rs
@@ -31,6 +31,12 @@ impl PublicKey {
     }
 }
 
+impl fmt::Display for PublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
 impl From<secp256k1::PublicKey> for PublicKey {
     fn from(key: secp256k1::PublicKey) -> Self {
         Self(bitcoin::PublicKey {
--- a/cnd/src/db/tables.rs
+++ b/cnd/src/db/tables.rs
@@ -1,14 +1,15 @@
 use crate::{
+    bitcoin,
     db::{
-        schema::{address_hints, halights, herc20s, secret_hashes, swaps},
+        schema::{address_hints, halights, hbits, herc20s, secret_hashes, swaps},
         wrapper_types::{
             custom_sql_types::{Text, U32},
-            Erc20Amount, EthereumAddress, LightningNetwork, Satoshis,
+            BitcoinNetwork, Erc20Amount, EthereumAddress, LightningNetwork, Satoshis,
         },
         Error, Sqlite,
     },
     identity, lightning,
-    swap_protocols::{halight, herc20, rfc003, HashFunction, Ledger, LocalSwapId, Role},
+    swap_protocols::{halight, hbit, herc20, rfc003, HashFunction, Ledger, LocalSwapId, Role},
 };
 use anyhow::Context;
 use diesel::{prelude::*, RunQueryDsl};
@@ -208,6 +209,72 @@ impl IntoInsertable for halight::CreatedSwap {
     }
 }
 
+#[derive(Associations, Clone, Copy, Debug, Identifiable, Queryable, PartialEq)]
+#[belongs_to(Swap)]
+#[table_name = "hbits"]
+pub struct Hbit {
+    id: i32,
+    swap_id: i32,
+    pub amount: Text<Satoshis>,
+    pub network: Text<BitcoinNetwork>,
+    pub hash_function: Text<HashFunction>,
+    pub redeem_identity: Option<Text<bitcoin::PublicKey>>,
+    pub refund_identity: Option<Text<bitcoin::PublicKey>>,
+    pub ledger: Text<Ledger>,
+}
+
+#[derive(Insertable, Clone, Copy, Debug)]
+#[table_name = "hbits"]
+pub struct InsertableHbit {
+    pub swap_id: i32,
+    pub amount: Text<Satoshis>,
+    pub network: Text<BitcoinNetwork>,
+    pub hash_function: Text<HashFunction>,
+    pub redeem_identity: Option<Text<bitcoin::PublicKey>>,
+    pub refund_identity: Option<Text<bitcoin::PublicKey>>,
+    pub ledger: Text<Ledger>,
+}
+
+impl InsertableHbit {
+    pub fn with_swap_id(&self, swap_id: i32) -> Self {
+        InsertableHbit {
+            swap_id,
+            amount: self.amount,
+            network: self.network,
+            hash_function: self.hash_function,
+            redeem_identity: self.redeem_identity,
+            refund_identity: self.refund_identity,
+            ledger: self.ledger,
+        }
+    }
+}
+
+impl IntoInsertable for hbit::CreatedSwap {
+    type Insertable = InsertableHbit;
+
+    fn into_insertable(self, swap_id: i32, role: Role, ledger: Ledger) -> Self::Insertable {
+        let redeem_identity = match role {
+            Role::Alice => Some(Text(self.identity)),
+            Role::Bob => None,
+        };
+        let refund_identity = match role {
+            Role::Alice => None,
+            Role::Bob => Some(Text(self.identity)),
+        };
+        assert!(redeem_identity.is_some() || refund_identity.is_some());
+
+        InsertableHbit {
+            swap_id,
+            amount: Text(self.amount.into()),
+            network: Text(self.network.into()),
+            hash_function: Text(HashFunction::Sha256),
+            redeem_identity,
+            refund_identity,
+            ledger: Text(ledger),
+        }
+    }
+}
+
 impl Insert<InsertableHerc20> for Sqlite {
     fn insert(
         &self,
@@ -229,6 +296,20 @@ impl Insert<InsertableHalight> for Sqlite {
         insertable: &InsertableHalight,
     ) -> anyhow::Result<()> {
         diesel::insert_into(halights::dsl::halights)
+            .values(insertable)
+            .execute(connection)?;
+
+        Ok(())
+    }
+}
+
+impl Insert<InsertableHbit> for Sqlite {
+    fn insert(
+        &self,
+        connection: &SqliteConnection,
+        insertable: &InsertableHbit,
+    ) -> anyhow::Result<()> {
+        diesel::insert_into(hbits::dsl::hbits)
             .values(insertable)
             .execute(connection)?;
 
@@ -468,6 +549,46 @@ impl Sqlite {
                     .first(connection)?;
 
                 Halight::belonging_to(&swap).first(connection).optional()
+            })
+            .await?
+            .ok_or(Error::SwapNotFound)?;
+
+        Ok(record)
+    }
+
+    pub async fn save_hbit(
+        &self,
+        swap_id: LocalSwapId,
+        data: &InsertableHbit,
+    ) -> anyhow::Result<()> {
+        self.do_in_transaction(|connection| {
+            let key = Text(swap_id);
+
+            let swap: Swap = swaps::table
+                .filter(swaps::local_swap_id.eq(key))
+                .first(connection)?;
+
+            let insertable = data.with_swap_id(swap.id);
+
+            diesel::insert_into(hbits::dsl::hbits)
+                .values(insertable)
+                .execute(&*connection)
+        })
+        .await?;
+
+        Ok(())
+    }
+
+    pub async fn load_hbit(&self, swap_id: LocalSwapId) -> anyhow::Result<Hbit> {
+        let record: Hbit = self
+            .do_in_transaction(|connection| {
+                let key = Text(swap_id);
+
+                let swap: Swap = swaps::table
+                    .filter(swaps::local_swap_id.eq(key))
+                    .first(connection)?;
+
+                Hbit::belonging_to(&swap).first(connection).optional()
             })
             .await?
             .ok_or(Error::SwapNotFound)?;
--- a/cnd/src/http_api/route_factory.rs
+++ b/cnd/src/http_api/route_factory.rs
@@ -177,8 +177,22 @@ pub fn create(
         .and(warp::path::param::<LocalSwapId>())
         .and(warp::path("refund"))
         .and(warp::path::end())
-        .and(facade)
+        .and(facade.clone())
         .and_then(http_api::routes::action_refund);
+
+    let hbit_herc20 = warp::post()
+        .and(warp::path!("swaps" / "hbit" / "herc20"))
+        .and(warp::path::end())
+        .and(warp::body::json())
+        .and(facade.clone())
+        .and_then(http_api::routes::post::post_hbit_herc20);
+
+    let herc20_hbit = warp::post()
+        .and(warp::path!("swaps" / "herc20" / "hbit"))
+        .and(warp::path::end())
+        .and(warp::body::json())
+        .and(facade)
+        .and_then(http_api::routes::post::post_herc20_hbit);
 
     preflight_cors_route
         .or(rfc003_get_swap)
@@ -198,6 +212,8 @@ pub fn create(
         .or(lightning_action_deploy)
         .or(lightning_action_redeem)
         .or(lightning_action_refund)
+        .or(hbit_herc20)
+        .or(herc20_hbit)
         .recover(http_api::unpack_problem)
         .with(warp::log("http"))
         .with(cors)
--- a/cnd/src/http_api/routes.rs
+++ b/cnd/src/http_api/routes.rs
@@ -1,5 +1,6 @@
 pub mod index;
 pub mod peers;
+pub mod post;
 pub mod rfc003;
 
 use crate::{
--- a/cnd/src/http_api/routes.rs
+++ b/cnd/src/http_api/routes.rs
@@ -139,7 +139,7 @@ pub async fn handle_get_halight_swap(
         LedgerState<asset::Erc20, htlc_location::Ethereum, transaction::Ethereum>,
     > = facade.alpha_ledger_states.get(&swap_id).await?;
 
-    let beta_ledger_state = facade.beta_ledger_states.get(&swap_id).await?;
+    let beta_ledger_state = facade.halight_states.get(&swap_id).await?;
 
     let created_swap = facade.get_created_swap(swap_id).await;
 
@@ -849,7 +849,7 @@ async fn handle_action_init(
         .ok_or_else(|| anyhow::anyhow!("alpha ledger state not found for {}", swap_id))?;
 
     let beta_ledger_state: halight::State = facade
-        .beta_ledger_states
+        .halight_states
         .get(&swap_id)
         .await?
         .ok_or_else(|| anyhow::anyhow!("beta ledger state not found for {}", swap_id))?;
@@ -902,7 +902,7 @@ async fn handle_action_deploy(
         .ok_or_else(|| anyhow::anyhow!("alpha ledger state not found for {}", swap_id))?;
 
     let beta_ledger_state: halight::State = facade
-        .beta_ledger_states
+        .halight_states
         .get(&swap_id)
         .await?
         .ok_or_else(|| anyhow::anyhow!("beta ledger state not found for {}", swap_id))?;
@@ -956,7 +956,7 @@ async fn handle_action_fund(
         .ok_or_else(|| anyhow::anyhow!("alpha ledger state not found for {}", swap_id))?;
 
     let beta_ledger_state: halight::State = facade
-        .beta_ledger_states
+        .halight_states
         .get(&swap_id)
         .await?
         .ok_or_else(|| anyhow::anyhow!("beta ledger state not found for {}", swap_id))?;
@@ -1017,7 +1017,7 @@ async fn handle_action_redeem(
         .ok_or_else(|| anyhow::anyhow!("alpha ledger state not found for {}", swap_id))?;
 
     let beta_ledger_state: halight::State = facade
-        .beta_ledger_states
+        .halight_states
         .get(&swap_id)
         .await?
         .ok_or_else(|| anyhow::anyhow!("beta ledger state not found for {}", swap_id))?;
@@ -1078,7 +1078,7 @@ async fn handle_action_refund(
         .ok_or_else(|| anyhow::anyhow!("alpha ledger state not found for {}", swap_id))?;
 
     let beta_ledger_state: halight::State = facade
-        .beta_ledger_states
+        .halight_states
         .get(&swap_id)
         .await?
         .ok_or_else(|| anyhow::anyhow!("beta ledger state not found for {}", swap_id))?;
--- a/cnd/src/http_api/routes/post.rs
+++ b/cnd/src/http_api/routes/post.rs
@@ -1,0 +1,164 @@
+use crate::{
+    asset,
+    db::{CreatedSwap, Save},
+    http_api::{problem, routes::into_rejection, Http},
+    identity,
+    network::{DialInformation, InitCommunication},
+    swap_protocols::{hbit, herc20, ledger, Facade, LocalSwapId, Role},
+};
+use serde::Deserialize;
+use warp::{http::StatusCode, Rejection, Reply};
+
+/// POST endpoints for the hbit/herc20 protocol pair.
+
+#[allow(clippy::needless_pass_by_value)]
+pub async fn post_hbit_herc20(
+    body: serde_json::Value,
+    facade: Facade,
+) -> Result<impl Reply, Rejection>
+where
+    Facade: Save<CreatedSwap<hbit::CreatedSwap, herc20::CreatedSwap>>
+        + InitCommunication<CreatedSwap<hbit::CreatedSwap, herc20::CreatedSwap>>,
+{
+    let body = Body::<Hbit, Herc20>::deserialize(&body)
+        .map_err(anyhow::Error::new)
+        .map_err(problem::from_anyhow)
+        .map_err(warp::reject::custom)?;
+
+    let swap_id = LocalSwapId::random();
+    let reply = warp::reply::reply();
+
+    let swap = hbit_herc20_created_swap_from_body(swap_id, body.clone());
+
+    facade
+        .save(swap.clone())
+        .await
+        .map_err(problem::from_anyhow)
+        .map_err(warp::reject::custom)?;
+
+    facade
+        .init_communication(swap_id, swap)
+        .await
+        .map(|_| {
+            warp::reply::with_status(
+                warp::reply::with_header(reply, "Location", format!("/swaps/{}", swap_id)),
+                StatusCode::CREATED,
+            )
+        })
+        .map_err(problem::from_anyhow)
+        .map_err(into_rejection)
+}
+
+#[allow(clippy::needless_pass_by_value)]
+pub async fn post_herc20_hbit(
+    body: serde_json::Value,
+    facade: Facade,
+) -> Result<impl Reply, Rejection>
+where
+    Facade: Save<CreatedSwap<herc20::CreatedSwap, hbit::CreatedSwap>>
+        + InitCommunication<CreatedSwap<herc20::CreatedSwap, hbit::CreatedSwap>>,
+{
+    let body = Body::<Herc20, Hbit>::deserialize(&body)
+        .map_err(anyhow::Error::new)
+        .map_err(problem::from_anyhow)
+        .map_err(warp::reject::custom)?;
+
+    let swap_id = LocalSwapId::random();
+    let reply = warp::reply::reply();
+
+    let swap = herc20_hbit_created_swap_from_body(swap_id, body.clone());
+
+    facade
+        .save(swap.clone())
+        .await
+        .map_err(problem::from_anyhow)
+        .map_err(warp::reject::custom)?;
+
+    facade
+        .init_communication(swap_id, swap)
+        .await
+        .map(|_| {
+            warp::reply::with_status(
+                warp::reply::with_header(reply, "Location", format!("/swaps/{}", swap_id)),
+                StatusCode::CREATED,
+            )
+        })
+        .map_err(problem::from_anyhow)
+        .map_err(into_rejection)
+}
+
+fn hbit_herc20_created_swap_from_body(
+    swap_id: LocalSwapId,
+    body: Body<Hbit, Herc20>,
+) -> CreatedSwap<hbit::CreatedSwap, herc20::CreatedSwap> {
+    CreatedSwap::<hbit::CreatedSwap, herc20::CreatedSwap> {
+        swap_id,
+        alpha: body.alpha.into(),
+        beta: body.beta.into(),
+        peer: body.peer.peer_id,
+        address_hint: None,
+        role: body.role.0,
+    }
+}
+
+fn herc20_hbit_created_swap_from_body(
+    swap_id: LocalSwapId,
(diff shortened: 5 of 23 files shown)


cnd/src/swap_protocols/hbit.rs
```diff
@@ -0,0 +1,242 @@
...
+    L: Clone + ledger::bitcoin::Bitcoin + ledger::bitcoin::Network,
+{
+    create_watcher::<_, _, _, _, htlc_location::Bitcoin, _, transaction::Bitcoin>(
+        connector.as_ref(),
+        ledger_states,
+        swap_id,
+        htlc_params,
+        Utc::now().naive_local(),
+    )
+    .instrument(tracing::error_span!("alpha_ledger", swap_id = %swap_id, role = %role))
```
thomaseizinger, line 82: Saw the same thing in Daniel's PR. I don''t know where this was copied from but at the time when I originally wrote these tracing statements, they made sense and this `alpha_ledger` was actually always true. This is no longer the case now.

As this needs some more thinking and probably refactoring to balance duplication and correctness, I would suggest to remove it for now.

cnd/src/swap_protocols/hbit.rs
```diff
@@ -0,0 +1,242 @@
...
+where
+    C: HtlcFunded<L, A, H, I, T>
+        + HtlcDeployed<L, A, H, I, T>
+        + HtlcRedeemed<L, A, H, I, T>
+        + HtlcRefunded<L, A, H, I, T>,
+    Deployed<H, T>: Clone,
+    Redeemed<T>: Clone,
+    Refunded<T>: Clone,
+{
+    let deployed = connector.htlc_deployed(&htlc_params, start_of_swap).await?;
```
thomaseizinger, line 167: Bitcoin HTLCs don't have a separate `deploy` step.

cnd/src/swap_protocols/hbit.rs
```diff
@@ -0,0 +1,242 @@
...
+                return;
+            }
+        }
+    }
+}
+
+/// Returns a future that waits for events to happen on a ledger.
+///
+/// Each event is yielded through the controller handle (co) of the coroutine.
+async fn watch_ledger<C, L, A, H, I, T>(
```
thomaseizinger, line 152: These type parameters are unnecessary now.

cnd/src/swap_protocols/hbit.rs
```diff
@@ -0,0 +1,242 @@
...
+
+            return Err(error);
+        }
+    }
+
+    Ok(())
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct HtlcParams<L, A, I> {
```
thomaseizinger, line 197: These are unnecessary as-well.

tcharding:
Thanks for the review @thomaseizinger.

> I appreciate your effort @tcharding but I am a bit hesitant if this is not too much duplication now.

We have to kill rfc003 _somehow_. I don't see a better way than duplication.  It is much easier than trying to fix stuff at the same time, for example, how I left all the deploy stuff in even though hbit should not have deploy - IMO this will be easier to remove once the code is duplicated and isolated.

> Wouldn't it be easier if we would first finish ERC20:LNBTC and let the abstractions settle and sink in for a bit and also reduce some of the duplication?

I don't believe so, but there is more than one way to skin a cat.  We are struggling to parallelise the LN work now with things so entangled, the hlbit/herc20 stuff can at least be done in parallel.

> I have the feeling things become almost unmanageable if we add another 1500 LoC that are basically copied from somewhere in the codebase :(

I have the feeling, as I have aired before, that the codebase is unmanageable now, basically a total mess and we should rewrite the whole thing from scratch. You have previously suggested you do not want to do this. I'm at the end of my mind with how tangled everything is, at least this duplication allows us to work on an isolated part of the code. I'm not the architect but I'm the one of the wombats who has to struggle against this clusterf**k everyday :)

cnd/src/bitcoin.rs
```diff
@@ -13,6 +13,50 @@ use serde::{
 };
 use std::{fmt, str::FromStr};
 
+/// The available Bitcoin networks.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Network {
```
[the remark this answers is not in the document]
tcharding: Well, since I was doing something totally new I thought I'd experiment with this - the three networks has been bothering me since we added them and no number of attempts at adding the Ethereum side has changed that (I've made 2 separate concerted efforts at it).

cnd/src/http_api/route_factory.rs
```diff
@@ -169,9 +169,23 @@ pub fn create(
         .and(warp::path::param::<LocalSwapId>())
         .and(warp::path("refund"))
         .and(warp::path::end())
-        .and(facade)
+        .and(facade.clone())
         .and_then(http_api::routes::action_refund);
 
+    let hbit_herc20 = warp::post()
+        .and(warp::path!("swaps" / "hbit" / "herc20"))
```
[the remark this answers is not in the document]
tcharding: word.

cnd/src/http_api/routes/post.rs
```diff
@@ -0,0 +1,188 @@
...
+    let reply = warp::reply::reply();
+
+    // FIXME: Make this clone less ugly.
+    let cloned = body.clone();
+    let swap = CreatedSwap::<hbit::CreatedSwap, herc20::CreatedSwap> {
+        swap_id,
+        alpha: cloned.alpha.into(),
+        beta: cloned.beta.into(),
+        peer: cloned.peer.peer_id,
+        address_hint: None, // FIXME: Add this.
```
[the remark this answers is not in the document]
tcharding: That sounds good to me, will work this in.

cnd/src/network/comit_ln.rs
```diff
@@ -628,14 +636,48 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
     }
 }
 
+#[async_trait]
+impl InitCommunication for ComitLN {
+    async fn init_communication<T>(&self, swap_id: LocalSwapId, params: T) -> anyhow::Result<()> {
+        // TODO: Implement generic communication methods based on
```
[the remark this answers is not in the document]
tcharding: I tried that and got compiler problems because of `Send`/`'sync`.  Caused apparently by the MDNS, this current attempt was the second failed attempt, any more suggestions appreciated.

cnd/src/swap_protocols/facade.rs
```diff
@@ -32,37 +37,56 @@ pub struct HanEtherereumHalightBitcoinCreateSwapParams {
...
+#[derive(Clone, Digest, Debug, PartialEq)]
+#[digest(hash = "SwapDigest")]
+pub struct Herc20HbitSwapParams {
+    #[digest(ignore)]
+    pub role: Role,
+    #[digest(ignore)]
+    pub peer: DialInformation,
+    #[digest(ignore)]
+    pub ethereum_identity: identity::Ethereum,
+    #[digest(prefix = "2001")]
```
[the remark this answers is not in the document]
tcharding: Done: https://github.com/comit-network/comit-rs/issues/2624

cnd/src/swap_protocols/hbit.rs
tcharding: Good catch, thanks.  I cut and pasted with no thought for the tracing - that is a habit I need to remedy :)

cnd/src/bitcoin.rs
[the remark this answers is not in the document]
thomaseizinger: Fair, maybe that kind of use of the type system is too much :)
Do you think we should turn around on this topic and go back to enums?

thomaseizinger:
> We have to kill rfc003 _somehow_. I don't see a better way than duplication. It is much easier than trying to fix stuff at the same time, for example, how I left all the deploy stuff in even though hbit should not have deploy - IMO this will be easier to remove once the code is duplicated and isolated.

I agree with killing it with duplication but I think as you duplicate stuff, you should at the same time make the changes to the code that you want to make (like removing `deploy`) otherwise there is no point in duplicating it, you might as-well use the old stuff.

> I have the feeling, as I have aired before, that the codebase is unmanageable now, basically a total mess and we should rewrite the whole thing from scratch. You have previously suggested you do not want to do this. I'm at the end of my mind with how tangled everything is, at least this duplication allows us to work on an isolated part of the code. I'm not the architect but I'm the one of the wombats who has to struggle against this clusterf**k everyday :)

Do you think it would help to take this as an opportunity to introduce a `comit` crate that holds all the protocols and btsieve and we slowly move things from cnd over and keep things "nice" over there? :)
This way, there would be a clear separation of code that we definitely want to keep, in a shape that we are happy with and the original modules in `cnd` are deprecated.

cnd/src/bitcoin.rs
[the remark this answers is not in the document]
tcharding: Yes, I do not think these three 'network' structs are a good use of the language features.  Also when we remove rfc003 I believe there original usecase is going to disappear (if I understood correctly why we added them in the first place). This PR shows that an enum serves us just nicely.

cnd/src/created_swap.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub alpha: A,
+    /// The parameters used on the beta ledger.
+    pub beta: B,
+    /// Peer ID of the swap counterparty.
+    pub peer: PeerId,
+    /// Role of the node in this swap, Alice or Bob.
+    pub role: Role,
+}
+
+pub type HbitHerc20CreatedSwap = CreatedSwap<hbit::CreatedSwap, herc20::CreatedSwap>;
```
luckysori, line 24: 🙃 Not too keen on this. Trading less characters for more indirection and basically the same name is not worth it for me.

cnd/src/db/created_swap.rs
```diff
@@ -142,14 +137,131 @@ impl Load<CreatedSwap<herc20::CreatedSwap, halight::CreatedSwap>> for Sqlite {
...
+        let hbit = InsertableHbit {
+            swap_id: 0, // FK, set during save.
+            amount: Text(created.alpha.amount.into()),
+            network: Text(created.alpha.network),
+            hash_function: Text(HashFunction::Sha256),
+            redeem_identity: redeem_identity.map(|pk| Text(pk.into())),
+            refund_identity: refund_identity.map(|pk| Text(pk.into())),
+            ledger: Text(Ledger::Alpha),
+        };
+        self.save_hbit(local_swap_id, &hbit).await?;
```
luckysori, line 178: 🤔 Could be extracted into a function to reduce duplication with function below. Or at least scoped. That way the comment on line 159 can go away too.

luckysori: Same for saving herc20 details.

cnd/src/http_api.rs
```diff
@@ -375,9 +352,9 @@ pub enum HttpAssetParams {
     Erc20(Erc20AssetParams),
 }
 
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
 pub struct BitcoinAssetParams {
-    quantity: Http<asset::Bitcoin>,
```
luckysori, line 380: ❓ Why the move away from the custom deserialisation?

cnd/src/http_api/route_factory.rs
```diff
@@ -133,6 +133,14 @@ pub fn create(
         .and(facade.clone())
         .and_then(http_api::routes::index::post_halight_bitcoin_herc20);
 
+    // This will replace get_halight_swap once implemented.
```
luckysori, line 136: 🔧 Seems obvious, could just remove these comments?

cnd/src/http_api/route_factory.rs
```diff
@@ -198,6 +220,9 @@ pub fn create(
         .or(lightning_action_deploy)
         .or(lightning_action_redeem)
         .or(lightning_action_refund)
+        .or(hbit_herc20)
+        .or(herc20_hbit)
+        //        .or(get_swaps)
```
luckysori, line 225: 🔧 Would remove.

cnd/src/http_api/routes/post.rs
```diff
@@ -0,0 +1,161 @@
...
+    db::Save,
+    http_api::{problem, routes::into_rejection, Http},
+    identity,
+    network::{DialInformation, InitCommunication},
+    swap_protocols::{hbit, herc20, Facade, LocalSwapId, Role},
+};
+use serde::Deserialize;
+use warp::{http::StatusCode, Rejection, Reply};
+
+/// POST endpoints for the hbit/herc20 protocol pair.
```
luckysori, line 13: ❓ Instead of having to clarify with this documentation string, should we just call the file something like `post_hbit_herc20.rs`. I guess eventually this all moves back to `index.rs` anyway.

luckysori: ❓ I don't immediately see why all this is not in `index.rs` already anyway. How does this duplication help us?

cnd/src/swap_protocols/facade.rs
```diff
@@ -34,6 +38,96 @@ pub struct Herc20HalightBitcoinCreateSwapParams {
     pub lightning_amount: asset::Bitcoin,
 }
 
+/// This represent the information available on a swap
```
luckysori, line 41: 🔧 
```suggestion
/// This represents the information available on a swap
```

luckysori: `Herc20HbitSwapParams` below is missing a similar docstring.

cnd/src/swap_protocols/hbit.rs
```diff
@@ -0,0 +1,204 @@
...
+use genawaiter::{
+    sync::{Co, Gen},
+    GeneratorState,
+};
+use std::sync::Arc;
+use tracing_futures::Instrument;
+
+/// Htlc Bitcoin atomic swap protocol.
+
+/// Data required to create a swap that involves Ether.
```
luckysori, line 41:
```suggestion
/// Data required to create a swap that involves Bitcoin.
```

cnd/src/swap_protocols/hbit/ledger_states.rs
```diff
@@ -0,0 +1,84 @@
+use crate::swap_protocols::{
+    hbit::{LedgerState, SwapEvent},
+    LocalSwapId,
+};
+use async_trait::async_trait;
+use std::collections::HashMap;
+use tokio::sync::Mutex;
+
+#[async_trait]
+pub trait Insert: Send + Sync + 'static {
```
luckysori, line 10: ❓ I could just be very far behind in terms of this codebase, but why are we not using the traits provided by `swap_protocols/state.rs`?

cnd/src/db/tables.rs
```diff
@@ -189,6 +189,46 @@ impl InsertableHalight {
     }
 }
 
+#[derive(Associations, Clone, Copy, Debug, Identifiable, Queryable, PartialEq)]
+#[belongs_to(Swap)]
+#[table_name = "hbits"]
+pub struct Hbit {
+    id: i32,
+    swap_id: i32,
+    pub amount: Text<Satoshis>, // FIXME: Do we want to store this as Satoshis?
```
bonomat, line 198: Would the alternative be to store it in Bitcoin? 
I think storing Satoshi makes sense to not have to convert Satoshi to Bitcoin and not use floating point numbers.

cnd/src/db/tables.rs
```diff
@@ -189,6 +189,46 @@ impl InsertableHalight {
...
+    pub network: Text<bitcoin::Network>,
+    pub hash_function: Text<HashFunction>,
+    pub redeem_identity: Option<Text<bitcoin::PublicKey>>,
+    pub refund_identity: Option<Text<bitcoin::PublicKey>>,
+    pub ledger: Text<Ledger>,
+}
+
+#[derive(Insertable, Clone, Copy, Debug)]
+#[table_name = "hbits"]
+pub struct InsertableHbit {
```
bonomat, line 208: Is this best-practice? 
I'd assume you can reuse the same type for inserting and loading.

cnd/src/http_api/routes.rs
```diff
@@ -202,7 +204,6 @@ pub async fn handle_get_halight_swap(
     }
 }
 
-// TODO: Refactor with make_finalized_swap_entity
```
bonomat, line 205: Instead of describing what someone needs to do, can you shortly explain why this should be refactored in this manner please? 
This might also indeed be better stored in a ticket as otherwise this will get lost here.

cnd/src/http_api/routes/get.rs
```diff
@@ -0,0 +1,25 @@
...
+    swap_protocols::{Facade, LocalSwapId},
+};
+use http_api_problem::HttpApiProblem;
+use warp::{Rejection, Reply};
+
+pub fn into_rejection(problem: HttpApiProblem) -> Rejection {
+    warp::reject::custom(problem)
+}
+
+#[allow(clippy::needless_pass_by_value)]
```
bonomat, line 12: Is my understanding correct that this will go away once we actually do the loading?

tcharding:
> I think it's okay to not tackle the unnecessary deploy of hbit (is that the new name?) with this PR, since I believe we still have it in rfc003. It's not like it's a step backwards. I would just remove those TODOs, we already know that we want that step to go away.

Cool, will do.  Thanks.  Oh and yes this is the new name bestowed upon us by the Gods i.e., me :)

cnd/src/created_swap.rs
tcharding: ok, I was indecisive about theses aliases, will remove.

cnd/src/http_api/routes/post.rs
```diff
@@ -0,0 +1,161 @@
...
+            warp::reply::with_status(
+                warp::reply::with_header(reply, "Location", format!("/swaps/{}", swap_id)),
+                StatusCode::CREATED,
+            )
+        })
+        .map_err(problem::from_anyhow)
+        .map_err(into_rejection)
+}
+
+#[allow(clippy::needless_pass_by_value)]
```
bonomat, line 52: Is this still valid?

cnd/src/db/created_swap.rs
tcharding: Nice, will do.

cnd/src/http_api.rs
tcharding: This was likely a mistake by me, have discussed with @thomaseizinger.  Am planning on removing this from the PR.

cnd/src/http_api/route_factory.rs
tcharding: Oh, you thought that was obvious?  I thought `get_halight_swap` was so mis-appropriately named that it was not obvious at all :)

cnd/src/http_api/route_factory.rs
tcharding: Can do.

cnd/src/http_api/routes/post.rs
tcharding: > Instead of having to clarify with this documentation string, should we just call the file something like post_hbit_herc20.rs.

I imagine _all_ the POST funtions to be in here.  This comment, that risks going stale, was to make it apparent that so far we only have this trading pair done - it feels like the `http_api/` is going to be hard to navigate because the naming of rfc003 files (`index.rs` etc.) is not being used now so its hard to tell which files are which.

> I don't immediately see why all this is not in index.rs already anyway. How does this duplication help us?

`index.rs` is going away, if I get my way, for 2 reasons. 1. because its not a useful name 2. Because it mostly contains halight code that needs rewriting. I did not want to break the LN code though so did not delete it yet.

cnd/src/swap_protocols/facade.rs
tcharding: Lol, I egotistically did not fix this because I don't like these structs :)  Will fix.

cnd/src/swap_protocols/hbit.rs
tcharding: Nice, thanks.

cnd/src/swap_protocols/hbit/ledger_states.rs
tcharding: Good and fair point, I will investigate.  Discussed already with @da-kami and @thomaseizinger and will either
- Use traits in `swap_protocols/state.rs`
- Remove the traits and just have methods.

cnd/src/swap_protocols/facade.rs
```diff
@@ -87,14 +166,12 @@ pub struct Facade {
     pub swarm: Swarm,
     // We currently only support Han-HALight, therefor 'alpha' is Ethereum and 'beta' is Lightning.
     pub alpha_ledger_states: Arc<LedgerStates>,
-    pub beta_ledger_states: Arc<halight::States>,
+    pub halight_states: Arc<halight::States>,
```
bonomat, line 169: Why was this renamed?

cnd/src/swap_protocols/hbit.rs
```diff
@@ -0,0 +1,204 @@
...
+pub struct HtlcParams {
+    pub network: bitcoin::Network,
+    pub asset: asset::Bitcoin,
+    pub redeem_identity: identity::Bitcoin,
+    pub refund_identity: identity::Bitcoin,
+    pub expiry: Timestamp,
+    pub secret_hash: SecretHash,
+}
+
+// TODO: Remove the Herc20 stuff i.e., no separate Deployed/Funded states.
```
bonomat, line 174: This should be tracked somewhere else. 
I propose to add a comment above at around line 135 explaining why we have this right now and creating a ticket to remove the unneeded swap events.

cnd/src/swap_protocols/hbit/events.rs
```diff
@@ -0,0 +1,77 @@
+use crate::{
+    asset, htlc_location,
+    swap_protocols::{hbit::HtlcParams, Secret},
+    transaction,
+};
+use chrono::NaiveDateTime;
+use serde::{Deserialize, Serialize};
+
+// TODO: Remove the Herc20 stuff i.e., no separate Deployed/Funded states.
```
bonomat, line 9: This TODO can go away if you resolve my comment in `cnd/src/swap_protocols/hbit.rs`

D4nte:
Lots of detailed reviews happening already. Not sure my input will help.
Please note I am confused by `hbit` name as it could both mean `halightbitcoin` and `hanbitcoin`.

cnd/src/bitcoin.rs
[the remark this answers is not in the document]
bonomat: Why is this introduced now and was not raised earlier? 

I'm afraid that introducing this here now means for sure that we have to refactor one or the other in the future. 
I'm not against this but we should stick to one solution and not jump back and forth.

cnd/src/swap_protocols/hbit/ledger_states.rs
```diff
@@ -0,0 +1,84 @@
...
+                tracing::warn!("Value not found for key {}", key);
+                return;
+            }
+        };
+
+        match event {
+            SwapEvent::Deployed(deployed) => ledger_state.transition_to_deployed(deployed),
+            SwapEvent::Funded(funded) => ledger_state.transition_to_funded(funded),
+            SwapEvent::Redeemed(redeemed) => {
+                // what if redeemed.secret.hash() != secret_hash in request ??
```
bonomat, line 61: Can you clarify how this can happen?

tcharding:
> Lots of detailed reviews happening already. Not sure my input will help.
> Please note I am confused by `hbit` name as it could both mean `halightbitcoin` and `hanbitcoin`.

'h' stands for HTLC, 'bit' stands for Bitcoin. `halight` will be relegated to our memories soon as we can come up with a name with at least a small amount of consensus, most likely `halbit`.

cnd/src/bitcoin.rs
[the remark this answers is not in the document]
thomaseizinger: > I'm afraid that introducing this here now means for sure that we have to refactor one or the other in the future.

We already have to do that because `Ethereum` is still using an enum :D
I am fine with moving back on this one, was a bad idea.

cnd/src/db/tables.rs
thomaseizinger: They differ in the way that `Insertable` doesn't have a primary key. Not sure if you can reuse them then?

cnd/src/http_api/routes/get.rs
thomaseizinger: No I don' think so. The problem is that `warp` passes us this by value so we actually cannot change the fn signature but clippy doesn't know that :(

cnd/src/db/created_swap.rs
tcharding: Had a play with this; it actually got me to somewhere I wanted to be without previously know how :)

We can add another type that groups with all the insertable/queryable types i.e.,
- `Foo`, `InsertableFoo`, and `FooData`

Open to a better name than `FooData`, the obvious is to use `Foo` and use `QueryableFoo` for the queryable type but this is out because of how the `belongs_to` macro works, well at least for the `Swap` type, and I'd like to be uniform.

Anyways, leaving this for a follow up PR because it touches all the db code.

cnd/src/db/tables.rs
bonomat: It won't work to make this field optional?

cnd/src/db/created_swap.rs
tcharding: What you have touched on here @lucas is actually a really big deal, fixing this makes the whole db code better. Easier to read, write, extend, and maintain. I was looking for this, thanks.

https://github.com/comit-network/comit-rs/issues/2635

cnd/src/swap_protocols/hbit/ledger_states.rs
tcharding: Removed the traits. Thanks.

cnd/src/http_api.rs
tcharding: So the upside of this is: left the custom serder in `bitcoin` module (because it is used by `strum_discriminants`/`HtlcState`), but did not use it in the http_api, instead reverted to using custom serde and wrapping in `Http` type. The `bitcoin` module serde stuff should be removed when we remove the `strum_discriminants` usage to create `HtlpState` (this relies on all types being serde).

Tracking issue: https://github.com/comit-network/comit-rs/issues/2636

cnd/src/asset/bitcoin.rs
```diff
@@ -31,6 +32,29 @@ impl fmt::Display for Bitcoin {
...
+        D: Deserializer<'de>,
+    {
+        let value = String::deserialize(deserializer)?;
+        let value =
+            u64::from_str(value.as_str()).map_err(<D as Deserializer<'de>>::Error::custom)?;
+        let amount = Bitcoin::from_sat(value);
+
+        Ok(amount)
+    }
+}
```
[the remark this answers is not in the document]
tcharding: Seen to in above comments, along with tracking issue.

cnd/src/swap_protocols/hbit/events.rs
tcharding: Removed, thanks.

cnd/src/swap_protocols/hbit/ledger_states.rs
tcharding: Ahhh, funny that should come up now - this is one of those places in the code that should have the label `FIXME`. I did not write this comment and have no clue what it means, hence I'd suggest we remove the CI rule and encourage devs to use
`FIXME (tcharding): blah blah`

tcharding: > Why is this introduced now and was not raised earlier?

It is introduced now because this PR does a bunch of code duplication that uses the structs that this enum replaces.

> I'm afraid that introducing this here now means for sure that we have to refactor one or the other in the future.

We definitely have a *lot* of refactoring to do to remove all the rfc003 kruft.

> I'm not against this but we should stick to one solution and not jump back and forth.

The stated aim of this PR was to write clean code _without_ breaking anyone else's code, hence we now have multiple abstractions for the same thing. (There is also an enum in the db layer that represents this also :)

cnd/src/db/tables.rs
da-kami: Please don'e try. We wasted enough time on trying to do such kind of optimisations. 

@bonomat we are not the only ones [discussing](https://www.reddit.com/r/rust/comments/9r0n1e/diesel_queryableinsertable/) this, but as far as I could see the typical way is still to use 2 different structs for insert and query.

cnd/src/http_api/routes.rs
D4nte: I Wrote this todo and actioned it. I wonder how it go in dev.

cnd/src/db/created_swap.rs
```diff
@@ -142,14 +137,131 @@ impl Load<CreatedSwap<herc20::CreatedSwap, halight::CreatedSwap>> for Sqlite {
...
+            Role::Alice => Some(created.alpha.identity),
+            Role::Bob => None,
+        };
+
+        let hbit = InsertableHbit {
+            swap_id: 0, // FK, set during save.
+            amount: Text(created.alpha.amount.into()),
+            network: Text(created.alpha.network),
+            hash_function: Text(HashFunction::Sha256),
+            redeem_identity: redeem_identity.map(|pk| Text(pk.into())),
```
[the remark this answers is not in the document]
tcharding: @lucas gave me an even better idea to clean all this constructor code up. Have opened a ticket for it so as to not spin for too long on perfecting the code. Thanks for the review.

cnd/src/swap_protocols/hbit/ledger_states.rs
thomaseizinger: @bonomat You commented on quite a few things that were not originally written in this PR but just copied over. I am not sure how helpful that is here :grimacing: 

We can debate how carefully one should be when copying code. I would argue quite careful but in this case, the main aim was to have a clean slate to work off from so I don't mind much :)

bonomat: How should I know this? It is green and marked as new, hence, it asked for a review ¯\_(ツ)_/¯ 

None of my comments are blocking, so don't worry too much about this.

thomaseizinger: > How should I know this? It is green and marked as new, hence, it asked for a review ¯_(ツ)_/¯

Fair point :)

cnd/src/swap_protocols/facade.rs
tcharding: Because the name is wrong, its an artifact of rfc003.

PR #2637, opened 2020-05-14
Save data we learn from the other party to the database

### LINKED ISSUE #2615: Save concern should be done in one transaction in the new db
## Problem

When implementing the new database schema we focussed on getting a working version fast (first iteration).
When saving the composition elements of `*Swap` structs into the database we need multiple save operations (e.g. insert into `swaps` table and insert into `address_hint` table). 

At the moment each insert is done in one transaction.

## Goal

Each save-concern (i.e. save operation from the outside consumer) should be within one transaction, i.e. if one save contains multiple table inserts those inserts should be done within one transaction.

thomaseizinger:
Also fixes #2615.

I had to merge #2640 into this because we cannot save data to the DB after the communication without saving from the controller first because we don't have a foreign key available in the DB.
The original code of this PR also had a bug where we messed up redeem/refund identities for the herc20 watcher by having them flipped. The new version should make clear, what is different per role before we spawn the watchers.


cnd/src/db.rs
```diff
@@ -190,26 +201,11 @@ pub struct CreatedSwap<A, B> {
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct InProgressSwap<A, B> {
     pub swap_id: LocalSwapId,
-    pub secret_hash: SecretHash,
```
thomaseizinger, line 193: Removed this because secret_hash is not present for Bob.

PR #2640, opened 2020-05-14
Store data only within a single transaction in the database

### LINKED ISSUE #2615: Save concern should be done in one transaction in the new db

thomaseizinger:
We also took the opportunity to make the `Save` impl for `CreatedSwap` generic over any kind of protocol, regardless of the side (alpha or beta).

Fixes #2615.

### DIFF
--- a/cnd/src/db.rs
+++ b/cnd/src/db.rs
@@ -99,7 +99,7 @@ impl Sqlite {
 
     async fn do_in_transaction<F, T, E>(&self, f: F) -> Result<T, E>
     where
-        F: Fn(&SqliteConnection) -> Result<T, E>,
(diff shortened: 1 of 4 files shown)


thomaseizinger:
Will be included in https://github.com/comit-network/comit-rs/pull/2637.

PR #2641, opened 2020-05-14
Go back to modelling Bitcoin ledgers as enums

luckysori:
Worth it.

### DIFF
--- a/cnd/src/comit_api.rs
+++ b/cnd/src/comit_api.rs
@@ -2,7 +2,7 @@ use crate::{
     asset::{self, AssetKind},
     libp2p_comit_ext::{FromHeader, ToHeader},
     swap_protocols::{
-        ledger::Ethereum,
+        ledger,
         rfc003::{messages::Decision, SwapId},
         SwapProtocol,
     },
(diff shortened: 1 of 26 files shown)


cnd/src/comit_api.rs
```diff
@@ -14,26 +14,26 @@ use std::str::FromStr;
 #[derive(Clone, Copy, derivative::Derivative, PartialEq)]
 #[derivative(Debug = "transparent")]
 pub enum LedgerKind {
```
tcharding, line 16: Doesn't need changing for this PR but I think we are overloading the term 'ledger', here we are referring to the network, right, perhaps `NetworkKind` would be better.  Also we started calling the protocol side 'ledger' also i.e., alpha/beta. I never commented on @thomaseizinger suggestion on slack but the more it sits with me the more I think 'side' might be a useful term. Anyways, just putting up some thoughts for later.

cnd/src/comit_api.rs
```diff
@@ -14,26 +14,26 @@ use std::str::FromStr;
...
 }
 
 impl FromHeader for LedgerKind {
     fn from_header(mut header: Header) -> Result<Self, serde_json::Error> {
         Ok(match header.value::<String>()?.as_str() {
             "bitcoin" => match header.take_parameter::<String>("network")?.as_ref() {
-                "mainnet" => LedgerKind::BitcoinMainnet,
-                "testnet" => LedgerKind::BitcoinTestnet,
-                "regtest" => LedgerKind::BitcoinRegtest,
+                "mainnet" => LedgerKind::Bitcoin(ledger::Bitcoin::Mainnet),
```
tcharding, line 25: Then these would become `network::Bitcoin::Mainnet` - and walah now I remember why this doesn't work, we cannot have another 'network' module in the root of cnd. Perhaps we need to submodule all the ledger stuff (e.g., identity, transaction, htlc_location, and now - network)?

cnd/src/comit_api.rs
thomaseizinger: What about just `Network` for this one?

cnd/src/comit_api.rs
thomaseizinger: I would see it as ledger specific: `bitcoin::Network::Mainnet`.
For Ethereum, the term is chain, so it would be: `ethereum::Chain::Ropsten`.

PR #2650, opened 2020-05-15
Make ComitLN more generic

### LINKED ISSUE #2388: Investigate alternative strategies for implementing the communication layer using the new oneshot protocols
## Problem

[Spike 23](https://github.com/comit-network/spikes/blob/master/0023-communication-protocol.adoc) defines a new set of libp2p protocols that can be orchestrated to achieve the communication necessary for starting  a swap.

An initial version of this was implemented by the LN Mob. This initial version is based on having a `NetworkBehaviour` per protocol and later composing those together with the derive of `NetworkBehaviour`.

This has several problems:

- Even if we don't see it, the code generated by the derived NetworkBehaviour is really complicated.
- We have to either maintain a [workaround](https://github.com/comit-network/comit-rs/blob/b9ca8f3659448f2499c314da8eb6b551c0a74ef7/cnd/src/network/oneshot_behaviour.rs#L35-L42) for not keep track of our connections or re-implement [the connection handling](https://github.com/comit-network/comit-rs/blob/b9ca8f3659448f2499c314da8eb6b551c0a74ef7/cnd/src/network/protocols/announce/behaviour.rs#L57-L103) in every behaviour.
- We have several layers of libp2p components that are pretty stupid and just forward things.
- The state management is pretty ugly and doesn't scale.
We have several hashmaps in this component where we store stuff keyed by the SwapId. We also maintain several boolean flags to keep track of what already happened.

## Goal

Investigate alternative ways of implementing the required functionality that are more elegant and hence, hopefully easier to maintain.
Scaling, adding other pairs for han/herc20/halight, should be kept in mind.

## Recommendation

It sounds promising to create a `ProtocolsHandler` that can handle more than a single protocol at once, similar to the `multi` handler: https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/protocols_handler/multi.rs

Unfortunately, the multi handler requires those protocols to have the same type, which is not the case for us. It seems unlikely that we can do this generically but we don't need to. A solution tailored to our codebase is fine.

A more modular design, if possible, is likely to be the ideal solution.

D4nte:
Hopefully inline comments help understand what is going on.

Kind of resolves #2388.


thomaseizinger:
Continued in #2869.

Issue #2626, opened 2020-05-13
Add `token_contract` to swap digest for Herc20 params


da-kami:
## Problem

`token_contract` should be part of the swap digest, as there can be multiple swaps using different tokens:
https://github.com/comit-network/comit-rs/blob/0b94ccc21ddd12229cf0cdd28ecb5af5bc3fb803/cnd/src/swap_protocols/facade.rs#L27-L28

When quickly changing the code form `Han` to `Herc20` this was not tackled.

I wanted to give this a quick fix, but ran into some (probably easy to fix) conversion problems when implementing `ToDigestInput` for `EthereumIdentity` - which us basically a `ethereum::Address` .

## Goal

`token_contract` is part of the digest:

PR #2661, opened 2020-05-18
Add token_contract to the digest for herc20 swaps

tcharding:
Swaps that involve the Herc20 protocol should include the token contract address in the digest since there may be multiple swaps including different tokens.

- Patch 1 updates `proc-macro2`
- Patch 2 removes 'pub' from H64 
- Patch 3 adds the `token_contract` field to digest

Fixes: #2626

### DIFF
--- a/cnd/src/swap_protocols/facade.rs
+++ b/cnd/src/swap_protocols/facade.rs
@@ -26,7 +26,7 @@ pub struct Herc20HalightBitcoinCreateSwapParams {
     pub ethereum_absolute_expiry: Timestamp,
     #[digest(prefix = "2002")]
     pub ethereum_amount: asset::Erc20Quantity,
-    #[digest(ignore)]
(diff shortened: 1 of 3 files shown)


cnd/src/swap_protocols/facade.rs
```diff
@@ -26,7 +26,7 @@ pub struct Herc20HalightBitcoinCreateSwapParams {
     pub ethereum_absolute_expiry: Timestamp,
     #[digest(prefix = "2002")]
     pub ethereum_amount: asset::Erc20Quantity,
-    #[digest(ignore)]
+    #[digest(prefix = "2003")]
     pub token_contract: identity::Ethereum,
```
da-kami, line 30: Would be nice if we can change `identity::Ethereum` to some other type (alias) sometimes, as semantically this is not an identity, but a token contract address. 
Can do that in another iteration.

PR #2671, opened 2020-05-19
Add missing dependencies to comit/Cargo.toml

luckysori:


tcharding:
Kicking off merge because this is super trivial, is non-invasive, and I'd like to have it in so I can run clippy in my editor again.

PR #2672, opened 2020-05-19
Actually build hbit

tcharding:
In Rust code is not built unless the module is declared in its parent module.

Add `hbit` to the root module so it is actually built. Fix the half baked effort I made when adding this module.

### DIFF
--- a/comit/src/hbit.rs
+++ b/comit/src/hbit.rs
@@ -1,9 +1,16 @@
 //! Htlc Bitcoin atomic swap protocol.
 
 use crate::{
-    asset, htlc_location, identity, timestamp::Timestamp, transaction, Secret, SecretHash,
+    asset,
+    btsieve::bitcoin::{BitcoindConnector, Cache},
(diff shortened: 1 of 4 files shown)


comit/src/hbit.rs
```diff
@@ -124,23 +133,32 @@ async fn watch_ledger(
...
+                    return Err(error);
+                }
+            }
         }
-        Err(either) => {
-            let (error, _other_future) = either.factor_first();
-            return Err(error);
+        Funded::Incorrectly { .. } => {
+            // FIXME: Is this what we want?
+            tracing::info!("Htlc was incorrectly funded, aborting watcher");
```
thomaseizinger, line 161: I would say no. This part of the code is not about making decisions, it is about observing.

Redeem and refund can still happen, even if incorrectly funded :)

tcharding: Cool, thanks. Will fix.

PR #2677, opened 2020-05-20
2648 validate table update

### LINKED ISSUE #2648: Verify number of updated rows for update statements
## Problem

Update statements can result in 0 affected rows without error here: https://github.com/comit-network/comit-rs/blob/c5b4b119f5053685ea7469ea166a5dd3b8a3f715/cnd/src/db/tables.rs#L330-L348

This is an error but won't bubble up.

## Goal

No silent errors

bonomat:
resolves #2648

### DIFF
--- a/cnd/src/db/tables.rs
+++ b/cnd/src/db/tables.rs
@@ -300,6 +300,22 @@ macro_rules! swap_id_fk {
     };
 }
 
+trait EnsureSingleRowAffected {
+    fn ensure_single_row_affected(self) -> anyhow::Result<usize>;
+}
+
+impl EnsureSingleRowAffected for usize {
+    fn ensure_single_row_affected(self) -> anyhow::Result<usize> {
+        if self != 1 {
+            return Err(anyhow::anyhow!(
+                "Expected rows to be updated should have been 1 but was {}",
+                self
+            ));
+        }
+        Ok(self)
+    }
+}
+
 impl Sqlite {
     pub async fn role(&self, swap_id: LocalSwapId) -> anyhow::Result<Role> {
         let swap = self.load_swap(swap_id).await?;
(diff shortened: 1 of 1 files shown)


cnd/src/db/tables.rs
```diff
@@ -399,8 +399,15 @@ impl Sqlite {
...
-        Ok(())
+        // There should be exactly one row updated, everything else is an error
+        if updated_rows == 1 {
+            Ok(())
+        } else {
+            Err(anyhow::anyhow!(
+                "Could not update halight refund identity for swap {}",
+                local_swap_id
+            ))
+        }
```
thomaseizinger: Personally, I mildly prefer handling error cases separately:

```suggestion
        // There should be exactly one row updated, everything else is an error
        if updated_rows != 1 {
            Err(anyhow::anyhow!(
                "Could not update halight refund identity for swap {}",
                local_swap_id
            ))
        }
        
        Ok(())
```

This way, the success path is visually different from the error path :)

bonomat:
> Not a huge fan of just extending the existing test but given that we don't have a concise way of constructing test data yet, it is understandable that you didn't want to duplicate all the setup code :)
> 

I had it as an extra test first and removed it again due to the duplication :)

> I think the extracted function makes things harder to understand to be honest.
> 
> If the duplication bothers you, you could look into making an extension trait that is implemented on the return type of `.execute()` so you can do:
> 

Ah, yes, I wanted to do this but didn't know how. If you think this is a good idea, I'd be happy to pair-program on this and learn :)

cnd/src/db/tables.rs
bonomat: Completely agree that the _return early_ best practice is way nicer.

thomaseizinger:
Thanks @bonomat :)

bonomat:
bors retry

PR #2680, opened 2020-05-20
Keep paving the way for handling multiple SwapKinds

luckysori:
...by implementing a generic way that works for the single protocol pair `herc20-halight`.

### DIFF
--- a/cnd/src/http_api/routes.rs
+++ b/cnd/src/http_api/routes.rs
@@ -21,6 +21,8 @@ use crate::{
 };
 use ::comit::{Protocol, Secret, SecretHash};
 use blockchain_contracts::ethereum::rfc003::{ether_htlc::EtherHtlc, Erc20Htlc};
+use comit::{identity, Never, Timestamp};
+use herc20::build_erc20_htlc;
 use http_api_problem::HttpApiProblem;
 use serde::Serialize;
 use std::collections::HashMap;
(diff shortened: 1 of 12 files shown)


cnd/src/http_api/routes.rs
```diff
@@ -684,19 +754,49 @@ impl RefundAction for AliceHerc20HalightBitcoinSwap {
     }
 }
 
+impl InitAction for BobHerc20HalightBitcoinSwap {
+    type Output = ();
```
thomaseizinger, line 758: Could use `comit::Never` here!

cnd/src/http_api/routes.rs
```diff
@@ -817,18 +912,12 @@ async fn handle_action_fund(id: LocalSwapId, facade: Facade) -> anyhow::Result<A
 
     let maybe_response = match role {
         Role::Alice => {
-            let state = facade.get_alice_herc20_halight_swap(id).await?;
-
-            state
-                .map(|state| state.fund_action().map(ActionResponseBody::from))
-                .flatten()
+            let swap = facade.get_alice_herc20_halight_swap(id).await?;
+            swap.fund_action().map(ActionResponseBody::from)
```
thomaseizinger, line 916: Future, potential optimization:

It might be worthwhile to change the return value of `fund_action` to a `Result` so we can `?` here and do:

```suggestion
            let action = facade.get_alice_herc20_halight_swap(id).await?.fund_action()?;
            
            ActionResponseBody::from(action)
```

We would need to return a dedicated error from this that can be mapped to a `404`.

thomaseizinger: That would also get rid of this line in every controller: `let response = maybe_response.ok_or(LndActionError::NotFound)?;`

cnd/src/swap_protocols/facade.rs
```diff
@@ -149,63 +150,110 @@ impl Facade {
         self.swarm.initiate_communication(id, swap_params).await
     }
 
-    pub async fn get_created_swap(
```
thomaseizinger, line 152: :tada:

cnd/src/swap_protocols/facade.rs
```diff
@@ -149,63 +150,110 @@ impl Facade {
...
         let beta_state = self.halight_states.get(&id).await?;
 
-        let (alpha_state, beta_state) = match (alpha_state, beta_state) {
+        let secret = self.seed.derive_swap_seed(id).derive_secret();
+
+        let (herc20_state, halight_state) = match (alpha_state, beta_state) {
             (Some(alpha_state), Some(beta_state)) => (alpha_state, beta_state),
-            _ => return Ok(None),
+            _ => {
+                // TODO: Don't load DisplaySwap, change database API
```
thomaseizinger, line 165: Yeah this is fine for now, we can optimize later :)

cnd/src/http_api/routes.rs
```diff
@@ -335,71 +298,97 @@ struct SwapResource {
...
 #[derive(Clone, Debug)]
-pub struct AliceHerc20HalightBitcoinSwap {
-    pub alpha_ledger_state: herc20::State,
-    pub beta_ledger_state: halight::State,
-
-    pub herc20_params: herc20::InProgressSwap,
-    pub halight_params: halight::InProgressSwap,
+pub enum AliceHerc20HalightBitcoinSwap {
+    Created {
+        // FIXME: this should be herc20::Asset
```
thomaseizinger, line 327: Does it have to be? I think `herc20::Asset` is particularly useful if you want to express it in a single type. Within this big one, it just seems to be unnecessary wrapping as we are already in the context of `herc20`? But maybe consistency is better :)

Given that we are going to remove `DisplaySwap`, do we even still need `herc20::Asset`? :D

thomaseizinger:
Thanks @luckysori and @da-kami !

This is great progress :)

cnd/src/http_api/routes.rs
luckysori: These are all valid questions! We were definitely targetting consistency with this. I am leaning towards removing them altogether for now.

da-kami: I'm totally OK with removing `herc20::Asset` again.

cnd/src/swap_protocols/facade.rs
```diff
@@ -149,63 +151,108 @@ impl Facade {
...
 
-        let herc20_params = Load::<herc20::InProgressSwap>::load(self, id).await?;
-        let halight_params = Load::<halight::InProgressSwap>::load(self, id).await?;
+        let (
+            herc20_asset,
+            herc20::Identities {
+                redeem_identity: herc20_redeem_identity,
+                refund_identity: herc20_refund_identity,
+            },
+            herc20_expiry,
```
tcharding, line 230: The tuple idea is quite nice but there is a downside to using tuples instead of types; now every dev has to look up the order of the three types in the tuple each time they write any code like this.

luckysori: I totally agree and we considered a few other possibilities. For now I am happy with this, especially since we avoid the ambiguity between the two identities by introducing the `{herc20,halight}::Identities` types.

PR #2688, opened 2020-05-21
Introduce `new` functions in protocol modules

thomaseizinger:
This allows us to conveniently create new instances of protocols without duplicating that code all over the place.

### DIFF
--- a/cnd/src/lib.rs
+++ b/cnd/src/lib.rs
@@ -18,7 +18,6 @@
 // an enum created by `strum_discriminants`.
 #[macro_use]
 extern crate strum_macros;
-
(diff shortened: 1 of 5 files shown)


thomaseizinger:
> Very cool! Not sure if the differences between the docs of `halight::new` and `herc20::new` are intentional though.

Was an oversight but I don't think it is critical :D

tcharding:
> > Very cool! Not sure if the differences between the docs of `halight::new` and `herc20::new` are intentional though.
> 
> Was an oversight but I don't think it is critical :D

I pushed a commit to make them uniform.

PR #2689, opened 2020-05-21
Improve the network module

tcharding:
Builds on top of @D4nte's draft PR re-working the network module to make it easier to add more protocols.

The idea is to make the `Comit` (previously named `ComitLN`) `NetworkBehaviour` agnostic to the swap protocol that it is involved in i.e., within `Comit` there is no alpha/beta only data to be communicated with the peer node. One layer up, at the swarm (`ComitNode` `NetworkBehaviour`) the protocol and role are known.

All reads and writes to the database, which imply knowledge of the protocol, happen in the swarm. For this we add a reference to the new `Storage`.

Please note: This PR touches the network module, and call sites to the network module entry point `initiate_communication`, this includes the `Facade` and `index.rs`. Also, since this PR removes the now stale 'new' way to enter the network module `InitCommunication` added during the `hbit` module PR. We leave the endpoint code as 'untouched' as possible by simply removing the calls to `init_communication` and adding `unimplemeneted!()`.

### DIFF
--- a/cnd/src/http_api/routes/index.rs
+++ b/cnd/src/http_api/routes/index.rs
@@ -3,12 +3,11 @@ use crate::{
     db::{CreatedSwap, Save},
     http_api::{problem, routes::into_rejection, DialInformation, Http},
     identity,
-    network::ListenAddresses,
-    swap_protocols::{
-        halight, herc20, ledger, Facade, Herc20HalightBitcoinCreateSwapParams, LocalSwapId,
-        Rfc003Facade, Role,
-    },
+    network::{Identities, ListenAddresses},
+    swap_protocols::{halight, herc20, ledger, Facade, LocalSwapId, Rfc003Facade, Role},
 };
+use comit::network::swap_digest;
+use digest::Digest;
 use http_api_problem::HttpApiProblem;
 use libp2p::{Multiaddr, PeerId};
 use serde::{Deserialize, Serialize};
@@ -66,28 +65,26 @@ pub async fn post_herc20_halight_bitcoin(
         .map_err(problem::from_anyhow)
         .map_err(warp::reject::custom)?;
 
-    let swap_params: Herc20HalightBitcoinCreateSwapParams = body.clone().into();
-
     let swap_id = LocalSwapId::default();
     let reply = warp::reply::reply();
 
-    let swap = CreatedSwap::<herc20::CreatedSwap, halight::CreatedSwap> {
-        swap_id,
-        alpha: body.alpha.into(),
-        beta: body.beta.into(),
-        peer: body.peer.into(),
-        address_hint: None,
-        role: body.role.0,
-    };
-
+    let swap = body.to_created_swap(swap_id);
     facade
         .save(swap)
         .await
         .map_err(problem::from_anyhow)
         .map_err(warp::reject::custom)?;
 
+    let identities = Identities {
+        ethereum_identity: Some(body.alpha.identity),
+        lightning_identity: Some(body.beta.identity),
+    };
+    let digest = swap_digest::Herc20Halight::from(body.clone()).digest();
+    let peer = body.peer.into();
+    let role = body.role.0;
+
     facade
-        .initiate_communication(swap_id, swap_params)
+        .initiate_communication(swap_id, peer, role, digest, identities)
         .await
         .map(|_| {
             warp::reply::with_status(
@@ -96,7 +93,7 @@ pub async fn post_herc20_halight_bitcoin(
             )
         })
         .map_err(problem::from_anyhow)
-        .map_err(into_rejection)
+        .map_err(warp::reject::custom)
 }
 
 // `warp::reply::Json` is used as a return type to please the compiler
@@ -127,20 +124,41 @@ pub struct Body<A, B> {
     pub role: Http<Role>,
 }
 
-impl From<Body<Herc20EthereumErc20, HalightLightningBitcoin>>
-    for Herc20HalightBitcoinCreateSwapParams
-{
+impl From<Body<Herc20EthereumErc20, HalightLightningBitcoin>> for swap_digest::Herc20Halight {
     fn from(body: Body<Herc20EthereumErc20, HalightLightningBitcoin>) -> Self {
         Self {
-            role: body.role.0,
-            peer: body.peer.into(),
-            ethereum_identity: body.alpha.identity,
             ethereum_absolute_expiry: body.alpha.absolute_expiry.into(),
-            ethereum_amount: body.alpha.amount,
-            lightning_identity: body.beta.identity,
+            erc20_amount: body.alpha.amount,
+            token_contract: body.alpha.contract_address,
             lightning_cltv_expiry: body.beta.cltv_expiry.into(),
             lightning_amount: body.beta.amount.0,
-            token_contract: body.alpha.contract_address,
+        }
+    }
+}
+
+trait ToCreatedSwap<A, B> {
+    fn to_created_swap(&self, id: LocalSwapId) -> CreatedSwap<A, B>;
+}
+
+impl ToCreatedSwap<herc20::CreatedSwap, halight::CreatedSwap>
+    for Body<Herc20EthereumErc20, HalightLightningBitcoin>
+{
+    fn to_created_swap(
+        &self,
+        swap_id: LocalSwapId,
+    ) -> CreatedSwap<herc20::CreatedSwap, halight::CreatedSwap> {
+        let body = self.clone();
+
+        let alpha = herc20::CreatedSwap::from(body.alpha);
+        let beta = halight::CreatedSwap::from(body.beta);
+
+        CreatedSwap {
+            swap_id,
+            alpha,
+            beta,
+            peer: body.peer.into(),
+            address_hint: None,
+            role: body.role.0,
         }
     }
 }
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -1,4 +1,5 @@
 pub mod comit;
+pub mod start_swap;
 #[cfg(test)]
 pub mod test_swarm;
 pub mod transport;
@@ -7,39 +8,38 @@ pub use ::comit::*;
 pub use transport::ComitTransport;
 
 use crate::{
-    asset::{AssetKind, Erc20},
-    btsieve::{
-        bitcoin::{self, BitcoindConnector},
-        ethereum::{self, Web3Connector},
-    },
+    asset::AssetKind,
+    btsieve::bitcoin::{self, BitcoindConnector},
     comit_api::LedgerKind,
     config::Settings,
-    db::{CreatedSwap, ForSwap, Rfc003Swap, Save, Sqlite},
+    db::{Rfc003Swap, Save, Sqlite},
     htlc_location,
-    http_api::LedgerNotConfigured,
     libp2p_comit_ext::{FromHeader, ToHeader},
-    network::comit::Comit,
-    seed::RootSeed,
+    network::comit::{Comit, LocalData},
+    protocol_spawner::ProtocolSpawner,
+    seed::{DeriveSwapSeed, RootSeed},
+    storage::Storage,
     swap_protocols::{
-        halight, hbit, herc20,
         rfc003::{
             self,
             messages::{Decision, DeclineResponseBody, Request, RequestBody, SwapDeclineReason},
             state::Insert,
-            LedgerState, SwapCommunication, SwapCommunicationStates, SwapId,
+            DeriveSecret, LedgerState, SwapCommunication, SwapCommunicationStates, SwapId,
         },
-        HashFunction, Herc20HalightBitcoinCreateSwapParams, LocalSwapId, Role, SwapProtocol,
+        HashFunction, LocalSwapId, Role, SharedSwapId, SwapProtocol,
     },
     transaction,
 };
-use ::comit::lnd::{LndConnectorAsReceiver, LndConnectorAsSender, LndConnectorParams};
+use ::comit::{
+    btsieve::ethereum::{self, Web3Connector},
+    network::protocols::announce::SwapDigest,
+};
 use anyhow::Context;
 use async_trait::async_trait;
-use chrono::Utc;
 use futures::{
     channel::oneshot::{self, Sender},
     stream::StreamExt,
-    Future, FutureExt,
+    Future,
 };
 use libp2p::{
     identity::{ed25519, Keypair},
@@ -79,14 +79,13 @@ impl Swarm {
         seed: RootSeed,
         bitcoin_connector: Arc<bitcoin::Cache<BitcoindConnector>>,
         ethereum_connector: Arc<ethereum::Cache<Web3Connector>>,
-        lnd_connector_params: Option<LndConnectorParams>,
         swap_communication_states: Arc<SwapCommunicationStates>,
         rfc003_alpha_ledger_states: Arc<rfc003::LedgerStates>,
         rfc003_beta_ledger_states: Arc<rfc003::LedgerStates>,
-        herc20_states: Arc<herc20::States>,
-        halight_states: Arc<halight::States>,
         database: &Sqlite,
         task_executor: tokio::runtime::Handle,
+        storage: Storage,
+        protocol_spawner: ProtocolSpawner,
     ) -> anyhow::Result<Self> {
         let local_key_pair = derive_key_pair(&seed);
         let local_peer_id = PeerId::from(local_key_pair.clone().public());
@@ -96,15 +95,14 @@ impl Swarm {
         let behaviour = ComitNode::new(
             bitcoin_connector,
             ethereum_connector,
-            lnd_connector_params,
             swap_communication_states,
             rfc003_alpha_ledger_states,
             rfc003_beta_ledger_states,
-            herc20_states,
-            halight_states,
             seed,
             database.clone(),
             task_executor.clone(),
+            storage,
+            protocol_spawner,
         )?;
 
         let mut swarm = SwarmBuilder::new(transport, behaviour, local_peer_id.clone())
@@ -129,19 +127,13 @@ impl Swarm {
     pub async fn initiate_communication(
         &self,
         id: LocalSwapId,
-        swap_params: Herc20HalightBitcoinCreateSwapParams,
+        peer: DialInformation,
+        role: Role,
+        digest: SwapDigest,
+        identities: Identities,
     ) -> anyhow::Result<()> {
         let mut guard = self.inner.lock().await;
-
-        guard.initiate_communication(id, swap_params)
-    }
-
-    pub async fn get_created_swap(
-        &self,
-        id: LocalSwapId,
-    ) -> Option<Herc20HalightBitcoinCreateSwapParams> {
-        let mut guard = self.inner.lock().await;
-        guard.get_created_swap(id)
+        guard.initiate_communication(id, peer, role, digest, identities)
     }
 }
 
@@ -201,8 +193,6 @@ pub struct ComitNode {
     pub bitcoin_connector: Arc<bitcoin::Cache<BitcoindConnector>>,
     #[behaviour(ignore)]
     pub ethereum_connector: Arc<ethereum::Cache<Web3Connector>>,
-    #[behaviour(ignore)]
-    lnd_connector_params: Option<LndConnectorParams>,
 
     #[behaviour(ignore)]
     pub seed: RootSeed,
@@ -210,6 +200,11 @@ pub struct ComitNode {
     pub db: Sqlite,
     #[behaviour(ignore)]
     task_executor: Handle,
+
+    #[behaviour(ignore)]
+    pub storage: Storage,
+    #[behaviour(ignore)]
+    pub protocol_spawner: ProtocolSpawner,
 
     // rfc003
     #[behaviour(ignore)]
@@ -220,11 +215,6 @@ pub struct ComitNode {
     pub rfc003_beta_ledger_states: Arc<rfc003::LedgerStates>,
     #[behaviour(ignore)]
     response_channels: Arc<Mutex<HashMap<SwapId, oneshot::Sender<libp2p_comit::frame::Response>>>>,
-
-    #[behaviour(ignore)]
-    pub herc20_states: Arc<herc20::States>,
-    #[behaviour(ignore)]
-    pub halight_states: Arc<halight::States>,
 }
 
 #[derive(Debug, Clone, Copy, thiserror::Error)]
@@ -249,15 +239,14 @@ impl ComitNode {
     pub fn new(
         bitcoin_connector: Arc<bitcoin::Cache<BitcoindConnector>>,
         ethereum_connector: Arc<ethereum::Cache<Web3Connector>>,
-        lnd_connector_params: Option<LndConnectorParams>,
         swap_communication_states: Arc<SwapCommunicationStates>,
         rfc003_alpha_ledger_states: Arc<rfc003::LedgerStates>,
         rfc003_beta_ledger_states: Arc<rfc003::LedgerStates>,
-        herc20_states: Arc<herc20::States>,
-        halight_states: Arc<halight::States>,
         seed: RootSeed,
         db: Sqlite,
         task_executor: Handle,
+        storage: Storage,
+        protocol_spawner: ProtocolSpawner,
     ) -> Result<Self, io::Error> {
         let mut swap_headers = HashSet::new();
         swap_headers.insert("id".into());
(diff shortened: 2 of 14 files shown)


tcharding:
@thomaseizinger in case you see this tonight; I have not done many of the suggested clean ups you made within `start_swap` (e.g. bring variants into scope). I'll play more in the morning.

cnd/src/http_api/routes/index.rs
```diff
@@ -127,20 +124,41 @@ pub struct Body<A, B> {
...
+            erc20_amount: body.alpha.amount,
+            token_contract: body.alpha.contract_address,
             lightning_cltv_expiry: body.beta.cltv_expiry.into(),
             lightning_amount: body.beta.amount.0,
-            token_contract: body.alpha.contract_address,
+        }
+    }
+}
+
+trait ToCreatedSwap<A, B> {
```
thomaseizinger, line 139: Are we using this trait somewhere for an abstraction? If no, then this fn can just go on the `Body` type.

cnd/src/network.rs
```diff
@@ -301,34 +304,27 @@ impl ComitNode {
...
-    }
-
-    fn init_herc20_hbit(
-        &mut self,
-        id: LocalSwapId,
-        swap: CreatedSwap<herc20::CreatedSwap, hbit::CreatedSwap>,
-    ) -> anyhow::Result<()> {
-        self.comit.init_herc20_hbit(id, swap)
-    }
+        let data = match role {
```
thomaseizinger, line 313: We match inside `initiate_communication` as-well right?
Could also just make two functions then it is not redundant :)

cnd/src/network.rs
```diff
@@ -868,130 +843,38 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<()> for ComitNode {
...
-                    self.herc20_states.clone(),
-                    self.ethereum_connector.clone(),
-                ));
+        let comit::BehaviourOutEvent {
+            local_swap_id,
+            remote_data,
+        } = event;
+
+        let storage = self.storage.clone();
+        // FIXME: These are in  storage but not accessible?
```
thomaseizinger, line 852: Help is on the way: https://github.com/comit-network/comit-rs/blob/respawn-swaps/cnd/src/protocol_spawner.rs

cnd/src/network/comit.rs
```diff
@@ -540,151 +514,225 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+    }
+}
+
+impl Set<SecretHash> for RemoteData {
+    fn set(&mut self, value: SecretHash) {
+        self.secret_hash = Some(value);
+    }
+}
+
+// #[cfg(test)]
```
thomaseizinger, line 610: Would be good to keep that one :)

cnd/src/storage.rs
```diff
@@ -39,6 +41,40 @@ impl Storage {
             halight_states,
         }
     }
+
+    pub async fn save_what_alice_learned_from_bob(
```
thomaseizinger, line 45: What do you need these for if you have the trait impl?

comit/src/lib.rs
```diff
@@ -27,6 +27,13 @@ pub use self::{
 };
 use digest::ToDigestInput;
 
+#[derive(Clone, Copy, Debug)]
+pub struct Swap {
```
thomaseizinger, line 31: I would refrain from adding things like these to the comit lib for now. If it turns out that we really need something like these in all the places, then we can always add it. But now is too early IMO.

cnd/src/http_api/routes/index.rs
tcharding: This will be used as soon as we add the other direction for this trading pair.  Since touching `index.rs` is already dangerously close to out of scope for this PR not adding this now.
```

impl ToCreatedSwap<halight::CreatedSwap, herc20::CreatedSwap>
    for Body<HalightLightningBitcoin, Herc20EthereumErc20>
{
    fn to_created_swap(
        &self,
        swap_id: LocalSwapId,
    ) -> CreatedSwap<halight::CreatedSwap, herc20::CreatedSwap> {
        let body = self.clone();

        let alpha = halight::CreatedSwap::from(body.alpha);
        let beta = herc20::CreatedSwap::from(body.beta);

        CreatedSwap {
            swap_id,
            alpha,
            beta,
            peer: body.peer.into(),
            address_hint: None,
            role: body.role.0,
        }
    }
}
```

cnd/src/network.rs
tcharding: Nice idea, implemented as suggested.

cnd/src/network.rs
tcharding: Link is broken

cnd/src/storage.rs
tcharding: I wanted to add helper functions, its either add them here as `Storage` methods or pass in an additional parameter `storage`, methods is nicer.

comit/src/lib.rs
tcharding: This is actually what got me out of bed to hack on this now, I've moved this new type to the db module. It is foreseeable that the `http_api::Swap` could be replaced now with this swap since this is arguably the top level view of a swap from the databases perspective and since this type is used solely to get the role and swap protocol instance (i.e. `Herc20Halight).

cnd/src/network.rs
tcharding: Oh my, that is nice. I have used it in this PR now :)

cnd/src/storage.rs
tcharding: Turns out these are no good anyways once we add the other trading pairs. Removed.

cnd/src/http_api/routes/index.rs
thomaseizinger: How is this different from an `impl Body<halight::CreatedSwap, ...>`?

comit/src/lib.rs
thomaseizinger: Happy to add it at the DB level.

cnd/src/http_api/routes/index.rs
tcharding: Oh I did not catch that idea fully when you first suggested it, will implement as above.

cnd/src/network/start_swap.rs
```diff
@@ -0,0 +1,90 @@
...
+                    local_swap_id: id,
+                    data: WhatAliceLearnedFromBob {
+                        redeem_ethereum_identity: ethereum_identity,
+                        refund_lightning_identity: lightning_identity,
+                    },
+                })
+                .await?;
+
+            let herc20_params: herc20::Params = storage.load(id).await?;
+            let halight_params: halight::Params = storage.load(id).await?;
```
thomaseizinger, line 52: Over in `respawn`, I am loading these in one call. Not sure what the better approach is.

As we can see here, doing them separately might allow us to create an API where you just pass `id`, load everything else from the DB and spawn. That could be a useful abstraction on top of `ProtocolSpawner`. Let's keep the idea in mind :)

cnd/src/protocol_spawner.rs
```diff
@@ -49,6 +52,27 @@ impl ProtocolSpawner {
             halight_states,
         }
     }
+
+    pub fn supports_halight(&self) -> anyhow::Result<()> {
```
thomaseizinger, line 56: This is done on line 110 already so we can delete it here.

cnd/src/storage.rs
```diff
@@ -292,3 +298,122 @@ impl Load<http_api::BobHerc20HalightBitcoinSwap> for Storage {
...
+        unimplemented!()
+    }
+}
+
+#[async_trait::async_trait]
+impl Load<herc20::Params> for Storage {
+    async fn load(&self, _swap_id: LocalSwapId) -> anyhow::Result<herc20::Params> {
+        unimplemented!()
+    }
+}
```
thomaseizinger, line 364: I think this is the reason for your e2e test failure.

cnd/src/protocol_spawner.rs
tcharding: Done, thanks.

cnd/src/storage.rs
tcharding: Awesome, thanks for finding this for me.

cnd/src/network/start_swap.rs
tcharding: Copied the code as suggested.

cnd/src/http_api/routes/index.rs
```diff
@@ -66,28 +65,26 @@ pub async fn post_herc20_halight_bitcoin(
...
-    let swap = CreatedSwap::<herc20::CreatedSwap, halight::CreatedSwap> {
-        swap_id,
-        alpha: body.alpha.into(),
-        beta: body.beta.into(),
-        peer: body.peer.into(),
-        address_hint: None,
-        role: body.role.0,
-    };
-
+    let swap = body.to_created_swap(swap_id);
```
D4nte, line 71: `ToCreatedSwap` trait is defined in this file. I understand it allows the conversion to a `db::CreatedSwap<A,B>` struct that can be saved in the db.

At first read, it is not clear why a `Body` is converted to a `CreatedSwap` using a local trait. Maybe keeping the `db` module in the usage path would help clarity: `db::CreatedSwap`.

Also, I am not sure the trait is necessary here, to me it just adds indirection. Instead, you can simply add a `fn to_create_swap` to `impl Block<Herc..,Halight..>` because `Body<A,B>` is defined in the same file (and hence you can have a `impl` block for it here).

cnd/src/network/comit.rs
```diff
@@ -66,14 +61,10 @@ pub struct Comit {
     #[behaviour(ignore)]
     swaps: Swaps<ReplySubstream<NegotiatedSubstream>>,
     #[behaviour(ignore)]
-    ethereum_identities: HashMap<SharedSwapId, identity::Ethereum>,
-    #[behaviour(ignore)]
-    lightning_identities: HashMap<SharedSwapId, identity::Lightning>,
+    remote_data: HashMap<SharedSwapId, RemoteData>,
     #[behaviour(ignore)]
+    // FIXME: Use plural for this hashmap field.
```
D4nte, line 66: Do it now?

cnd/src/network/comit.rs
```diff
@@ -97,143 +88,109 @@ impl Comit {
...
         // Communicate
-        self.ethereum_identity.send(
-            peer.clone(),
-            ethereum_identity::Message::new(shared_swap_id, create_swap_params.ethereum_identity),
-        );
-        self.lightning_identity.send(
-            peer,
-            lightning_identity::Message::new(shared_swap_id, create_swap_params.lightning_identity),
-        );
+        if let Some(ethereum_identity) = data.ethereum_identity {
```
D4nte, line 166: I think this trick is awesome 😃

cnd/src/http_api/routes/index.rs
tcharding: Implemented already, GitHub fail by me, Thomas and I had this conversation and I marked it as resolved before force pushing. Sorry for wasting your time.

cnd/src/network/comit.rs
```diff
@@ -540,90 +512,111 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+    // Currently we only support herc20/halight swap, as we add protocol
+    // combinations the identities will be added here.
+    pub secret_hash: Option<SecretHash>,
+    pub shared_swap_id: Option<SharedSwapId>,
+    pub ethereum_identity: Option<identity::Ethereum>,
+    pub lightning_identity: Option<identity::Lightning>,
+}
+
+impl LocalData {
+    pub fn for_alice(secret_hash: SecretHash, identities: Identities) -> Self {
```
D4nte, line 547: I am not convinced this is useful as I soon as you add a new swap (bitcoin onchain) then you have to create many of those or just trash it.
I'd recommend to remove these two functions and directly construct the `LocalData` struct where needed (thanks to the `pub` keywords on the fields).

cnd/src/network/comit.rs
tcharding: Bollocks, sloppy work Tobin. Thanks man.

cnd/src/network/comit.rs
tcharding: Sick huh, this is the key change that has made the **comit** module.

cnd/src/network/comit.rs
tcharding: I disagree, I added the next trading pair already (#2699) and these methods worked nicely. The make the code read nicely in my opinion.

tcharding: Can remove it if you feel super strongly though.

cnd/src/network/comit.rs
```diff
@@ -254,6 +211,49 @@ impl Comit {
...
+            .cloned()
+            .unwrap_or_default();
+        remote_data.set(value);
+        self.remote_data.insert(shared_swap_id, remote_data);
+    }
+
+    fn finalize(&mut self, peer: PeerId, shared_swap_id: SharedSwapId) {
+        let local_swap_id = match self.swaps.get_local_swap_id(shared_swap_id) {
+            Some(id) => id,
+            None => return,
```
luckysori, line 230: :question: Should we log something if we try to finalize a swap and we don't have a reference to it?

cnd/src/network/comit.rs
tcharding: Done.

cnd/src/network/comit.rs
```diff
@@ -254,6 +211,49 @@ impl Comit {
...
+        let mut remote_data = self
+            .remote_data
+            .get(&shared_swap_id)
+            .cloned()
+            .unwrap_or_default();
+        remote_data.set(value);
+        self.remote_data.insert(shared_swap_id, remote_data);
+    }
+
+    fn finalize(&mut self, peer: PeerId, shared_swap_id: SharedSwapId) {
```
luckysori, line 227: :thinking: This function is pretty tricky to understand for me.

cnd/src/network/comit.rs
```diff
@@ -540,90 +512,111 @@ impl NetworkBehaviourEventProcess<oneshot_behaviour::OutEvent<finalize::Message>
...
+                remote_data,
             });
         }
     }
 }
 
+/// All possible data to be sent to the remote node for any protocol
+/// combination.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct LocalData {
```
luckysori, line 536: :question: Do you expect this to eventually be parametrized by alpha identity and beta identity?

cnd/src/network/comit/swaps.rs
```diff
@@ -306,265 +316,58 @@ impl Default for Swaps<()> {
...
-
-    fn create_params() -> Herc20HalightBitcoinCreateSwapParams {
-        Herc20HalightBitcoinCreateSwapParams {
-            role: Role::Alice,
-            peer: DialInformation {
-                peer_id: PeerId::random(),
-                address_hint: None,
-            },
-            ethereum_identity: identity::Ethereum::random(),
+    // FIXME: Usage of this function relies on the fact that token_contract is
```
D4nte: Feel free to remove the comment without actioning.

cnd/src/network/comit.rs
D4nte: I can't find your usage of `for_alice` in the new PR. Can you point me to a line or commit please?

cnd/src/network/start_swap.rs
```diff
@@ -0,0 +1,89 @@
...
+    swap_protocols::{halight, herc20, LocalSwapId},
+};
+use ::comit::{
+    network::{WhatAliceLearnedFromBob, WhatBobLearnedFromAlice},
+    Protocol, Role, Side,
+};
+use chrono::offset::Utc;
+
+#[allow(clippy::too_many_arguments)]
+pub async fn start_swap(
```
D4nte, line 15: The missing piece!

cnd/src/network/start_swap.rs
```diff
@@ -0,0 +1,89 @@
...
+                })
+                .await?;
+
+            let herc20_params: herc20::Params = storage.load(id).await?;
+            let halight_params: halight::Params = storage.load(id).await?;
+
+            spawner.spawn(id, herc20_params, start_of_swap, Side::Alpha, role);
+            spawner.spawn(id, halight_params, start_of_swap, Side::Beta, role);
+        }
+        _ => todo!("handle the rest"),
```
D4nte, line 85: what rest?

cnd/src/network/comit/swaps.rs
```diff
@@ -306,265 +316,58 @@ impl Default for Swaps<()> {
...
-        assert_eq!(shared_swap_id, stored_shared_swap_id);
-        assert_eq!(create_params, stored_create_params);
-
-        let (stored_local_swap_id, stored_create_params) =
-            swaps.finalize_swap(&shared_swap_id).unwrap();
-
-        assert_eq!(local_swap_id, stored_local_swap_id);
-        assert_eq!(create_params, stored_create_params);
-        assert!(!swaps.swap_in_pending_hashmaps(&digest));
+    // FIXME: The same applies here as for digest() re property based testing.
```
luckysori: :wrench: You'll have to remove this or do it :stuck_out_tongue:

cnd/src/network/comit.rs
tcharding:
```
cnd/src/network.rs:303: let data = LocalData::for_alice(secret_hash, identities);
```

cnd/src/network/comit/swaps.rs
tcharding: Cool, I did that already :)

cnd/src/network/start_swap.rs
tcharding: Yep, fixing that up is todays task when this merges.

cnd/src/network/start_swap.rs
tcharding: This function will include all 4 supported swap protocols, currently its only herc20<->halight

cnd/src/network.rs
```diff
@@ -301,75 +289,41 @@ impl ComitNode {
...
-        &mut self,
-        id: LocalSwapId,
-    ) -> Option<Herc20HalightBitcoinCreateSwapParams> {
-        self.comit.get_created_swap(&id)
-    }
+        match role {
+            Role::Alice => {
+                let swap_seed = self.seed.derive_swap_seed(local_swap_id);
+                let secret = swap_seed.derive_secret();
+                let secret_hash = SecretHash::new(secret);
```
da-kami, line 302: 🙃  I think `secret.hash()` should be sufficient here.

cnd/src/network.rs
```diff
@@ -868,130 +822,20 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<()> for ComitNode {
...
-        }
+        let comit::BehaviourOutEvent {
+            local_swap_id,
+            remote_data,
+        } = event;
+
+        let storage = self.storage.clone();
+        let spawner = self.protocol_spawner.clone();
+
+        self.task_executor.spawn(start_swap::start_swap(
```
da-kami, line 833: this is sweet!

cnd/src/network/comit.rs
```diff
@@ -254,6 +210,49 @@ impl Comit {
...
+            Some(id) => id,
+            None => return,
+        };
+
+        let state = self.communication_states.get(&shared_swap_id);
+        let data = self.swaps.get_local_data(&local_swap_id);
+        let remote_data = self.remote_data.get(&shared_swap_id);
+
+        if let (Some(state), Some(data), Some(remote_data)) = (state, data, remote_data) {
+            let ethereum_sorted = if data.ethereum_identity.is_some() {
```
da-kami, line 238: ❓ I don't really understand the `_sorted` naming - what is sorted and sorted by what?

cnd/src/network/start_swap.rs
da-kami, line 85:
```suggestion
        _ => todo!("handle other protocol combinations"),
```

cnd/src/network.rs
thomaseizinger: That function no longer exists because we wanted to make `Secret` independent of `SecretHash`.

cnd/src/network/comit.rs
thomaseizinger: This "sorted" means "it is done", like "We've got this sorted" :)

Doesn't have anything to do with sorting things in a list.

cnd/src/network/start_swap.rs
```diff
@@ -0,0 +1,89 @@
...
+    storage::{Load, Storage},
+    swap_protocols::{halight, herc20, LocalSwapId},
+};
+use ::comit::{
+    network::{WhatAliceLearnedFromBob, WhatBobLearnedFromAlice},
+    Protocol, Role, Side,
+};
+use chrono::offset::Utc;
+
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
tcharding: Nice catch, that's out of date. Will remove.

cnd/src/network/comit.rs
tcharding: No, that's the big win with this PR - the comit module no longer knows about alpa and beta.

cnd/src/network/comit.rs
tcharding: I balked at this name also, this justifies a ticket IMO (or even an inline `TODO` ;)

cnd/src/network/start_swap.rs
tcharding: Thanks, this is going away anyways within a day or two if I have anything to do with it :)

tcharding:
Thanks for the reviews everyone.

cnd/src/protocol_spawner.rs
```diff
@@ -51,6 +51,18 @@ impl ProtocolSpawner {
     }
 }
 
+impl Clone for ProtocolSpawner {
```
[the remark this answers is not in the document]
tcharding: This is an artefact of earlier (erroneous) development also, thanks.

cnd/src/network/comit.rs
luckysori: Hmm, that's interesting. Why do you think it's better for it to know about optional concrete types rather than protocols?

There seems to be consensus around this being a positive change, but I'm always scared of structs with lots of optional fields. Maybe I don't understand the pain points of the other approach.

cnd/src/network/comit.rs
tcharding: This function is called so many times, right now I'm not keen to add any tracing to it. May be worth further investigation (especially with your comment on this function below :)

cnd/src/network/comit.rs
tcharding: I agree, I also am not a fan of this style of code. I intentionally did not put time into simplifying this for this PR. Will do it when I get some time allocated to clean up this module, adding it to my TODO list.

cnd/src/network.rs
da-kami: Okk, I guess I missed that 😅

cnd/src/network/comit.rs
da-kami: I'd propose `_received`, because I think that is the point - that we received information (identities, secret-hash...) from the other party.
Happy to do it in a follow up, if you tackle it let me know.

cnd/src/network/comit.rs
tcharding: Its because now `Comit` just handles `Option<identity::Lightning>`. by 'handles' we mean 'if this is Some value then send it to the peer'. `Comit` never needs to know _why_ there is a lightning identity i.e., `Comit` does not need extending for any protocol that includes a lightning identity. If its there we send it, if not we don't. Period.

In other words, once we add `Option<identity::hbit>` to the `Identities` struct then `Comit` can handle any combination of the three protocols without specifically ever knowing which protocol it is handling, this simplifies the code a whole bunch.

tcharding: `Comit` here refers to the comit network behaviour

luckysori: It does help! It makes sense now :)

cnd/src/network.rs
```diff
@@ -301,75 +289,58 @@ impl ComitNode {
...
-        self.comit.init_herc20_hbit(id, swap)
-    }
-
-    pub fn get_created_swap(
-        &mut self,
-        id: LocalSwapId,
-    ) -> Option<Herc20HalightBitcoinCreateSwapParams> {
-        self.comit.get_created_swap(&id)
-    }
+        // At this stage we do not know if the arguments passed to us make up a
```
D4nte, line 298: Why not trusting the controller?

tcharding: Because we make no assertions on the data received by the controller from the user, we _definitely_ don't trust the user do we?

D4nte: Shouldn't the controller do the assertion?

PR #2692, opened 2020-05-22
Random swap prefix

### LINKED ISSUE #2624: Digest macro uses arbitrary prefixes
## Problem 

Currently we use arbitrary prefixes for fields in the digestable struct
```
#[derive(Clone, Digest, Debug, PartialEq)]
#[digest(hash = "SwapDigest")]
pub struct HanEtherereumHalightBitcoinCreateSwapParams {
	...
    #[digest(prefix = "2001")]
    #[digest(prefix = "2002")]
    pub ethereum_amount: asset::Ether,
	...
```

## Goal

Make sense of these prefixes, also design what they should be as we add more structs.  Also of note do we require a digestable struct for each swap or just a single one for a portocol pair (since the fields are the same)?

bonomat:
resolves #2624

This PR changes the prefix to random values.
I chose 2 bytes length which would give us 65536 combinations. That should last for a while :)

image

just kidding, I went with: https://www.random.org/cgi-bin/randbyte?nbytes=2&format=h

```
#[digest(prefix = "99ef")]
pub ethereum_expiry: Timestamp,

#[digest(prefix = "f0ec")]
pub erc20_amount: asset::Erc20Quantity, 

#[digest(prefix = "944a")]

#[digest(prefix = "e6ca")]
pub bitcoin_expiry: Timestamp,

#[digest(prefix = "052d")]
pub bitcoin_amount: asset::Bitcoin,

 #[digest(prefix = "d07f")]
pub lightning_cltv_expiry: RelativeTime,

#[digest(prefix = "1475")]
pub lightning_amount: asset::Bitcoin,
```

eventually we should document these values in a specification. 

----- 
xkcd-source https://imgs.xkcd.com/comics/random_number.png


bonomat:
will re-open later as I've learned something new based on this comment: https://github.com/comit-network/comit-rs/issues/2624

PR #2697, opened 2020-05-22
Fix Swap rename

tcharding:
Recently we re-named the old `Swap` to `Rfc003Swap`, a call to the
constructor slipped passed me (tcharding) and the compiler some how.

Use the correct `Rfc003Swap::new()` constructor.

### DIFF
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -626,7 +626,7 @@ where
 {
     let id = swap_request.swap_id;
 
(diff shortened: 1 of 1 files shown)


tcharding:
> Huh? How did this ever work? :D

Funny huh, I'm guessing there is a Swap type in the same file with the same methods/fields that this one uses so this slipped past. We'll have to move to an even more tightly typed language if this happens too many times ;)

tcharding:
This change is in dev already.

PR #2698, opened 2020-05-22
Pass Storage to respawn

tcharding:
We do not need access to the `Facade` when respawning, `Storage` is a
top level type, no need to call through to it via the facade.

Pass `Storage` directly to the `respawn()` function instead of
delegating to the `Facade`.


thomaseizinger:
Thanks!
I have a feeling we either don't want to have a facade at all or have a scope-specific facade (i.e. one for http-api, one for respawn, one for network layer).

tcharding:
This is trivial, merging with a single review approval.

PR #2718, opened 2020-05-26
Several small improvements in the network module

thomaseizinger:


cnd/src/http_api/problem.rs
```diff
@@ -54,11 +53,6 @@ pub fn from_anyhow(e: anyhow::Error) -> HttpApiProblem {
         return HttpApiProblem::new("Swap not found.").set_status(StatusCode::NOT_FOUND);
     }
 
-    if e.downcast_ref::<SwapExists>().is_some() {
```
D4nte, line 57: Are we not sending an error specific to existing swap anymore?
Can we please re-introduce the functionality?

cnd/src/network/comit/swaps.rs
```diff
@@ -21,10 +21,12 @@ pub enum Error {
...
 
-/// Tracks the state of a swap in the communication phase, T is
-/// ReplySubstream<NegotiatedSubstream>
+/// Tracks the state of a swap in the communication phase.
+///
+/// We use a type parameter for the substream type so we can write unit tests
+/// without creating actual substreams.
 #[derive(Debug)]
-pub struct Swaps<T> {
+pub struct Swaps<T = ReplySubstream<NegotiatedSubstream>> {
```
D4nte, line 29: Nice trick.

cnd/src/http_api/problem.rs
thomaseizinger: The error struct was never used outside of this line so I deleted it. Meaning there is no change in functionality with this PR :)

I am not sure this PR is the best place to introduce this functionality again.

I am gonna re-open https://github.com/comit-network/comit-rs/issues/2490.

PR #2721, opened 2020-05-26
Reduce struct name and use fully qualified name `http_api::{protocol}::Finalized`

da-kami:
Import by fully qualified name (due to same protocol module names)

This is for you @tcharding - I have to say I am not really a fan of it. Due to having to use the fully qualified name most lines break. I don't think this patch really helps, but if the others feel strong about it we can merge it.

### DIFF
--- a/api_tests/src/actors/actor.ts
+++ b/api_tests/src/actors/actor.ts
@@ -1,5 +1,6 @@
 import {
     Cnd,
+    HalightLightningBitcoinHerc20EthereumErc20RequestBody,
     Herc20EthereumErc20HalightLightningBitcoinRequestBody,
     LedgerAction,
     Swap,
@@ -97,10 +98,10 @@ export class Actor {
      */
 
     /**
-     * Create a Swap
+     * Create a herc20-halight Swap
      * @param createSwapPayload
      */
-    public async createSwap(
+    public async createHerc20HalightSwap(
         createSwapPayload: Herc20EthereumErc20HalightLightningBitcoinRequestBody
     ) {
         this.alphaLedger = {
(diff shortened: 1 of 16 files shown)


tcharding:
> Import by fully qualified name (due to same protocol module names)
> 
> This is for you @tcharding - I have to say I am not really a fan of it. Due to having to use the fully qualified name most lines break. I don't think this patch really helps, but if the others feel strong about it we can merge it.

You are right @da-kami these qualified paths are horrible. I think this is something we need to resolve well because it is going to come up repeatedly now that we are moving to adding types in the modules that use them instead of having god types (implying that there is going to be other places that qualifying paths does not work).

Should we consider _not_ having protocol modules anywhere except `swap_protocols/`. @thomaseizinger do you have any ideas on how to solve this cleanly?

thomaseizinger:
> > Import by fully qualified name (due to same protocol module names)
> > This is for you @tcharding - I have to say I am not really a fan of it. Due to having to use the fully qualified name most lines break. I don't think this patch really helps, but if the others feel strong about it we can merge it.
> 
> You are right @da-kami these qualified paths are horrible. I think this is something we need to resolve well because it is going to come up repeatedly now that we are moving to adding types in the modules that use them instead of having god types (implying that there is going to be other places that qualifying paths does not work).
> 
> Should we consider _not_ having protocol modules anywhere except `swap_protocols/`. @thomaseizinger do you have any ideas on how to solve this cleanly?

One way of solving this would be to have protocol modules that are local to a given context and reexport the generic one from there. This way you can sort of "enhance" a module with more types and don't have to mess with long import paths.

Basically, put a `pub use cnd::swap_protocols::halight::*;` into the `cnd::http_api::halight` module.

da-kami:
> > > Import by fully qualified name (due to same protocol module names)
> > > This is for you @tcharding - I have to say I am not really a fan of it. Due to having to use the fully qualified name most lines break. I don't think this patch really helps, but if the others feel strong about it we can merge it.
> > 
> > 
> > You are right @da-kami these qualified paths are horrible. I think this is something we need to resolve well because it is going to come up repeatedly now that we are moving to adding types in the modules that use them instead of having god types (implying that there is going to be other places that qualifying paths does not work).
> > Should we consider _not_ having protocol modules anywhere except `swap_protocols/`. @thomaseizinger do you have any ideas on how to solve this cleanly?
> 
> One way of solving this would be to have protocol modules that are local to a given context and reexport the generic one from there. This way you can sort of "enhance" a module with more types and don't have to mess with long import paths.
> 
> Basically, put a `pub use cnd::swap_protocols::halight::*;` into the `cnd::http_api::halight` module.

But this would only partly solve the problem, because the `Finalized` struct would still have the same name. I am not sure they actually should have. They only do because I moved them into their respective module (which contains nothing else...). Probably better to just leave them protocol specific and move them into `protocol.rs`?

da-kami:
> > > Import by fully qualified name (due to same protocol module names)
> > > This is for you @tcharding - I have to say I am not really a fan of it. Due to having to use the fully qualified name most lines break. I don't think this patch really helps, but if the others feel strong about it we can merge it.
> > 
> > 
> > You are right @da-kami these qualified paths are horrible. I think this is something we need to resolve well because it is going to come up repeatedly now that we are moving to adding types in the modules that use them instead of having god types (implying that there is going to be other places that qualifying paths does not work).
> > Should we consider _not_ having protocol modules anywhere except `swap_protocols/`. @thomaseizinger do you have any ideas on how to solve this cleanly?
> 
> One way of solving this would be to have protocol modules that are local to a given context and reexport the generic one from there. This way you can sort of "enhance" a module with more types and don't have to mess with long import paths.
> 
> Basically, put a `pub use cnd::swap_protocols::halight::*;` into the `cnd::http_api::halight` module.

Thought about re-exporting again and actually like it. I think the tough part will be to keep doing it. Is there a simple way to avoid importing it from `swap_protocols` directly (enforce the import from the re-exported one in `http_api`)?

thomaseizinger:
> > > > Import by fully qualified name (due to same protocol module names)
> > > > This is for you @tcharding - I have to say I am not really a fan of it. Due to having to use the fully qualified name most lines break. I don't think this patch really helps, but if the others feel strong about it we can merge it.
> > > 
> > > 
> > > You are right @da-kami these qualified paths are horrible. I think this is something we need to resolve well because it is going to come up repeatedly now that we are moving to adding types in the modules that use them instead of having god types (implying that there is going to be other places that qualifying paths does not work).
> > > Should we consider _not_ having protocol modules anywhere except `swap_protocols/`. @thomaseizinger do you have any ideas on how to solve this cleanly?
> > 
> > 
> > One way of solving this would be to have protocol modules that are local to a given context and reexport the generic one from there. This way you can sort of "enhance" a module with more types and don't have to mess with long import paths.
> > Basically, put a `pub use cnd::swap_protocols::halight::*;` into the `cnd::http_api::halight` module.
> 
> Thought about re-exporting again and actually like it. I think the tough part will be to keep doing it. Is there a simple way to avoid importing it from `swap_protocols` directly (enforce the import from the re-exported one in `http_api`)?

Well, you are gonna get a compile error if you try to import both modules, `swap_protocols::halight` and `http_api::halight`.
I would say for now, lets try to keep an eye on it during review and while coding.

If we find ourselves violating these rules over and over, we can look into automated solutions like custom lints (if something like that exists).

da-kami:
will create new one

PR #2723, opened 2020-05-26
2624 random prefix

### LINKED ISSUE #2624: Digest macro uses arbitrary prefixes

bonomat:
This PR changes the prefix to random values.
I've choses 2 bytes length which would give us 65536 combinations. That should last for a while :)

image

just kidding, I went with: https://www.random.org/cgi-bin/randbyte?nbytes=2&format=h

There are some TODOs in the code, hence the draft PR. 

@thomaseizinger / @tcharding : please let me know what you think

Resolves #2624 

----- 
xkcd-source https://imgs.xkcd.com/comics/random_number.png

### DIFF
--- a/cnd/src/network/comit/swaps.rs
+++ b/cnd/src/network/comit/swaps.rs
@@ -310,9 +310,12 @@ mod tests {
     use crate::{
         asset::{self, ethereum::FromWei},
         identity,
-        network::LocalData,
+        network::comit::LocalData,
     };
-    use ::comit::network::{protocols::announce::SwapDigest, swap_digest::Herc20Halight};
+    use ::comit::network::{
+        protocols::announce::SwapDigest,
+        swap_digest::{Herc20Halight, SwapProtocol},
+    };
     use digest::Digest;
 
     // Usage of this function relies on the fact that token_contract is
@@ -324,6 +327,8 @@ mod tests {
             token_contract: identity::Ethereum::random(),
             lightning_cltv_expiry: 12345.into(),
             lightning_amount: asset::Bitcoin::from_sat(1_000_000_000),
+            alpha_protocol: SwapProtocol::Herc20,
+            beta_protocol: SwapProtocol::Halight,
         }
         .digest()
     }
--- a/cnd/src/network/comit/swaps.rs
+++ b/cnd/src/network/comit/swaps.rs
@@ -319,7 +319,7 @@ mod tests {
     // random. We should use property based testing.
     fn digest() -> SwapDigest {
         Herc20Halight {
-            ethereum_absolute_expiry: 12345.into(),
+            ethereum_expiry: 12345.into(),
             erc20_amount: asset::Erc20Quantity::from_wei(9_001_000_000_000_000_000_000u128),
             token_contract: identity::Ethereum::random(),
             lightning_cltv_expiry: 12345.into(),
--- a/cnd/src/swap_protocols/facade.rs
+++ b/cnd/src/swap_protocols/facade.rs
@@ -5,9 +5,10 @@ use crate::{
     network::{DialInformation, Identities, Swarm},
     storage::{Load, LoadAll, Storage},
     swap_protocols::{hbit, herc20, LocalSwapId, Role},
-    timestamp::Timestamp,
+    timestamp::{RelativeTime, Timestamp},
 };
 use ::comit::network::protocols::announce::SwapDigest;
+use comit::network::swap_digest::SwapProtocol;
 use digest::Digest;
 
 /// This represents the information available on a swap
@@ -33,6 +34,10 @@ pub struct HbitHerc20SwapParams {
     pub erc20_amount: asset::Erc20Quantity,
     #[digest(prefix = "3003")]
     pub token_contract: identity::Ethereum,
+    #[digest(ignore)]
+    pub alpha_protocol: SwapProtocol,
+    #[digest(ignore)]
+    pub beta_protocol: SwapProtocol,
 }
 
 impl From<CreatedSwap<hbit::CreatedSwap, herc20::CreatedSwap>> for HbitHerc20SwapParams {
@@ -52,6 +57,8 @@ impl From<CreatedSwap<hbit::CreatedSwap, herc20::CreatedSwap>> for HbitHerc20Swa
             ethereum_expiry: swap.beta.absolute_expiry.into(),
             erc20_amount: swap.beta.asset.quantity,
             token_contract: swap.beta.asset.token_contract,
+            alpha_protocol: SwapProtocol::Hbit,
+            beta_protocol: SwapProtocol::Herc20,
         }
     }
 }
@@ -77,6 +84,10 @@ pub struct Herc20HbitSwapParams {
     pub bitcoin_expiry: Timestamp,
     #[digest(prefix = "3002")]
     pub bitcoin_amount: asset::Bitcoin,
+    #[digest(ignore)]
+    pub alpha_protocol: SwapProtocol,
+    #[digest(ignore)]
+    pub beta_protocol: SwapProtocol,
 }
 
 impl From<CreatedSwap<herc20::CreatedSwap, hbit::CreatedSwap>> for Herc20HbitSwapParams {
@@ -96,6 +107,8 @@ impl From<CreatedSwap<herc20::CreatedSwap, hbit::CreatedSwap>> for Herc20HbitSwa
             bitcoin_identity: swap.beta.identity,
             bitcoin_expiry: swap.beta.absolute_expiry.into(),
             bitcoin_amount: swap.beta.amount,
+            alpha_protocol: SwapProtocol::Herc20,
+            beta_protocol: SwapProtocol::Hbit,
         }
     }
 }
--- a/cnd/src/swap_protocols/facade.rs
+++ b/cnd/src/swap_protocols/facade.rs
@@ -22,21 +22,21 @@ pub struct HbitHerc20SwapParams {
     pub peer: DialInformation,
     #[digest(ignore)]
     pub bitcoin_identity: identity::Bitcoin,
-    #[digest(prefix = "2001")]
+    #[digest(prefix = "e6ca")]
     pub bitcoin_expiry: Timestamp,
-    #[digest(prefix = "2002")]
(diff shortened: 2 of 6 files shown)


cnd/src/network/comit/swaps.rs
```diff
@@ -310,20 +310,25 @@ mod tests {
     use crate::{
         asset::{self, ethereum::FromWei},
         identity,
-        network::LocalData,
+        network::comit::LocalData,
```
tcharding, line 313: This change is unnecessary, network.rs exports everything in the comit module (`pub use comit::*`).

cnd/src/swap_protocols/facade.rs
```diff
@@ -5,9 +5,10 @@ use crate::{
     network::{DialInformation, Identities, Swarm},
     storage::{Load, LoadAll, Storage},
     swap_protocols::{hbit, herc20, LocalSwapId, Role},
-    timestamp::Timestamp,
+    timestamp::{RelativeTime, Timestamp},
 };
 use ::comit::network::protocols::announce::SwapDigest;
+use comit::network::swap_digest::SwapProtocol;
```
tcharding, line 11:
```suggestion
use comit::Protocol;
```

cnd/src/swap_protocols/facade.rs
```diff
@@ -21,18 +22,22 @@ pub struct HbitHerc20SwapParams {
...
+    #[digest(prefix = "f0ec")]
     pub erc20_amount: asset::Erc20Quantity,
-    #[digest(prefix = "3003")]
+    #[digest(prefix = "944a")]
     pub token_contract: identity::Ethereum,
+    #[digest(prefix = "41f4")]
+    pub alpha_protocol: SwapProtocol,
+    #[digest(prefix = "be74")]
+    pub beta_protocol: SwapProtocol,
 }
```
tcharding, line 41: I wouldn't touch the facade for this PR. The facade is creating two separate types where the name implies alpha/beta. With this PR these will be wrong and should be totally removed (and they should be over in `swap_digest.rs`, that change is in a PR at the moment.

comit/src/network/swap_digest.rs
```diff
@@ -1,19 +1,42 @@
...
+#[digest(hash = "SwapDigest")]
+pub enum SwapProtocol {
+    #[digest(prefix = "4b17")]
+    Hbit,
+    #[digest(prefix = "e5ec")]
+    Herc20,
+    #[digest(prefix = "c3d3")]
+    Halight,
+}
+
```
tcharding, line 15: We do not need this new type, we have a `Protocol` enum in `comit/src/lib.rs`.  You just need to add to that file
```
impl ToDigestInput for Protocol {
    fn to_digest_input(&self) -> Vec<u8> {
        let s = format!("{}", self);
        s.into_bytes()
    }
}
```
or some cleaner version of that.

comit/src/network/swap_digest.rs
```diff
@@ -1,19 +1,42 @@
...
+// TODO why is this necessary?
+impl ToDigestInput for SwapProtocol {
+    fn to_digest_input(&self) -> Vec<u8> {
+        self.clone().digest().to_digest_input()
+    }
+}
```
tcharding, line 22: You can remove this in favour of the suggestion above and _do not_ derive digest on the Protocol enum.

thomaseizinger: Yeah, `Digest` should only be derived on the root struct. Otherwise you are doing the algorithm several times, nested in each other.

cnd/src/swap_protocols/facade.rs
```diff
@@ -21,18 +22,22 @@ pub struct HbitHerc20SwapParams {
     pub peer: DialInformation,
     #[digest(ignore)]
     pub bitcoin_identity: identity::Bitcoin,
-    #[digest(prefix = "2001")]
+    #[digest(prefix = "e6ca")]
```
thomaseizinger, line 25: Would the macro work with constants @D4nte? Could aid documentation a lot to export this as constants in a single file so we can see, where they are reused.

comit/src/network/swap_digest.rs
```diff
@@ -1,19 +1,42 @@
 use crate::{asset, identity, network::protocols::announce::SwapDigest, RelativeTime, Timestamp};
-use digest::Digest;
+use digest::{Digest, ToDigestInput};
+
+// TODO should we use swap_protocols::SwapProtocol instead?
```
thomaseizinger, line 4: Yes + deriving `Digest` is not meant for leave nodes. You should only use it on the root struct. In other words, the call to `digest` should only happen once for a particular swap.

For enums like this, implement `ToDigestInput` manually and return static bytes for each variant :)

comit/src/network/swap_digest.rs
```diff
@@ -1,19 +1,42 @@
...
+pub enum SwapProtocol {
+    #[digest(prefix = "4b17")]
+    Hbit,
+    #[digest(prefix = "e5ec")]
+    Herc20,
+    #[digest(prefix = "c3d3")]
+    Halight,
+}
+
+// TODO why is this necessary?
```
D4nte, line 16: Because you assigned a prefix to a `SwapProtocol` field line 87 of facade.rs.

Also, we initially decided against the support of nested structs for digest with @thomaseizinger so I guess you hacked around that.

I would strongly suggest to update the digest crate to implement and document a streamlined way to support nested structs.

comit/src/network/swap_digest.rs
D4nte: And then you can remove the TODO line 16 https://github.com/comit-network/comit-rs/pull/2723/files#r430780519

cnd/src/swap_protocols/facade.rs
D4nte: We only support a string literal right now but shouldn't be too hard to change [this](https://github.com/comit-network/comit-rs/blob/80b019cde297d3d2ce88f3253b92678ceeeadf20/digest-macro-derive/src/lib.rs#L168) `if let` with a `match` to support more types.

comit/src/network/swap_digest.rs
thomaseizinger: > I would strongly suggest to update the digest crate to implement and document a streamlined way to support nested structs.

I agree with this but I would like us to first try and use the current version so we get a better grasp of what we actually need.
This usecase for example hints that we might need a 2nd derive for `ToDigestInput` that works on enums.

D4nte: Hum, it already works on enums. Not sure what is missing.

thomaseizinger: To my knowledge, we are only able to derive `Digest`? `ToDigestInput` needs to be implemented manually, no?

comit/src/network/swap_digest.rs
tcharding: `swap_protocols::SwapProtocol` is rfc003 related, I do not believe it should be used.

thomaseizinger: Ups yeah, use the new Protocol enum of course :)

bonomat:
Thanks for your reviews. 
I'm un-assigning myself from this as I won't be able to look into it before next week Tuesday. Please feel free to pick it up.

tcharding:
> Thanks for your reviews.
> I'm un-assigning myself from this as I won't be able to look into it before next week Tuesday. Please feel free to pick it up.

IMO it would be better to wait until hbit/herc20 is done before doing this. Current hbit digest code is based on the old digest code from her20 (that was remove/replaced with `comit/src/network/swap_digest.rs`. Doing this issue now is tricky because its not clear what to touch.

tcharding:
> This PR changes the prefix to random values.

I do not think the use of random prefixes resolves the issue that this PR is working towards resolving. We already have 'random' prefixes, just they are sequential. The 'randomness' is the problem, adding _more_ randomness is making this worse not better IMO.

thomaseizinger:
> > This PR changes the prefix to random values.
> 
> I do not think the use of random prefixes resolves the issue that this PR is working towards resolving. We already have 'random' prefixes, just they are sequential. The 'randomness' is the problem, adding _more_ randomness is making this worse not better IMO.

I think sequential numbers are not considered random :)

The anticipated value of moving to random prefixes is that documenting how to add a new field is easy and hence reduces mental load when maintaining this code: generate a random number and double-check for conflicts. Whereas with the current one, it looks like there is a strategy to it but it also not really thought through?

I would prefer not spending the time to think it through because I don't think there is a value over random prefixes.

tcharding:
Warning: My response contains bike shedding.

> > > This PR changes the prefix to random values.
> > 
> > 
> > I do not think the use of random prefixes resolves the issue that this PR is working towards resolving. We already have 'random' prefixes, just they are sequential. The 'randomness' is the problem, adding _more_ randomness is making this worse not better IMO.
> 
> I think sequential numbers are not considered random :)

What I meant was they are functionally random, they have no specific meaning.

> The anticipated value of moving to random prefixes is that documenting how to add a new field is easy and hence reduces mental load when maintaining this code: generate a random number and double-check for conflicts. Whereas with the current one, it looks like there is a strategy to it but it also not really thought through?

I can add that comment now and the code needs no change

// 'prefix' is a pseudo-random string of ASCII characters, no single struct may contain duplicate prefixes.

Our current sequential numbers are 100% pseudo-random, albeit from an incredibly  poor source of randomness :)

> I would prefer not spending the time to think it through because I don't think there is a value over random prefixes.

I agree, I want to close this PR to tidy up so felt obliged to comment before doing so :)

bonomat:
I feel like this needs more discussion. Given the changes are rather simple and can easy be replicated in a new PR I just close this for now.

PR #2728, opened 2020-05-27
Rename struct name to Finalized

da-kami:
Re-export swap_protocols::{protocol} in http_api

Note: Bit weird now, because we always refer to what is in `swap_protocols` module as `http_api::{protocol}` - especially in the `db` I find that a bit awkward (there is no distinction between what is coming from `http_api` and what from the `swap_protocols` now...).

### DIFF
--- a/cnd/src/http_api/halight.rs
+++ b/cnd/src/http_api/halight.rs
@@ -1,11 +1,13 @@
 use crate::swap_protocols::halight;
 use comit::{asset, identity, RelativeTime};
 
+pub use crate::swap_protocols::halight::*;
+
 #[derive(Clone, Copy, Debug)]
-pub struct HalightFinalized {
-    pub halight_asset: asset::Bitcoin,
-    pub halight_refund_identity: identity::Lightning,
-    pub halight_redeem_identity: identity::Lightning,
+pub struct Finalized {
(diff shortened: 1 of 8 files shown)


thomaseizinger:
> Re-export swap_protocols::{protocol} in http_api
> 
> Note: Bit weird now, because we always refer to what is in `swap_protocols` module as `http_api::{protocol}` - especially in the `db` I find that a bit awkward (there is no distinction between what is coming from `http_api` and what from the `swap_protocols` now...).

Hum, yeah I would have seen the DB as a separate context and only use the `http_api::{protocol}` module _inside_ `http_api`.

da-kami:
> > Re-export swap_protocols::{protocol} in http_api
> > Note: Bit weird now, because we always refer to what is in `swap_protocols` module as `http_api::{protocol}` - especially in the `db` I find that a bit awkward (there is no distinction between what is coming from `http_api` and what from the `swap_protocols` now...).
> 
> Hum, yeah I would have seen the DB as a separate context and only use the `http_api::{protocol}` module _inside_ `http_api`.

I changed `storage.rs` to use the fully qualified name for the `http_api` related  imports and use `halight` and `herc20` from `swap_protocols`.

PR #2729, opened 2020-05-27
Use Rust naming convention by avoiding `get` prefix.

D4nte:


D4nte:
> LGTM. I wonder if, for example, `ToAlphaParams` would be more correct, since we're going from a reference to an owned value.

I wondered the same thing.

D4nte:
> > LGTM. I wonder if, for example, `ToAlphaParams` would be more correct, since we're going from a reference to an owned value.
> 
> I wondered the same thing.

According to https://rust-lang.github.io/api-guidelines/naming.html:

```
borrowed -> owned (non-Copy types)
```
The issue is that we do not know if it is a Copy type

luckysori:
> According to https://rust-lang.github.io/api-guidelines/naming.html:
> 
> ```
> Prefix to_: borrowed -> owned (non-Copy types)
> ```

Thanks for looking into it! That's a helpful link too :smile: 

> The issue is that we do not know if it is a Copy type

Yep. It looks like the `to_` prefix is used to indicate high cost, and that will depend on the implementation of the trait. Let's leave it as is :wink:

PR #2730, opened 2020-05-27
Use a typed enum instead of string slice

D4nte:
Because this is not javascript.

### DIFF
--- a/cnd/src/http_api/routes.rs
+++ b/cnd/src/http_api/routes.rs
@@ -168,11 +168,11 @@ where
             entity.push_sub_entity(beta_state_sub);
 
             let maybe_action_names = vec![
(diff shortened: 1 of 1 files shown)


cnd/src/http_api/routes.rs
```diff
@@ -19,6 +19,7 @@ use crate::{
 };
 use ::comit::Protocol;
 use anyhow::bail;
+use bitcoin::hashes::core::fmt::Formatter;
```
luckysori, line 22: :wrench: 
```suggestion
use std::fmt::Formatter;
```

D4nte: Oops, CLion!

PR #2731, opened 2020-05-27
Guidelines for migrations to the cnd database

da-kami:
As discussed last week here is a first approach how we could deal with this.

* Consolidates all migrations related to the new db schema into one migration
* Adds a script that checks if a new migration is needed based on the git release tag. The script automatically set up a new version asking the dev for a name.


cnd/MIGRATIONS.md
```diff
@@ -0,0 +1,34 @@
+# Migrations in cnd
+
+## Context
+
+This README offers guidelines on how to manage database migrations in cnd.
+
+This document is motivated by the fact, that handling unnecessary migrations adds unnecessary complexity.
+Migrations are only relevant once a new version of cnd is released, because the prior motivation for migrations is ensuring downwards compatibility and tracking changes to achieve this compatibility.
```
tcharding, line 8: nit: I think you want 'backwards compatibility' :)

da-kami: Yes, thanks!

da-kami:
> Thanks for writing this up @da-kami !
> 
> Do you mind applying the current strategy right away to the current migrations?

Can do, it's the second point on the task-list in the description :)
I'll add to this PR until it's done.

@thomaseizinger could you handle the folder rename with timestamp for the automatic release process?

thomaseizinger:
> > Thanks for writing this up @da-kami !
> > Do you mind applying the current strategy right away to the current migrations?
> 
> Can do, it's the second point on the task-list in the description :)
> I'll add to this PR until it's done.

Excuse me for not properly reading your PR description!

> @thomaseizinger could you handle the folder rename with timestamp for the automatic release process?

Do we have to automate right away? I don't really have time unfortunately at the moment :(
Also, messing with the release action is cumbersome because the feedback loop is quite long.

Can we adapt the guidelines to maybe not require a change here?
I am thinking of simply dropping the `CURRENT` prefix and instead, have a separate file that tracks, which migrations have already been released. We could potentially write a script that uses `git` to check for the latest release tag and tells you whether or not the latest migration has been released.
If it has, you have to make a new one, otherwise you are free to modify the existing one.

Such check could even be part of CI.

da-kami:
> > > Thanks for writing this up @da-kami !
> > > Do you mind applying the current strategy right away to the current migrations?
> > 
> > 
> > Can do, it's the second point on the task-list in the description :)
> > I'll add to this PR until it's done.
> 
> Excuse me for not properly reading your PR description!
> 
> > @thomaseizinger could you handle the folder rename with timestamp for the automatic release process?
> 
> Do we have to automate right away? I don't really have time unfortunately at the moment :(
> Also, messing with the release action is cumbersome because the feedback loop is quite long.
> 
> Can we adapt the guidelines to maybe not require a change here?
> I am thinking of simply dropping the `CURRENT` prefix and instead, have a separate file that tracks, which migrations have already been released. We could potentially write a script that uses `git` to check for the latest release tag and tells you whether or not the latest migration has been released.
> If it has, you have to make a new one, otherwise you are free to modify the existing one.
> 
> Such check could even be part of CI.

Maybe @luckysori and me can just take care or it. I'd be a fan of handling this in the `draft-new-release` workflow rather than introducing another "extra" script.

da-kami:
After discussion with @thomaseizinger decision to use a script (triggering `Diesel CLI`) for handling the migrations locally, rather than integration into the github action. 

Working on the script now.

cnd/migrations.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+function navigate_to_last_folder() {
```
D4nte, line 3:
```suggestion
function navigate_to_first_folder() {
```
This navigates to the first folder by lexical order. I find the naming confusing.

da-kami: It is not the first, it's the last (using `tail`).
Maybe it can be done in a better way.

da-kami:
> Build failed:
> 
> * [build (ubuntu-latest)](https://github.com/comit-network/comit-rs/runs/752782670)

Note: Could not reproduce that one. I got the logs and Bob just gets a `404` from cnd. But I could not figure out how that would happen and could not reproduce it locally (neither on this branch, not on dev, nor on this branch rebased on dev).

cnd/migrations.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# check if the commit yields any tags
+TAG_LIST=$(git tag --contains $LATEST_COMMIT)
+
+if [[ $TAG_LIST == *"${RELEASE_VERSION}"* ]]; then
+  echo "A new migration is needed, please enter a name for the new migration based on your planned changes:"
+  echo "Naming according to Diesel conventions, separating words with '_'."
+  read MIGRATION_NAME
+  cd $MIGRATIONS_FOLDER
+  cargo install diesel_cli --no-default-features --features sqlite
```
thomaseizinger, line 27: Might want to check whether `diesel` is already installed with:

```
which diesel || cargo install diesel_cli --no-default-features --features sqlite
```

da-kami: we can add this - I feel it does not hurt. it just says `already installed` if it already is.

PR #2733, opened 2020-05-27
Add facade 'can_refund' methods for layer 1 connectors

tcharding:
We would like to be able to determine whether an expiry time has elapsed i.e., whether a refund transaction would be accepted by the relevant blockchain.

In order to do this we add a trait `HasPassed` and then implement it for the layer 1 blockchain connectors.

We then add a new `Connectors` struct and add an instance of this to the `Facade`. Also add methods on the facade to call through to the appropriate connector to expose the required functionality.

### DIFF
--- a/cnd/src/swap_protocols/facade.rs
+++ b/cnd/src/swap_protocols/facade.rs
@@ -1,5 +1,6 @@
 use crate::{
     asset,
+    connectors::Connectors,
     db::{CreatedSwap, Save, Sqlite},
     identity,
     network::{DialInformation, Identities, Swarm},
@@ -8,6 +9,7 @@ use crate::{
     timestamp::Timestamp,
 };
 use ::comit::network::protocols::announce::SwapDigest;
+use comit::HasPassed;
 use digest::Digest;
 
 /// This represents the information available on a swap
@@ -108,6 +110,7 @@ pub struct Facade {
     pub swarm: Swarm,
     pub db: Sqlite,
     pub storage: Storage,
+    pub connectors: Connectors,
 }
 
 impl Facade {
@@ -122,6 +125,16 @@ impl Facade {
         self.swarm
             .initiate_communication(id, peer, role, digest, identities)
             .await
+    }
+
+    /// True if expiry time has elapsed according to the Bitcoin connector.
+    pub async fn can_refund_bitcoin(&self, expiry: Timestamp) -> bool {
+        self.connectors.bitcoin.has_passed(expiry).await
+    }
+
+    /// True if expiry time has elapsed according to the Ethereum connector.
+    pub async fn can_refund_ethereum(&self, expiry: Timestamp) -> bool {
+        self.connectors.ethereum.has_passed(expiry).await
     }
 }
 
--- a/cnd/src/swap_protocols/facade.rs
+++ b/cnd/src/swap_protocols/facade.rs
@@ -128,12 +128,12 @@ impl Facade {
     }
 
     /// True if expiry time has elapsed according to the Bitcoin connector.
-    pub async fn can_refund_bitcoin(&self, expiry: Timestamp) -> bool {
+    pub async fn can_refund_bitcoin(&self, expiry: Timestamp) -> anyhow::Result<bool> {
         self.connectors.bitcoin.has_passed(expiry).await
     }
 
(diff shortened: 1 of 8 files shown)


comit/src/btsieve/bitcoin/cache.rs
```diff
@@ -69,3 +72,44 @@ where
         Ok(block)
     }
 }
+
+impl<C> Cache<C>
```
thomaseizinger, line 76: This goes against the original design of the cache being a proxy.

This functionality doesn't need the cache so should be implemented just on the abstraction of a connector.
If you pass a `Cache<Connector`, you won't loose any performance while computing this assuming the blocks are already cached.

comit/src/btsieve/ethereum/cache.rs
```diff
@@ -112,3 +114,29 @@ where
...
+#[async_trait]
+impl<C> HasPassed for Cache<C>
+where
+    C: LatestBlock<Block = Block>,
+{
+    async fn has_passed(&self, timestamp: Timestamp) -> bool {
+        match self.latest_block().await {
+            Err(e) => {
+                tracing::warn!("failed to get block time: {}", e);
+                false
```
thomaseizinger, line 130: This error handling feels quite premature to me, I'd let the caller deal with this.

cnd/src/swap_protocols/facade.rs
```diff
@@ -123,6 +126,16 @@ impl Facade {
             .initiate_communication(id, peer, role, digest, identities)
             .await
     }
+
+    /// True if expiry time has elapsed according to the Bitcoin connector.
+    pub async fn can_refund_bitcoin(&self, expiry: Timestamp) -> bool {
```
thomaseizinger, line 131: Nit: This function only calls through to `has_passed`, why is it named differently?

comit/src/btsieve/bitcoin/cache.rs
tcharding: > This goes against the original design of the cache being a proxy.

I don't understand this.

> This functionality doesn't need the cache so should be implemented just on the abstraction of a connector.

I don't understand this either, we want the cache so that if this function is called repeatedly (e.g. by a user polling the REST API) the blocks are cached.

> If you pass a `Cache<Connector`, you won't loose any performance while computing this assuming the blocks are already cached.

I don't get what you are suggesting here? Pass the `Cache<Connetor>` instead of implement this as a method on the `Cache` is what I understand but that makes no sense to me?

comit/src/btsieve/ethereum/cache.rs
tcharding: Cool, we can return `Result<bool>` instead. @D4nte are you happy with that?

cnd/src/swap_protocols/facade.rs
tcharding: The facade method has not been used yet; I was looking into the future and thinking that the caller of this method would like to see _why_ we are checking if the expiry has passed, hence this method name. The connector however just exposes lower level functionality to check if a time has passed, hence the trait name.

comit/src/btsieve/ethereum/cache.rs
tcharding: Implemented as suggested.

comit/src/btsieve/bitcoin/cache.rs
thomaseizinger: >> This goes against the original design of the cache being a proxy.

> I don't understand this.

A proxy, by design, forwards calls to something else and adds additional logic around it while adhering to the same interface that than the thing it forwards to.
This function does not forward to the connector but instead composes new functionality from the connector. A proxy must adhere to the same interface as the thing it is forwarding to for it to be a proxy.

> I don't understand this either, we want the cache so that if this function is called repeatedly (e.g. by a user polling the REST API) the blocks are cached.

This function can just have the signature: `median_time(connector: C)`. No trait, no impl block, just a function with a generic parameter.

If the caller of this fn wants it to be fast, they should pass a `Cache<Connector` instead of a plain connector but there is no need for this function to know about caching. It only depends on the interface of a connector.

If we stick to the cache being a proxy, the function doesn't need to care about which particular connector implementation you are passing and whether or not it is wrapped in a cache.

PR #2736, opened 2020-05-28
Recommend Unique Action

D4nte:
Refund (and only refund) is returned once it is possible to refund,
meaning that the HTLC is expired. This is done by checking the
blockchain.

### DIFF
--- a/cnd/src/http_api/routes.rs
+++ b/cnd/src/http_api/routes.rs
@@ -8,7 +8,8 @@ use crate::{
         action::ActionResponseBody,
         halight, herc20, problem,
         protocol::{AlphaEvents, AlphaParams, BetaEvents, BetaParams, GetRole},
-        route_factory, ActionNotFound, AliceSwap, BobSwap, Http, Swap,
+        route_factory, ActionNotFound, AliceSwap, BobSwap, Http, NextAction, RecommendedNextAction,
+        Swap,
     },
     storage::Load,
     swap_protocols::{
@@ -51,7 +52,7 @@ pub async fn handle_get_swap(
                 herc20::Finalized,
                 halight::Finalized,
             > = facade.load(swap_id).await?;
-            make_swap_entity(swap_id, swap)
+            make_swap_entity(swap_id, swap, &facade).await
         }
         Swap {
             alpha: Protocol::Herc20,
@@ -60,7 +61,7 @@ pub async fn handle_get_swap(
         } => {
             let swap: BobSwap<asset::Erc20, asset::Bitcoin, herc20::Finalized, halight::Finalized> =
                 facade.load(swap_id).await?;
-            make_swap_entity(swap_id, swap)
+            make_swap_entity(swap_id, swap, &facade).await
         }
         Swap {
             alpha: Protocol::Halight,
@@ -73,7 +74,7 @@ pub async fn handle_get_swap(
                 halight::Finalized,
                 herc20::Finalized,
             > = facade.load(swap_id).await?;
-            make_swap_entity(swap_id, swap)
+            make_swap_entity(swap_id, swap, &facade).await
         }
         Swap {
             alpha: Protocol::Halight,
@@ -82,24 +83,24 @@ pub async fn handle_get_swap(
         } => {
             let swap: BobSwap<asset::Bitcoin, asset::Erc20, halight::Finalized, herc20::Finalized> =
                 facade.load(swap_id).await?;
-            make_swap_entity(swap_id, swap)
+            make_swap_entity(swap_id, swap, &facade).await
         }
         _ => unimplemented!("other combinations not suported yet"),
     }
 }
 
-fn make_swap_entity<S>(swap_id: LocalSwapId, swap: S) -> anyhow::Result<siren::Entity>
+async fn make_swap_entity<S>(
+    swap_id: LocalSwapId,
+    swap: S,
+    facade: &Facade,
+) -> anyhow::Result<siren::Entity>
 where
     S: GetRole
         + AlphaParams
         + BetaParams
         + AlphaEvents
         + BetaEvents
-        + DeployAction
-        + InitAction
-        + FundAction
-        + RedeemAction
-        + RefundAction
+        + RecommendedNextAction
         + Clone,
 {
     let role = swap.get_role();
@@ -173,27 +174,18 @@ where
             );
             entity.push_sub_entity(beta_state_sub);
 
-            let maybe_action_names = vec![
-                swap.init_action().map(|_| ActionName::Init),
-                swap.deploy_action().map(|_| ActionName::Deploy),
-                swap.fund_action().map(|_| ActionName::Fund),
-                swap.redeem_action().map(|_| ActionName::Redeem),
-                swap.refund_action().map(|_| ActionName::Refund),
-            ];
+            if let Some(action_name) = swap.recommended_next_action(facade).await {
+                let siren_action = make_siren_action(swap_id, action_name);
+                entity = entity.with_action(siren_action);
+            }
 
-            Ok(maybe_action_names
-                .into_iter()
-                .filter_map(|action| action.ok())
-                .fold(entity, |acc, action_name| {
-                    let siren_action = make_siren_action(swap_id, action_name);
-                    acc.with_action(siren_action)
-                }))
+            Ok(entity)
         }
         _ => Ok(entity),
     }
 }
 
-fn make_siren_action(swap_id: LocalSwapId, action_name: ActionName) -> siren::Action {
+fn make_siren_action(swap_id: LocalSwapId, action_name: NextAction) -> siren::Action {
     siren::Action {
         name: action_name.to_string(),
         class: vec![],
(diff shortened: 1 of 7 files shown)


cnd/src/http_api/routes.rs
```diff
@@ -173,27 +174,18 @@ where
             );
             entity.push_sub_entity(beta_state_sub);
 
-            let maybe_action_names = vec![
-                swap.init_action().map(|_| ActionName::Init),
-                swap.deploy_action().map(|_| ActionName::Deploy),
-                swap.fund_action().map(|_| ActionName::Fund),
-                swap.redeem_action().map(|_| ActionName::Redeem),
-                swap.refund_action().map(|_| ActionName::Refund),
```
thomaseizinger, line 181: By removing this, you are removing the **intended** coupling between _can we construct the action_ and _should we tell the user that the action is available_.

The code between `NextRecommendedAction` and `XYZAction` can now potentially run out of sync which would result in the user seeing an action on the swap but when they actually hit the endpoint, it returns 404.

D4nte:
> Meta-comment: I am unsure how to deal with PR reviews of incomplete (shape-up style) projects :man_shrugging: You obviously seem to have a plan on how to implement stuff, yet the PRs so far only present individual bits. That makes it hard to give a holistic review.

I am not sure what you are talking about. This is the final PR.

tcharding:
> Meta-comment: I am unsure how to deal with PR reviews of incomplete (shape-up style) projects  You obviously seem to have a plan on how to implement stuff, yet the PRs so far only present individual bits. That makes it hard to give a holistic review.

This is an interesting point for shape up projects in general. (Not relevant to this shape up as @D4nte commented, there is only one PR coming out of this project.) I feel this is associated with a bigger problem we have, we have some conflicting ideas re PRs at the moment. On one hand we want to make PRs a single change because many team members review by PR not by patch, however this means making a patch series requires multiple draft PRs each stating that it relies on the others and is basically a cumbersome way to use GitHub. This has led us to merging code that is not used, widely accepted as bad practice in software developement. The current state of `hbit` is a good example. I believe much of this has come from the fact that we all resist rebasing because it is quite painful within our current environment for some reason. I have a suspicion that feature branches with proper discreet patches (i.e. each patch does one thing) would solve this, all except the rebasing difficulty we are seeing. 

Back to your original point, reviewers could then review the feature branch since each commit on the branch would be well formed and the series of commits on the branch would give an holistic view of the work being done. 

Re the rebasing difficulty. I believe we are just not leveraging the tools correctly (git and rust). Rebasing feature branches, or in other words patch series, is what git was made for.

> We are now duplicating crucial logic (which action is available) 

This is a fair point, happy to investigate further.

> and the `Facade` is not really meant to be passed around.

`can_refund_ethereum()` is implemented on the facade inline with other code, the caller does not know _who_ knows how to resolve this question - hence we use the facade. If you are referring to passing the facade into the trait `RecommendedNextAction` I agree it is not pretty. I would like to see all the logic (excluding getting the states and expiry out of the swap) to be in the comit crate, we cannot do that for hbit yet so we are stuck between do it this way, just do this PR for herc20/halight, or wait for hbit/herc20 to land before doing this work.

Perhaps there is another design that gets around these issues all together? 

> We also seem to lack tests for this code.

There are e2e tests. The 'has passed' code is too trivial to need unit tests. The 'which action to return' code is not good code to unit test because the only bugs will be logic bugs, if we write tests that test the same logic we implemented that gains us nothing - e2e tests test the logic is correct (does that make sense?).

D4nte:
Thanks @tcharding for your comment.

1. Dupe logic: we could look into refactoring
2. Facade passed around: The facade is already passed around, not sure how this PR is _too much_. In any case, we could look into having the facade return the current blockchain time and pass in the `RecommendedNextAction` trait instead.
3. There are 2 things to be tested:
	a. That only one action is returned at a given time, you can read that from the code, we call `with_action` only once. Nothing tricky here
	b. That the correct action is returned at a given time: when calling `fund()` on the SDK, it actually check if the action is available on the swap and then execute it. Which means this is already tested too.
	Hence, no new test are needed.

thomaseizinger:
> Facade passed around: The facade is already passed around, not sure how this PR is _too much_.

To my knowledge this is the only place in the code where we implement a trait on _data_ (`AliceSwap` / `BobSwap`) to pass in a long-lived "service" like struct that has knowledge about the database etc.
- From a dependency perspective, this is not great because you have circular dependencies between the structs `AliceSwap` and the `Facade`: The facade produces `AliceSwap` and `AliceSwap` uses the `Facade` to realize `RecommendedNextAction`.
- The design also makes it harder to move this code into the COMIT library because it depends on the `Facade`.

> In any case, we could look into having the facade return the current blockchain time and pass in the RecommendedNextAction trait instead.

This sounds like a much better alternative to me :)

> There are 2 things to be tested:
a. That only one action is returned at a given time, you can read that from the code, we call with_action only once. Nothing tricky here
b. That the correct action is returned at a given time: when calling fund() on the SDK, it actually check if the action is available on the swap and then execute it. Which means this is already tested too.
Hence, no new test are needed.

Making sure it works is not the only reason we are writing tests for. Tests are also essential to detect regressions (among other things).

Once we refactor all the action handling code so it can move into the COMIT crate, how can we assert that we didn't break the work you did for this project if there are no tests for it?

To me, this project is not complete if we don't have a test that is now green and turns red if we undo this change.

tcharding:
> To me, this project is not complete if we don't have a test that is now green and turns red if we undo this change.

I disagree with this, surely we are not following this in general - everytime one writes some functionality, no matter how trivial they have to write a passing regression test - no chance :)

thomaseizinger:
> > To me, this project is not complete if we don't have a test that is now green and turns red if we undo this change.
> 
> I disagree with this, surely we are not following this in general - everytime one writes some functionality, no matter how trivial they have to write a passing regression test - no chance :)

We are following this mostly, it is just that we haven't landed many end-user facing changes recently. The biggest one was changing what the POST request for creating a swap looks like and we did write end to end tests for that.
Everything else was mostly internal refactorings :man_shrugging: (by the definition of refactoring, there are no user visible changes to functionality)

The reason why I would like to have a least one test for this is because we've done this project _now_ with the expectation that the internal design is going to change completely one we move this to the COMIT library. Given that, it would be nice to have automated tests that make sure we are not breaking anything while doing that but keep the level of functionality that we have built here.

Now, I don't want to block your progress on this so I am gonna give my green mark here given that the functionality is there and (hopefully?) works as intended :)

D4nte:
We are doing a new PR to tackle the comments. (branch `rua`).

D4nte:
> Now, I don't want to block your progress on this so I am gonna give my green mark here given that the functionality is there and (hopefully?) works as intended :)

Please note that the e2e tests already verify that the expected action is returned.
In term of ensuring that this action is unique, it will be obvious from the code.

thomaseizinger:
> > Now, I don't want to block your progress on this so I am gonna give my green mark here given that the functionality is there and (hopefully?) works as intended :)
> 
> Please note that the e2e tests already verify that the expected action is returned.
> In term of ensuring that this action is unique, it will be obvious from the code.

My point is more that given the API's model, it is possible for multiple actions to be returned because the `actions` field in Siren is an array.
If we want to build software on top of that API that relies on invariants which are not captured by that model - that there is only ever max one action - then I think we should capture these invariants in tests to make sure that they are not broken in the future.

Now, if we were to go in the direction of re-modelling the API to better reflect these invariants, this would be a different story. I've mentioned some ideas in this regard in the [pitch discussion](https://3.basecamp.com/4403044/buckets/16478057/messages/2587793346#__recording_2620562167) as "Don't use Siren actions to fetch the data" but we agreed that it can be solved as a separate problem.

However, a consequence of that decision is that we are stuck with a sub-optimal model (an array) even though we only have either 0 or 1 elements in there at any point in time. Given that we want to write software (the app + SDK) on top of this model that relies on these invariants, I think the should be captured in tests until we fix the model.

tcharding:
> I think they should be captured in tests until we fix the model.

I'm going to open an issue for this.

D4nte:
> My point ...

Considering how we have done thing so far, wouldn't the correct way forward to update the SDK, examples and e2e tests?

thomaseizinger:
> > My point ...
> 
> Considering how we have done thing so far, wouldn't the correct way forward to update the SDK, examples and e2e tests (to use the SDK)?

I think we don't necessarily have to jump into updating the SDK. Updating the SDK is associated for me with "we are happy with this API, hence we make it available through the SDK". As long as we are iterating on it, I would keep things local in the e2e tests otherwise updating and releasing the SDK is just an added burden.

We chose to implement this functionality in a backwards compatible way, so I don't see a pressing need in updating the SDK.

D4nte:
> > > My point ...
> > 
> > 
> > Considering how we have done thing so far, wouldn't the correct way forward to update the SDK, examples and e2e tests (to use the SDK)?
> 
> I think we don't necessarily have to jump into updating the SDK. Updating the SDK is associated for me with "we are happy with this API, hence we make it available through the SDK". As long as we are iterating on it, I would keep things local in the e2e tests otherwise updating and releasing the SDK is just an added burden.
> 
> We chose to implement this functionality in a backwards compatible way, so I don't see a pressing need in updating the SDK.

I find it easier to update the SDK than to migrate part of the action logic to the e2e tests.

PR #2742, opened 2020-05-29
Align types herc20 halight

D4nte:
I think it makes the codebase slightly better (less than I hoped).

The idea is to remove duplication of data structs for each event and properly separate data & events concept.

### DIFF
--- a/comit/src/herc20.rs
+++ b/comit/src/herc20.rs
@@ -61,6 +61,7 @@ pub trait WaitForRefunded {
 
 /// Represents the events in the herc20 protocol.
 #[derive(Debug, Clone, PartialEq, strum_macros::Display)]
+#[allow(clippy::large_enum_variant)]
 pub enum Event {
     /// The protocol was started.
     Started,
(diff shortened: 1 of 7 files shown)


comit/src/herc20.rs
```diff
@@ -88,21 +92,20 @@ pub struct Deployed {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-pub enum Funded {
```
thomaseizinger, line 91: These structs were supposed to be _events_ that is why the carried only the data of this particular event.

Contrary to that, the `State` enum was an aggregation of all the data given a particular sequence of events.

I'd prefer to _not_ make this change.
The reason is that events and states are separate concepts. The events are the source of truth and you can always re-created the state from the events.

I envision to eventually only store the events in the database and re-create the states on the fly. That is also the reason why the states are still in `cnd` and not in `comit`.
Once you store the events in the DB, you would probably have a table per event. Having them accumulate data would lead to duplication, hence I think that design would be suboptimal.

I think the current design is okay-ish. Having the same fields in the `State` repeated is not great but I think all the alternatives have worse trade-offs. This way, the concepts are at least separate and not entangled with each other which gives us more modularity.

PR #2762, opened 2020-06-01
Modify how we bring types into scope

tcharding:
Currently we are struggling with rebase and merge conflicts when working on cnd. In an effort to have the tools work better from us this PR modifies the way we bring types into scope, this should result in less merge conflicts and/or easier to resolve merge conflicts.

The strategy is as follows:
- Define the module abstractions we want to know about, defined to be: `network`, `http_api`, `db`, and root. 
- Export types to the 'top level' of the required abstraction so users do not need to know where the code is i.e., `seed::RootSeed` can be brought into scope as just `RootSeed` if we export its types. (Further clarification; only export network types to the network module, not all the way up to the root module.)
- Export all comit crate types within cnd. This means that we no longer need to think about where types come from when writing code in cnd, we can just import them from `crate`.


cnd/src/network.rs
```diff
@@ -63,6 +63,9 @@ use std::{
 };
 use tokio::{runtime::Handle, sync::Mutex};
 
+// Export comit network types while maintaining the module abstraction.
+pub use ::comit::network::{WhatAliceLearnedFromBob, WhatBobLearnedFromAlice};
```
thomaseizinger, line 67: I wouldn't mind doing even this:

```suggestion
pub use ::comit::network::*;
```

thomaseizinger:
On a related note, I watched a talk yesterday that touched on this. Here is an interesting section about imports: https://youtu.be/FyCYva9DhsI?t=2027

cnd/src/network.rs
tcharding: Do you think that it is useful to have separate abstractions for network, http_api, db etc. I get the feeling it would be because it allows us to define types in each with the same name.

thomaseizinger: What exactly do you mean with "abstractions" here? If you are referring to modules then I would say yes and we already have them :)

PR #2764, opened 2020-06-02
Move protocol modules out of swap_protocols

tcharding:
Swap protocols is _all_ we do. We do not need a sub-module for this.
Also, `swap_protocols/` primarily has rfc003 code in it.

Move the 'new' protocol modules up to the root cnd module.


tcharding:
@luckysori and @da-kami please note that I intentionally did not move `hbit` because you guys are working on it. Please consider following the work done in this PR and moving hbit up to the root module while you work on it.

PR #2765, opened 2020-06-02
Use `nextAction` from SDK

D4nte:
Last PR for the Recommend Unique Action project.

### DIFF
--- a/api_tests/tests/erc20_halight.ts
+++ b/api_tests/tests/erc20_halight.ts
@@ -16,15 +16,15 @@ it(
         await alice.createHerc20HalightSwap(bodies.alice);
         await bob.createHerc20HalightSwap(bodies.bob);
 
-        await alice.init();
+        await alice.assertAndExecuteNextAction("init");
 
-        await alice.deploy();
-        await alice.fund();
+        await alice.assertAndExecuteNextAction("deploy");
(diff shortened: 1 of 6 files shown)


api_tests/tests/erc20_halight.ts
```diff
@@ -16,15 +16,15 @@ it(
         await alice.createHerc20HalightSwap(bodies.alice);
         await bob.createHerc20HalightSwap(bodies.bob);
 
-        await alice.init();
+        await expect(alice.next()).resolves.toBe("init");
```
thomaseizinger, line 19: Could we reduce the noise in writing tests here by introducing something like `alice.assertAndExecuteNext("init")`.

Naming of the function subject to be improved :)

D4nte: Yes, just wanted to check first that my peers agree on the assertion of the step.

thomaseizinger: Oh definitely, I actually wanted to make the above a question but forgot the question mark, lol :D

D4nte:
So Actually I shot myself in the foot because I cannot upgrade to the latest SDK as the latest SDK contain hbit changes that were not yet integrated in the e2e tests. I believe @thomaseizinger told me so...

PR #2766, opened 2020-06-02
Remove `WaitForXYZ` trait abstraction

thomaseizinger:
Modelling this as a trait unnecessarily couples it to a particular connector implementations. Instead, we can create simple free-functions for this that are actually an implementation detail of the `herc20::new` function that no one outside of this module needs to be concerned with.

This is a follow-up PR motivated by the blog post I wrote last week.

cc @comit-network/comit-devs.


thomaseizinger:
Some follow-up thoughts:

Now that these functions are implementation details, it becomes more relevant to ask: do we even need _these_ abstractions or can we write similarly expressive code just inside the `watch_ledger` function by refactoring away some of the verbosity inside the individual functions.

luckysori:
> Some follow-up thoughts:
> 
> Now that these functions are implementation details, it becomes more relevant to ask: do we even need _these_ abstractions or can we write similarly expressive code just inside the `watch_ledger` function by refactoring away some of the verbosity inside the individual functions.

I like the fact that we call functions named after the protocol steps, in the expected order. Maybe that's not a strong enough motivation to keep the current design (after this PR). What would be the benefit of, for example, removing `watch_for_deployed`?

thomaseizinger:
> > Some follow-up thoughts:
> > Now that these functions are implementation details, it becomes more relevant to ask: do we even need _these_ abstractions or can we write similarly expressive code just inside the `watch_ledger` function by refactoring away some of the verbosity inside the individual functions.
> 
> I like the fact that we call functions named after the protocol steps, in the expected order. Maybe that's not a strong enough motivation to keep the current design (after this PR). What would be the benefit of, for example, removing `watch_for_deployed`?

The motivation for clearly seeing the protocol steps in the expected order is definitely legit.
The question is, does this necessarily map to the technical concept of a function?

In this case, the function signatures are quite verbose because we have to type out the types + trait bounds.

If we manage to still clearly convey the steps without having to extract functions, we can save ourselves quite some code that needs to be written (and read).

PR #2782, opened 2020-06-03
Herc20 hbit refund

luckysori:
Almost exclusively TS changes after a little bit of debugging.

I would be happy to refine some e2e test suite APIs, because I probably modified things too fast.

Also, I believe there is no way to "refund" Halight, so I don't know if I should assert anything about lightning (or bitcoin) at the end of `alice-refunds-bob-refunds` tests.

### DIFF
--- a/api_tests/src/actors/actor.ts
+++ b/api_tests/src/actors/actor.ts
@@ -3,9 +3,7 @@ import {
     HalightHerc20RequestBody,
     Herc20HalightRequestBody,
     Herc20HbitRequestBody,
-    LedgerAction,
     Swap,
-    Transaction,
     Wallets as SdkWallets,
     HbitHerc20RequestBody,
 } from "comit-sdk";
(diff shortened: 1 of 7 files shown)


api_tests/src/actors/actor.ts
```diff
@@ -425,6 +425,19 @@ export class Actor {
         await this.assertRedeemed();
     }
 
+    /**
+     * Wait for and execute the refund action
+     */
+    public async refund() {
```
D4nte, line 431: 📣 use `assertAndExecuteNextAction` instead, do not add this new method.

api_tests/src/actors/actor.ts
```diff
@@ -630,6 +669,33 @@ export class Actor {
...
+                assetKey,
+                maximumFee
+            );
+            const expectedBalance = this.startingBalances.get(assetKey);
+            const currentWalletBalance = await wallet.getBalanceByAsset(
+                defaultAssetValue(asset, ledger)
+            );
+            const balanceInclFees = expectedBalance - maximumFee;
+            expect(currentWalletBalance).toBeGreaterThanOrEqual(
+                // @ts-ignore: Jest supports bigint, types to be fixed updated with
```
D4nte, line 692:
```suggestion
```

This is not needed any more.

api_tests/src/actors/actor.ts
D4nte: Also, in `hbit_herc20.ts`, do not use `redeem`, etc instead use `assertAndExecuteNextAction`.

luckysori: Oh, yeah. Nice! I'll remove all these actions. We weren't actually using them after the last patch.

api_tests/src/actors/swap_factory.ts
```diff
@@ -23,16 +23,16 @@ import { HbitRequestParams } from "comit-sdk/dist/src/cnd/swaps_payload";
 
 declare var global: HarnessGlobal;
 
-interface Ledgers {
-    alpha: keyof AllWallets;
-    beta: keyof AllWallets;
+interface SwapSettings {
+    ledgers?: { alpha: keyof AllWallets; beta: keyof AllWallets };
+    instantRefund?: boolean;
```
thomaseizinger, line 28: :joy:

luckysori: You hurt me, @thomaseizinger :sob:

thomaseizinger: Sorry, I didn't mean to :)
I was just positively amused!

PR #2788, opened 2020-06-04
Upgrade bitcoind to 0.20.0

D4nte:
Due to the lack of transactions on the local blockchain, bitcoind cannot
estimate a fee. Hence, fallback fee should be set.

See bitcoind 0.20.0 Release notes for details.


thomaseizinger:
Nice, that was a quick one @D4nte :D

PR #2789, opened 2020-06-04
Add halight::Network type

tcharding:
We currently have a few problems that are all intertwined:
- non-uniform usage of the terms ledger/network
- non-uniform usage of modelling `hbit::CreatedSwap` vs `asset::Bitcoin`
- non-uniform serde code in http_api and the db i.e., how we wrap types

This PR resolves all there of these for the halight 'network' type. We
choose 'network' over 'ledger' because if its common usage throughout
the Ethereum and Bitcoin communities to refer to the specific network
in use (mainnet, testnet etc.). Arguably this could also have been
'ledger' since each network _is_ a separate network.

Do:

- Add `halight::Network` enum to model the three networks.

- Use `halight::Network` throughout `http_api` module.
- Add serialization stability test for `Http<halight::Netwok>` (Follow
the module example and only do serialization test not deserialization
test).

- Use `halight::Network` throughout `db` module.
- Add a `Db` wrapper type that controls ser/der (actually `FromStr`,
  `Display` in the db). (Mimics the `Http` wrapper.)
- Add serialization/deserialization stability tests to the `db` module.

- Remove `ledger::Lightning` in favour of the newly added type.

### DIFF
--- a/cnd/src/db.rs
+++ b/cnd/src/db.rs
@@ -24,15 +24,18 @@ pub use self::{
 };
 
 use crate::{
-    db::wrapper_types::custom_sql_types::Text, swap_protocols::rfc003::SwapId, LocalSwapId,
-    Protocol, Role,
+    db::wrapper_types::custom_sql_types::Text, halight, swap_protocols::rfc003::SwapId,
+    LocalSwapId, Protocol, Role,
 };
 use async_trait::async_trait;
 use diesel::{self, prelude::*, sqlite::SqliteConnection};
 use libp2p::PeerId;
 use std::{
     ffi::OsStr,
+    fmt,
+    ops::Deref,
     path::{Path, PathBuf},
+    str::FromStr,
     sync::Arc,
 };
 use tokio::sync::Mutex;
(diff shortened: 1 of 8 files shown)


cnd/src/db.rs
```diff
@@ -254,4 +302,28 @@ mod tests {
         assert_that(&db).is_ok();
         assert_that(&path).exists();
     }
+
+    #[test]
+    fn stability_of_serialization_halight_network() {
```
thomaseizinger, line 307: We have a file with tests for these in the db module :)

comit/src/halight.rs
```diff
@@ -148,3 +148,11 @@ pub struct Settled {
...
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct Cancelled;
+
+/// Which layer 1 network we are running on (implicitly refers to the Bitcoin)
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum Network {
+    Mainnet,
+    Testnet,
+    Regtest,
+}
```
thomaseizinger, line 158: I would prefer to keep it modeled as `ledger::Lightning` but if you don't want to have the enum, you can have a field with a Bitcoin Network maybe?

cnd/src/db.rs
```diff
@@ -197,11 +200,56 @@ pub struct CreatedSwap<A, B> {
     pub role: Role,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct Db<I>(pub I);
```
thomaseizinger, line 204: Some rustdoc documentation would be nice for this one :)

cnd/src/db.rs
tcharding: Oh cool, I missed that. Thanks.

comit/src/halight.rs
tcharding: This change of model is the main concept of this draft PR. The reason I'm tackling this is explained in the PR message, can you explain _why_ you want to keep it as is and not unify all our models as `protocol::Type` instead of having half like that and half like this one `type::Protocol`?

cnd/src/db.rs
tcharding: Good point, will add.

PR #2776, opened 2020-06-03
Add herc20-hbit and hbit-herc20

da-kami:
resolves: https://github.com/comit-network/comit-rs/issues/2739

boom.

PR #2790, opened 2020-06-04
Respawn feature

### LINKED ISSUE #2771: Respawn tests for new routes

luckysori:
Fixes #2771.

### DIFF
--- a/cnd/src/db/schema.rs
+++ b/cnd/src/db/schema.rs
@@ -7,6 +7,7 @@ table! {
        local_swap_id -> Text,
        role -> Text,
        counterparty_peer_id -> Text,
+       start_of_swap -> Timestamp,
    }
 }
 
--- a/cnd/src/start_swap.rs
+++ b/cnd/src/start_swap.rs
@@ -1,31 +1,41 @@
 use crate::{
-    db::{ForSwap, Save, Swap},
+    db::{ForSwap, Save},
     halight, hbit, herc20,
-    network::{comit::RemoteData, WhatAliceLearnedFromBob, WhatBobLearnedFromAlice},
-    protocol_spawner::{ProtocolSpawner, Spawn},
-    storage::{Load, Storage},
-    LocalSwapId, Protocol, Role, Side,
+    network::{RemoteData, WhatAliceLearnedFromBob, WhatBobLearnedFromAlice},
+    Load, LocalSwapId, Protocol, ProtocolSpawner, Role, Side, Spawn, Storage,
 };
-use chrono::offset::Utc;
+use chrono::NaiveDateTime;
 
-pub async fn start_swap(
+#[derive(Clone, Copy, Debug)]
+pub struct DecisionSwap {
+    pub id: LocalSwapId,
+    pub role: Role,
+    pub alpha: Protocol,
+    pub beta: Protocol,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct Swap<A, B> {
+    pub role: Role,
+    pub alpha: A,
+    pub beta: B,
+    pub start_of_swap: NaiveDateTime,
+}
+
+pub async fn save_and_start_swap(
     storage: Storage,
     spawner: ProtocolSpawner,
     id: LocalSwapId,
     data: RemoteData,
-) -> anyhow::Result<()>
-where
-    ProtocolSpawner: Spawn<herc20::Params> + Spawn<halight::Params>,
-{
-    let start_of_swap = Utc::now().naive_local();
+) -> anyhow::Result<()> {
     let swap = storage.load(id).await?;
-
-    match (swap, data) {
+    match (&swap, data) {
         (
-            Swap {
+            DecisionSwap {
                 alpha: Protocol::Herc20,
                 beta: Protocol::Halight,
-                role: role @ Role::Alice,
+                role: Role::Alice,
+                ..
             },
             RemoteData {
                 ethereum_identity: Some(ethereum_identity),
@@ -42,18 +52,13 @@ where
                     },
                 })
                 .await?;
-
-            let herc20_params: herc20::Params = storage.load(id).await?;
-            let halight_params: halight::Params = storage.load(id).await?;
-
-            spawner.spawn(id, herc20_params, start_of_swap, Side::Alpha, role);
-            spawner.spawn(id, halight_params, start_of_swap, Side::Beta, role);
         }
         (
-            Swap {
+            DecisionSwap {
                 alpha: Protocol::Herc20,
                 beta: Protocol::Halight,
-                role: role @ Role::Bob,
+                role: Role::Bob,
+                ..
             },
             RemoteData {
                 ethereum_identity: Some(ethereum_identity),
@@ -72,18 +77,13 @@ where
                     },
                 })
                 .await?;
-
-            let herc20_params: herc20::Params = storage.load(id).await?;
-            let halight_params: halight::Params = storage.load(id).await?;
-
-            spawner.spawn(id, herc20_params, start_of_swap, Side::Alpha, role);
-            spawner.spawn(id, halight_params, start_of_swap, Side::Beta, role);
         }
         (
-            Swap {
+            DecisionSwap {
                 alpha: Protocol::Halight,
                 beta: Protocol::Herc20,
-                role: role @ Role::Alice,
+                role: Role::Alice,
+                ..
             },
             RemoteData {
                 ethereum_identity: Some(ethereum_identity),
@@ -100,18 +100,13 @@ where
                     },
                 })
                 .await?;
-
-            let halight_params: halight::Params = storage.load(id).await?;
-            let herc20_params: herc20::Params = storage.load(id).await?;
-
-            spawner.spawn(id, halight_params, start_of_swap, Side::Alpha, role);
-            spawner.spawn(id, herc20_params, start_of_swap, Side::Beta, role);
         }
         (
-            Swap {
+            DecisionSwap {
                 alpha: Protocol::Halight,
                 beta: Protocol::Herc20,
-                role: role @ Role::Bob,
+                role: Role::Bob,
+                ..
             },
             RemoteData {
                 ethereum_identity: Some(ethereum_identity),
@@ -130,18 +125,13 @@ where
                     },
                 })
                 .await?;
-
-            let halight_params: halight::Params = storage.load(id).await?;
-            let herc20_params: herc20::Params = storage.load(id).await?;
-
-            spawner.spawn(id, halight_params, start_of_swap, Side::Alpha, role);
-            spawner.spawn(id, herc20_params, start_of_swap, Side::Beta, role);
         }
         (
-            Swap {
+            DecisionSwap {
                 alpha: Protocol::Herc20,
                 beta: Protocol::Hbit,
-                role: role @ Role::Alice,
+                role: Role::Alice,
+                ..
             },
             RemoteData {
                 ethereum_identity: Some(ethereum_identity),
@@ -158,18 +148,13 @@ where
                     },
                 })
                 .await?;
-
-            let herc20_params: herc20::Params = storage.load(id).await?;
-            let hbit_params: hbit::Params = storage.load(id).await?;
-
-            spawner.spawn(id, herc20_params, start_of_swap, Side::Alpha, role);
-            spawner.spawn(id, hbit_params, start_of_swap, Side::Beta, role);
         }
         (
-            Swap {
+            DecisionSwap {
                 alpha: Protocol::Herc20,
                 beta: Protocol::Hbit,
-                role: role @ Role::Bob,
+                role: Role::Bob,
+                ..
             },
             RemoteData {
                 ethereum_identity: Some(ethereum_identity),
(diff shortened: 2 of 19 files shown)


cnd/src/network/start_swap.rs
```diff
@@ -1,31 +1,43 @@
...
+#[derive(Clone, Copy, Debug)]
+pub struct MetaSwap {
+    pub id: LocalSwapId,
+    pub role: Role,
+    pub alpha: Protocol,
+    pub beta: Protocol,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct Swap<A, B> {
```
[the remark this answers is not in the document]
luckysori: What if we expressed that this swap is "spawnable"?

cnd/src/storage.rs
```diff
@@ -198,23 +197,23 @@ impl Load<respawn::Swap<herc20::Params, hbit::Params>> for Storage {
...
+            start_of_swap: swap.start_of_swap,
         };
 
         Ok(swap)
     }
 }
 
 #[async_trait::async_trait]
-impl Load<respawn::Swap<hbit::Params, herc20::Params>> for Storage {
+impl Load<network::start_swap::Swap<hbit::Params, herc20::Params>> for Storage {
```
thomaseizinger, line 212: I think I would fit better if the `start_swap` module would be outside of `network` :)

cnd/src/db/schema.rs
```diff
@@ -7,6 +7,7 @@ table! {
        local_swap_id -> Text,
        role -> Text,
        counterparty_peer_id -> Text,
+       start_of_swap -> Timestamp,
```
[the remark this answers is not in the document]
thomaseizinger: I think this `Timestamp` is a SQL datatype and has nothing to do with our timestamp.

cnd/src/network/start_swap.rs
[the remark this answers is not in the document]
thomaseizinger: Not sure if that is such a good direction.

One of the things about generics in Rust are is that at the definition site, you don't care what is filled in.

Only when you attach functionality to it, it is important :)

cnd/src/db/schema.rs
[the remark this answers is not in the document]
thomaseizinger: Diesel has a `chrono` feature and can insert `NaiveDateTime`s directly, I think.

cnd/src/network/start_swap.rs
```diff
@@ -1,31 +1,43 @@
...
     protocol_spawner::{ProtocolSpawner, Spawn},
     storage::{Load, Storage},
     LocalSwapId, Protocol, Role, Side,
 };
-use chrono::offset::Utc;
+use chrono::NaiveDateTime;
 
-pub async fn start_swap(
+#[derive(Clone, Copy, Debug)]
+pub struct MetaSwap {
```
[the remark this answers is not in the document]
thomaseizinger: I get the feeling that they don't need to be different actually, we can just have a single `MetaSwap` in the DB module an be done with it :)

This is a good example for where we start out with a similar concept in several places of the codebase and once it all settles in, we can now see that all these places actually have the same requirements.

cnd/src/start_swap.rs
```diff
@@ -1,31 +1,41 @@
...
+
+#[derive(Clone, Copy, Debug)]
+pub struct Swap<A, B> {
+    pub role: Role,
+    pub alpha: A,
+    pub beta: B,
+    pub start_of_swap: NaiveDateTime,
+}
+
+pub async fn save_and_start_swap(
```
thomaseizinger, line 25: I think this function should be part of the `network` module and only `start_swap` should be outside.

Reason being that `RemoteData` is a network type so this will only called from within the `network` module anyway.

cnd/src/storage.rs
```diff
@@ -62,18 +62,18 @@ impl Storage {
 }
 
 #[async_trait::async_trait]
-impl Load<http_api::Swap> for Storage {
-    async fn load(&self, swap_id: LocalSwapId) -> anyhow::Result<http_api::Swap> {
+impl Load<http_api::DecisionSwap> for Storage {
+    async fn load(&self, swap_id: LocalSwapId) -> anyhow::Result<http_api::DecisionSwap> {
         self.db.load_meta_swap(swap_id).await
```
thomaseizinger, line 67: Still got the terminology of `meta` here.

cnd/src/start_swap.rs
```diff
@@ -246,12 +221,109 @@ where
...
+
+    Ok(())
+}
 
-            let hbit_params: hbit::Params = storage.load(id).await?;
-            let herc20_params: herc20::Params = storage.load(id).await?;
+pub async fn start_swap(
+    spawner: &ProtocolSpawner,
+    storage: &Storage,
+    meta_swap: DecisionSwap,
```
thomaseizinger, line 236: Still got the terminology of meta here.

cnd/src/start_swap.rs
```diff
@@ -246,12 +221,109 @@ where
                     },
                 })
                 .await?;
+        }
+        _ => tracing::info!("attempting to save for an unsupported swap"),
+    };
+
+    start_swap(&spawner, &storage, swap).await?;
```
thomaseizinger, line 228: We shouldn't start the swap if we hit this catch-all I reckon.

thomaseizinger: Better to bail from this function :)

cnd/src/storage.rs
```diff
@@ -201,23 +201,23 @@ impl Load<respawn::Swap<herc20::Params, hbit::Params>> for Storage {
...
+            start_of_swap: swap.start_of_swap,
         };
 
         Ok(swap)
     }
 }
 
 #[async_trait::async_trait]
-impl Load<respawn::Swap<hbit::Params, herc20::Params>> for Storage {
+impl Load<start_swap::Swap<hbit::Params, herc20::Params>> for Storage {
```
thomaseizinger, line 216: Should this module be called "spawn" maybe? :)

luckysori: Haha, @da-kami suggested it!

luckysori:
@thomaseizinger gonna tackle your suggestions in a follow-up PR. Already doing the same for your suggestions on PR #2776.

thomaseizinger:
> @thomaseizinger gonna tackle your suggestions in a follow-up PR. Already doing the same for your suggestions on PR #2776.

Sure thing :)

cnd/src/start_swap.rs
```diff
@@ -1,31 +1,41 @@
...
-    LocalSwapId, Protocol, Role, Side,
+    network::{RemoteData, WhatAliceLearnedFromBob, WhatBobLearnedFromAlice},
+    Load, LocalSwapId, Protocol, ProtocolSpawner, Role, Side, Spawn, Storage,
 };
-use chrono::offset::Utc;
+use chrono::NaiveDateTime;
 
-pub async fn start_swap(
+#[derive(Clone, Copy, Debug)]
+pub struct DecisionSwap {
```
D4nte, line 10: I don't understand the terminology used here? Why `Decision`?

comit/src/lnd.rs
```diff
@@ -42,6 +42,20 @@ pub enum PaymentStatus {
     Failed,
 }
 
+#[derive(Copy, Clone, Debug, thiserror::Error)]
+#[error("payment statuses don't match: expected {expected} but received {received}")]
+pub struct StatusMismatch {
```
D4nte, line 47:
```suggestion
pub struct PaymentStatusMismatch {
```
It's confusing to differentiate Payment and Invoice Mismatch error with `Status` vs `State`.

cnd/src/start_swap.rs
thomaseizinger: To my knowledge, it is because instances of this struct are used to make further decisions about swaps, i.e. which protocols and role it actually is :)

comit/src/lnd.rs
```diff
@@ -248,27 +258,44 @@ impl WaitForAccepted for LndConnectorAsSender {
...
             tokio::time::delay_for(Duration::from_millis(self.retry_interval_ms)).await;
-        }
 
-        Ok(Accepted)
+            let payment = match self.find_payment(params).await? {
+                Some(payment) => payment,
+                None => continue,
+            };
+
+            match payment.status {
```
D4nte, line 269: How is this necessary/related to this patch?

cnd/src/start_swap.rs
luckysori: These `DecisionSwap`s are only ever used to match and decide what information to load from the dabase.

comit/src/lnd.rs
luckysori: Will refine this in a separate PR.

cnd/src/start_swap.rs
da-kami: Because we purely use it for decision making. It only loads builds `meta` object to be able to reason about an actual `Swap. 

It was originally called `MetaSwap`, but we went for `Decision` instead because `Meta` is a very abstract term.

Note that I don't have a strong opinion about this. Do you @D4nte ? :)

comit/src/lnd.rs
thomaseizinger: The problem is, that we cannot "replay" halight protocols the same we do for other protocols because `lnd` only exposes the latest state.

So we need to actually explicitly match, which states are acceptable, taking into account that the desired state is already in the past, i.e. when looking for `Open`, `Accepted` is also good :)

thomaseizinger: To do that reliably, we needed to decouple fetching the invoice from matching against the desired state.

cnd/src/start_swap.rs
da-kami: 😄 we all answered at the same time - only noticed that now

tcharding: What they said :)

D4nte: Ok I see, we usually used the `meta` prefix which does not convey that much information but still better than `Decision` I think which is just plain confusing.

IMO, this is the metadata of the swap. It gives context to the content of the swap, the content being the asset amounts, token contract address, etc. Hence I think would be better to use `SwapMetadata` than `DecisionSwap`.

cnd/src/storage.rs
```diff
@@ -153,23 +153,23 @@ impl Load<respawn::Swap<halight::Params, herc20::Params>> for Storage {
...
+            start_of_swap: swap.start_of_swap,
         };
 
         Ok(swap)
     }
 }
 
 #[async_trait::async_trait]
-impl Load<respawn::Swap<herc20::Params, hbit::Params>> for Storage {
+impl Load<start_swap::Swap<herc20::Params, hbit::Params>> for Storage {
```
tcharding, line 168: nit: Feel free not to do this if you want to merge today but this should be pub used in network module then the usage can be `network::Swap` (same for `network::DecisionSwap`).

cnd/src/start_swap.rs
tcharding: Already used this joke today but here it is just for you @D4nte; I have some other data that relates to the `SwapMetadata`, I am going to call it `SwapMetaMetadata`.

D4nte: Isn't it `SwapMetadataMetadata`?

D4nte: Buffalo buffalo Buffalo buffalo buffalo buffalo Buffalo buffalo

da-kami:
Letting it merge - already working on the comments, will bring them in as follow-ups. Easier to review :)

cnd/src/start_swap.rs
thomaseizinger: > It gives context to the content of the swap

What about `SwapContext`?

> Hence I think would be better to use SwapMetadata than DecisionSwap.

Since we are already bikeshedding: `data` doesn't add any value and `SwapMeta` sounds funny. By flipping it around you end up at `MetaSwap` again :man_shrugging: 

A few more ideas based on what you said @D4nte :

- `SwapSetting`: The _setting_ in which the swap takes place (Role + Protocols)
- `SwapEnvironment`
- `SwapComposition`: The elements a swap is composed of

cnd/src/start_swap.rs
da-kami: https://github.com/comit-network/comit-rs/pull/2796

cnd/src/storage.rs
da-kami: https://github.com/comit-network/comit-rs/pull/2796

cnd/src/storage.rs
da-kami: https://github.com/comit-network/comit-rs/pull/2796

cnd/src/start_swap.rs
da-kami: https://github.com/comit-network/comit-rs/pull/2796

cnd/src/start_swap.rs
da-kami: Does this still apply to https://github.com/comit-network/comit-rs/pull/2796 ? (brain not able to think any more)

comit/src/lnd.rs
da-kami: https://github.com/comit-network/comit-rs/pull/2796

cnd/src/start_swap.rs
da-kami: https://github.com/comit-network/comit-rs/pull/2796

cnd/src/storage.rs
da-kami: Could you take care of that one @tcharding ? Would appreciate :)

PR #2791, opened 2020-06-04
Remove hard coded regtest network

tcharding:
We have a bunch of places where we hard code the regtest network, for all three ledgers. We already have this network information in the database, all we need to do is expose it via the `Finalized` structs.

Add 'network' field, ether `ledger::Bitcoin` for Bitcoin and Lightning swaps or `ChainId` for Ethereum, to the `Finalized` family of structs.

Use new fields and remove hard code ledger types.

- Patch 1 is clean up; removes the `ledger::Lightning` struct.
- Patch 2 is the PR content described above.

### DIFF
--- a/comit/src/herc20.rs
+++ b/comit/src/herc20.rs
@@ -33,7 +33,7 @@ lazy_static::lazy_static! {
 pub struct CreatedSwap {
     pub asset: asset::Erc20,
     pub identity: identity::Ethereum,
-    pub chain_id: u32,
+    pub chain_id: u32, // FIXME: Why is this not a ChainId?
     pub absolute_expiry: u32,
 }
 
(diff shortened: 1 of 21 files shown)


tcharding:
Sans epic codebase-wide refactoring ;)  (CC @thomaseizinger)

comit/src/herc20.rs
```diff
@@ -33,7 +33,7 @@ lazy_static::lazy_static! {
 pub struct CreatedSwap {
     pub asset: asset::Erc20,
     pub identity: identity::Ethereum,
-    pub chain_id: u32,
+    pub chain_id: u32, // FIXME: Why is this not a ChainId?
```
thomaseizinger, line 36: It should be :)

comit/src/ledger.rs
```diff
@@ -33,6 +33,33 @@ impl From<::bitcoin::Network> for Bitcoin {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum Lightning {
```
thomaseizinger, line 37: Are we actually using this for anything now?

tcharding: rebase fail, I remove this in the first patch and put it back in in the second one - seems like I'm not reading my own commits before raising PRs :(

PR #2795, opened 2020-06-05
Return ledger events in swap response for hbit and herc20

D4nte:
I looked into it and did not find any reason to include ledger events for halight.

### DIFF
--- a/api_tests/src/actors/actor.ts
+++ b/api_tests/src/actors/actor.ts
@@ -6,6 +6,8 @@ import {
     Swap,
     Wallets as SdkWallets,
     HbitHerc20RequestBody,
+    Step,
+    LedgerParameters,
 } from "comit-sdk";
 import { Logger } from "log4js";
 import { E2ETestActorConfig } from "../config";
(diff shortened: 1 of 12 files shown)


cnd/src/http_api/protocol.rs
```diff
@@ -84,74 +94,129 @@ pub trait BetaParams {
...
-    transactions: HashMap<String, String>,
+    events: HashMap<ActionName, String>,
     status: EscrowStatus,
 }
 
 impl LedgerEvents {
-    fn new(status: EscrowStatus) -> Self {
-        Self {
-            transactions: HashMap::new(), /* if we want transaction here, we should save the
-                                           * events to the DB */
```
thomaseizinger, line 97: There was the reason you mentioned in the PR description :)

I think we originally didn't have these here due to the `state` being different but that seems to have fixed itself up again :)

api_tests/src/actors/actor.ts
```diff
@@ -524,6 +534,40 @@ export class Actor {
         });
     }
 
+    private async assertLedgerEventPresence(
+        ledgerRel: "alpha" | "beta",
+        step: Step
+    ): Promise<void> {
+        await this.pollCndUntil(this.swap.self, (swapResponse) => {
```
thomaseizinger, line 541: We don't really need to `pollUntil` here because we are always call this after `assertLedgerStatus` which I think already uses `pollUntil`.

That would allow you to write this logic in a simpler manner and actually have an assertion in here :)
Strictly speaking, this doesn't assert anything.

It is tricky to balance the trade-offs here.

On the one hand, we don't want to have fixed sleeps because it makes the test flaky but just going once to `cnd` and asserting on the response would be nicer in the test.

One of the reasons for this is that our relationship with cnd is one-way: We only read from `cnd` but we never "write" to it and trigger state changes ourselves. Doing that would give us a fixed point after we can assert.

Probably not worth changing something about this now, but maybe you or someone else finds these thoughts interesting :)

PR #2809, opened 2020-06-09
Move time-related methods from cache to dedicated modules

thomaseizinger:
The `Cache` components are designed to be proxies. Hence, they
must not expose any APIs other than the ones from the component
they are proxying.

We achieve that by moving the calculation of Bitcoin's median-time
to the `bitcoin` module in the `comit` crate.

For Ethereum, the impl is so trivial that we just do it inline in
the facade.

### DIFF
--- a/cnd/src/facade.rs
+++ b/cnd/src/facade.rs
@@ -1,4 +1,5 @@
 use crate::{
+    btsieve::LatestBlock,
     connectors::Connectors,
     db::{Save, Sqlite},
     network::{DialInformation, Identities, Swarm},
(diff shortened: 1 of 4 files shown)


cnd/src/facade.rs
```diff
@@ -32,15 +34,24 @@ impl Facade {
             .await
     }
 
-    /// Returns the Bitcoin median time past, used for nLockTime and
-    /// CheckLockTimeVerify.
-    pub async fn bitcoin_median_time_past(&self) -> anyhow::Result<Timestamp> {
-        self.connectors.bitcoin.median_time_past().await
+    /// Returns the current Bitcoin median time past.
+    pub async fn bitcoin_median_time(&self) -> anyhow::Result<Timestamp> {
```
tcharding, line 38: @D4nte and I back and forth'ed on the '_past' part of median time past. I pushed for its inclusion because the term is from the bitcoin ecosystem, they refer to it as 'median time past' not as 'median time'.

thomaseizinger: Really? That is such a confusion term :D

Didn't know, will change back :)

Issue #2781, opened 2020-06-03
Assert side when loading HTLC parameters


luckysori:
Can we have a ticket for deciding on:

- Either loading these individually but with an expectation of which "side" we expect this to be
- Or only loading params in pairs where we assert that what we find matches the side the user requested it on

_Originally posted by @thomaseizinger in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3NDIzMzg2MjU1/pull_request_reviews/more_threads_

luckysori:
Was this already done with #2818 or a follow-up to that one? @tcharding

thomaseizinger:
Yes, this was done in #2818.

PR #2818, opened 2020-06-10
Clean up storage

tcharding:
Refactor the storage module, this one is kinda big. Patches are discreet and explained in their commit logs.

Patch 1 - Fix up use statements
Patch 2 - Shorten the qualified path usage
Patch 3 - Refactor load of the `spawn::Swap` type. This refactor uses the style of generics as done for `Save<CreatedSwap<A, B>>`. Adds a couple of macros also, specifically one is added  because the trait bounds for diesel trait `BelongingtoDsl` was too complicated to work out.
Patch 4 - Creates a sub-module `http_api` under `storage` and puts the load imlps for `AliceSwap` and `BobSwap` in it. Refactors heavily.
Patch 5 - trivial code re-order.

Patches 4 and 5 are the big ones, the diffs are not that easy to read - may be better to check out the branch locally and have a look at whats going on.

If there is anything I can do to make review easier for  future reviews of these big refactor PRs please say.

Thanks


cnd/src/storage.rs
```diff
@@ -56,22 +56,15 @@ impl Storage {
...
-#[async_trait::async_trait]
-impl Load<spawn::Swap<herc20::Params, halight::Params>> for Storage {
-    async fn load(
+impl Storage {
+    // Load swap for trading pair herc20/halight.
+    async fn load_tp_herc20_halight(
         &self,
         id: LocalSwapId,
-    ) -> anyhow::Result<spawn::Swap<herc20::Params, halight::Params>> {
+    ) -> anyhow::Result<(Swap, Herc20, Halight, Option<SecretHash>)> {
```
thomaseizinger, line 66: Could you make `Herc20` and `Halight` generic here with a trait-bound to make the [`belonging_to`](https://docs.rs/diesel/1.4.5/diesel/prelude/trait.BelongingToDsl.html) method available?

thomaseizinger: Even though it is just a private method, having a struct with field names would make it clearer that `.1` is `alpha` and `.2` is `beta`.

If we establish this and succeed in making this generic, this would be a nice place to do https://github.com/comit-network/comit-rs/issues/2781.

cnd/src/storage/http_api.rs
```diff
@@ -0,0 +1,497 @@
...
+            redeem_identity: herc20
+                .redeem_identity
+                .ok_or(db::Error::IdentityNotSet)?
+                .0
+                .into(),
+            expiry: herc20.expiry.0.into(),
+            state,
+        })
+    }
+}
```
thomaseizinger, line 474: This could just be a function analogous to `build_xyz_params` then you don't need the rather awkwardly named abstraction of `Herc20WithState` :)

cnd/src/storage.rs
thomaseizinger: As a follow-up, we can define traits for:

- derive_or_unwrap_secret_hash
- build_herc20_params
- build_halight_params
- build_hbit_params

and make one generic impl that loads `Params`, similar to what we did for `Save` !

thomaseizinger: Can all be a follow-up so we don't make this one too big.

tcharding: > Even though it is just a private method, having a struct with field names would make it clearer that .1 is alpha and .2 is beta.

These are not sides (alpha/beta), this function returns the trading pair - it is agnostic to the side.

cnd/src/storage.rs
```diff
@@ -1,18 +1,13 @@
 use crate::{
     asset,
     db::{
-        self,
-        tables::{Halight, Hbit, Herc20, SecretHash, Swap},
-        wrapper_types::custom_sql_types::Text,
-        ForSwap, NoHalightRedeemIdentity, NoHalightRefundIdentity, NoHerc20RedeemIdentity,
-        NoHerc20RefundIdentity, NoSecretHash, Save, Sqlite,
+        self, ForSwap, Halight, Hbit, Herc20, NoHalightRedeemIdentity, NoHalightRefundIdentity,
```
luckysori, line 6: :thinking: This patch doesn't introduce the problem, but I find the names `Halight`, `Hbit` and `Herc20` too general. Even if you're in the context of the `storage` module. I would find it clearer if we used them with a more qualified name, e.g `table::Halight` or even `db::table::Halight` (sorry I like long names [sometimes]).

luckysori: `db::<protocol-name>` would be better, but still a bit lacking imo.

cnd/src/storage.rs
```diff
@@ -1,18 +1,13 @@
...
+        NoHerc20RedeemIdentity, NoHerc20RefundIdentity, NoSecretHash, Save, SecretHash, Sqlite,
+        Swap, Text,
     },
     halight, hbit, herc20, http_api, identity,
     network::{WhatAliceLearnedFromBob, WhatBobLearnedFromAlice},
-    seed::RootSeed,
-    spawn,
-    swap_protocols::state::Get,
-    LocalSwapId, Protocol, Role, Side, SwapContext,
+    spawn, Get, LocalSwapId, Protocol, Role, RootSeed, Side, SwapContext,
```
luckysori: :upside_down_face: **Very** minor nit, but importing `Get` by itself is not great. Maybe it doesn't matter because it's just in the imports section, but I would rather see something like `state::Get`.

cnd/src/storage/http_api.rs
luckysori: > What I really wanted to do was implement this trait on a tuple - seems we can't do that in Rust ATM though.

Haha, I always try to do this :(

cnd/src/storage.rs
tcharding: ok, I'm good with that. Will do.

cnd/src/storage.rs
tcharding: I have to disagree with you here mate. Apart from the fact that I always favour short names; the storage module is different in that it fully 'owns' the database, so bringing types from the db into scope is the natural thing to do IMO. The fact that the new tables are in the tables module is not important, when I'm writing/reading code in the storage module it is just like we are in the db layer so all the types are implicitly db types. Does that make sense? We don't have to agree :) (on this point, that's why I originally brought `SecretHash` into scope from the db and used `comit::SecretHash` when needed, you did it the other way around and I left it as such when I rebased :)

luckysori: Yeah, I'm okay with this even if I don't love it :wink:

thomaseizinger: > the storage module is different in that it fully 'owns' the database

Would be nice to express that in our module hierachy.

cnd/src/storage.rs
```diff
@@ -158,197 +158,228 @@ impl Load<SwapContext> for Storage {
...
-
-        let (swap, halbit, herc20, secret_hash) = self
-            .db
-            .do_in_transaction::<_, _, anyhow::Error>(move |conn| {
-                let key = Text(id);
-
-                let swap: Swap = swaps::table
-                    .filter(swaps::local_swap_id.eq(key))
-                    .first(conn)?;
+pub trait LoadTables<A, B> {
```
thomaseizinger, line 131: Nit: Rename to `LoadProtocolTables`?

cnd/src/storage.rs
```diff
@@ -158,197 +158,228 @@ impl Load<SwapContext> for Storage {
...
+                            .first::<db::SecretHash>(conn)
+                            .optional()?;
+
+                        Ok((swap, alpha, beta, secret_hash))
+                    })
+                    .await
+                    .context(db::Error::SwapNotFound)?;
+
+                alpha.assert_side(Side::Alpha)?;
+                beta.assert_side(Side::Beta)?;
```
thomaseizinger, line 186: This is the only place where we use `assert_side` right?

Why not just:

```suggestion
	if alpha.side.0 != Side::Alpha || beta.side.0 != Side::Beta {
		anyhow::bail!("...")
	}
```

cnd/src/storage.rs
tcharding: Oh, I like that.

`db` and `seed` being private scope modules under `storage`. Will wait for the remove of rfc003 before doing this, adding to my TODO list.

cnd/src/storage.rs
tcharding: I had that first but it didn't build. Unless I'm mistaken, we don't know the types here so we need to use a trait.

cnd/src/storage.rs
tcharding: Yep can do.

tcharding: On second thoughts, this does not lead _just_ the protocol tables, it also includes the Swaps table and the SecretHash table. Admittedly this is a very generic name and will likely clash next time we need to load a set of different tables. Leaving as is for now.

cnd/src/storage.rs
thomaseizinger: This should work. You are inside a macro so the code will just expand. Isn't the field called `side` for all tables?

PR #2832, opened 2020-06-11
Use halbit in the CHANGELOG

tcharding:
Next release is the first to mention the new split protocols, lets not
muddy the water by using old terminology.

Use halbit in the CHANGELOG to refer to the new Lightning protocol.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
--   Support for `herc20-halight`, `halight-herc20`, `hbit-herc20` and `herc20-hbit` swaps.
+-   Support for `herc20-halbit`, `halbit-herc20`, `hbit-herc20` and `herc20-hbit` swaps.
 
 ## [0.7.3] - 2020-04-14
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 -   Fix windows build.
--   Return an error when creating a halight swap if lnd certificate or macaroon are unavailable instead of failing silently.
+-   Return an error when creating a halbit swap if lnd certificate or macaroon are unavailable instead of failing silently.
 -   Only return a single action on the get swaps endpoint. This is now the recommended action available to the user.
```
thomaseizinger, line 14: This one is not "fixed" compared to the previous version (0.7.3) because the endpoint this is addressing doesn't even exist in the previous version. For end users, this is the normal behaviour of the newly introduced endpoint.

Not sure if anyone actually reads our changelogs but that is something to keep in mind when writing them.

cc @D4nte

thomaseizinger: If we want to mention this somewhere, then I would suggest a release announcement blog post where we have more real estate to talk about this.

tcharding: Cool, will remove. Was not added by this patch but its worth cleaning up. Will do it as patch 0 on this PR and force push.

PR #2833, opened 2020-06-12
Make caches actually useful

thomaseizinger:
Turns out what I did in https://github.com/comit-network/comit-rs/pull/2830 was total garbage :D

This PR fixes that.


D4nte:
Ok so we need to use the Cache action *before* it's used and written, right?

PR #2834, opened 2020-06-12
Move Comit NetworkBehaviour to comit crate

tcharding:
The `NetworkBehaviour` which composes together all the communication protocols is called `Comit`; `Comit` should live in the comit crate so that users of that crate do not have to compose the communication protocols themselves.

Patch 1 and 2 are preparation
Patch 3 - Move `Comit` to the comit crate
Patch 4 and 5 are cleanup of use statements


comit/src/swap_id.rs
```diff
@@ -2,6 +2,55 @@ use serde::{Deserialize, Serialize};
 use std::{fmt, str::FromStr};
 use uuid::Uuid;
 
+/// This is an identifier created, and used locally, by a node to identify a
+/// swap to this node (as apposed to a ShareSwapId below).
+#[derive(Serialize, Deserialize, Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct LocalSwapId(Uuid);
```
thomaseizinger, line 8: I'd be curious if we can model things in a way that don't require `LocalSwapId`.

cnd/src/network.rs
```diff
@@ -837,7 +833,7 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<comit::BehaviourOutEvent> for C
         let save_and_start_swap = async move {
             let swap = storage.load(local_swap_id).await?;
             save_swap_remote_data(&storage, swap, remote_data).await?;
-            spawn(&spawner, &storage, swap).await?;
+            spawn::spawn(&spawner, &storage, swap).await?;
```
thomaseizinger, line 836: Why `spawn::spawn`? Seems pretty redundant to me :)

tcharding: Rust idiomatic ruling is to call functions that are from another module with the qualified path. We do not tend to do this but in this case it made the use statements less noisy so I favoured this syntax.

comit/src/swap_id.rs
tcharding: Just within the network module you mean, right?

thomaseizinger: Yes.

cnd/src/network.rs
thomaseizinger: > Rust idiomatic ruling is to call functions that are from another module with the qualified path.

Is this general statement actually true though? I never saw it as such a hard rule :)
I use qualified paths for functions if they add value to the name.

da-kami:
I missed that one - great seeing this moving!

My favourite patch: `Move Comit to comit crate` :)

PR #2838, opened 2020-06-12
Add a couple things to COMIT lib for Nectar

luckysori:
I think none of this should be contentious.

It turns out that, in Nectar, using the low-level API `watch_for_EVENT` enables better code composition than if we used `watch_ledger`. I don't think making it public could enable harmful patterns by consumers of the library. It appears to be a very useful building block.

Before patches https://github.com/comit-network/comit-rs/commit/45df400b8a6b39d6b91caf76cdf24ddf9e060d84 and https://github.com/comit-network/comit-rs/commit/6dd9fd48ca7b82e17d80a47ce678ee93847ffdb6, the action building logic lived in the HTTP API module of cnd (the only place where it was needed). Other applications will need this code, so moving it to the COMIT library is needed.

We don't need to get this merged soon, because we're just building against the branch anyway, but I want to know if there is consensus that these are good, correctly implemented changes.

### DIFF
--- a/comit/src/hbit.rs
+++ b/comit/src/hbit.rs
@@ -146,7 +146,7 @@ where
     Ok(())
 }
 
-async fn watch_for_funded<C>(
+pub async fn watch_for_funded<C>(
     connector: &C,
     params: &Params,
     start_of_swap: NaiveDateTime,
@@ -179,7 +179,7 @@ where
     Ok(event)
 }
 
-async fn watch_for_redeemed<C>(
+pub async fn watch_for_redeemed<C>(
     connector: &C,
     params: &Params,
     location: htlc_location::Bitcoin,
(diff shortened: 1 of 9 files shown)


comit/src/hbit.rs
```diff
@@ -221,14 +225,104 @@ where
...
+            fund_amount,
+            fund_location,
+            refund_address,
+            Box::new(move |htlc: BitcoinHtlc| htlc.unlock_after_timeout(secp, transient_refund_sk)),
+        )
+    }
+
+    pub fn build_redeem_action<C>(
+        &self,
+        secp: &'static Secp256k1<C>,
```
luckysori, line 271: Does anyone know if we can get rid of this `'static`?

luckysori: I can! No need for dynamic dispatch here (https://users.rust-lang.org/t/solved-understanding-why-a-closure-requires-static-lifetime/23887).

PR #2842, opened 2020-06-13
Pre rfc003 removal fixes

tcharding:
In preparation for removing rfc003 we have a bunch of places that need attention. 

We have artefacts from rfc003 remaining in the code, our e2e tests pass because we hit rfc003 endpoints and code paths. In order to remove rfc003 all this stuff needs to work with only the 'new' code.

Patches are discreet and self descriptive.

This PR does _most_ of the work, we still need to re-write the following test files to use the new actors
- `multiple_peers.ts`
- `peers_using_ip.ts`
- `siren_schema` (incl. `swap.schema.json`)


api_tests/tests/sanity.ts
```diff
@@ -68,6 +68,66 @@ describe("Sanity", () => {
         })
     );
 
+    it(
+        "returns-invalid-body-for-bad-json-herc20-halbit",
+        oneActorTest(async ({ alice }) => {
+            const promise = alice.createHerc20Halbit({
+                // @ts-ignore
+                garbage: true,
```
D4nte, line 27: We already have some sanity test for the new endpoint (search `create-herc20-halbit-returns-invalid-body`). Please merge them all together.

tcharding: Cool, thanks. Will do.

tcharding: Done. Two additional patches pushed onto this PR.

api_tests/src/actor_test.ts
```diff
@@ -8,7 +8,7 @@ import ProvidesCallback = jest.ProvidesCallback;
  * after the test, regardless if the test succeeded or failed.
  */
 function nActorTest(
-    actorNames: ["alice", "bob", "charlie"] | ["alice", "bob"] | ["alice"],
```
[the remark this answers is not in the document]
tcharding: Yes, I used the new thing of a PR targeting a branch then relying on that branch to be merged and github update the target brach to dev, which it has done. Not sure how this effects review though.

api_tests/tests/sanity.ts
```diff
@@ -19,40 +19,6 @@ describe("Sanity", () => {
         })
     );
 
-    it(
-        "bad-request-for-invalid-swap-combination",
```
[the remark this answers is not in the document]
tcharding: Sure thing, can do.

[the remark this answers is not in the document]
tcharding: Added a patch with some tests for this, is this what you meant?

api_tests/src/actor_test.ts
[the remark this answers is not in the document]
thomaseizinger: Weird, did you rebase the original branch after you opened this one?

api_tests/tests/sanity.ts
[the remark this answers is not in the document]
tcharding: `6e21c3052 e2e: Add sanity tests for invalid routes`

api_tests/src/actor_test.ts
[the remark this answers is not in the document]
thomaseizinger: Just checked, it is not the same commit hash than the one that was merged.

If you want to use the PR retargeting, I think you have to layer branches on top of each other and don't modify the original commits.

api_tests/tests/sanity.ts
```diff
@@ -6,6 +6,60 @@ import SwapFactory from "../src/actors/swap_factory";
 // ******************************************** //
 
 describe("Sanity", () => {
+    it(
+        "invalid-path-yields-405",
```
thomaseizinger, line 10: 405 means something entirely different and you are also not asserting what you are saying you are asserting :D

api_tests/tests/sanity.ts
```diff
@@ -6,6 +6,60 @@ import SwapFactory from "../src/actors/swap_factory";
...
+        oneActorTest(async ({ alice }) => {
+            // @ts-ignore (access to private field 'client')
+            const promise = alice.cnd.client.post("swaps/hbit/hbit", {});
+            await expect(promise).rejects.toHaveProperty("name", "Error");
```
thomaseizinger, line 14: These assertions are pretty bad. They will pass for whatever reason the code threw an error :/

It would be good to assert the actual full problem response that we get from `cnd` so we can assert the status and title.

api_tests/tests/sanity.ts
tcharding: I'll give you that I am not necessarily asserting its a 405 since I only check for 'name: Error' but curl tells me that its a 405 being returned. If that's the wrong thing to return that's upstream's business I believe since this is, if I understand correctly, a warp error that we do not control.

tcharding: or are you saying the term 'path' is wrongly used when talking about an endpoint/route? Please do correct me on HTTP stuff, I am not a web developer so I definitely need these things pointed out to me if I get them wrong.

api_tests/tests/sanity.ts
tcharding: I don't think we control this do we. FTR this is what we get back
```
    +   "code": "ECONNREFUSED",
    +   "columnNumber": undefined,
    +   "config": Object {
    +     "adapter": [Function httpAdapter],
    +     "baseURL": "http://0.0.0.0:36697",
    +     "data": "{}",
    +     "headers": Object {
    +       "Accept": "application/json, text/plain, */*",
    +       "Content-Length": 2,
    +       "Content-Type": "application/json;charset=utf-8",
    +       "User-Agent": "axios/0.19.2",
    +     },
    +     "maxContentLength": -1,
    +     "method": "post",
    +     "timeout": 0,
    +     "transformRequest": Array [
    +       [Function transformRequest],
    +     ],
    +     "transformResponse": Array [
    +       [Function transformResponse],
    +     ],
    +     "url": "swaps/hbit/hbit",
    +     "validateStatus": [Function validateStatus],
    +     "xsrfCookieName": "XSRF-TOKEN",
    +     "xsrfHeaderName": "X-XSRF-TOKEN",
    +   },
    +   "description": undefined,
    +   "fileName": undefined,
    +   "lineNumber": undefined,
    +   "message": "",
        "name": "Error",
    +   "number": undefined,
    +   "stack": undefined,
```

tcharding: And if you are interested, warp returns 405 because, again, if I understand it correctly, warp maintains a hierarchy of errors and returns the highest one. I don't have the tab open where I read that now though.

PR #2846, opened 2020-06-15
Re-export rust-bitcoin from COMIT lib

luckysori:
All current swaps involve Bitcoin and depend on the `rust-bitcoin` library. Consumers of COMIT lib can now rely on a specific version of `rust-bitcoin` through this re-exported dependency, avoiding problems with version incompatibility.

Is this all that needs to be done? Did I completely miss the point? :stuck_out_tongue:

### DIFF
--- a/comit/src/lib.rs
+++ b/comit/src/lib.rs
@@ -18,6 +18,9 @@ mod swap_id;
 mod timestamp;
 pub mod transaction;
 
+// Re-export rust-bitcoin to make our consumer's life easier.
+pub use ::bitcoin as bitcoin_lib;
+
 pub use self::{
     network::DialInformation,
     secret::Secret,
@@ -25,6 +28,7 @@ pub use self::{
     swap_id::{LocalSwapId, SharedSwapId},
     timestamp::{RelativeTime, Timestamp},
 };
+
 use digest::ToDigestInput;
 
 #[derive(


comit/src/bitcoin.rs
```diff
@@ -17,6 +17,8 @@ use serde::{
 };
 use std::{fmt, str::FromStr};
 
+pub use bitcoin::*;
```
D4nte, line 20: 📣 This should be in the `lib.rs` file. The consumer should be able to just do: `use comit::rust_bitcoin::Amount` and not `comit::bitcoin::bitcoin::Amount`.

thomaseizinger: `comit::rust_bitcoin::Amount` is pretty ugly :sweat_smile: 
I would like to avoid that if possible.

The current code would result in `comit::bitcoin::Amount` which is fine IMO.

thomaseizinger: The question is, do we even need to re-export `rust-bitcoin` at all?

Maybe we just shouldn't do that.

Which types do you actually need @luckysori?
Maybe we can find a better solution for the specific problems.

luckysori: > Which types do you actually need @luckysori?

I defer to @D4nte, since he requested these changes.

D4nte: > `comit::rust_bitcoin::Amount` is pretty ugly 😅
> I would like to avoid that if possible.
> 
> The current code would result in `comit::bitcoin::Amount` which is fine IMO.

As a library consumer, I expect the re-export to happen in `lib.rs`.

Having the re-export in a sub-module would guarantee the following scenario to happen:
- Want to use lib `foo`,
- Realises I need to use `lib` bar types,
- Check `lib.rs` and do not see re-export,
- Grunt and add `bar` to my `Cargo.toml`,
- Update libraries and manually sync my `bar` version and the `bar` version of `foo/Cargo.toml`, grunt more,
- Go do a PR on `foo` to re-export `bar`,
- Get told by maintainer that `bar` is re-exported in `bar::foo` module,
- Grunt more.

thomaseizinger: Which types do we actually need @D4nte? Aren't you gonna need `rust_bitcoin` anyway in `nectar` for the wallet and stuff?

Re-exporting libraries is useful if the COMIT lib would be the only lib that is bringing in rust-bitcoin and you only need to access those types because you want to interact with COMIT.

I am happy to re-export those, which then should go under the `bitcoin` module as it is happening now.

D4nte: My aim here is to avoid doing like rust web3 and making it a nightmare for comit library consumers to manage and coordinate dependencies.

Library that selectively re-export the types of a given dependency are worse than libraries that do not re-export at all because.

By selectively exporting some of the `bitcoin` types, you make a `comit` lib consumer believe that they do not need to add `bitcoin` to their `Cargo.toml`. Then, as soon as they want to do anything with `bitcoin` types, they realise they do not have access to some of the `bitcoin` type and have to import `bitcoin` in their `Cargo.toml.`

Now we are back to square 1 where the consumer has to coordinate between `comit` version and `bitcoin` version which is a pain, as we know from our `web3` experience.

I do not know how to make myself clearer: **I do not want to have to coordinate between the version of `bitcoin` and `comit` in the nectar `Cargo.toml` file.**

thomaseizinger: > I do not know how to make myself clearer: I do not want to have to coordinate between the version of bitcoin and comit in the nectar Cargo.toml file.

I fully support that but there is more than one way to resolve this problem.

I am concerned about the public API of the COMIT library that we are creating here and the consequences of such re-exporting decision.
If we just add re-exports of libraries as we please, maintaining backwards compatibility and avoiding breaking changes is going to be hard because effectively every dependency upgrade of a library is going to be a breaking change. Constant dependency upgrades are also very painful.

The other thing I am concerned about are the aesthetics of the public API. The more things we export, the more cognitive load we impose on clients in terms of what they should use and when.

Generally, we have two ways of going about this:

1. We don't expose any types of dependencies in our public APIs. This way, which dependencies (and which versions) we are using is an implementation detail and we don't need to re-export dependencies.
2. We expose types of our dependencies in public APIs and re-export the dependency.

For some libraries it makes the most sense to go for (2). This is typically the case if the library stays within the same domain and just enhances it with more functionality. `web3` is a good example here. Re-exporting is the much better choice because the dependency `primitive-types` was only extracted so parity could re-use those types between substrate and parity-ethereum. One could argue that they should have never been separate crates in the first place but rather find other ways of reducing that duplication or simply duplicate. In the end, parity-ethereum and substrate have little reason to actually share the same types here.

For us, I think it is the better choice to make our dependency on `rust-bitcoin` an implementation detail instead of re-exporting it because:

- We already have new-types for representing things like `PublicKey`s and `Amount`s so there is probably fairly little work left to hide it completely from the public API.
- `nectar` has its own, completely separate requirements for depending on `rust-bitcoin`. Relying on `comit` to re-export `rust-bitcoin` creates coupling that is not strictly necessary.

Keeping versions manually in sync is only necessary if the library in questions releases breaking changes. Otherwise cargo will happily resolve a compatible version between the specification in both `Cargo.toml` files. This means our efforts should be directed in properly encapsulating our dependencies so we can actually stop this bubbling up of breaking changes to our consumers instead of exposing it to them.

If we decide to keep a very small foot-print of `rust-bitcoin` within our public API (like a `From` implementation), then how painful upgrades are is still a function of how many breaking changes `rust-bitcoin` releases and how fast we upgrade to the most recent version.

PR #2850, opened 2020-06-15
Remove multiple peers tests

tcharding:
We currently have a few tests that attempt to test the ability to do swaps with multiple peers concurrently. There are a number of issues with these tests that have led me to the conclusion that the best way forward is to delete them and create tickets for there re-introduction. The problems are:

- They use rfc003 Actor class
- They use rfc003 swaps
- They only test the swap occurs in a 'shallow' manner
 
The functionality these tests set out to test is totally useful and we would like to test it. Our ability to write these tests using the new swap protocols is hampered by:

- Actor implicitly assumes there is only a single swap
- newSwap assumes it is called once and once only
- theeActorTest does not work for the new swap protocols

While these things can all be overcome, they are non-trivial and will require heavy modification to the test infrastructure. Having the current rfc003 multiple peers tests in place adds no value to either the new swap protocols or to the developer who has to implement multiple peer tests for the new swap protocols.

Delete the rfc003 multiple peers tests. Done as three separate patches with explanation for each test removed in the commit log of each patch.

Tracking issue for re-introduction of multiple peers tests: https://github.com/comit-network/comit-rs/issues/2852


api_tests/tests/peers_using_ip.ts
```diff
@@ -30,38 +30,6 @@ async function assertPeersAvailable(actor: Actor, peers: Actor[]) {
 }
 
 describe("Peers using IP tests", () => {
-    it(
-        "alice-send-request-wrong-peer-id",
```
thomaseizinger, line 41: This test actually doesn't need `Carol`, should we port it to using `twoActorTest`?

tcharding: AFAICT we have not defined the behaviour we want for peer ids for the new swap protocols. It is not possible to port this test because of that reason. Should we be defining this now? That question is the reason I elected to delete it, I don't think we should but I also don't think that that should block the removal of rfc003 because it is totally unrelated. rfc003 should not be used as a proxy todo list so lets delete this and track the issues somewhere and not rely on rfc003 code to show us what we still need to do.

tcharding: Tracking issue: https://github.com/comit-network/comit-rs/issues/2856

PR #2854, opened 2020-06-16
Remove the ethereum::Bytes wrapper

### LINKED ISSUE #2416: The `ethereum::Bytes` type is not useful in any way
## Problem

We inherited a struct called `Bytes` from the `web3` library that simply wraps a `Vec<u8>` and serializes and deserializes to hex.
The struct doesn't add any value:

- The name couldn't be more generic. Everything is a byte in engineering.
- There are no useful functions on this type.

## Goal

No unnecessary abstractions.

## Recommendation

Remove this particular type and replace all the current usage sites with something more descriptive, which could just be a `Vec<u8>` if nothing else is more appropriate.

thomaseizinger:
The main purpose of this wrapper was serialization to and from hex.
This is a concern of the respective API layer and should hence be
co-located with the consuming code in the REST API.

Resolves #2416.

### DIFF
--- a/cnd/src/http_api/action.rs
+++ b/cnd/src/http_api/action.rs
@@ -51,7 +51,7 @@ pub enum ActionResponseBody {
         min_median_block_time: Option<Timestamp>,
     },
     EthereumDeployContract {
-        data: crate::ethereum::Bytes,
+        data: EthereumData,
         amount: asset::Ether,
         gas_limit: crate::ethereum::U256,
         chain_id: ChainId,
@@ -59,7 +59,7 @@ pub enum ActionResponseBody {
     EthereumCallContract {
         contract_address: identity::Ethereum,
         #[serde(skip_serializing_if = "Option::is_none")]
-        data: Option<crate::ethereum::Bytes>,
+        data: Option<EthereumData>,
         gas_limit: crate::ethereum::U256,
         chain_id: ChainId,
         #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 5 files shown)


luckysori:
@thomaseizinger: From experience this is a breaking change :stuck_out_tongue: Could you add it to the changelog.

thomaseizinger:
Huh? You mean on the REST API?

luckysori:
> Huh? You mean on the REST API?

No, I meant the disappearance of the `Bytes` type. We were relying on its deserialization in Nectar.

thomaseizinger:
> > Huh? You mean on the REST API?
> 
> No, I meant the disappearance of the `Bytes` type. We were relying on its deserialization in Nectar.

Oh, yes from a COMIT lib perspective, this is a breaking change.
I am a bit reluctant to add these things to the changelog now already. I guess we are gonna make many breaking changes over the upcoming weeks? Documenting all of those seems a bit over the top no?

How exactly were you relying on it? Because the _deserialization_ from web3 is still the same, just with one less level of nesting.

Thinking about it, removing this may have been a wrong call given that there is actually a canonical serialization for this concept of "data" in the Ethereum ecosystem: https://eth.wiki/json-rpc/API

I am gonna make a new ticket to fix this.

luckysori:
> I am a bit reluctant to add these things to the changelog now already. I guess we are gonna make many breaking changes over the upcoming weeks? Documenting all of those seems a bit over the top no?

You're probably right :)

> How exactly were you relying on it? Because the deserialization from web3 is still the same, just with one less level of nesting.

Yep, I just had to add `#[serde(with = "SerHexSeq::<StrictPfx>")]` wherever I needed to deserialize bytes.

> Thinking about it, removing this may have been a wrong call given that there is actually a canonical serialization for this concept of "data" in the Ethereum ecosystem: https://eth.wiki/json-rpc/API

It would make sense to reintroduce it then, cool!

PR #2855, opened 2020-06-16
Siren for new swap protocols

tcharding:
The current siren validation is based on rfc003. In preparation for removing rf003 translate the schema to the new swap protocols.

- Patch one adds a schema for the root endpoint. This one is reasonably complete.
- Patch two removes the rfc003 "/swaps/:id" schema and adds a trivial replacement.

I spent a while on trying to write a better schema for "/swaps/:id" but was not able to come up with anything decent. It is questionable whether this is even valuable to have so I elected to not spend any more time on it.


api_tests/tests/siren_schema.ts
```diff
@@ -24,7 +30,7 @@ async function assertValidSirenDocument(
     const swapResponse = await alice.cnd.fetch(selfLink);
     const swapEntity = swapResponse.data as siren.Entity;
 
-    expect(swapEntity).toMatchSchema(sirenJsonSchema);
+    expect(swapEntity).toMatchSchema(sirenSwapJsonSchema);
```
thomaseizinger, line 33: There is a slight mixup here :)

This line used to validate that the whole response is valid Siren, ignoring the properties of the specific entity. We should keep that validation :)

api_tests/tests/siren_schema.ts
```diff
@@ -24,7 +30,7 @@ async function assertValidSirenDocument(
...
     expect(swapEntity.properties).toMatchSchema(swapPropertiesJsonSchema);
```
thomaseizinger, line 34: This is the one that you can actually delete given that we no longer provide these properties.

api_tests/tests/siren_schema.ts
```diff
@@ -1,19 +1,25 @@
...
 import * as swapPropertiesJsonSchema from "../swap.schema.json";
-import { Rfc003Actor } from "../src/actors/rfc003_actor";
+import { SwapResponse } from "../src/payload";
 
 // ******************************************** //
 // Siren Schema tests                                 //
 // ******************************************** //
 
-async function assertValidSirenDocument(
+async function assertValidSirenSwapDocument(
```
thomaseizinger, line 22: If you look closely on what this function does, I think you will come to the conclusion that it is pretty useless because we already have the link to the swap outside now, thanks to the new API :)

thomaseizinger: I think we can just inline the single assertion that the document is valid siren.

api_tests/swap.schema.json
```diff
@@ -1,375 +1,7 @@
...
-                            "title": "The Status Schema",
-                            "examples": ["SENT"]
-                        }
-                    }
-                }
-            }
-        }
-    }
+    "title": "Swap",
+    "description": "/swaps/:id schema"
```
thomaseizinger, line 6: I would just delete this schema, it doesn't really serve a purpose :D

tcharding:
Did some pair hacking with @thomaseizinger on this, merging without approval.

PR #2858, opened 2020-06-16
Few fixes to migrations.sh

tcharding:
A few fixes to the `migrations.sh` script. The first minor the second fundamental, please review carefully to check I'm not missing something. 

- Patch one fixes the fact that the script does not run on Linux using `./migrations.sh`

- Patch two fixes the fact that the script does not get the current version correctly, it currently returns v0.7.3 and we are on v0.8.0

### DIFF
--- a/cnd/migrations.sh
+++ b/cnd/migrations.sh
@@ -4,7 +4,7 @@ function navigate_to_last_folder() {
   cd $(ls -d */|tail -n 1)
 }
 
-RELEASE_TAG=$(git describe --tags)
+RELEASE_TAG=$(git tag | tail -1)
 RELEASE_VERSION=$(echo $RELEASE_TAG | cut -d'-' -f 1)
 echo "Checking for current release version ${RELEASE_VERSION}"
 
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Interesting, we seem to be doing something wrong with our release branch merging. `git describe` _should_ work!

thomaseizinger:
I think a problem with your approach could be that it doesn't take into account the current HEAD. It just gives you the latest release regardless of where you are in tree.

tcharding:
> I think a problem with your approach could be that it doesn't take into account the current HEAD. It just gives you the latest release regardless of where you are in tree.

Oh, you are right. Easy fix; we could add a check and warn if running the script on a branch other than dev?

E.g.
"You are not on dev, if there is a tagged commit in dev that is not in this branch script output will be incorrect"

thomaseizinger:
Do you mind looking into why `git describe` doesn't cut it?

tcharding:
> Do you mind looking into why `git describe` doesn't cut it?

Sure thing, can do.

PR #2865, opened 2020-06-17
Remove the swap_protocols module

tcharding:
Swap protocols is _all_ we do, we do not need this layer of abstraction.

Move the `state` module to the root cnd module.

Move the `swap_protocols::actions` module code into the `crate::actions` module.

Remove the `swap_protocols` module altogether.


cnd/src/actions.rs
```diff
@@ -1,5 +1,17 @@
 pub use comit::actions::*;
 
+/// Common interface across all protocols supported by COMIT
+///
+/// This trait is intended to be implemented on an Actor's state and return
+/// the actions which are currently available in a given state.
+pub trait Actions {
```
thomaseizinger, line 7: This is an rfc003 artifact.

PR #2872, opened 2020-06-17
Refactor finalize logic

tcharding:
Attempt to simplify the `finalize` logic by doing an early return any time a condition is not met (incl. code comment as to why).

If none of the return statements are hit then we are ready to finalize.

### DIFF
--- a/comit/src/network/comit.rs
+++ b/comit/src/network/comit.rs
@@ -210,32 +210,34 @@ impl Comit {
         let remote_data = self.remote_data.get(&shared_swap_id);
 
         if let (Some(state), Some(data), Some(remote_data)) = (state, data, remote_data) {
-            let ethereum_sorted = if data.ethereum_identity.is_some() {
-                remote_data.ethereum_identity.is_some() && state.ethereum_identity_sent
-            } else {
-                true
-            };
-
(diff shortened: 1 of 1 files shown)


comit/src/network/comit.rs
```diff
@@ -210,32 +210,34 @@ impl Comit {
...
+                && (remote_data.lightning_identity.is_none() || !state.lightning_identity_sent)
+            {
+                // Swap not yet finalized, Lightning identities not synced.
+                return;
+            }
+
+            if data.bitcoin_identity.is_some()
+                && (remote_data.bitcoin_identity.is_none() || !state.bitcoin_identity_sent)
+            {
+                // Swap not yet finalized, Bitcoin identities not synced.
```
thomaseizinger, line 230: I really like the notion of `synced`! Been struggling with finding a nice terminology here but `sync` could work really well :)

"Before a swap can take place, the parties first use the announce protocol verify essential properties of the swap like amounts and expiry times. After that, they synchronize technical execution parameters like secret-hash and identities."

tcharding: Cheers, yeah seems like a good term for what we are doing.

tcharding:
No one is reviewing, I'm kicking this one in

bors r+

PR #2898, opened 2020-06-26
Remove panicing functions from public interfaces of Address and Hash

thomaseizinger:


comit/src/ethereum.rs
```diff
@@ -108,25 +111,11 @@ impl From<Hash> for [u8; 32] {
...
-        h256.0.copy_from_slice(src);
-        h256
-    }
-
     pub fn as_bytes(&self) -> &[u8] {
         &self.0
     }
 }
 
-impl FromStr for Hash {
```
thomaseizinger, line 122: This was never used besides the test :)

PR #2902, opened 2020-06-26
An initial attempt to split out the announce protocol from the execution parameter exchange

thomaseizinger:
10% conceptual review please. The last two patches are the most interesting.

For the orderbook, we are introducing a sister protocol to the current announce protocol: The taker order protocol.
When the orderbook is in use, the taker order protocol is used instead of the announce protocol to agree on a `SharedSwapId`.

I am therefore imagining the following refactoring to the current network layer:

1. Split away the announce protocol from `Comit`, now renamed to `ExecutionParameters`.
2. Move functionality regarding announcement actually into the announce network behaviour. At the moment, the NB is pretty dumb and only forwards the events. We can actually move all of the logic regarding swap digests, timeouts etc into there and make this layer more useful.

In the end, I think we should have four NetworkBehaviours next to each other in `cnd`:

- `Announce`
- `TakeOrder`
- `Orderbook`,
- `ExecutionParameters`

The overall goal would be to make composing them low effort. At the moment, most of the logic of announce is actually inside the `Comit` and `Swaps`. My plan with (2) is to move it out of there and co-locate it in the `Announce` protocol.

Not all of the codebase compiles currently but you can run the test an it passes. If people agree, I would move forward with more tests similar as the one added in patch 3 and remove that functionality from the `ExecutionParameters` network behaviour.

There is one aspect which I wasn't quite sure about how to tackle it:

At the moment, Bob just creates a `SharedSwapId` ad-hoc if his caller gave him a `SwapDigest` that Alice is trying to confirm. Should the generation of this ID maybe be moved up to the caller? In other words, should `Announce::await_announcement` take the `SharedSwapId` that should be used for confirming as a parameter?
It seems a little odd that the caller doesn't have any control over this. On the other hand, I don't feel like there is a need to control it.


tcharding:
> There is one aspect which I wasn't quite sure about how to tackle it:

> At the moment, Bob just creates a SharedSwapId ad-hoc if his caller gave him a SwapDigest that Alice is trying to confirm. Should the generation of this ID maybe be moved up to the caller? In other words, should Announce::await_announcement take the SharedSwapId that should be used for confirming as a parameter?
It seems a little odd that the caller doesn't have any control over this. On the other hand, I don't feel like there is a need to control it.

Since the `SharedSwapId` is only ever used internally within the network layers I don't see any real benefit in exposing this to the caller.

tcharding:
The concepts in this PR seem good to me. I am missing knowledge of the order book though so I may be somethings.

thomaseizinger:
> > There is one aspect which I wasn't quite sure about how to tackle it:
> 
> > At the moment, Bob just creates a SharedSwapId ad-hoc if his caller gave him a SwapDigest that Alice is trying to confirm. Should the generation of this ID maybe be moved up to the caller? In other words, should Announce::await_announcement take the SharedSwapId that should be used for confirming as a parameter?
> > It seems a little odd that the caller doesn't have any control over this. On the other hand, I don't feel like there is a need to control it.
> 
> Since the `SharedSwapId` is only ever used internally within the network layers I don't see any real benefit in exposing this to the caller.

The "caller" in this case would be the piece of code that mediates between `Announce` and `ExecutionParameters`.

Meaning, it will be exposed one way or another. Either by passing it in or by receiving it as a result of the confirmation that is emitted by `Announce`.
In other words, it is not really a question of encapsulation but only precisely about the construction of `SharedSwapId`.

The argument I had in my had was something like this:

- Not exposing it allows us to delay the creation of the `SharedSwapId` for as long as possible, until we actually need it.
- Not exposing it also makes the protocol symmetric for both parties: You only learn the `SharedSwapId` once the swap is confirmed.
- Given that Bob's node chooses the `SharedSwapId`, it does feel a bit odd that this happens inside `Announce`.

Conclusio: I was able to write a test that doesn't need the `SharedSwapId` to be passed in, hence I didn't bother and just create it on the spot.

rishflab:
Is this new execution paramters network behaviour for exchanging execution/swap parameters (identities, expiries etc) or would these exchanged earlier and then confirmed in this behaviour?

thomaseizinger:
> Is this new execution paramters network behaviour for exchanging execution/swap parameters (identities, expiries etc) or would these exchanged earlier and then confirmed in this behaviour?

This is a subset of what used to be named `Comit`. It exchanges secret-hash and identities with a finalize in the end.

Expiries should be part of the order :)

thomaseizinger:
Closing this in favor of a smaller PR, thanks for the input!

PR #2912, opened 2020-07-01
Only cache target dir on ubuntu

thomaseizinger:
Our caches are so big that they invalidate them constantly. Ubuntu
takes the longest to run with the e2e tests, so it would benefit
the most from caching.


thomaseizinger:
Before: 
image

After:
image

PR #2917, opened 2020-07-02
Enable cnd to run over Tor

tcharding:
We recently discovered that it is possible to run libp2p traffic over Tor. Enable this for cnd by doing:

- Add a tor/ directory with example config files for tor and cnd
- Add a Tor libp2p Transport
- Add config options to enable tor
- Add a second transport builder function for Tor
- Conditionally build the transport based on tor config option
- Add a README to describe cnd usage over Tor


tor/README.md
```diff
@@ -0,0 +1,19 @@
...
+1. Configure an onion service using the Tor run file.
+
+You can use the exaple `./torrc` and run `tor` using `sudo /usr/bin/tor --defaults-torrc tor-service-defaults-torrc -f torrc --RunAsDaemon 0`
+
+2. Set the onion address in the cnd config file.
+
+Once you have run `tor` for the first time get the onion address from the `hostname` file (if you used the `tor` invocation above this will be in `/var/lib/tor/hidden_service/hostname`).
+See `./cnd.toml` for an example config file.
+
+3. Before starting cnd to do a swap you will need a running Tor instance. Currently we use the default SOCKS5 Tor port 9050.
```
thomaseizinger, line 19: What does the dependency on SOCKS5 this mean for running `cnd` on MacOS and Windows?

tor/cnd.toml
```diff
@@ -0,0 +1,34 @@
+[network]
```
thomaseizinger, line 1: In the past, config files that we checked into source control were outdated very quickly.

What is gonna prevent this one from being outdated?

tor/tor-service-defaults-torrc
```diff
@@ -0,0 +1,15 @@
+DataDirectory /var/lib/tor
```
thomaseizinger, line 1: This is the default right? Why do we have to provide it then?

tor/torrc
```diff
@@ -0,0 +1,192 @@
+## Configuration file for a typical Tor user
```
thomaseizinger, line 1: Same here, if this is the default, why do we have to provide it?

cnd/src/network.rs
```diff
@@ -54,7 +57,18 @@ impl Swarm {
...
         tracing::info!("Starting with peer_id: {}", local_peer_id);
 
-        let transport = transport::build_comit_transport(local_key_pair)?;
+        let transport = if settings.network.use_tor {
+            let mut map = HashMap::new();
+            if settings.network.listen.len() != 1 {
+                anyhow::bail!("To use Tor please configure a single onion address")
+            }
+            for addr in settings.network.listen.iter() {
+                map.insert((*addr).clone(), 9939u16);
```
thomaseizinger, line 66: Does this mean we are gonna listen on 9939 locally and Tor has to be configured to route traffic do this port?

How does Tor know about this?

cnd/src/config.rs
```diff
@@ -23,6 +23,7 @@ pub struct Data {
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct Network {
     pub listen: Vec<Multiaddr>,
+    pub use_tor: bool,
```
thomaseizinger, line 26: At least at the moment, this is redundant. You could detect that from the Multiaddress that is configured.

However, if we were to give `cnd` more control over this by configuring Tor itself by the use of the cookie file or something, then we'd probably need a configuration option. Actually, the presence of the cookie file could be the enabling bit :)

tor/README.md
tcharding: > What does the dependency on SOCKS5 this mean for running `cnd` on MacOS and Windows?

I have no clue how SOCKS5 works on MacOS, its UNIX so I assumed it would work the same as Linux. The default port is set by Tor. I did not add configuration for the port to cnd since we have not run into the problem of needing it yet.

We don't support Windows anyways so ...

Or am I missing something in your question?

tor/tor-service-defaults-torrc
tcharding: There are a few things about this file that I don't fully understand the reason for. This is the created from the default set up when installing Tor on Ubuntu 18. I elected to just leave it as is instead of deleting things that I didn't see why they put there in the first place. I don't actually see why Tor uses this file at all and not just the torrc - see why I was hesitant to change it :)

tor/cnd.toml
tcharding: Cool, will remove.

tor/torrc
tcharding: Oh this is not the default any more I modified it. We can actually just have a file (or comment in code) on the hidden service config parts. This is the only bits I added and that users need to add
```
HiddenServiceDir /var/lib/tor/hidden_service/
HiddenServicePort 9939 127.0.0.1:9939
```

tcharding: My intention with adding these files was to make running Tor as easy as possible in the 'typical' case since you expressed before that we want zero config.

cnd/src/network.rs
tcharding: Yes. That is configured in the torrc.
```
HiddenServicePort 9939 127.0.0.1:9939
```

The first port number is the virtual port (the port used as part of the onion address).

Previously I did this using the Tor control protocol but this is, in my opinion, a better way to do it.

cnd/src/config.rs
tcharding: Good thinking re asserting use of tor based on the address. Will add.

tor/tor-service-defaults-torrc
tcharding: I have removed the whole `tor/` directory now.

tor/torrc
tcharding: Remove this file and added a section to the main README describing how to use Tor.

cnd/src/config.rs
tcharding: Done as suggested.

tor/torrc
thomaseizinger: Ah okay, now I get it!

Yes only the relevant part is way clearer I think, thank you!

tor/README.md
thomaseizinger: > Or am I missing something in your question?

Nope! :)
Just wanted to make sure that what we are doing here doesn't lock us out from supporting swaps on several operating systems.

I assume if our precondition is "configure a hidden service with Tor on port 9939" then this is also gonna work on Windows, however it is done over there (if it is even possible).

README.md
```diff
@@ -62,6 +62,21 @@ To run individual end-to-end tests, use `yarn` inside the `api_tests` folder:
 - `yarn run fix`: run prettier and linter to fix format
 - `yarn run check`: run tsc (to check validity of TypeScript code) and verify format
 
+## Tor
+
+If you would like to maintain anonymity while using cnd for atomic swaps we support running cnd over Tor.
+You will need to configure an onion service (previously hidden service), for example:
```
thomaseizinger, line 68: The "for example" might be misleading in that it suggests that you can freely choose the port whereas in reality, cnd is hardcoded against 9939 for the hidden service.

tcharding: Good point, I was only referring to the virtual port. Will make it clearer, thanks.

PR #2928, opened 2020-07-06
Update changelog with new config dir

thomaseizinger:
Caused by https://github.com/comit-network/comit-rs/pull/2923.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
--  **Breaking Change** Remove support for RFC003 swaps
+-   **Breaking Change** Remove support for RFC003 swaps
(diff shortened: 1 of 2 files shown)


thomaseizinger:
> Was not that easy to find the actual validation for this one :)
> 
> [developer.apple.com/library/archive/documentation/General/Conceptual/MOSXAppProgrammingGuide/AppRuntime/AppRuntime.html](https://developer.apple.com/library/archive/documentation/General/Conceptual/MOSXAppProgrammingGuide/AppRuntime/AppRuntime.html)
> 
> Note that this states that this should be data that is not needed to run the application.
> 
> On stackoverflow there are different opinions: [stackoverflow.com/questions/15677388/default-location-for-configuration-files-macos](https://stackoverflow.com/questions/15677388/default-location-for-configuration-files-macos)
> 
> :D

Interesting, thanks for digging this up @da-kami !
I actually didn't do that much research. Given what you posted, I am wondering why the `directories` library changed their implementation :thinking:

thomaseizinger:
@bonomat @D4nte I am gonna merge this to make sure our changelog is up to date.
If we want to do something about this change, I would suggest that we discuss that in a separate ticket.

bors r+

PR #2929, opened 2020-07-06
Remove dead code and take measures to prevent it in the future

thomaseizinger:
Rustc has detection against dead code being present in the codebase. However, it can only detect dead code if it can prove that actually nothing could ever use it.

To make better use of this, we first get rid of the `lib.rs` file in `cnd`. This prevents rustc from thinking that any of this code will ever be used by another crate.
As a next step, we make as most of the modules inside `cnd` private.

This results in rustc emitting quite a few dead code warnings that haven't been there before.


cnd/src/network/tor.rs
```diff
@@ -88,12 +88,6 @@ impl TorTokioTcpConfig {
         }
     }
 
-    /// Sets the TTL to set for opened sockets.
-    pub fn ttl(mut self, value: u32) -> Self {
-        self.ttl = Some(value);
-        self
-    }
-
```
tcharding, line 96: Interesting PR, I was actually thinking about this dead code issue recently when I added the `pub` to this module so that this very function didn't throw a clippy warning :)

thomaseizinger:
> @thomaseizinger I am guessing rustc detects deadcode by default?

By default, `dead_code` is a lint on the warn level if that is what you are asking.
However, in order for the lint to trigger, it has to be provable that the item is never used.

Any public code from a crate can potentially be used by other crates so that lint never triggers. The idea behind this PR was to move all the code to the binary (all modules are now declaring in `main.rs`). Meaning if they are not used from there, there is no way they could ever be used, hence rustc can be more aggressive here :)

luckysori:
> @luckysori can we do that in nectar if we move `test_harness` as a separate crate?

We should do something similar, yeah.

PR #2932, opened 2020-07-07
Encapsulate announce logic within Announce network behaviour

thomaseizinger:
First two patches are cleanup.

Third patch is the actual meat of the PR.

Previously, the logic around announcing was spread out over several parts of the code making the whole thing the opposite of modular.
This moves all of the logic around announcing a swap into the Announce behaviour.
At the same time, we remove a lot of code by re-implementing it on top of the new `request-response` abstraction provided by libp2p. That is an implementation detail though. We could move away from that any time and the API shouldn't change.


cnd/src/network/peer_tracker.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        assert!(alice_swarm.connected_peers.is_empty());
+        assert!(bob_swarm.connected_peers.is_empty());
+
+        connect(&mut alice_swarm, &mut bob_swarm).await;
+
+        assert_eq!(
+            alice_swarm.connected_peers.get(&bob_id),
+            Some(&vec![bob_address])
+        );
```
tcharding, line 110: Perhaps we can add this here also
```
        // We only track connections in the dialer.
        assert!(bob_swarm.connected_peers.is_empty());
```

cnd/src/network.rs
```diff
@@ -88,13 +91,33 @@ impl Swarm {
     pub async fn initiate_communication(
         &self,
         id: LocalSwapId,
-        peer: DialInformation,
+        DialInformation {
+            peer_id,
+            address_hint,
+        }: DialInformation,
```
tcharding, line 97: This syntax is super ugly. I get the feeling from the clunkyness of `DialInformation` that the fields are misnamed or the whole type is not useful.

Perhaps changing the `peer_id` field name to `id`
```

#[derive(Clone, Debug, PartialEq)]
pub struct DialInformation {
    pub id: PeerId,
    pub address_hint: Option<Multiaddr>,
}
```
This allows more ergonomic usage as such
```
fn foo(peer: DialInformation) {
    some_fn(peer.id)
    ...
}
```

tcharding: I think though that a better solution would be to remove `DialInformation` all together and use two arguments when they are needed.

cnd/src/network.rs
```diff
@@ -88,13 +91,33 @@ impl Swarm {
...
         role: Role,
         digest: SwapDigest,
         identities: Identities,
     ) -> anyhow::Result<()> {
         let mut guard = self.inner.lock().await;
-        guard.initiate_communication(id, peer, role, digest, identities)
+
+        // best effort attempt to establish a connection to the other party
+        if let Some(address_hint) = address_hint {
```
tcharding, line 105: Correct me if I'm wrong but if the user of cnd POSTS a swap as Bob but includes an IP address (easy mistake to make) the we will dial from Bob's node to Alice's node as well as listen and receive a connection request from Alice's node so we have two TCP connections. This doesn't seem like what we want?

cnd/src/network.rs
```diff
@@ -213,17 +236,29 @@ fn derive_key_pair(seed: &RootSeed) -> Keypair {
...
     task_executor: Handle,
-
+    /// Here is where we track the "LocalData" that we need for the execution
+    /// parameter phase whilst the announce protocol is in progress.
+    #[behaviour(ignore)]
+    local_data: HashMap<LocalSwapId, LocalData>,
+    /// Here is where we track the what the local swap id for a given shared
+    /// swap id is. The execution parameter exchange only knows about
+    /// SharedSwapIds, so we need to map this back to a local one to save the
+    /// data correctly to the database.
```
tcharding, line 259:
```suggestion
    /// The execution parameter exchange only knows about SharedSwapIds, so we
    /// need to map this back to a LocalSwapId to save the data correctly to
    /// the database.
```

cnd/src/network.rs
```diff
@@ -422,18 +460,30 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<()> for ComitNode {
...
+                shared_swap_id,
                 remote_data,
             } => {
                 let storage = self.storage.clone();
                 let spawner = self.protocol_spawner.clone();
 
+                let local_swap_id = match self.local_swap_ids.remove(&shared_swap_id) {
+                    Some(local_swap_id) => local_swap_id,
+                    None => {
+                        tracing::warn!("inconsistent data within ComitNode");
```
tcharding, line 482:
```suggestion
                        tracing::warn!("inconsistent data, missing local swap id");
```

cnd/src/network.rs
thomaseizinger: I think it is not that critical.

libp2p can easy handle multiple connections to the same peer and if we move the logic to the outside here, it is truly a concern of cnd instead of the comit lib.

I can guard it with role == Alice if you want.

thomaseizinger:
@tcharding All comments actioned :)

tcharding:
> @tcharding All comments actioned :)

Thanks man!

PR #2941, opened 2020-07-09
Make SwapContext query a view

thomaseizinger:
This allows us to avoid inline SQL and instead use diesel's built-in abstractions to query and deal with this query.

Initially, I thought the benefit we would be getting out of this is that sqlite would complain if we make a mistake in the SQL query, for example, when a column is misspelled.

Unfortunately, that is not the case but I only realized after the change was done so I thought I put it up anyway :D

### DIFF
--- a/cnd/src/storage.rs
+++ b/cnd/src/storage.rs
@@ -341,10 +341,10 @@ pub struct SwapContext {
     pub beta: Protocol,
 }
 
-impl From<SwapContextRow> for SwapContext {
-    fn from(row: SwapContextRow) -> Self {
+impl From<tables::SwapContext> for SwapContext {
+    fn from(row: tables::SwapContext) -> Self {
         SwapContext {
-            id: row.id.0,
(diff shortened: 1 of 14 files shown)


cnd/src/storage.rs
```diff
@@ -355,16 +355,20 @@ impl From<SwapContextRow> for SwapContext {
...
+        use self::schema::swap_contexts;
+
         let context = self
             .db
-            .swap_contexts()
+            .do_in_transaction(|connection| {
+                swap_contexts::table
+                    .filter(swap_contexts::local_swap_id.eq(Text(swap_id)))
+                    .get_result::<tables::SwapContext>(connection)
+                    .optional()
```
da-kami, line 366: ❓ Why do we need this `optional()` here?

thomaseizinger: Because it could not exist potentially so I do `ok_or` a couple of lines down.

PR #2947, opened 2020-07-13
Replace LocalSwapId in Announce protocol with generic context

### LINKED ISSUE #2945: Removing `LocalSwapId` from the comit library
Reflecting over the current design of the comit library and the network layer, I would like to make a proposal of removing `LocalSwapId` from the comit library.

The only usecase within `comit-rs` is in the `Announce` protocol but it could easily be removed from there.
I checked nectar and saw it is is also used there although I assume, it could easily be removed or replaced there as well. (cc @D4nte @luckysori @da-kami)
From my understanding, the current changes to the orderbook also don't require a `LocalSwapId`. (cc @tcharding @rishflab)

I think removing it makes sense because there is no abstraction within the `comit` lib that requires a `LocalSwapId`. In fact, the existence of `LocalSwapId` is an artifact of `cnd` needing to assign an identifier to the swap given to the REST API.

If I remember correctly, the only reason why we moved it to the `comit` lib was because the `Swaps` abstraction used it to keep track of state and we wanted to move that to the `comit` lib. However, that is now gone with the recent refactoring of the `Announce` protocol.

thomaseizinger:
Resolves #2945.

### DIFF
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -213,7 +213,7 @@ fn derive_key_pair(seed: &RootSeed) -> Keypair {
 #[derive(NetworkBehaviour)]
 #[allow(missing_debug_implementations)]
 pub struct ComitNode {
-    announce: Announce,
+    announce: Announce<LocalSwapId>,
     comit: Comit,
     peer_tracker: PeerTracker,
 
--- a/cnd/src/network.rs
+++ b/cnd/src/network.rs
@@ -513,13 +513,15 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<::comit::network::comit::Behavi
     }
 }
 
-impl libp2p::swarm::NetworkBehaviourEventProcess<announce::BehaviourOutEvent> for ComitNode {
-    fn inject_event(&mut self, event: announce::BehaviourOutEvent) {
+impl libp2p::swarm::NetworkBehaviourEventProcess<announce::BehaviourOutEvent<LocalSwapId>>
+    for ComitNode
+{
+    fn inject_event(&mut self, event: announce::BehaviourOutEvent<LocalSwapId>) {
         match event {
             announce::BehaviourOutEvent::Confirmed {
                 peer,
                 shared_swap_id,
-                local_swap_id,
+                context: local_swap_id,
             } => {
                 let data = match self.local_data.remove(&local_swap_id) {
                     Some(local_data) => local_data,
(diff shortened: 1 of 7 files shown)


cnd/src/network.rs
```diff
@@ -513,13 +513,15 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<::comit::network::comit::Behavi
...
+impl libp2p::swarm::NetworkBehaviourEventProcess<announce::BehaviourOutEvent<LocalSwapId>>
+    for ComitNode
+{
+    fn inject_event(&mut self, event: announce::BehaviourOutEvent<LocalSwapId>) {
         match event {
             announce::BehaviourOutEvent::Confirmed {
                 peer,
                 shared_swap_id,
-                local_swap_id,
+                context: local_swap_id,
```
D4nte, line 524: Does this change the way the message is formed over the network?

thomaseizinger: The context is never sent over the network! :)

This is just an identifier that is stored locally and re-emitted with the event so the caller can treat the event accordingly.

thomaseizinger:
> Are we saying that the caller can choose any arbitrary string as a context?

If that is what your application uses to keep track of swaps, then yes :)

comit/src/network/protocols/announce.rs
```diff
@@ -23,31 +23,38 @@ use std::{
...
+/// associate submitted announce requests with the events emitted by this
+/// NetworkBehaviour.
 #[derive(NetworkBehaviour)]
-#[behaviour(out_event = "BehaviourOutEvent", poll_method = "poll")]
+#[behaviour(out_event = "BehaviourOutEvent<C>", poll_method = "poll")]
 #[allow(missing_debug_implementations)]
-pub struct Announce {
+pub struct Announce<C>
+where
+    C: fmt::Display + Send + 'static,
```
tcharding, line 35: Previously you told me that you use qualified paths out of lazyness but here you it seems you went up to the import statements to add `fmt`, why not add `fmt::Display`?

thomaseizinger: Because I am being inconsistent with myself! :D

Using `fmt::Display` is a habit from implementing the trait because you need `fmt::Formatter` and `fmt::Result` for it, so I usually just import `std::fmt` and I am good to go.

That is not needed here so `Display` would be better yes.

thomaseizinger:
bors retry

thomaseizinger:
> Interesting solution.

Interesting as in, what the f*** is he doing? :D
I was trying to find a solution that doesn't require adding another ID to our system :sweat_smile:

PR #2951, opened 2020-07-14
Add serialization/deserialization for Bitcoin network

tcharding:
The bitcoin networks have canonical names (bitcoin, testnet, regtest). As such we can implement serialization/deserialization on them within COMIT and users can simply use these implementations.

Done in preparation for adding networks to the orderbook.

Based on information learned in @thomaseizinger recent post about serialization. If I have mis-understood you Thomas please say so.

### DIFF
--- a/cnd/src/http_api.rs
+++ b/cnd/src/http_api.rs
@@ -174,43 +174,6 @@ impl Serialize for Http<PeerId> {
     }
 }
 
-impl Serialize for Http<ledger::Bitcoin> {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        let str = match self {
-            Http(ledger::Bitcoin::Mainnet) => "mainnet",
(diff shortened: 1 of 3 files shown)


D4nte:
> The bitcoin networks have canonical names (bitcoin, testnet, regtest). As such we can implement serialization/deserialization on them within COMIT and users can simply use these implementations.

I think you mean _mainnet, testnet, regtest_ (at least according to the code).

tcharding:
Oh dear, this is wrong - as noted above we specifically do _not_ use the canonical form so IIUC, according to the new knowledge base we are working from, this ser/deser stuff should not be in `ledger.rs`?

D4nte:
> Oh dear, this is wrong - as noted above we specifically do _not_ use the canonical form so IIUC, according to the new knowledge base we are working from, this ser/deser stuff should not be in `ledger.rs`? We can resolve this by using "bitcoin" instead of "mainnet" but that is going to be a breaking API change in cnd.

Please do NOT use "bitcoin" for mainnet. We only use this network for bitcoin ledgers, mainnet is mainnet.

PR #2952, opened 2020-07-14
Add configurable ledger to ordebook

tcharding:
Currently we hardcode regtest in the orderbook. We would like to somehow be able to configure the network so this works on mainnet.

Support mainnet by doing:
- Add a field for both bitcoin and ethereum ledgers on the cnd order creation endpoint.
- Add the ledgers to the orderbook within comit crate.
- Add an assertion method on `NewOrder` to assert a valid ledger pair, defined as the 'same' network kind. Add infrastructure for this assertion.
- Add ser/deser tests for the `ledger` module.

### DIFF
--- a/cnd/src/http_api/orderbook.rs
+++ b/cnd/src/http_api/orderbook.rs
@@ -1,7 +1,5 @@
 use crate::{
-    asset,
-    ethereum::ChainId,
-    hbit, herc20,
+    asset, hbit, herc20,
     http_api::problem,
     identity, ledger,
     storage::{CreatedSwap, Save},
@@ -23,18 +21,23 @@ struct MakeHerc20HbitOrderBody {
     // struct is a breaking API change so the e2e tests will break.
     #[serde(with = "asset::bitcoin::sats_as_string")]
     buy_quantity: asset::Bitcoin,
+    bitcoin_ledger: ledger::Bitcoin,
     sell_token_contract: ethereum::Address,
     sell_quantity: asset::Erc20Quantity,
+    ethereum_ledger: ledger::Ethereum,
     absolute_expiry: u32,
     refund_identity: bitcoin::Address,
     redeem_identity: identity::Ethereum,
 }
 
 impl MakeHerc20HbitOrderBody {
+    // TODO: This should implement From
     fn to_order(&self) -> NewOrder {
         NewOrder {
             buy: self.buy_quantity,
+            bitcoin_ledger: self.bitcoin_ledger,
             sell: asset::Erc20::new(self.sell_token_contract, self.sell_quantity.clone()),
+            ethereum_ledger: self.ethereum_ledger,
             absolute_expiry: self.absolute_expiry,
         }
     }
@@ -103,7 +106,7 @@ pub async fn post_take_herc20_hbit_order(
         alpha: herc20::CreatedSwap {
             asset: order.sell,
             identity: refund_identity,
-            chain_id: ChainId::regtest(),
+            chain_id: order.ethereum_ledger.chain_id,
             absolute_expiry: order.absolute_expiry,
         },
         beta: hbit::CreatedSwap {
@@ -155,6 +158,11 @@ pub async fn post_make_herc20_hbit_order(
 
     let reply = warp::reply::reply();
     let order: NewOrder = body.to_order();
+
+    order
+        .assert_valid_ledger_pair()
+        .map_err(problem::from_anyhow)
+        .map_err(warp::reject::custom)?;
 
     // TODO: We need to save the bitcoin address here else it is lost.
     let swap_id = LocalSwapId::default();
@@ -298,14 +306,17 @@ mod tests {
         let json = r#"
         {
             "sell_token_contract": "0xB97048628DB6B661D4C2aA833e95Dbe1A905B280",
+            "bitcoin_ledger": "regtest",
             "buy_quantity": "300",
             "sell_quantity": "200",
+            "ethereum_ledger": {"chain_id":2},
             "absolute_expiry": 600,
             "refund_identity": "1F1tAaz5x1HUXrCNLbtMDqcw6o5GNn4xqX",
             "redeem_identity": "0x00a329c0648769a73afac7f9381e08fb43dbea72",
             "maker_addr": "/ip4/127.0.0.1/tcp/39331"
         }"#;
 
-        let _body: MakeHerc20HbitOrderBody = serde_json::from_str(json).unwrap();
+        let _body: MakeHerc20HbitOrderBody =
+            serde_json::from_str(json).expect("failed to deserialize order");
     }
 }
--- a/comit/src/ledger.rs
+++ b/comit/src/ledger.rs
@@ -1,5 +1,18 @@
 use crate::ethereum::ChainId;
+use fmt::Display;
 use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt;
+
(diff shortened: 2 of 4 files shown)


cnd/src/http_api/orderbook.rs
```diff
@@ -156,6 +159,11 @@ pub async fn post_make_herc20_hbit_order(
     let reply = warp::reply::reply();
     let order: NewOrder = body.to_order();
 
+    order
+        .assert_valid_ledger_pair()
+        .map_err(problem::from_anyhow)
+        .map_err(warp::reject::custom)?;
```
thomaseizinger, line 165: Should this check be in the order book instead of the controller?

cnd/src/http_api/orderbook.rs
```diff
@@ -298,8 +306,10 @@ mod tests {
         let json = r#"
         {
             "sell_token_contract": "0xB97048628DB6B661D4C2aA833e95Dbe1A905B280",
+            "bitcoin_ledger: regtest"
             "buy_quantity": "300",
             "sell_quantity": "200",
+            "ethereum_ledger: "
```
thomaseizinger, line 312: That is not valid JSON.

comit/src/network/protocols/orderbook.rs
```diff
@@ -338,6 +344,18 @@ impl Order {
...
 
+impl NewOrder {
+    pub fn assert_valid_ledger_pair(&self) -> anyhow::Result<()> {
+        let a = self.bitcoin_ledger;
+        let b = self.ethereum_ledger;
+
+        if ledger::is_valid_ledger_pair(a, b) {
+            return Ok(());
+        }
+        Err(anyhow::anyhow!("invalid ledger pair {}/{}", a, b))
```
thomaseizinger, line 355: If you are only allowing "valid" ledger pairs, why don't you just have a single field of type `Kind` in the `Order`?
Then you can't even run into this error.

comit/src/ledger.rs
```diff
@@ -1,5 +1,18 @@
 use crate::ethereum::ChainId;
+use fmt::Display;
 use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt;
+
+/// The ledger network kind. We define this as a cross blockchain domain term.
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum Kind {
```
thomaseizinger, line 8: I like this!

Maybe we can even go further in the future and instead of just classifying the networks of other ledgers, actually establish a direct mapping of networks on COMIT to networks of other ledgers.

comit/src/network/protocols/orderbook.rs
tcharding: I thought about doing that. The current solution allows folks to stipulate their preferred Ethereum testnet network.

comit/src/ledger.rs
tcharding: I've wanted to introduce this for pretty much as long as I can remember while working at CoBloX :)

cnd/src/http_api/orderbook.rs
tcharding: How'd that get in there? Its hard to get good help :) Thanks for flagging this for me.

cnd/src/http_api/orderbook.rs
tcharding: I'm inclined to leave the orderbook being permissive. Following the principle of keeping policy out of the library and in the application. (I don't actually know if this is a thing in application development, in the kernel they talk about this - policy stays in userspace)

tcharding:
> Question: Did you spend some thoughts on how you want to do this on a gossipsub level? Are we gonna filter out orders from other networks or just always return all of them?

I'm inclined to see this as an optimisation. When we have the problem that users are complaining that the network is being saturated with testnet/devnet swaps then this should be solved. Right now we have just two nodes so this is not an issue IMO.

cnd/src/http_api/orderbook.rs
tcharding: This is now
```
        let json = r#"
        {
            "sell_token_contract": "0xB97048628DB6B661D4C2aA833e95Dbe1A905B280",
            "bitcoin_ledger": "regtest",
            "buy_quantity": "300",
            "sell_quantity": "200",
            "ethereum_ledger": {"chain_id":2},
            "absolute_expiry": 600,
            "refund_identity": "1F1tAaz5x1HUXrCNLbtMDqcw6o5GNn4xqX",
            "redeem_identity": "0x00a329c0648769a73afac7f9381e08fb43dbea72",
            "maker_addr": "/ip4/127.0.0.1/tcp/39331"
        }"#;
```

D4nte:
> Are we gonna filter out orders from other networks or just always return all of them?

To me, 2 comit nodes that are connected to different networks should not be able to talk to each other. Meaning, the libp2p stream should not even open.

cnd/src/http_api/orderbook.rs
thomaseizinger: Interesting viewpoint. What does that mean for compatibility between nodes though?
If the policy is important for compatibility, it should go into the (not yet existent) spec which means it should also go into the library that implements the spec (the `comit` crate).

What do you think?

comit/src/network/protocols/orderbook.rs
thomaseizinger: Fair point although I wonder if that flexibility doesn't harm us in the long run :)
Happy to move forward with this and see where we end up.

comit/src/ledger.rs
thomaseizinger: I reckon we should have done that way earlier but it always felt a bit weird without lifting the abstraction level of COMIT on layer up (from swaps to trading).

thomaseizinger:
> > Are we gonna filter out orders from other networks or just always return all of them?
> 
> To me, 2 comit nodes that are connected to different networks should not be able to talk to each other. Meaning, the libp2p stream should not even open. From a backend dev point of view, it's like if you have UAT nodes talking to a Prod DB. It should not even be possible from an infrastructure point of view.

If that is our goal, we could go the bitcoin way and listen on different ports?

tcharding:
> To me, 2 comit nodes that are connected to different networks should not be able to talk to each other.

I disagree. One cannot trust any other node to do what it says it is doing. We should assume folks are going to try to convince our nodes user to swap its real coins (mainnet) for their testnet coins.

> If that is our goal, we could go the bitcoin way and listen on different ports?

I like this, then we can have multiple comit networks mapped to 'typical' underlying ledger networks i.e., a mainnet network for mainnet swaps and a testnet network for testnet swaps.

Again, this seems to me like early optimisation since we only currently support two nodes. Two nodes do not, IMO, make a network.

thomaseizinger:
> Again, this seems to me like early optimisation since we only currently support two nodes. Two nodes do not, IMO, make a network.

That is not quite true. With the orderbook, we actually have a network.

thomaseizinger:
> I like this, then we can have multiple comit networks mapped to 'typical' underlying ledger networks i.e., a mainnet network for mainnet swaps and a testnet network for testnet swaps.

Yep, that is what I was thinking. You couldn't arbitrarily choose the underlying network then though.
Like, if we say COMIT testnet means Ropsten for Ethereum then that is it, you can't just use a different one then.

We could get around this by making the mapping only fixed for mainnet and testnet. The COMIT devnet could be configurable with an arbitrary mapping. So you could launch a network of COMIT nodes with an arbitrary mapping.

PR #2953, opened 2020-07-14
Add serialization tests for MakerId

tcharding:
The `MakerId` only exists so we can implement serialization/deserialization on it. 

- Adds stability test
- Adds a roundtrip test
- Adds `From` and `Display` impls (not so sure if we want these)?

### DIFF
--- a/comit/src/network/protocols/orderbook.rs
+++ b/comit/src/network/protocols/orderbook.rs
@@ -496,6 +496,7 @@ mod tests {
         network::test::{await_events_or_timeout, new_connected_swarm_pair},
     };
     use libp2p::Swarm;
+    use spectral::prelude::*;
 
     fn create_order(id: PeerId) -> Order {
         Order::new(id, NewOrder {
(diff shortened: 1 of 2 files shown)


comit/src/network/protocols/orderbook.rs
```diff
@@ -118,7 +119,7 @@ impl Orderbook {
     pub fn peer_id_for_order(&self, order_id: OrderId) -> Option<PeerId> {
```
thomaseizinger, line 119: This method should maybe even actually return the `MakerId` instead of the `PeerId`. Would make it more descriptive that we are actually getting the maker of the order here.

thomaseizinger: The fn should probably also be private :)

tcharding: Good ideas, implemented as suggested.

comit/src/network/protocols/orderbook.rs
```diff
@@ -273,10 +272,21 @@ pub type OrderId = Uuid;
...
+
+impl From<MakerId> for PeerId {
+    fn from(id: MakerId) -> Self {
+        id.0
+    }
+}
+
+impl Display for MakerId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0.to_string())
```
D4nte, line 289:
```suggestion
        write!(f, "{}", self.0)
```
This should be enough :)

PR #2955, opened 2020-07-15
Add trade direction to trading pair

rishflab:
50% PR

Added buy/sell direction to order facilitating move towards asset centric model.

The sell path has not yet been added.

### DIFF
--- a/cnd/src/facade.rs
+++ b/cnd/src/facade.rs
@@ -59,7 +59,7 @@ impl Facade {
         Ok(timestamp)
     }
 
-    pub async fn take_btc_dai_buy_order(
+    pub async fn take_btc_dai_sell_order(
         &mut self,
         order_id: OrderId,
         swap_id: LocalSwapId,
@@ -73,11 +73,11 @@ impl Facade {
             .await;
 
         self.swarm
-            .take_btc_dai_buy_order(order_id, swap_id, redeem_identity, refund_identity)
+            .take_btc_dai_sell_order(order_id, swap_id, redeem_identity, refund_identity)
             .await
     }
 
-    pub async fn make_btc_dai_buy_order(
+    pub async fn make_btc_dai_sell_order(
         &self,
         order: NewOrder,
         swap_id: LocalSwapId,
@@ -85,7 +85,7 @@ impl Facade {
         refund_identity: crate::bitcoin::Address,
     ) -> anyhow::Result<OrderId> {
         self.swarm
-            .make_btc_dai_buy_order(order, swap_id, redeem_identity, refund_identity)
+            .make_btc_dai_sell_order(order, swap_id, redeem_identity, refund_identity)
             .await
     }
 
--- a/cnd/src/facade.rs
+++ b/cnd/src/facade.rs
@@ -59,7 +59,7 @@ impl Facade {
         Ok(timestamp)
     }
 
-    pub async fn take_herc20_hbit_order(
+    pub async fn take_btc_dai_buy_order(
         &mut self,
         order_id: OrderId,
         swap_id: LocalSwapId,
@@ -73,11 +73,11 @@ impl Facade {
             .await;
 
         self.swarm
-            .take_herc20_hbit_order(order_id, swap_id, redeem_identity, refund_identity)
+            .take_btc_dai_buy_order(order_id, swap_id, redeem_identity, refund_identity)
             .await
     }
 
-    pub async fn make_herc20_hbit_order(
+    pub async fn make_btc_dai_buy_order(
         &self,
         order: NewOrder,
         swap_id: LocalSwapId,
@@ -85,7 +85,7 @@ impl Facade {
         refund_identity: crate::bitcoin::Address,
     ) -> anyhow::Result<OrderId> {
         self.swarm
-            .make_herc20_hbit_order(order, swap_id, redeem_identity, refund_identity)
+            .make_btc_dai_buy_order(order, swap_id, redeem_identity, refund_identity)
             .await
     }
 
--- a/cnd/src/http_api/orderbook.rs
+++ b/cnd/src/http_api/orderbook.rs
@@ -4,6 +4,7 @@ use crate::{
     hbit, herc20,
     http_api::problem,
     identity, ledger,
+    network::TradeDirection,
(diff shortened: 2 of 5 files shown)


cnd/src/facade.rs
```diff
@@ -59,7 +59,7 @@ impl Facade {
         Ok(timestamp)
     }
 
-    pub async fn take_herc20_hbit_order(
+    pub async fn take_btc_dai_buy_order(
```
tcharding, line 62: Currently the orderbook supports a single COMIT protocol, the herc20-hbit protocol. In this protocol Bob is swapping the BTC he has for the Erc20 tokens Alice has. Mapping this to the trading pair BTC/DAI means, from Bob's perspective he is 'selling' the trading pair.

If, as I believe it is from my brief research, a trade order is named from the perspective of the maker then this function should be called `take_btc_dai_sell_order`. Bob made a 'sell' order and Alice is taking it.

cnd/src/http_api/orderbook.rs
```diff
@@ -18,30 +19,32 @@ use uuid::Uuid;
 use warp::{http, http::StatusCode, Rejection, Reply};
 
 #[derive(Deserialize)]
-struct MakeHerc20HbitOrderBody {
+struct MakeBtcDaiBuyOrderBody {
```
tcharding, line 22: As comment above this should be 'Sell'.

cnd/src/http_api/orderbook.rs
```diff
@@ -18,30 +19,32 @@ use uuid::Uuid;
...
     // TODO: Fix these names, buy/sell are incorrectly used here. Remember that changing this
     // struct is a breaking API change so the e2e tests will break.
```
tcharding, line 24: You can remove this comment now :)

cnd/src/http_api/orderbook.rs
```diff
@@ -18,30 +19,32 @@ use uuid::Uuid;
...
 }
 
-impl MakeHerc20HbitOrderBody {
+impl MakeBtcDaiBuyOrderBody {
     fn to_order(&self) -> NewOrder {
         NewOrder {
-            buy: self.buy_quantity,
-            sell: asset::Erc20::new(self.sell_token_contract, self.sell_quantity.clone()),
+            pair: TradingPair::BtcDai,
+            direction: TradeDirection::Buy,
```
tcharding, line 38: Is this naming coming from Nectar? (Referring to the use of the word 'direction'.)

cnd/src/facade.rs
rishflab: yes you are right, ill swap it

comit/src/network/protocols/orderbook.rs
```diff
@@ -229,37 +229,36 @@ pub enum OrderbookError {
...
-    pub fn to_topic(&self, peer: &PeerId) -> Topic {
-        let trading_pair_topic = TradingPairTopic {
-            peer: PeerId::into_bytes(peer.clone()),
-            buy: self.buy,
-            sell: self.sell,
-        };
-        trading_pair_topic.to_topic()
+    pub fn to_topic(&self) -> Topic {
+        match self {
+            TradingPair::BtcDai => Topic::new("BTC_DAI".to_string()),
```
tcharding, line 246: Remember how we got burned last week from using 'magic' strings?

comit/src/network/protocols/orderbook.rs
```diff
@@ -229,37 +229,36 @@ pub enum OrderbookError {
...
+        }
     }
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
-pub struct TradingPairTopic {
-    peer: Vec<u8>,
-    buy: SwapType,
-    sell: SwapType,
+pub struct TradingPairFromMaker {
```
tcharding, line 252: Can I suggest that you just delete all this code that maps peer ids of a maker to trading pairs they are offering. Correct me if I have misunderstood the code but
- We only support two topics 1. "makers" 2. "BTC_DAI"
- We only support two nodes, there is no 'network'
- Any logic that needs to be done before taking an order (e.g. whitelisting/blacklisting of peers) can be done on a order level, it is not done on a topic level.

What use is mapping peer id to topic?

So unless I'm missing something, and please correct me if I am, this mapping code actually achieves nothing.

comit/src/network/protocols/orderbook.rs
```diff
@@ -323,18 +326,16 @@ impl Order {
         Order {
             id: Uuid::new_v4(),
             maker: MakerId(peer_id),
-            buy: new_order.buy.as_sat(),
-            sell: new_order.sell,
+            pair: TradingPair::BtcDai,
+            direction: new_order.direction,
+            dai_quantity: new_order.dai_quantity,
+            bitcoin_quantity_sat: new_order.bitcoin_quantity.as_sat(),
```
tcharding, line 332: Ok, this is starting to look nice. Are we on to 90% review yet? If so, I'd like to see us migrate  to using `btc` instead of `bitcoin`. If we are moving to being asset-centric in cnd this makes more sense and its less characters to type with no loss of meaning.

cnd/src/http_api/orderbook.rs
thomaseizinger: Nectar uses "position" for buy/sell.

comit/src/network/protocols/orderbook.rs
thomaseizinger: Pardon me of chiming in here but:

> We only support two nodes, there is no 'network'

The orderbook supports more than two nodes. The orderbook is in-fact a network of nodes gossiping around orders.

> Any logic that needs to be done before taking an order (e.g. whitelisting/blacklisting of peers) can be done on a order level, it is not done on a topic level.

If we don't go for explicit subscriptions to makers, then the whole "makers" topic is also redundant because you can just subscribe to "BTC_DAI" directly.

The only thing I'd ask for if you want to drop this functionality is that you actually make it an implementation detail, i.e. no part of this design must leak out of the orderbook NB and a transition to separate topics should not require an API change :)

rishflab:
resolved with https://github.com/comit-network/comit-rs/pull/2958

PR #2956, opened 2020-07-15
Refactor order code

tcharding:
Refactor the order code (`Order` and `NewOrder`) by doing:

- Patch 1: Add an `order` module to the orderbook (move `Order`, `NewOrder`, `OrderId` into it)
- Patch 2: Use a struct instead of a type alias for `OrderId`
- Patch 3: Move the `NewOrder` type to cnd (it is only used to interface between the controller and the network layer in cnd)
- Patch 4: Use `asset::Bitcoin` in the `Order` struct instead of a u64

### DIFF
--- a/cnd/src/facade.rs
+++ b/cnd/src/facade.rs
@@ -1,13 +1,13 @@
 use crate::{
     btsieve::LatestBlock,
     connectors::Connectors,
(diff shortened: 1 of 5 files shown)


tcharding:
This is just refactor. Removing reviewers except @rishflab and kicking bors.

PR #2958, opened 2020-07-15
ordebook: field names and position

tcharding:
A bunch of clean up work and additional features for the orderbook.  This work conflicts with https://github.com/comit-network/comit-rs/pull/2959 mainly due to different names, happy to find some common ground in order to get one of these in.

Does:

- Removes subscription handling
- Use domain term `Position` enabling trades in both directions
- Adds both expiries
- Removes a non-actionalble todos
- Removes unwraps (and associated todos)
- Adds ser/deser unit tests

Turns e2e test into a skeleton.

### DIFF
--- a/comit/src/network/protocols/orderbook.rs
+++ b/comit/src/network/protocols/orderbook.rs
@@ -95,16 +95,16 @@ impl Orderbook {
 
     /// Create and publish a new 'make' order. Called by Bob i.e. the maker.
     pub fn make(&mut self, order: Order) -> anyhow::Result<OrderId> {
-        let order_id = order.id;
         let ser = bincode::serialize(&Message::CreateOrder(order.clone()))?;
-        let topic = order.topic();
+        let topic = Topic::new(order.topic());
+        let id = order.id;
 
         self.gossipsub.publish(&topic, ser);
-        tracing::info!("published order: {}", order_id);
+        tracing::info!("published order: {}", id);
 
-        self.orders.insert(order_id, order);
+        self.orders.insert(id, order);
 
-        Ok(order_id)
+        Ok(id)
     }
 
     /// Take an order, called by Alice i.e., the taker.
@@ -367,6 +367,7 @@ impl NetworkBehaviourEventProcess<GossipsubEvent> for Orderbook {
 
             match decoded {
                 Message::CreateOrder(order) => {
+                    // TODO: Should we just blindly insert here?
                     self.orders.insert(order.id, order);
                 }
                 Message::DeleteOrder(order_id) => {
@@ -455,11 +456,13 @@ mod tests {
         Order {
             id: OrderId::random(),
             maker: MakerId::from(id),
-            buy: asset::Bitcoin::from_sat(100),
+            trade: Trade::Sell,
+            btc: asset::Bitcoin::from_sat(100),
             bitcoin_ledger: ledger::Bitcoin::Regtest,
-            sell: Erc20 {
-                token_contract: Default::default(),
+            dai: Erc20 {
                 quantity: Erc20Quantity::max_value(),
+                // TODO: Use a more sane value?
+                token_contract: Default::default(),
             },
             ethereum_ledger: ledger::Ethereum::default(),
             absolute_expiry: 100,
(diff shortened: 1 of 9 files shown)


comit/src/network/protocols/orderbook/order.rs
```diff
@@ -35,27 +34,38 @@ impl FromStr for OrderId {
...
     }
+
+    pub fn topic(&self) -> String {
+        BTC_DAI.to_string()
+    }
+}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum Trade {
```
thomaseizinger, line 60: Nectar calls this a `Position` IIRC.

What is the motivation for diverging here?

comit/src/network/protocols/orderbook.rs
```diff
@@ -160,39 +163,43 @@ impl Orderbook {
...
 
     /// Subscribe to a trading pair topic from peer.
-    pub fn subscribe(&mut self, peer: PeerId, trading_pair: TradingPair) -> anyhow::Result<()> {
-        let topic = TradingPairTopic::new(peer, trading_pair);
-        self.trading_pairs.insert(topic.clone());
-        self.gossipsub.subscribe(topic.to_topic());
-        Ok(())
+    /// Returns true if the subscription worked. Returns false if we were
+    /// already subscribed.
+    pub fn subscribe(&mut self, topic: String) -> bool {
```
thomaseizinger, line 173: Why do we have `String`-based APIs in a strongly typed language? :)

Why are topics even exposed on the public API? How does this go together with making the use of gossipsub and how we design the topics an implementation detail?

comit/src/network/protocols/orderbook/order.rs
tcharding: I was not aware of that term. Will use it.

comit/src/network/protocols/orderbook.rs
tcharding: You scratched a good itch here. We actually do not need to handle subscriptions at all, gossipsub does this for us. All this code is now removed.

comit/src/network/protocols/orderbook/order.rs
```diff
@@ -35,27 +34,38 @@ impl FromStr for OrderId {
...
-            buy: SwapType::Hbit,
-            sell: SwapType::Herc20,
-        }
-        .to_topic(&peer_id)
+    pub fn tp(&self) -> TradingPair {
+        TradingPair::BtcDai
     }
+
+    pub fn topic(&self) -> String {
+        BTC_DAI.to_string()
```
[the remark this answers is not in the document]
thomaseizinger: To not make your life to hard, I think you'd want the topic to at least define the format of the messages statically, i.e. the same way an HTTP route should map 1-to-1 to a body.

If you go for a COMIT topic then the internal structure of the order will be a lot more complex.

comit/src/network/protocols/orderbook.rs
```diff
@@ -476,12 +456,12 @@ mod tests {
...
-            buy: asset::Bitcoin::from_sat(100),
+            trade: Trade::Sell,
+            bitcoin_amount: asset::Bitcoin::from_sat(100),
             bitcoin_ledger: ledger::Bitcoin::Regtest,
-            sell: Erc20 {
-                token_contract: Default::default(),
-                quantity: Erc20Quantity::max_value(),
-            },
+            // TODO: Use a more sane value?
+            ethereum_amount: Erc20Quantity::max_value(),
```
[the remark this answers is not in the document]
tcharding: I added static values for now, all the ser/deser testing needs a thorough going through.

comit/src/network/protocols/orderbook/order.rs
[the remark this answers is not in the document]
tcharding: All the topic design needs re-thinking IMO. Leaving as is for now. Can fix function scope.

tcharding:
We need to align on the field names. @rishflab has stated that he is not fussed so I'm going to kick this one in.

Please note that the `Position` stuff and expiries were also added by @rishflab in parallel to this in https://github.com/comit-network/comit-rs/pull/2959, me merging this is making some of his work become obscured. Flagging that here for his benefit and for full disclosure to the team.

PR #2959, opened 2020-07-15
Add swap execution to orderbook e2e test

rishflab:
The orderbook e2e test executes the comit protocol to ensure
that swaps are correctly created from the order. Alpha and
beta ledger expiries have been added to the order to facilitate
this change.

### DIFF
--- a/api_tests/src/actors/actor.ts
+++ b/api_tests/src/actors/actor.ts
@@ -237,7 +237,7 @@ export class Actor {
         };
         this.betaAsset = {
             name: AssetKind.Erc20,
-            quantity: create.alpha.amount,
+            quantity: create.beta.amount,
             ledger: LedgerKind.Ethereum,
             tokenContract: create.beta.token_contract,
         };
--- a/api_tests/src/actors/actor.ts
+++ b/api_tests/src/actors/actor.ts
@@ -237,7 +237,7 @@ export class Actor {
         };
         this.betaAsset = {
             name: AssetKind.Erc20,
-            quantity: create.beta.amount,
+            quantity: create.alpha.amount,
             ledger: LedgerKind.Ethereum,
             tokenContract: create.beta.token_contract,
         };
@@ -252,6 +252,38 @@ export class Actor {
             new SdkWallets({
                 bitcoin: this.wallets.bitcoin.inner,
                 ethereum: this.wallets.ethereum.inner,
+            })
+        );
+    }
+
+    public async initOrderbookTest(swapUrl: string, create: Herc20HbitPayload) {
+        this.alphaLedger = {
+            name: LedgerKind.Ethereum,
+            chain_id: create.alpha.chain_id,
+        };
+        this.betaLedger = {
+            name: LedgerKind.Bitcoin,
+            network: create.beta.network,
+        };
+        this.alphaAsset = {
+            name: AssetKind.Erc20,
+            quantity: create.alpha.amount,
+            ledger: LedgerKind.Ethereum,
+            tokenContract: create.alpha.token_contract,
+        };
+        this.betaAsset = {
+            name: AssetKind.Bitcoin,
+            quantity: create.beta.amount,
+            ledger: LedgerKind.Bitcoin,
+        };
+        await this.setStartingBalances();
+
+        this.swap = new Swap(
+            this.cnd,
+            swapUrl,
+            new SdkWallets({
+                ethereum: this.wallets.ethereum.inner,
+                bitcoin: this.wallets.bitcoin.inner,
             })
         );
     }
--- a/api_tests/src/actors/order_factory.ts
+++ b/api_tests/src/actors/order_factory.ts
@@ -18,11 +18,10 @@ interface Ethereum {
 }
 
 export default class OrderFactory {
-    public static bobOrderfromSwap(
+    public static bobOrderFromSwap(
         makerAddr: string,
         swap: Herc20HbitPayload
     ): Herc20HbitOrder {
-        console.log(swap);
         return {
             btc_quantity: swap.alpha.amount,
             bitcoin_ledger: swap.beta.network,
--- a/api_tests/src/actors/order_factory.ts
+++ b/api_tests/src/actors/order_factory.ts
@@ -1,12 +1,15 @@
+import { Herc20HbitPayload } from "../payload";
+
 export interface Herc20HbitOrder {
-    buy_quantity: string;
+    btc_quantity: string;
     bitcoin_ledger: string;
-    sell_token_contract: string;
-    sell_quantity: string;
+    erc20_token_contract: string;
+    erc20_quantity: string;
     ethereum_ledger: Ethereum;
-    absolute_expiry: number;
-    refund_identity: string;
-    redeem_identity: string;
+    alpha_expiry: number;
+    beta_expiry: number;
+    bob_refund_identity: string;
+    bob_redeem_identity: string;
     maker_addr: string;
 }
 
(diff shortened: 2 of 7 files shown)


api_tests/tests/orderbook.ts
```diff
@@ -79,6 +80,8 @@ it(
         const aliceSwapResponse = await alice.cnd.client.get(
             aliceTakeOrderResponse.headers.location
         );
+        console.log(aliceTakeOrderResponse.headers.location);
```
tcharding, line 83: Please remove this before you do the final PR, tests should be quiet when they are pass - console.log makes them noisy.

api_tests/tests/orderbook.ts
```diff
@@ -95,8 +98,52 @@ it(
...
+            })
+        ).herc20Hbit;
+
+        await bob.initOrderbookTest(
+            linkToBobSwap.href,
+            bodies.bob.beta.network,
+            bodies.bob.alpha.chain_id,
+            bodies.bob.beta.amount,
+            bodies.bob.alpha.amount,
+            bodies.bob.alpha.token_contract
```
tcharding, line 123: Why not just pass `bodies.bob` to this function?

api_tests/src/actors/actor.ts
```diff
@@ -232,12 +232,12 @@ export class Actor {
         };
         this.alphaAsset = {
             name: AssetKind.Bitcoin,
-            quantity: create.alpha.amount,
+            quantity: create.beta.amount,
```
tcharding, line 235: Why are doing this?

api_tests/src/actors/actor.ts
```diff
@@ -256,6 +256,46 @@ export class Actor {
...
+    public async initOrderbookTest(
+        swapUrl: string,
+        network: string,
+        chainId: number,
+        btcAmount: string,
+        erc20Amount: string,
+        erc20TokenContract: string
+    ) {
+        this.alphaLedger = {
+            name: LedgerKind.Bitcoin,
```
tcharding, line 268: Please tell me this was a mistake and that you understand that you are doing a herc20-hbit swap and that the alpha ledger is Ethereum.

api_tests/src/actors/actor.ts
rishflab: fixed

api_tests/tests/orderbook.ts
thomaseizinger: Usually, having assertions in the middle of a test is not that great but within e2e tests, it can be useful to do so to catch errors early. Hence, if you care about the `location` header here, you could just assert against it :)

Even `toBeDefined` can be a useful assertion.

api_tests/src/actors/actor.ts
```diff
@@ -237,7 +237,7 @@ export class Actor {
         };
         this.betaAsset = {
             name: AssetKind.Erc20,
-            quantity: create.beta.amount,
+            quantity: create.alpha.amount,
```
tcharding, line 240: Why are you changing this here?

api_tests/src/actors/order_factory.ts
```diff
@@ -15,18 +18,23 @@ interface Ethereum {
 }
 
 export default class OrderFactory {
-    public static newHerc20HbitOrder(makerAddr: string): Herc20HbitOrder {
+    public static bobOrderfromSwap(
```
tcharding, line 21: Missing capitalisation of the 'f' in 'from'.

api_tests/tests/orderbook.ts
```diff
@@ -7,10 +7,21 @@ import OrderFactory from "../src/actors/order_factory";
 import { twoActorTest } from "../src/actor_test";
 import { Entity, Link } from "comit-sdk/dist/src/cnd/siren";
 import { sleep } from "../src/utils";
+import SwapFactory from "../src/actors/swap_factory";
 
 it(
```
tcharding, line 12: Why is there no 'describe' here?

api_tests/tests/orderbook.ts
```diff
@@ -24,7 +35,10 @@ it(
         /// Wait for alice to accept an incoming connection from Bob
         await sleep(1000);
 
-        const bobMakeOrderBody = OrderFactory.newHerc20HbitOrder(bobAddr[0]);
+        const bobMakeOrderBody = OrderFactory.bobOrderfromSwap(
+            bobAddr[0],
+            bodies.bob
+        );
         // @ts-ignore
```
tcharding, line 42: Do we need this?

api_tests/src/actors/actor.ts
thomaseizinger, line 240: That seems like a bug no? The field this is assigned to is named `this.betaAsset`, why are we assigning the `alpha` amount to it?

cnd/src/http_api/orderbook.rs
```diff
@@ -52,10 +55,11 @@ struct TakeHerc20HbitOrderBody {
...
     #[serde(with = "asset::bitcoin::sats_as_string")]
-    buy_quantity: asset::Bitcoin,
-    sell_token_contract: ethereum::Address,
-    sell_quantity: asset::Erc20Quantity,
-    absolute_expiry: u32,
+    btc_quantity: asset::Bitcoin,
+    erc20_token_contract: ethereum::Address,
+    erc20_quantity: asset::Erc20Quantity,
+    alpha_expiry: u32,
+    beta_expiry: u32,
```
tcharding, line 62: My gut feeling is that using alpha/beta here instead of associating the expiries with the asset is going to bite us. Once we add position (buy/sell), devs are going to have to constantly map these expiry fields to bitcoin/ethereum depending on position. I'd prefer to see these named `bitcoin_absolute_expiry`  and `ethereum_absolute_expiry` to make the code easier to maintain and less fragile.

cnd/src/network.rs
```diff
@@ -238,9 +238,7 @@ pub struct ComitNode {
     #[behaviour(ignore)]
     bitcoin_addresses: HashMap<identity::Bitcoin, crate::bitcoin::Address>,
     #[behaviour(ignore)]
-    order_swap_ids: HashMap<OrderId, LocalSwapId>,
-    #[behaviour(ignore)]
-    confirmed_order_peers: HashMap<OrderId, PeerId>,
+    order_to_swap_id: HashMap<OrderId, LocalSwapId>,
```
tcharding, line 241: Uniformity in the code is really important to increase maintainability and reduce cognitive burden on future readers of the code. In this instance all the other hash maps use plural, please either be uniform or do a PR that changes all of them (with justification).

cnd/src/network.rs
```diff
@@ -618,8 +617,6 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<orderbook::BehaviourOutEvent> f
                     }
                 });
 
-                self.confirmed_order_peers.insert(order_id, peer_id);
```
tcharding, line 621: Its a little hard to review this PR because there is more than one thing happening, I know there is a lot of clean up work being done in the orderbook so this is kind of expected however removal of this line is strange - can you please explain why it was removed?

cnd/src/network.rs
```diff
@@ -629,26 +626,42 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<orderbook::BehaviourOutEvent> f
...
+                    None => {
+                        tracing::error!(
+                            "order {} specified in take order confirmation not found in local store",
+                            order_id
+                        );
+                        return;
+                    }
+                };
+
+                tracing::info!("take order confirmation received, starting comit protocol");
```
tcharding, line 663:
```suggestion
                tracing::info!("take order confirmation received, starting communication protocols");
```

api_tests/tests/orderbook.ts
```diff
@@ -7,10 +7,21 @@ import OrderFactory from "../src/actors/order_factory";
 import { twoActorTest } from "../src/actor_test";
 import { Entity, Link } from "comit-sdk/dist/src/cnd/siren";
 import { sleep } from "../src/utils";
+import SwapFactory from "../src/actors/swap_factory";
 
 it(
-    "orderbook_bob_makes_order_alice_takes_order",
+    "orderbook_bob_makes_order_alice_takes_order_both_redeem",
```
thomaseizinger, line 13: This test is pretty long and verbose. It is quite hard to understand what is being tested here.

Ideally, the test reads exactly according to the mental model of what you are testing. For orderbook + doing a swap, that is:

- Bob makes an order
- Alice looks at her orders and takes one
- Alice funds
- Bob funds
- Alice redeems
- Bob redeems

It is hard to avoid all boilerplate due to the need for setting up stuff etc but we should at least aim as high as we can!
Conceptually, this test is not a whole lot different to the others. Just instead of saying `actor.create` we use the orderbook to `make` and `take` orders.
Hence, our target should be to make this test roughly as concise as the other e2e tests.

Let me know if you need help with that :)

api_tests/tests/orderbook.ts
tcharding: This comment refers to the `// @ts-ignore`

rishflab:
replaced with:
https://github.com/comit-network/comit-rs/pull/2966 and another soon to be created PR to clean up the test code

PR #2964, opened 2020-07-16
Some more cleanup to the orderbook

tcharding:
This PR does:

- Add todos from previous PR review
- Improve documentation of the order and orderbook
- Removes the announce trading pair logic from cnd and the orderbook

Reason for announce trading pair removal: We only support a single trading pair, all the announce logic is not needed, it adds unnecessary complexity, code to maintain and reason about.

### DIFF
--- a/cnd/src/http_api/route_factory.rs
+++ b/cnd/src/http_api/route_factory.rs
@@ -155,15 +155,8 @@ pub fn create(facade: Facade, allowed_origins: &AllowedOrigins) -> BoxedFilter<(
         .and(warp::path!("dial"))
         .and(warp::path::end())
         .and(warp::body::json())
-        .and(facade.clone())
-        .and_then(orderbook::post_dial_peer);
-
-    let post_announce_trading_pair = warp::post()
-        .and(warp::path!("announce"))
-        .and(warp::path::end())
-        .and(warp::body::json())
         .and(facade)
-        .and_then(orderbook::post_announce_trading_pair);
+        .and_then(orderbook::post_dial_peer);
 
     preflight_cors_route
         .or(get_peers)
@@ -184,7 +177,6 @@ pub fn create(facade: Facade, allowed_origins: &AllowedOrigins) -> BoxedFilter<(
         .or(get_orders)
         .or(get_order)
         .or(post_dial_addr)
-        .or(post_announce_trading_pair)
         //.or(get_makers)
         //.or(subscribe)
         .recover(http_api::unpack_problem)
--- a/comit/src/network/protocols/orderbook.rs
+++ b/comit/src/network/protocols/orderbook.rs
@@ -29,6 +29,9 @@ use std::{
 
 pub use self::order::*;
 
+// TODO: Audit function scope
+// TODO: Rethink topic logic
+
 /// String representing the BTC/DAI trading pair.
 const BTC_DAI: &str = "BTC/DAI";
 
--- a/comit/src/network/protocols/orderbook.rs
+++ b/comit/src/network/protocols/orderbook.rs
@@ -97,9 +97,9 @@ impl Orderbook {
     pub fn make(&mut self, order: Order) -> anyhow::Result<OrderId> {
         let ser = bincode::serialize(&Message::CreateOrder(order.clone()))?;
         let topic = order.tp().topic();
-        let id = order.id;
-
         self.gossipsub.publish(&topic, ser);
+
+        let id = order.id;
         self.orders.insert(id, order);
 
         Ok(id)
--- a/comit/src/network/protocols/orderbook.rs
+++ b/comit/src/network/protocols/orderbook.rs
@@ -183,17 +183,6 @@ impl Orderbook {
 #[error("order not found in orderbook: {0:?}")]
 pub struct OrderNotFound(OrderId);
 
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
-pub enum TradingPair {
-    BtcDai,
-}
-
-impl TradingPair {
-    pub fn topic(&self) -> Topic {
-        Topic::new(BTC_DAI.to_string())
-    }
-}
-
 /// MakerId is a PeerId wrapper so we control serialization/deserialization.
 #[derive(Debug, Clone, PartialEq)]
 pub struct MakerId(PeerId);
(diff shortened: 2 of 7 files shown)


comit/src/network/protocols/orderbook/order.rs
```diff
@@ -78,6 +81,16 @@ fn meaningless_expiry_value() -> u32 {
     100
 }
 
+/// The position the maker takes in this order. A BTC/DAI buy order,
+/// also described as an order that buys the trading pair BTC/DAI,
+/// means that the order maker buys the base currency (in this case
```
thomaseizinger, line 103:
```suggestion
/// means that the maker buys the base currency (in this case
```

cnd/src/http_api/route_factory.rs
```diff
@@ -155,15 +155,8 @@ pub fn create(facade: Facade, allowed_origins: &AllowedOrigins) -> BoxedFilter<(
         .and(warp::path!("dial"))
         .and(warp::path::end())
         .and(warp::body::json())
-        .and(facade.clone())
-        .and_then(orderbook::post_dial_peer);
-
-    let post_announce_trading_pair = warp::post()
```
thomaseizinger, line 161: Removing this will break the e2e test, please adapt accordingly.

comit/src/network/protocols/orderbook.rs
```diff
@@ -179,19 +179,9 @@ impl Orderbook {
...
-    IdentitiesForOrderNotFound(OrderId),
-    #[error("could not take order because not found")]
-    OrderNotFound(OrderId),
-    #[error("could not subscribe to all peers for topic")]
-    Subscribe,
-    #[error("could not unsubscribe to all peers for topic")]
-    UnSubscribe,
-}
+#[derive(PartialEq, Clone, Copy, Debug, thiserror::Error)]
+#[error("order not found in orderbook: {0:?}")]
```
thomaseizinger, line 183:
```suggestion
#[error("order {0} not found in orderbook")]
```

1. `Display` instead of `Debug` makes for a nicer output
2. Putting the ID in the sentence is closer to an English sentence

Thoughts?

comit/src/network/protocols/orderbook/order.rs
```diff
@@ -60,6 +61,22 @@ impl Order {
...
+// this and the calls to order.tp().topic() to make it explicit that
+// there is only a single trading pair and the trading pair is
+// defined by the order struct.
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub enum TradingPair {
+    BtcDai,
+}
+
+impl TradingPair {
+    pub fn topic(&self) -> Topic {
```
thomaseizinger, line 75:
```suggestion
    pub fn to_topic(&self) -> Topic {
```

`to_XYZ` would convey the allocation that is happening as part of this function call.

comit/src/network/protocols/orderbook/order.rs
tcharding: Here I am specifically trying to link the maker of the order with the position i.e., the position is from the perspective of person who makes the order.

cnd/src/http_api/route_factory.rs
tcharding: Not the e2e test on dev :) Thanks for the heads up though I'll keep in mind when working on the e2e test.

comit/src/network/protocols/orderbook.rs
tcharding: Typically I prefer to have the bit of an error message that changes at the end so its easier to visually parse the logs. That might just be a personal thing though. It makes it easier to write log statements that are uniform since so one can scan down the logs and assume any time a id appears it will be at the end. Although now I write this I realise that with line wrapping maybe this is not a real benefit? I'm not fussed which way we go but I'd like to be uniform.

tcharding: Display vs Debug, you are definitely correct. thanks

comit/src/network/protocols/orderbook/order.rs
thomaseizinger: Ok, now I get it but I find it quite confusing to read TBH :sweat_smile: 

It is sort of implied that they are making an order and not cake so I would still drop it :)

comit/src/network/protocols/orderbook/order.rs
tcharding: Whats the convention for consuming self vs not consuming self? I thought `to_xyz` functions consumed self. Would definitely like to convey the allocation though.

comit/src/network/protocols/orderbook/order.rs
tcharding: lol, I'll take some cake, make cake, take cake - definitely take.

tcharding: Will try to better English this whole comment.

tcharding: Is this clearer?
```
/// The position of the maker for this order. A BTC/DAI buy order,
/// also described as an order that buys the trading pair BTC/DAI,
/// means that the maker buys the base currency (in this case BTC) in
/// return for DAI. A sell order means that the maker sells BTC and
/// receives DAI.
///
/// Please note: we do not set the base currency to 1 and use rate
/// (i.e., quote currency) and amount as is commonly done in Forex
/// trading. We use the amounts of each currency to determine the
/// rate.
```

@lucas, you are the resident English expert - can you read over this comment please and give your input?

comit/src/network/protocols/orderbook/order.rs
tcharding: `std::string::ToString` takes `&self` and returns `String` implying that your suggestion is correct. Is this the canonical trait for such things?

comit/src/network/protocols/orderbook/order.rs
thomaseizinger: That one is clear to me :)

comit/src/network/protocols/orderbook.rs
thomaseizinger: We are talking about errors here and not general log statements though. For logs, I'd say the bit that changes should ideally be in a tracing context otherwise you are likely going to repeat it over several statements.

Since errors cross function boundaries, they are usually logged outside of the scope they were created in and then it helps if they are a self-contained and easily readable message.

comit/src/network/protocols/orderbook/order.rs
thomaseizinger: > Whats the convention for consuming self vs not consuming self? I thought `to_xyz` functions consumed self. Would definitely like to convey the allocation though.

If self is consumed, then the method should be called `into_xyz`. See here: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

Issue #2927, opened 2020-07-06
Find a way to run Rust tests on MacOS and Windows


thomaseizinger:
## Problem

At the moment, we are only compiling the code on MacOS and Windows. Tests are only run under Linux due to the docker dependency of `testcontainers`.

## Goal

Find a way to do as many checks as possible on MacOS and Windows during CI.

Motivated by: https://github.com/comit-network/comit-rs/pull/2928

thomaseizinger:
Fixed by https://github.com/comit-network/comit-rs/commit/24d5a1b02d5b826b9e234f8476f27ce0433e69b8.

thomaseizinger:
Ah, we never merged that :D

D4nte:
@thomaseizinger are we running tests or not now?

thomaseizinger:
Nope, we are just running them on Linux atm.

PR #2970, opened 2020-07-17
Run Rust tests on MacOS in CI

thomaseizinger:
Fixes https://github.com/comit-network/comit-rs/issues/2927.

### DIFF
--- a/comit/Cargo.toml
+++ b/comit/Cargo.toml
@@ -34,6 +34,9 @@ serdebug = "1"
 sha2 = "0.9"
 strum = "0.18.0"
 strum_macros = "0.18.0"
+# dev-dependencies can't be optional :(
+# https://github.com/rust-lang/cargo/issues/1596
+testcontainers = { version = "0.9", optional = true }
 thiserror = "1"
 tokio = { version = "0.2.21", features = ["sync"] }
 tracing = "0.1.16"
@@ -47,7 +50,6 @@ libp2p = { version = "0.21", default-features = false, features = ["secio", "yam
 log = { version = "0.4", features = ["serde"] }
 proptest = "0.10.0"
 spectral = { version = "0.6", default-features = false }
-testcontainers = "0.9"
 tokio = { version = "0.2.21", features = ["macros"] }
 tracing = { version = "0.1", features = ["attributes"] }
 tracing-core = "0.1"
@@ -58,3 +60,7 @@ tracing-subscriber = "0.2"
 [features]
 default = []
 test = ["libp2p/secio", "libp2p/yamux"]
+
+[[test]]
+name = "bitcoin_transaction_pattern_e2e"
+required-features = ["testcontainers"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,10 +97,14 @@ jobs:
       - name: Build docs
         run: make doc
 
-      # Ignore tests on macos due to missing docker
       - name: Run unit tests
-        if: matrix.os == 'ubuntu-latest'
         run: make test
+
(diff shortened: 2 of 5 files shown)


thomaseizinger:
> so now we need to run `cargo test --features testcontainers` locally to ensure that all test pass. Can't even run `make test` for that. Can we have at least a makefile shortcut?

I didn't adapt the Makefile because you argued in the past that the Makefile should do the same thing on CI and locally.

We are generally erring on the side of using the Makefile to optimize the local development environment. I am happy to change the Makefile but that implies for me that we shouldn't use it in CI anymore if we want to be consistent.

I would prefer that to be honest. Primarily because CI and local have very different needs in terms of optimiztion and compared to languages like C or C++, cargo is a powerful enough package manager to act as a solid base layer that makes sure everything is compiled accordingly.

That is, as long as we stick to cargo's conventions as much as possible. We do that already, hence our Makefile is just a thin wrapper around cargo anyway.

> I prefer the way I did with nectar where one can `deactivate` the docker tests if they can't run docker. I understand this is not possible to do the feature way (as you mentioned that dev-dependencies cannot be optional).

Using features this way can be tricky though. While it may not be relevant in this specific case, deactivating default features only works if every crate in the dependency chain opts out of the default features.
Hence, I'd prefer if we stick to cargo's conventions here and make sure that building and testing works on every platform _by default_ and users can opt-in to more things.

An unsolved problem of not using the Makefile in CI is how we want to deal with applying and checking the formatting without too much duplication. I don't have an answer to this.

Generally, I'd happy if we can merge this before solving all these problems.
The presented patch makes our CI more robust at the expense of a minor annoyance if you want to run that **one** additional testcontainers test.

comit/Cargo.toml
```diff
@@ -34,6 +34,9 @@ serdebug = "1"
 sha2 = "0.9"
 strum = "0.18.0"
 strum_macros = "0.18.0"
+# dev-dependencies can't be optional :(
```
bonomat, line 37: `can't be optional` but `optional=true`? I don't understand :)

D4nte:
> Hence, I'd prefer if we stick to cargo's conventions here and make sure that building and testing works on every platform by default and users can opt-in to more things.

I'd be curious to have a source to this statement. If it were the case then `std` would be opt-in and not opt-out in most crates.

thomaseizinger:
> > Hence, I'd prefer if we stick to cargo's conventions here and make sure that building and testing works on every platform by default and users can opt-in to more things.
> 
> I'd be curious to have a source to this statement. If it were the case then `std` would be opt-in and not opt-out in most crates.

I can't find a specific source for it but it is basically the result of logical deduction around how features are designed in cargo.

Features are accumulated across the dependency graph, which means as soon as one crate opts in to a feature, it is impossible to remove it from the graph.

Most crates have a "no-std" feature that actually _opts-in_ to not having `std`.

I can't find the thread at the moment but I've read somewhere that the Rust team actually intends to modularise `std` with cargo features so there is a more fine-grained choice in what you get. The reasoning behind that is `no_std` removes too much and it doesn't clearly map to certain environments. It works for micro-controllers but then there is also cases where you actually can allocate but you don't have other things that you get with `std`.

Once that is achieved, crates opt-in to only what they need and expose more features to add more functionality instead of adding a feature that removes all of `std`.

comit/Cargo.toml
thomaseizinger: Read the comment again :)

"dev-dependencies can't be optional" - hence this dependency is under `[dependencies]` although it is just used for tests ;)

bonomat: aahhh... got it :D

PR #2971, opened 2020-07-17
Run e2e tests on MacOS in CI

thomaseizinger:


.github/workflows/ci.yml
```diff
@@ -57,6 +57,8 @@ jobs:
         include:
           - os: ubuntu-latest
             e2e: true
+          - os: macos-latest
```
bonomat, line 60: Simpler would be to explicitly *not* run it on windows-latest

tcharding:
Kicked bors without second approval while trying to debug bors crash.

.github/workflows/ci.yml
thomaseizinger: Agreed but I didn't want to mess with the config too much.

Ideally, I would run them on windows too but for that we need to change how we download the blockchain nodes.

PR #2979, opened 2020-07-20
bidirectional e2e test for orderbook 

rishflab:
Added bidirectional test path in the orderbook e2e test. Refactored the test to improve code reuse and make the test less noisy and easier to understand.

In order to get bidirectional trades working happen the orderbook API was changed. Refund and redeem identity has been replaced bitcoin and ethereum identity. Derivation of transient identities has also changed to take into consideration the direction of the trade.

### DIFF
--- a/api_tests/src/actors/actor.ts
+++ b/api_tests/src/actors/actor.ts
@@ -353,7 +353,6 @@ export class Actor {
             const aliceTakeOrderResponse = await this.cnd.executeSirenAction(
                 aliceOrderTakeAction,
                 async (field) => {
-                    // this could be wrong
                     if (field.name === "bitcoin_identity") {
                         // @ts-ignore
                         return Promise.resolve(bitcoinIdentity);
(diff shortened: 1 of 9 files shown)


api_tests/src/actors/actor.ts
```diff
@@ -256,50 +254,176 @@ export class Actor {
...
+            );
+            // Alice executes the siren take action extracted in the previous line
+            // The resolver function fills the refund and redeem address fields required
+            // "POST /orders/63c0f8bd-beb2-4a9c-8591-a46f65913b0a/take"
+            // Alice receives a url to the swap that was created as a result of taking the order
+            // @ts-ignore
+            const aliceTakeOrderResponse = await this.cnd.executeSirenAction(
+                aliceOrderTakeAction,
+                async (field) => {
+                    // this could be wrong
```
tcharding, line 356: Stale comment?

rishflab: yep

PR #2988, opened 2020-07-22
Remove old take_order module

tcharding:
This module is not built in, it is an artefact that should have been
removed.  Remove it.


tcharding:
This is a trivial patch, the compiler does all the review that is required. Merging with single review.  bors r+

PR #2990, opened 2020-07-23
Pre-work from the forex terminology PR

tcharding:
This PR is some the initial pre-work patches from https://github.com/comit-network/comit-rs/pull/2983

This stuff is useful on its own even though the motivation comes from the forex work.

Please review carefully the functions added/modified in the `ethereum` module.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -317,7 +317,7 @@ version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4152116fd6e9dadb291ae18fc1ec3575ed6d84c29642d97890f4b4a3417297e4"
 dependencies = [
- "generic-array 0.14.2",
+ "generic-array 0.14.3",
 ]
 
 [[package]]
@@ -397,12 +397,9 @@ dependencies = [
 
 [[package]]
 name = "bytes"
-version = "0.5.5"
+version = "0.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "118cf036fbb97d0816e3c34b2d7a1e8cfc60f68fcf63d550ddbe9bd5f59c213b"
-dependencies = [
- "loom",
-]
+checksum = "0e4cec68f03f32e44924783795810fa50a7035d8c8ebe78580ad7e6c703fba38"
 
 [[package]]
 name = "c_linked_list"
@@ -447,15 +444,6 @@ dependencies = [
  "textwrap",
  "unicode-width",
  "vec_map",
-]
-
-[[package]]
-name = "clear_on_drop"
-version = "0.2.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c9cc5db465b294c3fa986d5bbb0f3017cd850bff6dd6c52f9ccff8b4d21b7b08"
-dependencies = [
- "cc",
 ]
 
 [[package]]
@@ -640,9 +628,9 @@ checksum = "b3a71ab494c0b5b860bdc8407ae08978052417070c2ced38573a9157ad75b8ac"
 
 [[package]]
 name = "cpuid-bool"
-version = "0.1.0"
+version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d375c433320f6c5057ae04a04376eef4d04ce2801448cf8863a78da99107be4"
+checksum = "8aebca1129a03dc6dc2b127edd729435bbc4a37e1d5f4d7513165089ceb02634"
 
 [[package]]
 name = "crossbeam-utils"
@@ -781,7 +769,7 @@ version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d3dd60d1080a57a05ab032377049e0591415d2b31afd7028356dbf3cc6dcb066"
 dependencies = [
- "generic-array 0.14.2",
+ "generic-array 0.14.3",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


comit/src/asset/ethereum/erc20.rs
```diff
@@ -22,11 +27,18 @@ impl Erc20Quantity {
...
 
-    pub fn from_wei_dec_str(str: &str) -> Result<Self, Error> {
+    pub fn try_from_wei_dec_str(str: &str) -> Result<Self, Error> {
         let int = BigUint::from_str_radix(str, 10)?;
         Ok(Self::try_from_wei(int)?)
     }
 
+    /// Create a erc20 quantity from a float string. E.g., this can be used to
+    /// create a DAI quantity from the ergonomic string "1234.56".
+    pub fn try_from_float(eth: &str) -> anyhow::Result<Self> {
```
D4nte, line 37:
```suggestion
    pub fn try_from_eth_str(eth: &str) -> anyhow::Result<Self> {
```
Let's try to keep the function names similar.

comit/src/asset/ethereum/erc20.rs
```diff
@@ -22,11 +27,18 @@ impl Erc20Quantity {
         self.0.to_str_radix(10)
     }
 
-    pub fn from_wei_dec_str(str: &str) -> Result<Self, Error> {
+    pub fn try_from_wei_dec_str(str: &str) -> Result<Self, Error> {
```
D4nte, line 30:
```suggestion
    pub fn try_from_wei_dec_str(wei: &str) -> Result<Self, Error> {
```
Just to make it clearer as the type is already expressed in the signature.

comit/src/float_math.rs
```diff
@@ -0,0 +1,223 @@
+use bitcoin::hashes::core::cmp::Ordering;
```
D4nte, line 1: So actually the name of the file is incorrect 😁 We are doing fixed point arithmetic here.
The aim of the module is to brought some safety in handling fractional numbers and doing maths/arithmetic with them. Some ideas for a better name:
- fraction
- fractional number
- maths
- arithmetic
- fixed point

comit/src/float_math.rs
```diff
@@ -0,0 +1,223 @@
...
+use std::str::FromStr;
+
+/// Truncate the float's mantissa to length `precision`.
+pub fn truncate(float: f64, precision: u16) -> f64 {
+    let s = float.to_string();
+    truncate_str(&s, precision)
+}
+
+/// Truncate the float's mantissa to length `precision`.
+pub fn truncate_str(float: &str, precision: u16) -> f64 {
```
D4nte, line 12: I think it would make more sense for it to return a string as it took a string slice as input.  the conversion back in a float can be done in `truncate`.

```suggestion
pub fn truncate_str(number: &str, precision: u16) -> f64 {
```

comit/src/float_math.rs
```diff
@@ -0,0 +1,223 @@
...
+use num::{BigUint, Zero};
+use std::str::FromStr;
+
+/// Truncate the float's mantissa to length `precision`.
+pub fn truncate(float: f64, precision: u16) -> f64 {
+    let s = float.to_string();
+    truncate_str(&s, precision)
+}
+
+/// Truncate the float's mantissa to length `precision`.
```
D4nte, line 11:
```suggestion
/// Truncate the number's mantissa to length `precision`.
```

I have loosely used the word `float` when I actually mean `number` or more precisely `fractional number`.

comit/src/float_math.rs
```diff
@@ -0,0 +1,223 @@
...
+pub fn truncate_str(float: &str, precision: u16) -> f64 {
+    let mut string = float.to_string();
```
D4nte, line 13: Because this accepts a string, there is no guarantee the string actually represents a number.
We should check that and also return a result.
Also, we should proptest this function with any arbitrary string value and ensure it does not panic/overflows.

comit/src/float_math.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    let index = string.find('.');
+
+    match index {
+        None => f64::from_str(float).expect("This has no decimal, should convert"),
```
D4nte, line 18: This would panic if the string passed is not a number.

comit/src/float_math.rs
```diff
@@ -0,0 +1,223 @@
...
+/// Call truncate(float) before using this function if you need to.
+// TODO: Enforce the above in code.
+pub fn multiply_pow_ten_str(float: &str, pow: u16) -> anyhow::Result<BigUint> {
+    let mut float = float.to_string();
+    let decimal_index = float.find('.');
+
+    match decimal_index {
+        None => {
+            let zeroes = "0".repeat(pow as usize);
+            Ok(BigUint::from_str(&format!("{}{}", float, zeroes)).expect("an integer"))
```
D4nte, line 57: It accepts any string so that can lead to a panic. Validate the string and proptest looking for panics.

comit/src/network/protocols/orderbook.rs
```diff
@@ -29,8 +30,10 @@ use std::{
 
 pub use self::order::*;
 
-/// String representing the BTC/DAI trading pair.
-const BTC_DAI: &str = "BTC/DAI";
+lazy_static::lazy_static! {
+    /// The Topic used to publish BTC/DAI orders.
+    static ref BTC_DAI_TOPIC: Topic = Topic::new("BTC/DAI".to_string());
```
D4nte, line 35: So, I guess I should start a mailing list discussion about the topic?

tcharding:
@D4nte, I did not go over yet your ideas re foat_math module. I have removed it from this PR simply because we shouldn't push anything into COMIT lib that is not spot on. This PR is meant to be simple pre-work, probably should never have included the float_math stuff. I also split it into two separate PRs , one for ethereum stuff and one for orderbook stuff. I'd like to flag this PR somehow so your suggestions are not lost, I've saved it as a bookmark - any better workflow ideas welcome.

tcharding:
Kicking in the stripped down PR with just the approved patches: bors r+

comit/src/asset/ethereum/erc20.rs
tcharding: Well picked up (the inconsistency), thanks. Will keep in mind (the ethereum stuff is on hold until float_math goes in).

comit/src/asset/ethereum/erc20.rs
tcharding: Better, thanks.

comit/src/float_math.rs
tcharding: +1 for `fraction`

comit/src/float_math.rs
tcharding: Aggree.

comit/src/float_math.rs
tcharding: How about we favour the term `fraction` in identifiers if we choose to use that as the filename?

comit/src/float_math.rs
tcharding: Oh, you mean we should actually write the API properly instead of me just hacking ups some shit and trying to merge it into the comit lib - if this is what you mean you are totally correct. Sloppy work here from me. I think we should move comit to a separate crate and enforce much stricter code practices on it.

comit/src/network/protocols/orderbook.rs
tcharding: As I've said in other places, the whole gossipsub needs rethinking/redesigning IMNSHO. Please do hit the mailing list, I'm interested in you view, I have some ideas also.

PR #2991, opened 2020-07-23
Subscribe after we have a valid connection

tcharding:
Making draft while debugging other stuff.

If this issue on github is correct

https://github.com/libp2p/rust-libp2p/issues/1671#issuecomment-662709729

then we should have an open connection to another node on the network
before subscribing.

Add a method to the orderbook and all the layes above it to facilitate
subscribing. Call it after a succesfull dial is made.

### DIFF
--- a/api_tests/src/actors/order_factory.ts
+++ b/api_tests/src/actors/order_factory.ts
@@ -29,15 +29,16 @@ interface Ethereum {
 
 export default class OrderbookFactory {
     public static async connect(alice: Actor, bob: Actor) {
-        // Get alice's listen address
-        const aliceAddr = await alice.cnd.getPeerListenAddresses();
-
-        // Bob dials alices
+        // Alice dials Bob, must be this way around in order to ensure
+        // Alice subscribes to the BTC/DAI topic _after_ she has a connection.
+        const addr = await bob.cnd.getPeerListenAddresses();
         // @ts-ignore
-        await bob.cnd.client.post("dial", { addresses: aliceAddr });
+        await alice.cnd.client.post("dial", { addresses: addr });
 
         /// Wait for alice to accept an incoming connection from Bob
         await sleep(1000);
+
+        // TODO: This could be wait until a peer id is returned on the /peers endpoint.
     }
 
     public static async initWalletsForBtcDaiOrder(alice: Actor, bob: Actor) {
--- a/cnd/src/http_api/dial_addr.rs
+++ b/cnd/src/http_api/dial_addr.rs
@@ -20,5 +20,12 @@ pub async fn post_dial_addr(
     for addr in body.addresses {
         facade.dial_addr(addr).await;
     }
+
+    // Best effort, assume we got a connection.
+    let sub = facade.subscribe().await;
+    if !sub {
+        tracing::warn!("failed to subscribe to orderbook gossipsub for BTC/DAI");
+    }
+
     Ok(warp::reply::reply())
 }
--- a/comit/src/network/protocols/orderbook.rs
+++ b/comit/src/network/protocols/orderbook.rs
@@ -72,7 +72,7 @@ impl Orderbook {
             config,
         );
 
-        let mut orderbook = Orderbook {
+        let orderbook = Orderbook {
             peer_id,
             gossipsub,
             take_order: behaviour,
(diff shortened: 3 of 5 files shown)


api_tests/src/actors/order_factory.ts
```diff
@@ -29,15 +29,16 @@ interface Ethereum {
...
 export default class OrderbookFactory {
     public static async connect(alice: Actor, bob: Actor) {
-        // Get alice's listen address
-        const aliceAddr = await alice.cnd.getPeerListenAddresses();
-
-        // Bob dials alices
+        // Alice dials Bob, must be this way around in order to ensure
+        // Alice subscribes to the BTC/DAI topic _after_ she has a connection.
+        const addr = await bob.cnd.getPeerListenAddresses();
         // @ts-ignore
```
D4nte, line 35: It's always good to justify the `@ts-ignore` so one can quickly whether it still applies.

cnd/src/http_api/dial_addr.rs
```diff
@@ -20,5 +20,12 @@ pub async fn post_dial_addr(
     for addr in body.addresses {
         facade.dial_addr(addr).await;
     }
+
+    // Best effort, assume we got a connection.
```
D4nte, line 24: I don't understand, isn't the whole idea of this patch to _ensure_ we got a connection and not _assume_ it?

comit/src/network/protocols/orderbook.rs
```diff
@@ -72,20 +72,27 @@ impl Orderbook {
...
-            .subscribe(Topic::new(BTC_DAI.to_string()));
+    }
 
-        orderbook
+    /// Subscribe to the BTC/DAI gossipsub topic. Caller is responsible for
+    /// ensuring that there is an open connection to another node in the network
+    /// before calling.
+    // Currently we call subscribe after dialing, this implies that in order to
+    // receive orders on the gossipsub network one must dial to another node. For a
+    // node acting as a maker only it probably does not matter if we do not receive
```
D4nte, line 91: I believe @bonomat had a strong opinion on the subject that all takers and makers must participate to the gossip network.
I understand this workaround is temporary and being worked on upstream, right?

D4nte:
Wait a minute.

Didn't we say that if the maker `denies` an order being taken then it closes the connection? Or is that on a different socket? When do we open orderbook connections? What closes them?

api_tests/src/actors/order_factory.ts
tcharding: Good point, thanks. I have been asking if these ts-ignores are needed since this work started but got no answer.

cnd/src/http_api/dial_addr.rs
tcharding: We can't ensure we have a connection, the aim of this patch is to subscribe at a time when we can guess that we might have a connection. This patch tries to apply a bandaid solution to the problem that subscription before connection may be problematic, if it is not a good bandaid, or a fully thought out solution is deemed necessary, then we should totally do that. The issue that triggered this work is under discussion on github, I'm following it, that's why this is draft.

Thanks for asking though because I'm not convince that the gossipsub stuff is fully thought out in general - at least its not fully thought out by me.

comit/src/network/protocols/orderbook.rs
tcharding: It is my opinion that we need a sit down with the folks invested in the orderbook to discuss the design, the goals, and what not. @bonomat, @rishflab, and myself have a meeting today about it but that does not include anyone from Nectar. I personally do not know why the orderbook was designed the way it is, I do not fully understand what the goal is with it, with Nectar, or with cnd's usage. Perhaps its just me?

tcharding:
> Wait a minute.
> 
> Didn't we say that if the maker `denies` an order being taken then it closes the connection? Or is that on a different socket? When do we open orderbook connections? What closes them?

Denying closes the substream not the connection.

api_tests/src/actors/order_factory.ts
D4nte: Remove them and see if you can run the e2e tests :) but yeah, need to justify exactly for the reason you are saying! Maybe they are here because the types of an upstream lib are not up to date, meaning they can be removed one the upstream lib is updated.
Or maybe they are here because we are breaking the type system in purpose, in this case it's good to explain why.

comit/src/network/protocols/orderbook.rs
D4nte: We actually inadvertly started the discussion on the mailing so let's see how it goes there.

da-kami:
@tcharding to clarify what is the intention of this PR: 

Is the intention of this PR to change the initial connection behaviour, so `Bob` does not dial `Alice` anymore, but `Alice` would just subscribe to the gossipsub topic?

As far as I know so far `Bob` had to dial `Alice` (the dial endpoint on the `http_api` passed `Alice`'s multiaddress to `Bob` and then he dialed on the swarm. Then `Alice` could receive `Orders`. - With this change we expect the same outcome but we don't need the `dial` from `Bob` to `Alice` anymore?

api_tests/src/actors/order_factory.ts
tcharding: Cool thanks, fixed: https://github.com/comit-network/comit-rs/pull/3009

api_tests/src/actors/order_factory.ts
```diff
@@ -29,15 +29,16 @@ interface Ethereum {
...
-        await bob.cnd.client.post("dial", { addresses: aliceAddr });
+        await alice.cnd.client.post("dial", { addresses: addr });
```
tcharding, line 36: @bonomat, evidence that I patched this at least once. Did I do this in review prior - solo dios sabe.

tcharding:
Closing this, seems that subscribing on startup _should_ work. Probably good to keep this in mind if we get intermittent subscription failures (i.e. missing messages) but since we have an integration test hopefully we catch this before it gets to the e2e tests.

PR #2993, opened 2020-07-24
Add support for partial order taking

rishflab:
The taker can specify a quantity in the take order request allowing the taker to partialy fill an order. Ethereum amount in the Order struct was replaced with a DAI to BTC conversion rate. The ethereum amount is calculated on the fly using this rate and the Bitcoin amount when a partial take request is made and received. This method avoids rounding issues that may occur if the taker specifies a take amount that does not divide "cleanly" with the erc20 and btc amounts.


api_tests/tests/orderbook.ts
```diff
@@ -10,21 +10,40 @@ import { getIdentities } from "../src/actors/defaults";
 
 // todo: move test initialisation into single mega function to reduce noise
 describe("orderbook", () => {
+    // direct quote
+    // BTC/DAI
+    // position: buy
+    // quote: 1.71 (DAI/BTC)
```
tcharding, line 16: I finally worked out what you mean here 'DAI per BTC', this is a really confusing usage because two assets with a slash is used two lines above to mean a totally different thing i.e., the currency pair (is is customary).

Please write something like: 'quoted in units of DAI'

tcharding: or  'DAI per unit of BTC'

rishflab:
> The name of this PR is totally misleading - this PR in now way adds 'support for partial order taking' since it does not even touch Rust code.

will be added shortly in subsequent commits

api_tests/tests/orderbook.ts
```diff
@@ -7,16 +7,32 @@ import { twoActorTest } from "../src/actor_test";
 import OrderFactory from "../src/actors/order_factory";
 import { sleep } from "../src/utils";
 
-// todo: move test initialisation into single mega function to reduce noise
 describe("orderbook", () => {
+    // direct quote
```
tcharding, line 11: This is an 'indirect' quote, not a direct quote.

api_tests/tests/orderbook.ts
```diff
@@ -7,16 +7,32 @@ import { twoActorTest } from "../src/actor_test";
...
+    // pair: "BTC/DAI"
+    // position: buy
+    // rate: 100 (1000 wei for 1 satoshi)
+    // amount: 900,000 satoshi BTC
```
tcharding, line 15: Please use 'price' and 'quantity' as outlined here: https://3.basecamp.com/4403044/buckets/15109376/messages/2878398232

api_tests/tests/orderbook.ts
```diff
@@ -7,16 +7,32 @@ import { twoActorTest } from "../src/actor_test";
...
         "btc_dai_buy_order",
         twoActorTest(async ({ alice, bob }) => {
             await alice.connect(bob);
-            const order = await OrderFactory.newBtcDaiOrder(alice, bob, "buy");
+            const order = await OrderFactory.newBtcDaiOrder(
+                alice,
+                bob,
+                "buy",
+                100,
+                "900000"
```
tcharding, line 30: Did you mean to write price of 9000?

api_tests/tests/orderbook.ts
```diff
@@ -35,20 +51,37 @@ describe("orderbook", () => {
             await bob.assertBalancesAfterSwap();
         })
     );
+    // direct quote
+    // pair: "BTC/DAI"
+    // position: sell
+    // rate: 100 (1000 wei for 1 satoshi)
+    // amount: 900,000 satoshi BTC
```
tcharding, line 58: 'satoshi BTC'?

cnd/src/http_api/orderbook.rs
```diff
@@ -226,11 +226,11 @@ pub async fn get_orders(facade: Facade) -> Result<impl Reply, Rejection> {
 #[derive(Clone, Debug, Deserialize)]
 struct MakeOrderBody {
     position: Position,
+    rate: Rate,
```
tcharding, line 229: This is no good, please review @D4nte's work on floating point math (in project Nectar).

api_tests/tests/orderbook.ts
rishflab: The nectar code base is using rate and amount so maintained consistency with nectar

cnd/src/http_api/orderbook.rs
rishflab: I have looked at the code in project nectar. I don't understand why comit-rs requires the floating point conversion? I will add decimal string deserialisation but I still don't see why we need floating point conversions

api_tests/src/actors/actor.ts
```diff
@@ -337,7 +338,7 @@ export class Actor {
     /**
      * Takes a BtcDai sell order (herc20-hbit Swap)
      */
-    public async takeOrder() {
+    public async takeOrder(partialTakeAmount: string) {
```
thomaseizinger, line 341: I think this parameter should be optional, otherwise one has to unnecessarily pass in the full amount again if you want to take the full order.

api_tests/src/actors/actor.ts
```diff
@@ -470,6 +481,57 @@ export class Actor {
...
+                await this.setStartingBalance([
+                    this.alphaAsset,
+                    {
+                        ...this.betaAsset,
+                        quantity: "0",
+                    },
+                ]);
+                this.expectedBalanceChanges.set(
+                    toKey(this.betaAsset),
+                    BigInt(Number(this.order.rate) * Number(partialTakeAmount))
```
thomaseizinger, line 510: Is this math really necessary?

If I am taking only the `partialTakeAmount`, shouldn't my balance change exactly by this amount? :)

cnd/src/http_api/orderbook.rs
```diff
@@ -199,14 +191,22 @@ pub async fn get_orders(facade: Facade) -> Result<impl Reply, Rejection> {
...
         };
 
+        let ethereum_amount = siren::Field {
+            name: "bitcoin_amount".to_string(),
+            class: vec!["ethereum".to_string(), "address".to_string()],
+            _type: None,
+            value: None,
+            title: None,
+        };
+
```
thomaseizinger, line 201: So the variable is called `ethereum_amount`, the field's name is `bitcoin_amount` and the classes are `ethereum` and `address`? :D What is going on here?

comit/src/asset/ethereum/erc20.rs
```diff
@@ -50,6 +50,12 @@ impl FromWei<U256> for Erc20Quantity {
     }
 }
 
+impl From<BigUint> for Erc20Quantity {
+    fn from(quantity: BigUint) -> Self {
+        Erc20Quantity(quantity)
+    }
+}
```
thomaseizinger, line 57: :mega: This breaks encapsulation of the `Erc20Quantity` type because it allows construction of it without checking the invariant that quantity must not exceed `max_value`!

comit/src/network/orderbook.rs
```diff
@@ -102,21 +103,43 @@ impl Orderbook {
 
     /// Take an order, called by Alice i.e., the taker.
     /// Does _not_ remove the order from the order book.
-    pub fn take(&mut self, order_id: OrderId) -> anyhow::Result<()> {
+    pub fn take(
+        &mut self,
+        order_id: OrderId,
+        partial_take_amount: asset::Bitcoin,
```
thomaseizinger, line 109: I think this parameter could be optional, in case you didn't want to take the order fully.

thomaseizinger: Either we make the field on the wire also optional or we default to the full amount.

comit/src/network/orderbook.rs
```diff
@@ -251,6 +278,8 @@ pub enum BehaviourOutEvent {
     TakeOrderRequest {
         /// The peer from whom request originated.
         peer_id: PeerId,
+        /// The amount of the order the taker wishes to fill
+        amount: asset::Bitcoin,
```
thomaseizinger, line 282: Adding this field with a type of `asset::Bitcoin` makes our `take-order` protocol now specific to BTC/DAI orders.

Unless we take any measures against this, I would propose to change our protocol name to reflect this and make it specific to taking BTC/DAI orders.

comit/src/network/orderbook/order.rs
```diff
@@ -50,6 +52,26 @@ impl FromStr for OrderId {
...
+/// rate.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum Position {
+    Buy,
+    Sell,
+}
+/// The amount of DAI in wei for one satoshi of BTC
+// todo: deserialise from decimal strings
+pub type Rate = u64;
```
thomaseizinger, line 73: When you come across `Rate` in the code, it gives you a false impression of safety that is associated with newtypes whereas in reality, any u64 will do and the compiler won't complain.

I find this use of type aliases does more harm than good. Type aliases are useful if you want to shorten a long type but this kind of "translation" between primitive types and domain types should either happen as part of a newtype or not at all.

api_tests/tests/orderbook.ts
thomaseizinger: I feel adopting a new terminology should be its own PR, otherwise we will never get this in :grimacing:

api_tests/tests/orderbook.ts
```diff
@@ -35,20 +51,37 @@ describe("orderbook", () => {
...
+    // The take request becomes a swap where Bob receives for 2,000,000 wei for  20,000 satoshi
     it(
         "btc_dai_sell_order",
         twoActorTest(async ({ alice, bob }) => {
             await alice.connect(bob);
-            const order = await OrderFactory.newBtcDaiOrder(alice, bob, "sell");
+            const order = await OrderFactory.newBtcDaiOrder(
+                alice,
+                bob,
+                "buy",
```
thomaseizinger, line 71: The testname still says "sell" but the code does "buy"?

rishflab:
> Thanks for working on this Rishab! :)
> 
> I've made some comments inline on things I noticed!
> 
> Generally, I'd be curious if we necessarily had to change to "rate" terminology as part of this functionality?
> 
> I feel like that is causing more friction because it is sort of half-way through adopting new terminology but also not really.
> Is there maybe a subset of this PR that ships the partial order taking while sticking to the "two amounts" setting?

The problem with using 2 amounts for partial order taking is that alice can specify a partial take amount that leads to rounding errors eg.

ethereum quantity: 100.482
bitcoin_quantity: 4.534

calculate the rate from the quantities: 22.1618876.....(repeating decimals)

alice specifies a partial take amount in bitcoin of: 2.341
The corresponding ethereum quantity is 51.88097887...(repeating decimals)
This number cannot be represented by a erc20 quantity because of the repeating decimals

Potential solutions
1. Implement a default rounding method so that alice and bob agree on the quantities to be swapped?
2. Throw an error if alice specifies a bitcoin quantity that results in a ethereum quantity with repeating decimals?
3 . Switch to "rate" terminology which prevents

I went with option 3 because we are going to move to rate terminology anyway and any work done to implement option 1 or 2 will go to waste.

comit/src/network/orderbook.rs
rishflab: String would be the most flexible

thomaseizinger:
> > Thanks for working on this Rishab! :)
> > I've made some comments inline on things I noticed!
> > Generally, I'd be curious if we necessarily had to change to "rate" terminology as part of this functionality?
> > I feel like that is causing more friction because it is sort of half-way through adopting new terminology but also not really.
> > Is there maybe a subset of this PR that ships the partial order taking while sticking to the "two amounts" setting?
> 
> The problem with using 2 amounts for partial order taking is that alice can specify a partial take amount that leads to rounding errors eg.
> 
> ethereum quantity: 100.482
> bitcoin_quantity: 4.534
> 
> calculate the rate from the quantities: 22.1618876.....(repeating decimals)
> 
> alice specifies a partial take amount in bitcoin of: 2.341
> The corresponding ethereum quantity is 51.88097887...(repeating decimals)
> This number cannot be represented by a erc20 quantity because of the repeating decimals
> 
> Potential solutions
> 
>     1. Implement a default rounding method so that alice and bob agree on the quantities to be swapped?
> 
>     2. Throw an error if alice specifies a bitcoin quantity that results in a ethereum quantity with repeating decimals?
>        3 . Switch to "rate" terminology which prevents
> 
> 
> I went with option 3 because we are going to move to rate terminology anyway and any work done to implement option 1 or 2 will go to waste.

Fair enough, thanks for the explanation :)

api_tests/tests/orderbook.ts
tcharding: I agree that the switch to forex terminology should be a separate PR. The source of truth is the basecamp post, all other projects should by [slowly] migrated.

cnd/src/http_api/orderbook.rs
tcharding: Even without the floating point comparison (assuming satoshi or wei integers) we still need to be able to multiply them together, this requires shifting decimal places to account for the conversion to/from sats and wei. This is what is being worked on in `float_math.rs`

tcharding:
> ethereum quantity: 100.482
> bitcoin_quantity: 4.534
> 
> calculate the rate from the quantities: 22.1618876.....(repeating decimals)
> 
> alice specifies a partial take amount in bitcoin of: 2.341
> The corresponding ethereum quantity is 51.88097887...(repeating decimals)
> This number cannot be represented by a Erc20Quantity because of the repeating decimals

I thought you said above that you don't see why we need floating point math in `comit-rs`?

rishflab:
> > ethereum quantity: 100.482
> > bitcoin_quantity: 4.534
> > calculate the rate from the quantities: 22.1618876.....(repeating decimals)
> > alice specifies a partial take amount in bitcoin of: 2.341
> > The corresponding ethereum quantity is 51.88097887...(repeating decimals)
> > This number cannot be represented by a Erc20Quantity because of the repeating decimals
> 
> I thought you said above that you don't see why we need floating point math in `comit-rs`

We dont. We can and should convert decimal strings to BigUint and u64 and do the maths on these types. I will push a commit that adds this.

comit/src/network/orderbook.rs
thomaseizinger: It is not just about the datatype here though but what the format on the wire is and what the other party can "parse" this into.

However, instead of spending time on designing a protocol now that can be reused across several trading pairs, I'd opt to just make it specific now and we can then consolidate later :)

api_tests/src/actors/actor.ts
rishflab: the partialTakeAmount is specified in bitcoin, the corresponding dai amount for the swap needs to be calculated using by partialTakeAmount * BTCtoDAIConversionRate

comit/src/network/orderbook/order.rs
```diff
@@ -1,9 +1,38 @@
...
 use uuid::Uuid;
 
+/// An order, created by a maker (Bob) and shared with the network via
+/// gossipsub.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Order {
+    pub id: OrderId,
+    pub maker: PeerId,
+    pub position: Position,
+    pub price: BtcDaiRate,
```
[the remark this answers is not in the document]
rishflab: I dont think price is a useful name for a type. Price can be expressed as either an absolute quantity ($10.25) or a rate ($5.45/kg). I think price is a good field name but is too ambiguous for a type name.

cnd/src/network.rs
```diff
@@ -526,6 +534,7 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<orderbook::BehaviourOutEvent> f
         match event {
             orderbook::BehaviourOutEvent::TakeOrderRequest {
                 peer_id,
+                quantity: amount,
```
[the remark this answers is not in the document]
rishflab: yep

comit/src/asset/bitcoin.rs
```diff
@@ -36,6 +37,54 @@ impl fmt::Display for Bitcoin {
...
 }
 
+impl Serialize for Bitcoin {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let precision = 8;
+        let base: u64 = 10;
+        let remainder = self.as_sat() % base.pow(precision);
```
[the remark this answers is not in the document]
rishflab: The remainder here is numbers after the decimal point deser. If you have a sat value of 123_456_789 the remainder will be 9 which becomes the decimals in the serialised string eg. "12_345_678.9" BTC

comit/src/asset/bitcoin.rs
```diff
@@ -36,6 +37,54 @@ impl fmt::Display for Bitcoin {
     }
 }
 
+impl Serialize for Bitcoin {
```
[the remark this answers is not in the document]
rishflab: This ser/deser to/from decimal strings. We were doing it as satoshis which is not user friendly.
If you specify a btc quantity of "123.456" it will convert it to 12_345_600_000 satoshi and vice versa.

rishflab:
> Others have already commented but I'll re-state in case you missed it. Please introduce the forex terminology in a single PR, NOT in the same PR as the partial order taking. There is too much going on in this PR.

I don't think it is worth the time/effort to split it up now. The end result will be the same. I originally intended to do partial take without introducing forex terminology but ran into rounding issues which were solved by switching to forex terminology.

comit/src/asset/bitcoin.rs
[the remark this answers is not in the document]
thomaseizinger: Uhm, but `123456789` sats are `1.23456789` BTC and not `12345678.9` :grimacing:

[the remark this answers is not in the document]
rishflab: it is implemented how you have suggested. corrected my comment
```
  #[test]
    fn btc_serialization_success() {
        let expected = "\"0.12345678\"".to_string();
        let bitcoin = Bitcoin::from_sat(12345678);
        let actual = serde_json::to_string(&bitcoin).expect("failed to serialise bitcoin");
        assert_that(&actual).is_equal_to(expected);
    }
```

tcharding:
> > Others have already commented but I'll re-state in case you missed it. Please introduce the forex terminology in a single PR, NOT in the same PR as the partial order taking. There is too much going on in this PR.
> 
> I don't think it is worth the time/effort to split it up now. The end result will be the same. I originally intended to do partial take without introducing forex terminology but ran into rounding issues which were solved by switching to forex terminology.

Please close this PR then, I will not be reviewing any further.

comit/src/asset/bitcoin.rs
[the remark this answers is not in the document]
thomaseizinger: Why are satoshis not user-friendly? They can be formatted into BTC by the UI if it wishes to. On the API level, being unambiguous and correct is more important than values being formatted for human readability :)

Also, we already have a module that formats a Bitcoin asset in a satoshi string. If we were to add another serialization form, it should be added as another module :)

I should have commented that at first already, sorry for only coming up with this only now :grimacing: 
Can we remove this decimal string stuff again in favor of satoshis please?

comit/src/asset/bitcoin.rs
```diff
@@ -36,6 +37,54 @@ impl fmt::Display for Bitcoin {
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, <D as Deserializer<'de>>::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let precision = 8;
+        let string = String::deserialize(deserializer)?;
+        let v: Vec<&str> = string.as_str().split('.').collect();
+
+        let integer = *v.first().unwrap();
+        let decimals = *v.last().unwrap();
```
thomaseizinger, line 67: These panics can easily be triggered by users if they supply a value that doesn't have a dot in it.

Just mentioning it as general feedback, I don't think we need these deserializers.

comit/src/network/orderbook/order.rs
```diff
@@ -55,33 +84,94 @@ impl NewOrder {
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, <D as Deserializer<'de>>::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let precision = 10;
+        let string = String::deserialize(deserializer)?;
+        let v: Vec<&str> = string.as_str().split('.').collect();
+
+        let integer = *v.first().unwrap();
+        let decimals = *v.last().unwrap();
```
thomaseizinger, line 132: As noted in the other one, this can easily be triggered by user code and should be handled properly. You can instantiate a new error through `<D as Deserializer<'de>>::Error::custom` with a new error message.

comit/src/network/orderbook/take_order.rs
```diff
@@ -8,16 +8,23 @@ use serde::{Deserialize, Serialize};
...
     }
 }
 
 #[derive(Clone, Copy, Debug, Default)]
-pub struct TakeOrderCodec;
+pub struct TakeBtcDaiOrderCodec;
+
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+pub struct TakeBtcDaiOrderRequest {
+    pub(crate) order_id: OrderId,
```
thomaseizinger, line 24: Any particular reason for making these `pub(crate)`?

cnd/src/network.rs
```diff
@@ -582,26 +600,39 @@ impl libp2p::swarm::NetworkBehaviourEventProcess<orderbook::BehaviourOutEvent> f
...
                 let order_id = order.id;
 
+                let hbit = hbit::CreatedSwap {
+                    amount: partial_take_quantity,
+                    final_identity: final_identity.into(),
+                    network: order.bitcoin_ledger,
+                    absolute_expiry: order.bitcoin_absolute_expiry,
+                };
+
+                let ethereum_quantity = match order.ethereum_quantity(partial_take_quantity) {
```
thomaseizinger, line 610: As a temporary measure, `token_quantity` might be a good name. We are only doing BTC-DAI but much of the code is still generic of the given ERC20 contract. "Ethereum quantity" is a bit too generic I think. What does that even refer to? :)

rishflab:
Closing this because we are handling it as a part of broader changes being made in project tantalus

rishflab:
Closing this because we are handling it as a part of broader changes being made in project tantalus

PR #2994, opened 2020-07-24
Nectar new orderbook

D4nte:
Small changes to allow comit lib integration in nectar.

### DIFF
--- a/comit/src/network/protocols/orderbook.rs
+++ b/comit/src/network/protocols/orderbook.rs
@@ -1,5 +1,5 @@
 mod order;
-mod take_order;
+pub mod take_order;
 
 use crate::{
     network::protocols::orderbook::take_order::{Response, TakeOrderCodec, TakeOrderProtocol},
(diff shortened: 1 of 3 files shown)


comit/src/network/protocols/orderbook.rs
```diff
@@ -1,5 +1,5 @@
 mod order;
-mod take_order;
+pub mod take_order;
```
rishflab, line 2: Why do you guys need access to take order? Afaik the functionality is exposed through the orderbook module?

thomaseizinger, line 2: Making this public exposes the internal structure of the orderbook which makes it harder to maintain a stable API of this module.

Re-exporting whatever you need from `take_order` should do the trick as well.

luckysori: So the reason why I made the module public was because there's a struct `take_order::Response` which I wanted to use just like that. Re-exporting just `Response` directly from `orderbook` seems wrong. Would it be okay to re-export as `TakeOrderResponse`? Is there a better alternative?

thomaseizinger: > Is there a better alternative?

Can't think of anything :)

PR #2995, opened 2020-07-24
Use libp2p-tokio-socks5 crate for transport

tcharding:
We just published the Tor transport to `crates.io`.

https://crates.io/crates/libp2p-tokio-socks5

We can now depend on this crate instead of having the transport code in our codebase.


tcharding:
> noice!

Thanks man

cnd/src/network/transport.rs
```diff
@@ -66,7 +66,7 @@ fn build_tor_transport(
     let mut map = HashMap::new();
     map.insert(addr, PORT);
 
-    let transport = TorTokioTcpConfig::new().nodelay(true).onion_map(map);
+    let transport = Socks5TokioTcpConfig::default().nodelay(true).onion_map(map);
```
thomaseizinger, line 69: Hmm, previously this was called `TorTokioTcpConfig` so it made sense for it to expose a function called `onion_map`.

If this is now a general-purpose `Socks5TokioTcpConfig`, what does that mean for the `onion_map` configuration?

Are we abstracting too early here?

tcharding: The name was the result of the thread on libp2p. There is a problem with the crate outlined in the README that while it is generic to any SOCKS5 proxy it currently only works with tor because of the onion address handling. Thanks for noticing this combinator, I did specifically think about the name of it although I think its ok at this stage since it makes explicit that there is _some_ tor specific stuff here. There has to be proxy specific address handling and that _could_ be implemented with the help of different combinators so this may stay for ever.

thomaseizinger: Interesting, I did read the libp2p thread after making this comment as it popped up in my notifications :D

I am curious of how this could be solved but for now it doesn't really harm us. The worst that can happen is some user being confused of why this think is named socks5 but actually only works with tor-socks5 :)

PR #2996, opened 2020-07-24
One chat to bring them all

D4nte:
and in the blockchain swap them.


D4nte:
Actually, best to spend 5min to create a badge using https://github.com/vector-im/logos/tree/master/matrix and https://shields.io/category/chat

PR #3006, opened 2020-07-27
Add a DAI type alias

tcharding:
We are heavily invested in code that manipulates DAI tokens, saying `Erc20Quantity` is cumbersome.

Add a DAI type alias to enable writing `asset::Dai`.

This PR adds code that is not used. I've written this code 10 times in the last two weeks and its still not in, just add it already.

### DIFF
--- a/comit/src/asset.rs
+++ b/comit/src/asset.rs
@@ -3,5 +3,5 @@ pub mod ethereum;
 
 pub use self::{
     bitcoin::Bitcoin,
-    ethereum::{Erc20, Erc20Quantity, Ether},
+    ethereum::{Dai, Erc20, Erc20Quantity, Ether},
 };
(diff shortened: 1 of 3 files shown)


tcharding:
> I am afraid this is wrong. An Erc20Quantity is not DAI. In order to express DAI, we would need to hardcode the address of the token contract on the respective Ethereum network (mainnet is different from testnet).
> 
> However, it is unfortunately not as simple as adding a type alias. I did put some thought into this [here](https://github.com/coblox/pitches/blob/cnd-trading-api/cnd_trading_api.adoc#24-whitelisting-erc20-tokens) though, happy for you to join the discussion :)

I had the hardcoded address in the next PR to come, I did not think of the different address per network though :)

I'll close this one and leave it to whoever works on that project you link. Thanks.

PR #3007, opened 2020-07-27
Add a dream API call for creating orders

tcharding:
This is purely fictional but this is how I'd like to call to cnd to create a limit order. I assume I will later call cnd again supplying identities to use when I confirm the take order request that would come in if another node tried to take my order.

(Initial patch is for [closed PR](https://github.com/coblox/pitches/blob/cnd-trading-api/cnd_trading_api.adoc#24-whitelisting-erc20-tokens) adding DAI type alias, see the PR for explanation of why this is incomplete :)


tcharding:
Closing. This was an exploratory PR done during cooldown, I see no value in leaving it open.

PR #3010, opened 2020-07-28
Separate take_order NB out of orderbook

tcharding:
10% review please. This PR does not patch cnd, will do after review if this idea has legs (slang explanation: means if this idea is likely to go forward)

Currently the take_order NB is a part of the orderbook. On a centralized exchange it makes sense to think of taking an order as part of the orderbook because the exchange does the order matching and the order execution. I am of the opinion that for a decentralized exchange it makes less sense because we do not control order matching or order execution. By control I mean that the user must be involved in the order execution and will likely be involved in the matching (e.g. matching multiple orders or partial orders) . It is therefore both conceivable and likely that the user will want to make some decision before doing a take order request (and accepting take request) and also before initiating execution.

For the above reasons (happy to flesh out the reasoning more if required) it is my opinion that the take_order NB should _not_ be a part of the orderbook. The orderbook should be a thin layer around gossipsub. This has the following benefits
- Makes the orderbook module super simple
- Keeps policy out of the orderbook, see below for comments on 'policy'
- Contains the orderbooks concern to just the gossiping, this may seen like an brain dead observation but because of the underlying complexity of libp2p and gossipsub I believe this is very real benefit
- reduces the number of layers that we have to call down through and bubble events up, users of the NBs have to handle the events anyways so this adds no extra complexity to the user code

## On Policy

We currently have two users of the comit library, cnd and Nectar. Both have different requirements although both are coupled together because they currently will take the two sides of all trades. It is likely, that eventually devs external to CoBloX will write their own software that uses comit lib to make/take orders. Even before this different folks may wish to hack different strategies into Nectar when running it.

For the above reasons I am of the belief that we should keep policy _out_ of the comit library. By policy I mean things that the user may or may not want to do based on their trading strategy.

### Concrete example: 

When receiving a take_order request we could conceivable bubble all and any such request to the user. Alternatively we could only bubble up take order requests for an order that was created by the node receiving the request. These are both totally valid. The second option introduces policy into the comit library because users of the library can no longer get _all_ take order requests.


thomaseizinger:
My primary argument for including it is that I think the _implementation_ of everything related to the orderbook should be encapsulated to form a stable API for users.

Whether we use gossip-sub and a single request-response protocol today or something completely handrolled tomorrow should not matter to the API that we are exposing.

Having two modules / structs here is 100% an artifact of how rust-libp2p is designed by means of using `derive(NetworkBehaviour)` and composing those things together.
If it were designed differently, then I don't see a reason why it should be two different modules.

Regarding policy:

I am not so sure if we can or should apply this to our software (yet). If there is anything we've learned in the past two years then it is that no one will come around and extend our software. Hence, leaving all kinds of possibilities open in how our software can be used just harms our maintainability because it prevents us from putting our finger down on a functionality and say: "This is how it should behave".

Therefore my stance would be to make as many things as possible an implementation detail. If someone comes around and says "I'd like this to work differently", then we can talk about opening things up and buy into that additional complexity.

tcharding:
> My primary argument for including it is that I think the _implementation_ of everything related to the orderbook should be encapsulated to form a stable API for users.

Spliting orderbook NB and take_order NB apart need not change the API because the API is the surface between orderbook and user - this boundry is still the same its just within the API implementation that we would now have two NBs. 

> Whether we use gossip-sub and a single request-response protocol today or something completely handrolled tomorrow should not matter to the API that we are exposing.

As above.

> Having two modules / structs here is 100% an artifact of how rust-libp2p is designed by means of using `derive(NetworkBehaviour)` and composing those things together.
> If it were designed differently, then I don't see a reason why it should be two different modules.
>
> Regarding policy:
> 
> I am not so sure if we can or should apply this to our software (yet). If there is anything we've learned in the past two years then it is that no one will come around and extend our software. Hence, leaving all kinds of possibilities open in how our software can be used just harms our maintainability because it prevents us from putting our finger down on a functionality and say: "This is how it should behave".

Oh, so the comit library should totally be considered to have just two users - cnd and Nectar. I miss-understood this. I reverse my views of policy then, actually comit should totally enforce any policy that we want to ensure between the two users.

> Therefore my stance would be to make as many things as possible an implementation detail. If someone comes around and says "I'd like this to work differently", then we can talk about opening things up and buy into that additional complexity.

Thanks for the response

thomaseizinger:
> Spliting orderbook NB and take_order NB apart need not change the API it just changes which field the user calls the API on. E.g. self.orderbook.confirm() vs self.take_order.confirm() - this does not seem like a 'breaking API change' to me.

Removing a function (i.e. `confirm`) from a component is a breaking change.

@tcharding do you mind responding to this point:

> Whether we use gossip-sub and a single request-response protocol today or something completely handrolled tomorrow should not matter to the API that we are exposing.

If we change the implementation of the orderbook to something that doesn't use gossip-sub and but has a message built in for taking orders, how are you planning on realizing that if it is split up?

Not that we are going to do this but it is an interesting thought-experiment to validate that it is a "good" idea to combine them: Splitting them leaks that the current implementation uses two different libp2p concepts although it only fulfills a single job: agreeing on an a trade with someone else.

> Thanks for the response, I still think the take_order NB should be separate to the orderbook for all the other benefits apart from policy i.e., it makes the orderbook a thin layer.

I would actually use that as a counter argument. Anything that is just a thin layer should be looked at with very critical eyes. What value does it add if it is just a thin layer around something else? Why not use that something else directly?

tcharding:
> > Spliting orderbook NB and take_order NB apart need not change the API it just changes which field the user calls the API on. E.g. self.orderbook.confirm() vs self.take_order.confirm() - this does not seem like a 'breaking API change' to me.
> 
> Removing a function (i.e. `confirm`) from a component is a breaking change.
> 
> @tcharding do you mind responding to this point:
> 
> > Whether we use gossip-sub and a single request-response protocol today or something completely handrolled tomorrow should not matter to the API that we are exposing.

Oh but it does, the orderbook is not a single component but rather it is a network behaviour, as such its interface is intertwined with libp2p - we cannot escape that. My point is that whether it is one NB or two NBs makes no difference to usage of it apart from _how_ is is plugged in to libp2p by its users.  

> If we change the implementation of the orderbook to something that doesn't use gossip-sub and but has a message built in for taking orders, how are you planning on realizing that if it is split up?

It would be different if the orderbook where a module with an API that we fully control but we don't because of libp2p.

> Not that we are going to do this but it is an interesting thought-experiment to validate that it is a "good" idea to combine them: Splitting them leaks that the current implementation uses two different libp2p concepts although it only fulfills a single job: agreeing on an a trade with someone else.

Having a single 'component' leaks that we use a libp2p NB - this is a mute argument IMO.
 
> > Thanks for the response, I still think the take_order NB should be separate to the orderbook for all the other benefits apart from policy i.e., it makes the orderbook a thin layer.
> 
> I would actually use that as a counter argument. Anything that is just a thin layer should be looked at with very critical eyes. What value does it add if it is just a thin layer around something else? Why not use that something else directly?

Interesting argument. The gossipsub stuff is thick enough to warrant a module/NB because there is _some_ logic (e.g. messages, topics, in memory store of received and/or published orders). I'm referring to all this as the 'thin' layer. It is easier to reason about this code if it is separate and it reduces the layers of abstraction that we have to call down through and bubble events up out of making the code easier to write and cheaper to maintain.

thomaseizinger:
> It would be different if the orderbook where a module with an API that we fully control but we don't because of libp2p.

We don't fully control the whole API but we **do** fully control the API that is relevant to users:

- `confirm`
- `publish`
- `deny`

> Oh but it does, the orderbook is not a single component but rather it is a network behaviour, as such its interface is intertwined with libp2p - we cannot escape that. My point is that whether it is one NB or two NBs makes no difference to usage of it apart from _how_ is is plugged in to libp2p by its users.

It is a single component from a user PoV. Users (like nectar) [never](https://github.com/coblox/nectar/blob/97d8900662c819037e4f9ba9799d0b325b6111b8/src/network.rs#L168-L188) interact with the `libp2p` part of the component, they only interact with our API.

What I am arguing for is that there is value in hiding the details of how this API is realized. By splitting the API up into two components, we expose that there are two different components in the works for fulfilling this API. This completely removes our ability to ever do something different here.

This is the same reasoning for introducing the `Storage` abstraction in `cnd`. We wanted to hide the details behind what information comes from the database and what comes from the root seed file. Callers of the `Storage` module shouldn't need to worry about this.

> It is easier to reason about this code if it is separate and it reduces the layers of abstraction that we have to call down through and bubble events up out of making the code easier to write and cheaper to maintain.

Would the following structure also fulfill your requirement of making it easier to reason about?

```
- Orderbook
	- TakeOrder
		- RequestResponse
	- Orders
		- GossipSub
```

tcharding:
> Would the following structure also fulfill your requirement of making it easier to reason about?
> 
> ```
> - Orderbook
> 	- TakeOrder
> 		- RequestResponse
> 	- Orders
> 		- GossipSub
> ```

Yes! I like that, thanks for your patience so we got here. And now the new 'get_orders' (or whatever name it ends up being) would be on the same level as `TakeOrder` and `Orders` - I like it. (We are overloading the term `Orders` with the introduction of the `Orders` ADT, just a side note.)

thomaseizinger:
> > Would the following structure also fulfill your requirement of making it easier to reason about?
> > ```
> > - Orderbook
> > 	- TakeOrder
> > 		- RequestResponse
> > 	- Orders
> > 		- GossipSub
> > ```
> 
> Yes! I like that, thanks for your patience so we got here. And now the new 'get_orders' (or whatever name it ends up being) would be on the same level as `TakeOrder` and `Orders` - I like it. (We are overloading the term `Orders` with the introduction of the `Orders` ADT, just a side note.)

Alright, nix we fleshed this one out. Given that we are overhauling things anyway, I think we can close this one though. Thanks for your efforts!

PR #3012, opened 2020-07-28
Change the cache key

tcharding:
We are getting strange CI errors. The only know way to clear the cache
is by changing the cache key.


tcharding:
> We are using these caches in the release job as well I think.

Too quick for me, I added them shortly after raising the PR. Thanks

PR #3013, opened 2020-07-28
Make bors listen on all CI jobs

thomaseizinger:
This prevents weird timeout errors in bors in case the macos job fails and therefore cancels the ubuntu build.

It does mean that we are now requiring all of those to pass before we can merge but I reckon that is ok. If we don't actually care about these jobs, we should remove them altogether.


thomaseizinger:
bors cancel

PR #3018, opened 2020-07-29
debug - do not merge

tcharding:
This is a test PR to try and prove the cause of the possible race condition caused by subscribing on startup. Race condition solutions are hard to 'prove' so leaving open without merging for now.

ref: https://github.com/libp2p/rust-libp2p/issues/1671

### DIFF
--- a/comit/src/network/orderbook.rs
+++ b/comit/src/network/orderbook.rs
@@ -278,7 +278,7 @@ pub enum BehaviourOutEvent {
 
 impl NetworkBehaviourEventProcess<GossipsubEvent> for Orderbook {
     fn inject_event(&mut self, event: GossipsubEvent) {
-        if let GossipsubEvent::Message(_peer_id, _message_id, message) = event {
+        if let GossipsubEvent::Message(_peer_id, message_id, message) = event {
             let decoded: Message = match bincode::deserialize(&message.data[..]) {
                 Ok(msg) => msg,
                 Err(e) => {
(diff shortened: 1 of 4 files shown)


comit/src/network/orderbook.rs
```diff
@@ -80,6 +80,8 @@ impl Orderbook {
             events: VecDeque::new(),
         };
 
+        std::thread::sleep(std::time::Duration::from_secs(2));
```
da-kami, line 83: @rishflab just pointed me to this debug-PR.

when trying to connect taker-ui (behind a cnd node) and nectar I ran into the problem that Nectar never properly registers the taker once we dial.

In a setup with 2 cnd nodes (one for the taker and one for the maker) when the taker dials the maker, we see the registration of the taker on the maker's side:

```
Jul 30 14:41:20.065  INFO libp2p_gossipsub::behaviour: New peer connected: PeerId("12D3KooWJdBwGAohCJbaZDFurkFkVkLuXCQmSa7wJ7cVMgujLJRR")    
Jul 30 14:41:20.130  INFO libp2p_gossipsub::behaviour: SUBSCRIPTION: Adding peer: PeerId("12D3KooWJdBwGAohCJbaZDFurkFkVkLuXCQmSa7wJ7cVMgujLJRR") to topic: TopicHash { hash: "BTC/DAI" } 
```

However, when using Nectar we never see this - the subscription is somehow not done.
The incoming connection is noticed in Nectar (when actively monitoring `libp2p::swarm::SwarmEvent::IncomingConnection` events), but we subscription is never done,

Thought I report this here.

tcharding: Thanks for the report @da-kami.

tcharding:
We are not testing gossipsub now, we are implementing get_orders directly from the node.  Closing this for a later date when we start testing the gossipsub stuff we have remaining.

PR #3020, opened 2020-07-29
Upgrade to latest comit-sdk

D4nte:
To hopefully fix the lightning CI error `OpenStatusUpdate is malformed`.

Also fixed the `cannot find serde_derive crate` issue.

See https://github.com/comit-network/comit-js-sdk/pull/255

### DIFF
--- a/api_tests/package.json
+++ b/api_tests/package.json
@@ -32,7 +32,7 @@
         "bitcoin-core": "^3.0.0",
         "bitcoinjs-lib": "^5.1.10",
         "chmod": "^0.2.1",
-        "comit-sdk": "^0.17.0",
+        "comit-sdk": "^0.17.1",
         "download": "^8.0.0",
         "ethers": "^5.0.7",
         "find-cache-dir": "^3.3.1",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
         with:
           path: ~/.cargo/registry
           # The registry cache is useful as long as we need the same dependencies as another job, regardless of the Rust version and operating system.
-          key: cargo-registry-${{ hashFiles('Cargo.lock') }}-v2
+          key: cargo-registry-${{ hashFiles('Cargo.lock') }}-v3
 
       - name: Cache cargo binaries
         uses: actions/cache@v2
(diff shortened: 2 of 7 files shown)


D4nte:
@tcharding FYI

thomaseizinger:
> Edit: Ok this strategy cannot work as comit-sdk is a TS package. Will need to release to test in the CI. let's do that tomorrow.

To reduce the release churn, @da-kami just imported the whole SDK into the app and modifies things as he needs. Should we do the same thing for the e2e tests and call the SDK a bad idea (for now)?

1. Some APIs in the SDK are horribly outdated and no longer work.
2. The only thing that is really common between the app and the e2e tests is actually _Siren_ specific, like converting an action into an HTTP request. We could extract a siren-toolkit out of that if we fancy it to be DRY.
3. We've now several times run into situations where the e2e tests need a lot more control over what is happening within the e2e tests.
4. This PR is another situation where the code in the SDK being separate is hurting us more than it helps.

My vote goes down for inlining the SDK into the e2e tests (and optionally archiving the SDK repository for now).

api_tests/package.json
```diff
@@ -32,7 +32,7 @@
         "bitcoin-core": "^3.0.0",
         "bitcoinjs-lib": "^5.1.10",
         "chmod": "^0.2.1",
-        "comit-sdk": "^0.17.0",
+        "comit-sdk": "comit-network/comit-js-sdk#0df2d5a0",
```
thomaseizinger, line 35: You could use https://github.com/whitecolor/yalc if you wanted to.

D4nte: Would you use that in the comit-rs CI?

thomaseizinger:
@D4nte seems like your fix doesn't work all the time :(

D4nte:
> @D4nte seems like your fix doesn't work all the time :(

Well, the serde_derive fix seems to work at least :D

thomaseizinger:
> > @D4nte seems like your fix doesn't work all the time :(
> 
> Well, the serde_derive fix seems to work at least :D

That is nice! What do you think of cherrypicking that into its own PR so we can at least get that out of the way before fixing the lnd issue?

D4nte:
> @D4nte seems like your fix doesn't work all the time :(

As per the extra `console.log`. The data sent back from lnd would not lead to an error thrown. I believe it is a different issue and that my fix does stop the lnd error.

D4nte:
bors retry

D4nte:
bors retry

D4nte:
Last time I retry, next failure  I'll get the serde_derive patch in separately

PR #3038, opened 2020-08-05
Add Recommended Safe Action module

tcharding:
Add a skeleton module that defines the API to the new Recommended Safe Action (RSA) logic. Use the API in cnd.

Note: I am introducing in this PR a new acronym: RSA, this was chosen to be uniform with the existing acronym RUA (Recommended Unique Action).

### DIFF
--- a/comit/src/rsa.rs
+++ b/comit/src/rsa.rs
@@ -1,0 +1,129 @@
+//! Recommended Safe Actions
+//!
+//! This module provides logic for recommending when it is safe/unsafe to take a
+//! particular action for the COMIT protocol.
+
+//! 'Safe' is defined as:
+//! 1 - You will not loose money.
+//! 2 - The counterparty is likely to continue with the swap because it is safe
+//!     for them to do so.
+
+use crate::Timestamp;
+use time::Duration;
+
+/// SwapTime enables recommendations on whether it is safe to take an action.
+#[derive(Debug, Clone, Copy)]
+pub struct SwapTime {
+    alpha_ledger: Ledger,
+    beta_ledger: Ledger,
+    alpha_expiry: Timestamp,
+    beta_expiry: Timestamp,
+}
+
+impl SwapTime {
+    /// Construct a new SwapTime from the given expiries.
+    pub fn new(
+        alpha_ledger: Ledger,
+        beta_ledger: Ledger,
+        alpha_expiry: Timestamp,
+        beta_expiry: Timestamp,
+    ) -> Self {
+        SwapTime {
+            alpha_ledger,
+            beta_ledger,
+            alpha_expiry,
+            beta_expiry,
+        }
+    }
+
+    /// Calculates a pair of valid expiries using the given parameters.
+    /// The created SwapTime will return true from calls to
+    /// is_time_for_swap_to_complete() for the duration of `good_for`.
+    pub fn with_calculated_expiries(
+        _alpha_ledger: Ledger,
+        _beta_ledger: Ledger,
+        _good_for: Duration,
+    ) -> Result<Self, InsufficientTime> {
+        unimplemented!()
+    }
+
+    /// Helper method so we do not need to know which action is first (since
+    /// that is ledger specific).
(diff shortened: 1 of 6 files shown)


comit/src/rsa.rs
```diff
@@ -0,0 +1,115 @@
...
+        _alpha_expiry: Timestamp,
+        _beta_expry: Timestamp,
+    ) -> Self {
+        unimplemented!()
+    }
+
+    /// Calculates a pair of valid expiries using the given parameters.
+    /// The created SwapTime will return true from calls to
+    /// is_time_for_swap_to_complete() for the duration of `good_for`.
+    pub fn with_calculated_expiries(
```
thomaseizinger, line 39: From the name I assume this should take `self` by value.

comit/src/rsa.rs
```diff
@@ -0,0 +1,115 @@
...
+//! This module provides logic for recommending when it is safe/unsafe to take a
+//! particular action for the COMIT protocol.
+//!
+//! 'Safe' is defined as:
+//! 1 - You will not loose money.
+//! 2 - The counterparty is likely to continue with the swap because it is safe
+//!     for them to do so.
+
+use crate::Timestamp;
+use time::Duration;
```
thomaseizinger, line 12: Why is `std::time::Duration` not good enough?

comit/src/rsa.rs
```diff
@@ -0,0 +1,115 @@
...
+        unimplemented!()
+    }
+}
+
+/// Supported ledgers.
+#[derive(Debug, Clone, Copy)]
+pub enum Ledger {
+    Bitcoin,
+    Ethereum,
+}
```
thomaseizinger, line 111: Why do we need a new enum for this?

comit/src/rsa.rs
```diff
@@ -0,0 +1,115 @@
...
+//! 1 - You will not loose money.
+//! 2 - The counterparty is likely to continue with the swap because it is safe
+//!     for them to do so.
+
+use crate::Timestamp;
+use time::Duration;
+
+/// SwapTime enables recommendations on whether it is safe to take an action.
+#[derive(Debug, Clone, Copy)]
+pub struct SwapTime {
```
thomaseizinger, line 16: What is the reasoning behind the name? I don't find it particularly descriptive for the functionality it exposes :see_no_evil:

thomaseizinger:
> Looks good!
> 
> I would like to avoid the RSA acronym because it means something very different in the crypto space.
> We don't have RUA anywhere as an acronym so I don't see a reason for "consistency" :)

Also, we should be aware that only recommending safe actions should really be the default. For someone not familiar with the history of the code, seeing a module that is dedicated to recommending "safe" actions would presumably be really confusing as it implies that there is also a non-safe way of doing things.

I think our history shouldn't drive our designs too much :)

tcharding:
> Looks good!
> 
> I would like to avoid the RSA acronym because it means something very different in the crypto space.
> We don't have RUA anywhere as an acronym so I don't see a reason for "consistency" :)

No worries, will stick with what is in the doc 'Safely Recommend Actions' (SRA)

comit/src/rsa.rs
tcharding: Good point, thanks.

comit/src/rsa.rs
tcharding: This type will be hold time on a minute granularity, `std::time::Duration` seems designed for smaller periods, specifically the docs say:
"
A Duration type to represent a span of time, typically used for system timeouts.
"

This will be blocktimes, and 'time to act' (possibly even measured in hours).

comit/src/rsa.rs
tcharding: Fair point, thanks.

comit/src/rsa.rs
tcharding: Before this I had `Foo` :) I'm always open to suggestions for better names.

tcharding:
Burying this one in favour of: https://github.com/comit-network/comit-rs/pull/3055

PR #3046, opened 2020-08-07
SRA add document

tcharding:
## Safely Recommend Action

This PR adds a document that describes the problem space and solution around expiry times and safe actions. This work has been dubbed 'Safely Recommend action' or SRA.

Please note, we add `docs/` directory to the root directory of `comit-rs` to hold the document because this document will hopefully be the source of truth for understanding the issue, now and in times to come. Although, if we make the correct decisions, users should not need to know this - as soon as anyone wishes to tune Nectar or the App they will likely want to understand this topic.

### Request for Review

The aim of this document is to be concise, thorough, and correct. Please review both the correctness and the writing style. I'm here to learn.

### DIFF
--- a/docs/actions.adoc
+++ b/docs/actions.adoc
@@ -1,0 +1,173 @@
+= COMIT Protocol Actions
+Tobin C. Harding <tobin@coblox.tech>;
+:toc:
+:revdate: 2020-08-07
++ Created on: {revdate} +
+
+== Introduction
+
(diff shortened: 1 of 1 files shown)


D4nte:
I recommend to move it to `comit-network/spikes`.

tcharding:
Moved to spikes: https://github.com/comit-network/spikes/pull/64

PR #3055, opened 2020-08-12
Standardized COMIT Expiry Times

tcharding:
Add `expiries` module.

Ready for merge but not fully functional, reasoning is that this is a library module and the library module _is_ the project deliverable. I will of course attempt to use it after merge and before the project ends. I'd like to clearly separate the two phases described below, merge is how I propose to do that.

This PR completes the first phase, of primary concern are the two `transition_period` methods. The individual transition functions (TTI, TTL) are intended to be opaque (currently `todo!()`).  

The second phase is implementing all the period functions (represented in the spike by TTL, TTI, TTF etc.).

## Notes: 

We would like to be able to manipulate timestamps and durations together. We use `time::Duration` because it is more suited to durations of the timeframes we expect i.e., hour long durations (`std::time` is better suited, according to the docs, to system times i.e., smaller timeframes). We modify our `Timestamp` module to handle `Duration`s.

### DIFF
--- a/comit/src/expiries.rs
+++ b/comit/src/expiries.rs
@@ -1,0 +1,641 @@
+//! Standardized COMIT Expiry Times
+//!
+//! The COMIT protocols rely on two expiry times in order for the swap to be
+//! atomic. One expiry for the alpha ledger and one for the beta ledger. The
+//! beta ledger expiry time must elapse before the alpha ledger expiry time.
+
+use crate::timestamp::{self, Timestamp};
+use time::{prelude::*, Duration};
+
+// TODO: Currently we ignore deploy for Erc20.
+
+// Note on expiry types: Any `Duration` expiry is a relative expiry, any
+// `Timestamp` expiry is an absolute expiry. We can use relative expiries to
+// create absolute expiries when, for example, swap execution starts.
+
+/// Current time as a UNIX timestamp from the perspective of the implementer.
+///
+/// Intended for getting the current time from the underlying blockchain.
+/// The definition `current time` varies depending on the blockchain, this
+/// always refers to the time used by the op codes used in the COMIT contracts.
+pub trait CurrentTime {
+    fn current_time(&self) -> Timestamp; // TODO: This will need to be async.
+}
+
+/// Data that defines and manipulates expiry times.
+///
+/// Language note: We use the term 'expiries' (plural of expiry) to make it
+/// explicit that there are two expiry times however we refer it as singular
+/// since it is the struct, or pair of expiries, that we are referring to.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Expiries<A, B> {
+    /// Configuration values for calculating transition periods.
+    config: Config,
+    /// Alpha blockchain connector for getting the current time.
+    alpha_connector: A,
+    /// Beta blockchain connector for getting the current time.
+    beta_connector: B,
+    /// The alpha ledger relative expiry time.
+    alpha_expiry: Duration,
+    /// The beta ledger relative expiry time.
+    beta_expiry: Duration,
+}
+
+impl<A, B> Expiries<A, B>
+where
+    A: CurrentTime,
+    B: CurrentTime,
+{
+    /// Construct a pair of relative expiries values.
+    pub fn new(protocol: Protocol, alpha: A, beta: B) -> Expiries<A, B> {
+        let config = protocol.config();
+
+        let alice_needs = happy_path_swap_period_for_alice(&config);
+        let bob_needs = happy_path_swap_period_for_bob(&config);
+
+        Expiries {
+            config,
+            alpha_connector: alpha,
+            beta_connector: beta,
+            // Bob redeems on alpha ledger so needs time to act before the alpha expiry.
+            alpha_expiry: bob_needs,
+            // Alice redeems on beta ledger so needs time to act before the beta expiry.
+            beta_expiry: alice_needs,
+        }
+    }
+
+    /// Returns true if both:
+    /// 1. Alice has time to transition from initial state to `Done`.
+    /// 2. Bob has time to transition from initial state to `Done`.
+    pub fn is_useful(&self) -> bool {
+        let alice_needs = happy_path_swap_period_for_alice(&self.config);
+        let bob_needs = happy_path_swap_period_for_bob(&self.config);
+
+        let alice_has_enough_time = self.beta_expiry >= alice_needs;
+        let bob_has_enough_time = self.alpha_expiry >= bob_needs;
+
+        alice_has_enough_time && bob_has_enough_time
+    }
+
+    /// Convert relative expiries to absolute expiries.
+    pub fn to_absolute(&self) -> (Timestamp, Timestamp) {
+        let alpha_now = self.alpha_connector.current_time();
+        let alpha_expiry = alpha_now.add_duration(self.alpha_expiry);
+
+        let beta_now = self.beta_connector.current_time();
+        let beta_expiry = beta_now.add_duration(self.beta_expiry);
+
+        (alpha_expiry, beta_expiry)
+    }
+
+    /// True if Alice has time to complete a swap (i.e. transition to done)
+    /// before the expiry time elapses.
+    fn alice_can_complete_from(&self, current_state: AliceState, beta_expiry: Timestamp) -> bool {
+        let period = period_for_alice_to_complete(&self.config, current_state);
+        let now = self.beta_connector.current_time();
+
+        // Alice redeems on beta ledger so is concerned about the beta expiry.
+        let end_time = now.add_duration(period);
+        end_time < beta_expiry
+    }
+
+    /// True if Bob has time to complete a swap (i.e transition to done)
+    /// before the expiry time elapses.
+    fn bob_can_complete_from(&self, current_state: BobState, alpha_expiry: Timestamp) -> bool {
+        let period = period_for_bob_to_complete(&self.config, current_state);
+        let now = self.alpha_connector.current_time();
+
+        // Bob redeems on alpha ledger so is concerned about the alpha expiry.
+        let end_time = now.add_duration(period);
+        end_time < alpha_expiry
+    }
+
+    /// If Alice's next action is not taken within X minutes the expiries will
+    /// become un-useful. Returns X.
+    pub fn alice_should_act_within(
+        &self,
+        current_state: AliceState,
+        beta_expiry: Timestamp,
+    ) -> Duration {
+        let period = period_for_alice_to_complete(&self.config, current_state);
+        let start_time = beta_expiry.sub_duration(period);
+        let now = self.beta_connector.current_time();
+
+        timestamp::duration_between(now, start_time)
+    }
+
+    /// If Bob's next action is not taken within X minutes the expiries will
+    /// become un-useful. Returns X.
+    pub fn bob_should_act_within(
+        &self,
+        current_state: BobState,
+        alpha_expiry: Timestamp,
+    ) -> Duration {
+        let period = period_for_bob_to_complete(&self.config, current_state);
+        let start_time = alpha_expiry.sub_duration(period);
+        let now = self.alpha_connector.current_time();
+
+        timestamp::duration_between(now, start_time)
+    }
+
+    /// Returns the recommended next action that Alice should take.
+    pub fn next_action_for_alice(
+        &self,
+        current_state: AliceState,
+        alpha_expiry: Timestamp,
+        beta_expiry: Timestamp,
+    ) -> AliceAction {
+        if current_state == AliceState::Done {
+            return AliceAction::NoFurtherAction;
+        }
+
+        let funded = current_state.has_sent_fund_transaction();
+
+        if self.alpha_expiry_has_elapsed(alpha_expiry) {
+            if funded {
+                return AliceAction::Refund;
+            }
+            return AliceAction::Abort;
+        }
+
+        let both_parties_can_complete = self.alice_can_complete_from(current_state, beta_expiry)
+            && self.bob_can_complete_from(current_state.into(), alpha_expiry);
+
+        if !both_parties_can_complete {
+            if funded {
+                return AliceAction::WaitForRefund;
+            }
+            return AliceAction::Abort;
+        }
+
+        let (next_action, _state) = current_state.next();
+        next_action
+    }
+
+    /// Returns the recommended next action that Bob should take.
+    pub fn next_action_for_bob(
+        &self,
+        current_state: BobState,
+        alpha_expiry: Timestamp,
+        beta_expiry: Timestamp,
+    ) -> BobAction {
+        if current_state == BobState::Done {
+            return BobAction::NoFurtherAction;
+        }
+
+        // If Alice has redeemed Bob's only action is to redeem irrespective of expiry
+        // time.
+        if current_state == BobState::RedeemBetaTransactionSeen {
+            return BobAction::RedeemAlpha;
+        }
+
+        let funded = current_state.has_sent_fund_transaction();
+
+        if self.beta_expiry_has_elapsed(alpha_expiry) {
+            if funded {
+                return BobAction::Refund;
+            }
+            return BobAction::Abort;
+        };
+
+        let both_parties_can_complete = self
+            .alice_can_complete_from(current_state.into(), beta_expiry)
+            && self.bob_can_complete_from(current_state, alpha_expiry);
+
+        if !both_parties_can_complete {
+            if funded {
+                return BobAction::WaitForRefund;
+            }
+            return BobAction::Abort;
+        }
+
+        let (next_action, _state) = current_state.next();
+        next_action
+    }
+
+    fn alpha_expiry_has_elapsed(&self, expiry: Timestamp) -> bool {
+        let now = self.alpha_connector.current_time();
+        now > expiry
+    }
+
+    fn beta_expiry_has_elapsed(&self, expiry: Timestamp) -> bool {
+        let now = self.beta_connector.current_time();
+        now > expiry
+    }
+}
+
+/// Duration for a complete happy path swap for Alice.
+fn happy_path_swap_period_for_alice(config: &Config) -> Duration {
+    period_for_alice_to_complete(&config, AliceState::None)
+}
+
+/// Duration for a complete happy path swap for Bob.
+fn happy_path_swap_period_for_bob(config: &Config) -> Duration {
+    period_for_bob_to_complete(&config, BobState::Created)
+}
+
+/// The minimum time we should allow for Alice to transition from
+/// `current_state` to done.
+fn period_for_alice_to_complete(config: &Config, current_state: AliceState) -> Duration {
+    // Tail call recursion with an accumulator.
+    fn period_to_complete(config: &Config, state: AliceState, acc: Duration) -> Duration {
+        if state == AliceState::Done {
+            return acc;
+        }
+
+        let (_action, next_state) = state.next();
+        let transition_period = state.transition_period(config);
+
+        period_to_complete(config, next_state, acc + transition_period)
+    }
+
+    period_to_complete(config, current_state, Duration::zero())
+}
+
+/// The minimum time we should allow for Bob to transition from
+/// `current_state` to done.
+fn period_for_bob_to_complete(config: &Config, current_state: BobState) -> Duration {
+    // Tail call recursion with an accumulator.
+    fn period_to_complete(config: &Config, state: BobState, acc: Duration) -> Duration {
+        if state == BobState::Done {
+            return acc;
+        }
+
+        let (_action, next_state) = state.next();
+        let transition_period = state.transition_period(config);
+
+        period_to_complete(config, next_state, acc + transition_period)
+    }
+
+    period_to_complete(config, current_state, Duration::zero())
+}
+
+/// Happy path states for Alice.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum AliceState {
+    /// Initial state, swap not yet created.
+    None,
+    /// Swap has been created.
+    Created,
+    /// The fund alpha transaction has been sent to the network.
+    FundAlphaTransactionSent,
+    /// Implies fund alpha transaction has reached finality.
+    AlphaFunded,
+    /// Implies fund beta transaction has reached finality.
+    BetaFunded,
+    /// The redeem beta transaction has been sent to the network.
+    RedeemBetaTransactionSent,
+    /// Implies beta redeem transaction has reached finality.
+    Done,
+}
+
+/// Possible next action for Alice.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum AliceAction {
+    // Happy path actions for Alice.
+    Create,
+    FundAlpha,
+    WaitForFundTransactionFinality,
+    WaitForBetaFunded,
+    RedeemBeta,
+    WaitForRedeemBetaTransactionFinality,
+    NoFurtherAction, // Implies swap is done from Alice's perspective.
+    // Cancel path.
+    Abort,         // Implies HTLC not funded, abort but take no further action.
+    WaitForRefund, // Implies alpha ledger HTLC funded but expiry not yet elapsed.
+    Refund,        // Implies HTLC funded and expiry time elapsed.
+}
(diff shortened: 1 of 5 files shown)


comit/src/expiries.rs
```diff
@@ -0,0 +1,488 @@
...
+    alpha_expiry: Timestamp,
+    /// The beta ledger expiry time.
+    beta_expiry: Timestamp,
+}
+
+impl<C> Expiries<C> {
+    /// Create an expiries that will be useful for the duration of `ttl`.
+    pub fn create(ttl: Duration, protocol: Protocol, connector: C) -> Expiries<C>
+    where
+        C: CurrentTime,
```
thomaseizinger, line 40: You are repeating yourself all over the place here. Moving this bound to the impl block should fix that.

Type variables that are defined on the actual type will usually also need their bounds to be defined there (as in `impl<C> Expiries<C> where C: ...`.

You can always have multiple `impl` blocks if the bounds differ per function.

comit/src/expiries.rs
```diff
@@ -0,0 +1,488 @@
...
+        timestamp::duration_between(now, start_time)
+    }
+
+    /// Returns the recommended next action that Alice should take.
+    pub fn next_action_for_alice(&self, current_state: AliceState) -> AliceAction
+    where
+        C: CurrentTime,
+    {
+        if current_state == AliceState::Done {
+            return AliceAction::NoFurtherAction;
```
thomaseizinger, line 145: This variant could be replaced with the use of an `Option`. Not sure if it is better though :)

comit/src/expiries.rs
```diff
@@ -0,0 +1,488 @@
...
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum AliceState {
+    // FIXME: Do we need a pre-created state in order to give Alice one more TTA period
+    // before `created`?
+    Created,
+    FundAlphaTransactionSent,
+    AlphaFunded, // Implies alpha fund transaction confirmed configured number of times.
+    BetaFunded,  // Implies beta fund transaction confirmed configured number of times.
+    RedeemBetaTransactionSent,
+    Done, // Implies beta redeem transaction confirmed configured number of times.
```
thomaseizinger, line 260: I reckon these comments should go as rust-docs on the variants :)

comit/src/expiries.rs
```diff
@@ -0,0 +1,488 @@
...
+        let (next_action, _next_state) = self.next();
+        if next_action == NoFurtherAction {
+            return Duration::zero();
+        }
+
+        unimplemented!()
+    }
+
+    /// True if Alice has funded the alpha HTLC.
+    fn includes_fund(&self) -> bool {
```
thomaseizinger, line 307:
```suggestion
    fn has_funded_alpha(&self) -> bool {
```

comit/src/expiries.rs
```diff
@@ -0,0 +1,488 @@
...
+        match self {
+            AliceState::Created => false,
+            // Be conservative because whether or not to try and refund depends on this value.
+            AliceState::FundAlphaTransactionSent => true,
+            AliceState::AlphaFunded => true,
+            AliceState::BetaFunded => true,
+            AliceState::RedeemBetaTransactionSent => true,
+            AliceState::Done => true,
```
thomaseizinger, line 315:
```suggestion
            AliceState::FundAlphaTransactionSent | AliceState::AlphaFunded | AliceState::BetaFunded | AliceState::RedeemBetaTransactionSent | AliceState::Done=> true,
```

comit/src/expiries.rs
```diff
@@ -0,0 +1,488 @@
...
+}
+
+/// To<Role>State returns either our own state or our best guess at
+/// the counterparties state. Returns the latest possible state that
+/// we can guarantee assuming the counterparty has the same blockchain
+/// knowledge as we do.
+
+pub trait AsAliceState {
+    fn as_alice_state(&self) -> AliceState;
+}
```
thomaseizinger, line 404: `AsXYZ` usually returns a reference by convention. If you want that, you could also implement `AsRef` :)

Otherwise your doc comment says the right thing: It should be `ToXYZ`.

comit/src/expiries.rs
```diff
@@ -0,0 +1,488 @@
...
+use time::Duration;
+
+// TODO: Currently we ignore deploy for Erc20.
+
+/// Current time as a UNIX timestamp as viewed by each of the underlying
+/// blockchains.
+pub trait CurrentTime {
+    fn alpha_current_time(&self) -> Timestamp;
+    fn beta_current_time(&self) -> Timestamp;
+}
```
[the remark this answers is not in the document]
tcharding: Good observation, I was aware of the problem you mention but not the solution - thanks.

comit/src/expiries.rs
tcharding: Awesome, thanks.

comit/src/expiries.rs
tcharding: Yeah, I had that at first - I kinda like this solution, not married to it though.

comit/src/expiries.rs
tcharding: Can do.

comit/src/expiries.rs
tcharding: Nice.

comit/src/expiries.rs
tcharding: lol, so obvious now I see it on the screen :)

comit/src/expiries.rs
tcharding: I researched this last time we discussed it, the conclusion I came to was that `AsXYZ` is for 'cheap' things and 'ToXYZ' is for 'expensive' things. With cheap and expensive being defined the same as for `Copy` vs `Clone`. Whether or not it takes/returns a reference is explicit in the method. 

So in this case, the `AliceState` struct implements `Copy` so I elected for `AsAliceState`.

tcharding:
Thanks for your review @thomaseizinger, insightful comments.

comit/src/expiries.rs
```diff
@@ -0,0 +1,581 @@
...
+/// explicit that there are two expiry times however we refer it as singular
+/// since it is the struct, or pair of expiries, that we are referring to.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Expiries<C> {
+    /// Configuration values for calculating transition periods.
+    config: Config,
+    /// Alpha blockchain connector for getting the current time.
+    alpha_connector: C,
+    /// Beta blockchain connector for getting the current time.
+    beta_connector: C,
```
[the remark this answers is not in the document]
tcharding: Face palm

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+            alpha_expiry,
+            beta_expiry,
+        }
+    }
+
+    /// We define 'useful' to mean:
+    /// 1. Alice has time to transition from current state to `Done`.
+    /// 2. Bob has time to transition from current state to `Done`.
+    ///
+    /// An actor only knows there own state but can reason about the state of
```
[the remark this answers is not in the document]
tcharding: Yes, it should be `their`.  Thanks

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+    }
+
+    /// True if Bob has time to complete a swap (i.e transition to done).
+    fn bob_has_time_to_complete(&self, current_state: BobState) -> bool {
+        let period = period_for_bob_to_complete(&self.config, current_state);
+        let now = self.alpha_connector.current_time();
+
+        // Bob redeems on alpha ledger so is concerned about the alpha expiry.
+        let end_time = now.add_duration(period);
+        end_time < self.alpha_expiry
```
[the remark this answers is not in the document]
tcharding: Thanks for looking at this @da-kami.

>  Does this take `Refund` into account? (I assume yes, then it is correct.)

This calculates if Bob has time to complete the happy path only, does not take refund into account.

> I am a bit confused about what the state `Done` stands for, because as far as I saw it does not take `Refund` into account. I guess because if depicts the happy-path. If we cannot reach `Done` we either decide to not fund at all or we go for `Refund` if we already funded as Bob.

Correct, done is happy path complete i.e., 'swapped'.

> Out of the context, I wonder if above condition is correct:

I believe the condition is correct because it is determining whether Bob has time to complete all his actions hence it is against alpha expiry.

> Bob can only redeem till `beta_expiry` is reached! After that he has to go for refund. The time window between `beta_expiry` and `alpha_expiry` is his refund window. (Alice can refund after `alpha_expiry` is reached, she is protected by the secret, Bob is not.)

You are correct in that as soon as beta expiry is available Bob should take it, this logic is within `next_action_for_bob` in this module. The check on expiry is done _before_ we call the code mentioned above. Oh you inadvertently found a bug for me, I was checking the wrong expiry in both the `next_action_for_...` methods!  

> I think it should hence be `beta_expiry` here, if you only talk about `Redeeming` in `Done`.
> 
> Please correct me if I have a thinking mistake :)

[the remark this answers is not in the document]
tcharding: Man, on further thought you might be onto something here. Even if Bob has time to act before his alpha expiry the beta expiry overrides that. Will think more about this, thanks.

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+//! beta ledger expiry time must elapse before the alpha ledger expiry time.
+
+use crate::timestamp::{self, Timestamp};
+use time::{prelude::*, Duration};
+
+// TODO: Currently we ignore deploy for Erc20.
+
+/// Current time as a UNIX timestamp from the perspective of the implementer.
+///
+/// Intended for getting the current time from the underlying blockchain.
```
D4nte, line 14:
```suggestion
/// Intended for getting the current time from the underlying blockchain.
/// The definition `current time` varies depending on the blockchain, this always
/// refers to the time used by the op codes used in the COMIT contracts.
```
Maybe something to clarify here.

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+    /// The beta ledger expiry time.
+    beta_expiry: Timestamp,
+}
+
+impl<A, B> Expiries<A, B>
+where
+    A: CurrentTime,
+    B: CurrentTime,
+{
+    /// Create an expiries that will be useful for the duration of `ttl`.
```
D4nte, line 43: Please define `ttl` somewhere in the code. I think at the start of this file would be good.

D4nte: As discussed, the presence of a configurable `ttl` is dangerous.

When deciding the `time to act` for Alice and Bob. You mentioned that for Alice it's 1 hour and Bob 15 minutes.

You also mentioned that the absolute expiries are calculated and included in published order.

Now, if a `ttl` is configurable and set to 12 hours, then the time to act for Alice is actually `1 hour + 12 hours - age of order`. which could be 12 or 13 hours.

This invalidates the previous statement (Alice's time to act is 1 hour).

The issue stems from the fact that the absolute expiries are included in the published orders, which is a bad choice.

Instead, relative expiries should be included in the published orders, allowing the validity of an order to be orthogonal to the execution expiries.

The absolute expiries must be agreed and communicated when the `time to act` stopwatch starts for Alice. This is when `nectar` replies `ok` to an order being taken.
Then, Alice can confirm that the absolute expiries matches expectation and she then has the agreed `time to act` to fund.

I understand it means changing quite a bit of thing on the orderbook part but this is not part of this project :)

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+        let funded = current_state.has_funded_beta();
+
+        if self.beta_expiry_has_elapsed() {
+            if funded {
+                return BobAction::Refund;
+            }
+            return BobAction::Abort;
+        };
+
+        if !self.bob_has_time_to_complete(current_state) {
```
D4nte, line 175: 📣 This does not take in account if Alice has redeemed. If she has redeemed, Bob should redeem, not matter the current time.

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+        period_to_complete(config, next_state, acc + transition_period)
+    }
+
+    period_to_complete(config, current_state, Duration::zero())
+}
+
+/// Happy path states for Alice.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum AliceState {
+    // FIXME: Do we need a pre-created state in order to give Alice one more TTA period
```
D4nte, line 236: I believe our discussion answers this question.

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+    // before `created`?
+    /// Initial state, swap has been created.
```
D4nte, line 238: `swap is created` is not a terminology we commonly used before. Maybe `None` or `Init` or `Start` would a better definition.

Realistically, this state is when the communication is done and we are starting the stopwatch.

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+    Created,
+    /// The fund alpha transaction has been sent to the network.
+    FundAlphaTransactionSent,
+    /// Implies fund alpha transaction has reached finality.
+    AlphaFunded,
```
D4nte, line 243: `Reach finality` is not a terminology we used before or commonly used. I would use "confirmed" instead.
```suggestion
    /// Implies fund alpha transaction has reached the target confirmation number
    FundAlphaTransactionConfirmed,
```

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+
+/// Happy path states for Alice.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum AliceState {
+    // FIXME: Do we need a pre-created state in order to give Alice one more TTA period
+    // before `created`?
+    /// Initial state, swap has been created.
+    Created,
+    /// The fund alpha transaction has been sent to the network.
+    FundAlphaTransactionSent,
```
D4nte, line 241: Can we confirm we only assume a local state to reach this variant? e.g. Bob never considers Alice state to be this variant?

comit/src/expiries.rs
tcharding: @bonomat pointed out to me early in discussion that 'finality' is used. Here is an example search result: https://academy.binance.com/glossary/finality

comit/src/expiries.rs
tcharding: Yep, cool. I went with `None`. thanks

comit/src/expiries.rs
tcharding: Nice catch! Thanks man

comit/src/expiries.rs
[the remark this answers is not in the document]
da-kami: Yes, it should be correct. I took a look at [the HTLC expiry documentation](https://comit.network/docs/core-concepts/atomic-swap-htlc/#htlc-expiries) and I think my statement was actually not completely correct.

Bob has time till `alpha_expiry` is reached to redeem or refund. Alice has to redeem until beta expiry is reached. But Bob can actually redeem even after beta expiry is reached as long as he has still enough time to get his transaction confirmed before `alpha-expiry` is reached.

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+        }
+    }
+
+    /// We define 'useful' to mean:
+    /// 1. Alice has time to transition from current state to `Done`.
+    /// 2. Bob has time to transition from current state to `Done`.
+    ///
+    /// An actor only knows their own state but can reason about the state of
+    /// the counterparty.
+    // BUG: Assuming an actor uses their own state to calculate usefulness to
```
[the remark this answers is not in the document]
bonomat: Same here please.

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+    // themselves and their guess at the couterparties state to calculate
+    // usefulness to the counterparty we can find ourselves in a position where
+    // we continue because the swap is useful to us but the counterparty
+    // immediately cancels because they believe the swap to be not-useful to us.
+    // This is due to the fact that we must be conservative in our guess of the
+    // counterparties state, hence we may include additional unnecessary
+    // transitions in our calculation. This will only become an issue as we approach
+    // the time cutoff for usefulness i.e., while actors act comfortably within
+    // their 'time to act' window we should not hit this bug.
+    pub fn is_useful<S>(&self, current_state: S) -> bool
```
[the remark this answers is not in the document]
tcharding: I don't quite get what this comment means. Are you referring to the function name `is_useful` or to the logic?

comit/src/expiries.rs
D4nte: TIL.

comit/src/expiries.rs
```diff
@@ -0,0 +1,591 @@
...
+                // Transition from None to Created
+                c.alice_to_act()
+            }
+            FundAlpha => {
+                // Transition from Created to FundAlphaTransactionSent
+                c.alice_to_act()
+            }
+            WaitForFundTransactionFinality => {
+                // Transition from FundAlphaTransactionSent to AlphaFunded
+                c.include_alpha_fund_transaction() + c.finality_alpha()
```
D4nte, line 332:
```
c.include_alpha_fund_transaction()
```
I am guessing this is the time we assume it would take to mine the transaction? `mine` terminology would be better. Also, this is not specific to a `fund` transaction, or is it?

comit/src/expiries.rs
```diff
@@ -0,0 +1,591 @@
...
+                // Transition from RedeemBetaTransactionSent to Done
+                c.include_beta_redeem_transaction() + c.finality_beta()
+            }
+            // Transitioning to any of the cancel actions takes, be definition, zero time.
+            NoFurtherAction | Abort | WaitForRefund | Refund => Duration::zero(),
+        }
+    }
+
+    /// True if Alice has funded the alpha HTLC. We return true as soon as the
+    /// fund transaction has been sent since any time after attempting to refund
```
D4nte, line 352:
```suggestion
    /// fund transaction has been sent since any time after attempting to refund
```
I think it'd be good to reflect that in the name of the function as we have several states for a transaction (sent, mined, finalised).

comit/src/expiries.rs
```diff
@@ -0,0 +1,591 @@
...
+        }
+    }
+}
+
+/// Happy path states for Bob.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum BobState {
+    // Bob does not have a None state because swap creation time is implicit in Bob's first time
+    // to act after Alice funds.
+    /// Initial state, swap has been created.
```
D4nte, line 371: Again, I feel like `Started` might be better. `swap has been created.` created where?

comit/src/expiries.rs
```diff
@@ -0,0 +1,591 @@
...
+}
+
+/// Configuration values used during transition period calculations.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Config {
+    alpha_required_confirmations: usize,
+    beta_required_confirmations: usize,
+}
+
+// TODO: Calculate _real_ values instead of just returning 1 hour.
```
D4nte, line 505: Wasn't this done as part of the spike?

comit/src/expiries.rs
```diff
@@ -0,0 +1,591 @@
...
+
+    /// The duration of time it takes for the beta redeem transaction to be
+    /// included in the blockchain.
+    pub fn include_beta_redeem_transaction(&self) -> Duration {
+        1.hours()
+    }
+
+    /// The duration of time it takes for a transaction to reach finality on the
+    /// alpha ledger.
+    pub fn finality_alpha(&self) -> Duration {
```
D4nte, line 543: This value is actually a number of blocks. Block mining time may vary.

Once a transaction is confirmed, do you still use this value or do you provide a more exact time depending on the number of remaining confirmations?

comit/src/expiries.rs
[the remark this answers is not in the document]
D4nte: I am referring to the naming. I skipped a key word in my comment :) now edited.

comit/src/expiries.rs
tcharding: I do not fully understand the last line but I used your text as is.

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+
+        // Alice redeems on beta ledger so needs time to act before the beta expiry.
+        let alice_needs = period_for_alice_to_complete(&config, AliceState::Created);
+        let beta_now = beta.current_time();
+        let beta_start_time = beta_now.add_duration(ttl);
+        let beta_expiry = beta_start_time.add_duration(alice_needs);
+
+        // Bob redeems on alpha ledger so needs time to act before the alpha expiry.
+        let bob_needs = period_for_bob_to_complete(&config, BobState::Created);
+        let alpha_now = alpha.current_time();
```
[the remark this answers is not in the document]
tcharding: You are correct that they are orthogonal, each parties state transitions however fully include steps taken by the counterparty.

comit/src/expiries.rs
D4nte: Well, don't use it if you don't understand it... What I am saying is that the definition of `current time` differs from blockchain to blockchain and that what we define as `current time` here is what is considered as `current time` by the op codes of EVM or Bitcoin Script (meaning latest block time and median time).

Maybe it's an unnecessary clarification that just confuses more than it helps.

comit/src/expiries.rs
tcharding: Not sure what you are asking. Code now includes the following rustdoc
```
/// From<'role'State> converts a state object into a best guess at the
/// counterparties state. We cannot know for sure what state the counterparty is
/// in because there are state transitions which rely on local knowledge. We
/// make a conservative guess at the counterparties state using only public
/// knowledge.
```

comit/src/expiries.rs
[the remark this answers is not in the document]
tcharding: > Nice work!
> I add the `no-mergify` label just in case, seems you are still working in review comments.

Cool thanks, I forgot that :)

comit/src/expiries.rs
D4nte: So are you saying we are going to look into the Bitcoin and Ethereum mempool to be able to transition the counterpart state to `FundAlphaTransactionSent`?

comit/src/expiries.rs
```diff
@@ -0,0 +1,631 @@
...
+
+    /// The minimum time we need to allow to transition to the next state.
+    fn transition_period(&self, c: &Config) -> Duration {
+        use self::AliceAction::*;
+
+        let (next_action, _next_state) = self.next();
+
+        match next_action {
+            FundAlpha => {
+                // Transition from Created to FundAlphaTransactionSent
```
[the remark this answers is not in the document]
tcharding: The action variants are not named the same as the states they transition to/from. Hence the comment on each.

comit/src/expiries.rs
tcharding: Cool, you and @thomaseizinger now both prefer 'mine' over 'include' - will use. Yes this is specific to the fund, I see it as possible, in the future, that we may want to influence the mine latency of the fund transaction differently to the redeem transaction (lower fees for fund for example).

comit/src/expiries.rs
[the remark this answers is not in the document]
D4nte: Ah yeah, thanks.

comit/src/expiries.rs
tcharding: Used `has_sent_fund_transaction`

comit/src/expiries.rs
tcharding: No, that is the reason the spike is not merged :)

comit/src/expiries.rs
tcharding: As per the spike this function represents the minimum time we should wait for a transaction on the alpha ledger to reach finality. It is computed from the number of confirmations and the average block time using a Poisson distribution. See[ open draft PR](https://github.com/comit-network/comit-rs/pull/3058) if you are interested in how I am thinking of doing this.

tcharding:
> Good job I guess 
> Please add some test cases though.
> At least I was not able to verify the correctness by just looking at the code :)
> 
> Since you like theory: you could look up the concept of `Boundary testing and Equivalence partitioning` for this ;)

Cool cheers, I have no clue right now how to write useful tests for this :)

PR #3058, opened 2020-08-14
SEP time period functions

tcharding:
We model the swap state transitions by describing individual time periods for actions that must occur in order to transition from one state to the other.

Add a config module to provide the logic for calculating these individual time periods based on pre-configured values.

**Note:** After much messing around with Poisson distributions it was decided that simply using `average block time` * `n blocks` is just as good a way to calculate the expected time to wait for a n blocks to be mined as using the complex Poisson distribution calculation. This is typical in the ecosystem. The biggest issue we face when calculating this time period is that average block time is not known, this issue exists whether we use Poisson distribution or not.

### DIFF
--- a/comit/src/expiries.rs
+++ b/comit/src/expiries.rs
@@ -4,8 +4,15 @@
 //! atomic. One expiry for the alpha ledger and one for the beta ledger. The
 //! beta ledger expiry time must elapse before the alpha ledger expiry time.
 
-use crate::timestamp::{self, Timestamp};
-use time::{prelude::*, Duration};
+mod config;
+
+use crate::{
+    asset,
+    timestamp::{self, Timestamp},
+};
+use time::Duration;
+
+use self::config::{Config, Protocol};
 
 // TODO: Currently we ignore deploy for Erc20.
 
@@ -20,6 +27,10 @@ use time::{prelude::*, Duration};
 /// always refers to the time used by the op codes used in the COMIT contracts.
 pub trait CurrentTime {
     fn current_time(&self) -> Timestamp; // TODO: This will need to be async.
+}
+
+pub trait GasPrice {
+    fn gas_price(&self) -> asset::Ether;
 }
 
 /// Data that defines and manipulates expiry times.
@@ -226,7 +237,9 @@ where
 
 /// Duration for a complete happy path swap for Alice.
 fn happy_path_swap_period_for_alice(config: &Config) -> Duration {
-    period_for_alice_to_complete(&config, AliceState::None)
+    let p = period_for_alice_to_complete(&config, AliceState::None);
+    println!("period for alice to complete: {}", p.whole_seconds());
+    p
 }
 
 /// Duration for a complete happy path swap for Bob.
@@ -246,6 +259,8 @@ fn period_for_alice_to_complete(config: &Config, current_state: AliceState) -> D
         let (_action, next_state) = state.next();
         let transition_period = state.transition_period(config);
 
+        let d = acc + transition_period;
+        println!("acc {}", d.whole_seconds());
         period_to_complete(config, next_state, acc + transition_period)
     }
 
@@ -521,94 +536,6 @@ impl From<BobState> for AliceState {
             }
         }
     }
-}
-
-// TODO: Add support for lightning.
-#[derive(Clone, Copy, Debug)]
-pub enum Protocol {
-    Herc20Hbit,
-    HbitHerc20,
-}
-
-impl Protocol {
-    fn config(&self) -> Config {
-        match self {
-            Protocol::Herc20Hbit => Config {
-                alpha_required_confirmations: ethereum_required_confirmations(),
-                beta_required_confirmations: bitcoin_required_confirmations(),
-            },
-            Protocol::HbitHerc20 => Config {
-                alpha_required_confirmations: bitcoin_required_confirmations(),
-                beta_required_confirmations: ethereum_required_confirmations(),
-            },
-        }
-    }
-}
-
-/// Configuration values used during transition period calculations.
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub struct Config {
-    alpha_required_confirmations: usize,
-    beta_required_confirmations: usize,
-}
-
-// TODO: Calculate _real_ values instead of just returning 1 hour.
-impl Config {
-    /// The duration of time it takes for Alice to act.
-    pub fn alice_to_act(&self) -> Duration {
-        1.hours()
-    }
-
-    /// The duration of time it takes for Bob to act.
-    pub fn bob_to_act(&self) -> Duration {
-        1.hours()
-    }
-
-    /// The duration of time it takes for the alpha fund transaction to be
-    /// mined into the blockchain.
-    pub fn mine_alpha_fund_transaction(&self) -> Duration {
-        1.hours()
-    }
-
-    /// The duration of time it takes for the beta fund transaction to be
-    /// mined into the blockchain.
-    pub fn mine_beta_fund_transaction(&self) -> Duration {
-        1.hours()
-    }
-
-    /// The duration of time it takes for the alpha redeem transaction to be
-    /// mined into the blockchain.
-    pub fn mine_alpha_redeem_transaction(&self) -> Duration {
-        1.hours()
-    }
-
-    /// The duration of time it takes for the beta redeem transaction to be
-    /// mined into the blockchain.
-    pub fn mine_beta_redeem_transaction(&self) -> Duration {
-        1.hours()
-    }
-
-    /// The duration of time it takes for a transaction to reach finality on the
-    /// alpha ledger.
-    pub fn finality_alpha(&self) -> Duration {
-        1.hours()
-    }
-
-    /// The duration of time it takes for a transaction to reach finality on the
-    /// beta ledger.
-    pub fn finality_beta(&self) -> Duration {
-        1.hours()
-    }
-}
-
-fn bitcoin_required_confirmations() -> usize {
-    // TODO: Add documentation on _why_ we picked this value.
-    6
-}
-
-fn ethereum_required_confirmations() -> usize {
-    // TODO: Add documentation on _why_ we picked this value.
-    12
 }
 
 #[cfg(test)]
--- a/comit/src/expiries.rs
+++ b/comit/src/expiries.rs
@@ -237,9 +237,7 @@ where
 
 /// Duration for a complete happy path swap for Alice.
 fn happy_path_swap_period_for_alice(config: &Config) -> Duration {
-    let p = period_for_alice_to_complete(&config, AliceState::None);
-    println!("period for alice to complete: {}", p.whole_seconds());
-    p
+    period_for_alice_to_complete(&config, AliceState::None)
 }
 
 /// Duration for a complete happy path swap for Bob.
@@ -259,8 +257,6 @@ fn period_for_alice_to_complete(config: &Config, current_state: AliceState) -> D
         let (_action, next_state) = state.next();
         let transition_period = state.transition_period(config);
 
-        let d = acc + transition_period;
-        println!("acc {}", d.whole_seconds());
         period_to_complete(config, next_state, acc + transition_period)
     }
 
--- a/comit/src/expiries/config.rs
+++ b/comit/src/expiries/config.rs
@@ -1,0 +1,210 @@
+//! This module provides functionality for calculating the duration of actions
+//! required to determine the transition period from one swap state to the next.
+
+use time::{prelude::*, Duration};
+
+// TODO: From somewhere within the system we need to return to the
+// user a transaction fee to use for each of the transactions (deploy,
+// fund, refund, redeem). In this module we assume the fee is set to
+// the suggested amount. We rely on this assumption for the
+// confirmation time calculations in this module to be correct.
+//
+// For Ethereum:
+//
+// gas limit: blockchain-contracts already implements gas limit functionality
+//            for each Ethereum transaction type.
+// gas price: current gas price can be learned from the Ethereum blockchain,
+//            for example from geth or infura.
+// modifier:  We could use a pre-configured modifier, as is widely done in the
+//            Ethereum ecosystem for slow, medium, fast.
+//
+// For Bitcoin:
+//
+// With a configured 'include within N blocks' we can use bitcoind via
+// bitcoin-cli, by calling:
+//
+//  `bitcoin-cli estimatesmartfee N_BLOCKS`
+//
+// CAVEAT: The fee estimator relies on an active mempool with good uptime to
+// watch fee activity on the network, that in turn implies that our calculations
+// herein are only as good as the fee estimator of the Bitcoin connector.
+
+// We use specific integer types to limit the upper bound, this reduces the need
+// to turn off lints.
+
+const BITCOIN_BLOCK_TIME_SECS: u16 = 600; // 10 minutes, average Bitcoin block time.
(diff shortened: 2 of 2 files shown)


comit/src/expiries/config.rs
```diff
@@ -0,0 +1,289 @@
...
+// bitcoin-cli, by calling:
+//
+//  `bitcoin-cli estimatesmartfee N_BLOCKS`
+//
+// CAVEAT: The fee estimator relies on an active mempool with good uptime to
+// watch fee activity on the network, that in turn implies that our calculations
+// herein are only as good as the fee estimator of the Bitcoin connector.
+
+const BITCOIN_BLOCK_TIME_SECS: usize = 600; // 10 minutes, average Bitcoin block time.
+const ETHEREUM_BLOCK_TIME_SECS: usize = 20; // Conservative Ethereum block time.
```
D4nte, line 34: Why using a conservative and not average/target block time?

comit/src/expiries/config.rs
```diff
@@ -0,0 +1,289 @@
...
+const BITCOIN_MINE_WITHIN_N_BLOCKS: usize = 3; // Value arbitrarily picked.
+const ETHEREUM_MINE_WITHIN_N_BLOCKS: usize = 3;
```
D4nte, line 36: Is this not arbitrarily picked?

comit/src/expiries/config.rs
```diff
@@ -0,0 +1,289 @@
...
+
+// TODO: Add support for lightning.
```
D4nte, line 38: I never thought of lightning by the way...

comit/src/expiries/config.rs
```diff
@@ -0,0 +1,289 @@
...
+#[derive(Clone, Copy, Debug)]
+pub enum Protocol {
+    Herc20Hbit,
+    HbitHerc20,
+}
+
+impl Protocol {
+    pub fn config(&self) -> Config {
+        // Since we are targeting Nectar and the App we can set these.
+        let alice = ActorType::Human;
```
D4nte, line 48: It's an interesting concept that you came up with. However, I am not sure it's appropriate. As @thomaseizinger mentioned on the mailing list, we want to move the app to a full automation.
Which means (to my understanding) that once you take an order, you only need to sign once with your hardware wallet and the rest will be handled by the app (bot).

More importantly than knowing who is behind the screen (or not), I'd be keen to have the time to act fair and balanced in the context of the "option problem". Each party should be able to watch the market and decide to proceed or abort for the same period of time.

Interestingly, Alice has two periods to do so:
A1. Before funding (no funds locked)
A2. Before redeeming (A & B funds locked)

And Bob only has one:
B1. Before funding (A funds locked)

It'd be interesting to say for example:

```
A1+A2 = B1: 
```
.e.g. A1 = 30min, A2 = 30min and B1 = 60min.

However, in terms of advantages/penalty, you could think:
- A1 gives least advantages as no funds are locked
- A2 gives a bit of advantage has both parties have funds locked, Alice has the choice but Bob has his funds locked for the least amount of time
- B1 gives the most advantage as it is risk free for Bob but Alice would be penalised by having her funds locked

It's basically some game theory (and I know nothing about it) and the question of what is the more advantageous:
- having the upper hand
- having more time to decide
- having funds locked for shorter time

etc...

I am guessing there is no easy answer unless we have stats that shows how many time people abort and in which conditions.

comit/src/expiries/config.rs
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+}
+
+/// Configuration values used during transition period calculations.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Config {
+    alice: ActorType,
+    bob: ActorType,
+    alpha_required_confirmations: usize,
```
D4nte, line 85: `usize` is an odd choice. Are you sure your not shooting yourself on the foot by using a platform dependent type?

comit/src/expiries/config.rs
```diff
@@ -0,0 +1,289 @@
...
+            ActorType::Human => 60.minutes(),
+        }
+    }
+}
+
+impl Config {
+    // Optimisation: finality_alpha/beta never change, we should cache the value so
+    // we only do the calculation once.
+
+    // Open Question: Is time_to_mine_n_blocks() _really_ better than just doing:
```
D4nte, line 114: Yes because on Bitcoin the difficulty changes every 2016 blocks (2 weeks). Which means that if there is a big influx or withdraw of hashing power, the average block time could drastically change which would then influence this figure.

comit/src/expiries/config.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub fn finality_beta(&self) -> Duration {
+        let n = self.beta_required_confirmations;
+        let block_time = self.beta_block_time;
+
+        time_to_mine_n_blocks(n, block_time)
+    }
+}
+
+fn bitcoin_required_confirmations() -> usize {
+    6 // This is the accepted value within the Bitcoin ecosystem.
```
D4nte, line 186: Maybe worth adding those as constants at the start of the file.

comit/src/expiries/config.rs
tcharding: Interesting thoughts. I was under the impression that only A2 constitutes the option problem.

comit/src/expiries/config.rs
tcharding: I thought `usize` was the default size to use for an unsigned integer in Rust. This value is always going to small (less than 2 bytes probably) so the integer size does not matter here.

comit/src/expiries/config.rs
tcharding: We use average block time of 10 minutes so the code is still subject to the weakness you point out - hence the question.

comit/src/expiries/config.rs
tcharding: Will do, thanks for prodding me.

tcharding:
Thanks for the reveiw @D4nte

comit/src/expiries/config.rs
D4nte: > Interesting thoughts. I was under the impression that only A2 constitutes the option problem.

Hum, ok maybe it's not a real "option problem" but B1 definitely gives Bob some upper hand as he can decide not to proceed depending on market direction and Alice would end up with assets locked for a long time.

But then, of course if he does decide to proceed as the market is going in the right direction for him, Alice may decide not to redeem as the market may be going in the bad direction for her.

So more Bob waits before funding (assume he eventually funds), more he risks for Alice to actually not proceed as she would also be watching the market.
Alice can watch the market during Atot = A1 + B1 + A2.
If Bob makes B1 long, then yes he gives himself some advantage to see where the market go but he also increases Atot allowing Alice to have even more time to watch the market (and likely decide in the opposite direction than Bob).

I hope it makes sense :) it's hard to make it clear.

comit/src/expiries/config.rs
D4nte: If you need to cast the `usize` variable to `u32` down the line due to another API that takes a `u32` then you risk a truncation that is only protected by the functional behaviour of the code ("less than 2 bytes probably") which is never good.
I don't think you do such cast but why not just using `u16` or `u32` instead if you think it's "less than 2 bytes probably"?

Also, as per https://doc.rust-lang.org/std/primitive.usize.html, usize is:

> The pointer-sized unsigned integer type.

> The size of this primitive is how many bytes it takes to reference any location in memory. For example, on a 32 bit target, this is 4 bytes and on a 64 bit target, this is 8 bytes.

This is not the "default size to use for an unsigned integer in Rust". It is the type to use for referencing location in memory.

Which means you should use `usize` if a variable is used to index memory location. For example, to track the location of an item in a `Vec` or other heaped container.

comit/src/expiries/config.rs
D4nte: It's still more precise and if our hardcoding of 10min becomes an issue in the future we can optimise.

If you decide to move away from `time_to_mine_n_blocks` I think it's ok-ish too. We just need to take in account that 10min is around 30% approx. It'd be good to have a quick look at the historical because hashpower shift because has encountered (I think it was 30% at the last halving , hence my suggestion).

comit/src/expiries/config.rs
tcharding: The only information I could find suggested that Ethereum average block time is typically between 10 and 20 seconds. Since this is so small (compared to Bitcoin) I figured just take a conservative position and use 20.

comit/src/expiries/config.rs
tcharding: :), added the comment here also.

comit/src/expiries/config.rs
tcharding: That's clear, it is for the reason you outline (i.e., Bob waiting helps and hinders) that I do not think we need consider B1. IMO we should limit our concern to the true option problem and forget the rest, just assume honest actors who start the swap as soon as possible and act in a 'rational' manner. Re: human/bot, perhaps these should be renamed to app/nectar but I feel that it is still useful to have two values here, even if we decide to set them to the same value for now. This is the same reasoning behind having two time to mine values (one for fund and one for redeem). The idea is to make sure we have all the knobs we need to correctly model the expiries, not to make sure we have the values correct.

comit/src/expiries/config.rs
tcharding: Thanks, I consider myself schooled. Here is another quote by a Rust guy that backs up what you said. Thanks, this was a useful lesson.

> Use usize and isize when it’s related to memory size – the size of an object, or indexing a vector, for instance. It will be a 32-bit number on 32-bit platforms, as that’s the limit of memory they can address, and likewise for 64-bit.
>
> Use u32 and i32 when you just want numbers.

ref: https://users.rust-lang.org/t/i32-vs-isize-u32-vs-usize/22657

comit/src/expiries.rs
```diff
@@ -4,8 +4,15 @@
...
-use crate::timestamp::{self, Timestamp};
-use time::{prelude::*, Duration};
+mod config;
+
+use crate::{
+    asset,
+    timestamp::{self, Timestamp},
+};
+use time::Duration;
+
```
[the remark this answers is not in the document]
tcharding: I left this space on purpose because the use of `self` is unusual for our codebase.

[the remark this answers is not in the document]
tcharding: `use self::config::{Config, Protocol};`

comit/src/expiries/config.rs
D4nte: Oh, I thought it was officially 17s.

comit/src/expiries/config.rs
thomaseizinger: Thanks for tagging me on this @D4nte !

Maybe the `ActorType` abstraction is just too broad? Given the idea about more automation on the mailing list, only one out of two actions is performed by a human, hence maybe this should be broken down to _per action_?

What we also may want to consider (in addition to automating redeeming) is the idea of "pre-signing" the fund transaction and let cnd broadcast it at the right point in time (assuming the fees are still useful). This would allow users to trade against each other (i.e. no nectar) without having to attend the UI all the time. In that case, all transactions would be broadcasted by a machine and hence time-to-act is not that relevant anymore.

For now, I'd be interested in how big of difference this actually makes?
How big is TTA compared to the whole swap duration? Because if it is rather small, we might as well just ignore it for now in the face of the idea that we will eventually automate most of it.

comit/src/expiries/config.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /// The duration of time it takes for Bob to act.
+    pub fn bob_to_act(&self) -> Duration {
+        self.bob.period_to_act()
+    }
+
+    /// The duration of time it takes for the alpha fund transaction to be
+    /// mined into the blockchain.
+    pub fn mine_alpha_fund_transaction(&self) -> Duration {
```
thomaseizinger, line 129: Terminology nit-pick: "mined" to me implies a single confirmation. Might want to change this to "confirm_alpha_fund_transaction".

comit/src/expiries/config.rs
```diff
@@ -0,0 +1,215 @@
...
+
+fn bitcoin_required_confirmations() -> u8 {
+    BITCOIN_CONFIRMATIONS
+}
+
+fn ethereum_required_confirmations() -> u8 {
+    ETHEREUM_CONFIRMATIONS
+}
+
+// Time to mine n blocks is governed by a Poisson distribution. As an
```
thomaseizinger, line 190: This is outdated now, no?

thomaseizinger (commented):
Nice progress!

comit/src/expiries/config.rs
tcharding: No its not outdated, its still governed by this we just do not calculate it. I include the note in case we wish to calculate this at some later date.

comit/src/expiries/config.rs
tcharding: This specifically refers to a single confirmation.

comit/src/expiries.rs
```diff
@@ -226,7 +237,9 @@ where
 
 /// Duration for a complete happy path swap for Alice.
 fn happy_path_swap_period_for_alice(config: &Config) -> Duration {
-    period_for_alice_to_complete(&config, AliceState::None)
+    let p = period_for_alice_to_complete(&config, AliceState::None);
+    println!("period for alice to complete: {}", p.whole_seconds());
```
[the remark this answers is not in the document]
tcharding: My bad, there should not be any output in this  - that's from debugging.

comit/src/expiries/config.rs
tcharding: I am working on an updated spike version that has the time to act modelled separately for each transaction.

thomaseizinger: But the implementation uses `time_to_mine_n_blocks`? :)

If it is specifically one confirmation, can't you hardcode `n` to `1` then?

thomaseizinger: @tcharding I am not blocking your merge but I would like an answer to this question :)

tcharding: I'm not sure what you are asking @thomaseizinger. `time_to_mine_n_blocks()` calculates a time period that is big enough that we can hope to mine N blocks within this period. It is called from `mine_alpha_fund_transaction` with N set to the same N we should use when calculating the transaction fee. Said differently the transaction fee is calculated to ensure a transaction gets mined within N blocks, we use `time_to_mine_n_blocks` to calculate the time period.

Does this answer your question?

thomaseizinger: No :sweat_smile: 

My point exactly is that this function assumes more than one block is being mined, which is usually referred to as "confirmations".

If the function is called "mine_alpha_fund_transaction" though. Hence, I would assume it gives me the time until the alpha fund transaction is mined which - as I said in my first comment - usually refers to single confirmation aka block but not N blocks which is what this function is actually doing.

Initially it was just a nit but I feel there is a major misunderstanding here:

- "Confirmed" implies to me that N blocks have been mined with N being my confirmation target
- "Mined" implies to me that a single block has been mined that includes this transaction

Hence, for what the function does, I believe it should be called "confirm_alpha_fund_transaction" as it calculates the time until it is confirmed with N blocks and not just included within one.

tcharding: This function does not calculate the time until it is confirmed N times. It calculates the time it would take to mine N blocks because we expect that our transaction will be included _within_ N blocks - therefore we must wait at least N blocks to hope that our transaction is within one of these blocks. Hence it is called 'time to mine'. This is distinct from the time we wait for N confirmations even though both functions rely on the time to mine N blocks.

tcharding:
This PR is superseded with: https://github.com/comit-network/comit-rs/pull/3070

PR #3071, opened 2020-08-20
Infer DAI token from Ethereum chain ID

thomaseizinger:
We are going to use this to avoid including the token contract address in the orders that we are sending around.

This introduces the ethereum::Address type into our config::File
struct. Unfortunately, this opens a can of worms with a bug in
the serde-hex library. See [0] for more details.

We work around this by removing the dependency on serde-hex from our
Ethereum deserialization code and instead hand-roll everything for
our usecase.

This also has the advantage that we can now again just use our structs
directly in the route handlers of warp instead of serde_json::Value.
See [1] for more on this.

[0]: https://github.com/fspmarshall/serde-hex/pull/8
[1]: https://github.com/comit-network/comit-rs/pull/2405


cnd/src/config.rs
```diff
@@ -144,8 +158,16 @@ impl Ethereum {
 
         let chain_id = ethereum.chain_id;
         let geth = ethereum.geth.unwrap_or_else(Geth::new);
+        let tokens = ethereum.tokens.map_or_else(
+            || Tokens::from_config_or_chain_id(None, chain_id),
+            |tokens| Tokens::from_config_or_chain_id(tokens.dai, chain_id),
+        )?;
```
tcharding, line 164: Ha! Took me a while to figure out this pattern but its kinda nice :)

comit/src/ethereum.rs
```diff
@@ -227,6 +198,7 @@ pub struct ChainId(u32);
 impl ChainId {
     pub const MAINNET: Self = ChainId(1);
     pub const ROPSTEN: Self = ChainId(3);
```
tcharding, line 200: Out of interest, is there a reason we don't have a token contract for Ropsten?

tcharding (approved):

comit/src/ethereum.rs
thomaseizinger: Yes, because there is no offical one :(

https://developer.makerdao.com/dai/1/api/

cnd/src/config.rs
thomaseizinger: Yeah, there is a lot going on here but I like how concise it is, especially with the error handling as well :)

thomaseizinger: One unfortunate thing is that the arguments are in a weird order. The first one is the default `or_else` and the second one is the `map`. There is a reason though: https://github.com/rust-lang/rust/issues/53268#issuecomment-429988655

thomaseizinger: This actually made me notice a certain inconsistency in how these functions are named :)

comit/src/ethereum.rs
D4nte: There is: `0x31F42841c2db5173425b5223809CF3A38FEde360`
See https://github.com/makerdao/developerguides/blob/804bb1f4d1ea737f0287cbf6480a570b888dd547/dai/dai-token/dai-token.md

thomaseizinger: Nice!

Thank you @D4nte.

But where did you get this address from? For me it says: 0x4F96Fe3b7A6Cf9725f59d353F723c1bDb64CA6Aa

thomaseizinger: NVM, I mixed it up :)

thomaseizinger:
@D4nte and @tcharding I pushed two follow-up commits that were necessary to fix the e2e tests and also to include the ropsten address :)

PR #3084, opened 2020-08-25
Always initialize a wallet for each ledger within the Actor

thomaseizinger:
This allows us to write DRY tests that for example only require an
address. It is also cleaner in the way that `Actor` is now less
mutable because everything gets initialized from the start.


thomaseizinger:
@D4nte this is almost the same as #3075 except that I am creating the stub wallets in a less verbose way :)

PR #3085, opened 2020-08-25
Use datetime type that includes timezone

D4nte:
While one could argue that the domain dictate a use of UTC, we already
demonstrated that we are not impervious to mistakes coblox/nectar#111.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   **Breaking Change** Remove support for RFC003 swaps
 -   **Breaking Change** Config directory for MacOS changed from `/Users/<user>/Library/Preferences/comit/` to `/Users/<user>/Library/Application Support/comit/`.
+-   **Breaking Change comit lib API**: Use `DateTime<Utc>` instead of `NaiveDateTime` to remove ambiguity on the timezone.
 
 ## [0.8.0] - 2020-06-12
 
--- a/cnd/src/hbit.rs
+++ b/cnd/src/hbit.rs
@@ -6,7 +6,7 @@ use crate::{
     LocalSwapId, Role, Side,
 };
 use bitcoin::{Address, Block, BlockHash};
-use chrono::NaiveDateTime;
+use chrono::{DateTime, Utc};
 use comit::{asset, htlc_location, transaction, LockProtocol, Secret};
 pub use comit::{hbit::*, identity};
 use futures::TryStreamExt;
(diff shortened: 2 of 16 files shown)


thomaseizinger (commented):
If we care about the timezone, then I'd suggest that we use the type provided by chrono that carries a timezone:

https://docs.rs/chrono/0.4.15/chrono/struct.DateTime.html

D4nte:
> If we care about the timezone, then I'd suggest that we use the type provided by chrono that carries a timezone:
> 
> [docs.rs/chrono/0.4.15/chrono/struct.DateTime.html](https://docs.rs/chrono/0.4.15/chrono/struct.DateTime.html)

I don't think we care about it as the domain is such as only UTC is relevant but what we care is helping consumers not making mistakes.

thomaseizinger:
> > If we care about the timezone, then I'd suggest that we use the type provided by chrono that carries a timezone:
> > [docs.rs/chrono/0.4.15/chrono/struct.DateTime.html](https://docs.rs/chrono/0.4.15/chrono/struct.DateTime.html)
> 
> I don't think we care about it as the domain is such as only UTC is relevant but what we care is helping consumers not making mistakes.

Hmm fair but it still sounds like enforcing that through the type system might be better than just changing the variable name? :grimacing:

tcharding:
> While one could argue that the domain dictate a use of UTC, we already
> demonstrated that we are not impervious to mistakes [coblox/nectar#111](https://github.com/coblox/nectar/pull/111).

Seems to me that `NaiveDateTime` is explicitly used for UTC

From the [docs:](https://docs.rs/chrono/0.4.15/chrono/naive/struct.NaiveDateTime.html)
> ISO 8601 combined date and time without timezone.

So the mistake in Nectar was using `DateTime` in the first place. I don't see this PR as a fix to that. If we want timezones we should use `DateTime` if we do not we should just use `NaiveDateTime`. Or am I missing something?

thomaseizinger:
> Seems to me that `NaiveDateTime` is explicitly used for UTC

That doesn't seem like a right conclusion to me. According to Wikipedia, ISO8601 formats _without_ timezone information imply local time: https://en.wikipedia.org/wiki/ISO_8601#Local_time_(unqualified)

> If we want timezones we should use `DateTime` if we do not we should just use `NaiveDateTime`.

I agree with that but would go further in saying: It seems like we want timezones, in particular, we want UTC.

Also, it seems like the type system could have prevented this bug as `DateTime` has a generic parameter for the timezone. That is a good enough reason by itself for me that we should be using `DateTime`.

thomaseizinger (approved):
Thank you @D4nte !

tcharding (approved):

PR #3100, opened 2020-08-30
Minor changes for nectar upgrade to latest comit

rishflab:
needed for https://github.com/coblox/nectar/pull/112

### DIFF
--- a/comit/src/network/orderbook.rs
+++ b/comit/src/network/orderbook.rs
@@ -67,6 +67,10 @@ impl Orderbook {
         &self.orderpool
     }
 
+    pub fn clear_own_orders(&mut self) {
+        self.orderpool.clear_own_orders();
+    }
+
     fn my_poll<BIE>(
         &mut self,
         _: &mut Context<'_>,


comit/src/asset/ethereum/erc20.rs
```diff
@@ -100,6 +100,12 @@ impl From<Erc20Quantity> for blockchain_contracts::ethereum::TokenQuantity {
     }
 }
 
+impl From<BigUint> for Erc20Quantity {
```
thomaseizinger, line 103: NACK. There is a `TryFrom` conversion that you can use which maintains the MAX_VALUE invariant.

thomaseizinger (changes requested):

comit/src/expiries.rs
```diff
@@ -326,9 +326,9 @@ impl<A, B> fmt::Display for Expiries<A, B> {
 
 // Its super easy to mix these up, add types so the compiler saves us.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub struct AlphaOffset(Duration);
+pub struct AlphaOffset(pub Duration);
```
thomaseizinger, line 329: Hmm, not sure if these should be public. What do you think @tcharding?

There exist `From` conversions, so one could argue that if I can convert any `Duration` into one of these, then they might as well be public.

thomaseizinger (approved):

comit/src/expiries.rs
D4nte: As a general rule I'd say that if you need conversion, then use the `From`/`Into` traits.
Conversion is not a good enough excuse to expose internals.

thomaseizinger: @rishflab Can you use the conversion functions please instead of making these public?

rishflab: doen

PR #3101, opened 2020-08-30
Mark orders as settling once setup-swap succeeds

thomaseizinger:


cnd/src/storage/db/tables.rs
```diff
@@ -378,22 +383,65 @@ impl Order {
...
+    ///
+    /// Whilst we don't have partial order matching, this simply means updating
+    /// the percent of:
+    ///
+    /// - `open` to `0`
+    /// - `settling` to `100`
+    ///
+    /// Once we implement partial order matching, this will need to get more
+    /// sophisticated.
+    pub fn mark_as_settling(conn: &SqliteConnection, order: &Order) -> Result<()> {
```
rishflab, line 397: Where you thinking of using the satoshi quantity as the settling number?

thomaseizinger: No, the idea was that the "state" values are a percentage of how much is settled / closed / etc.

Not sure if it is the best design. I think for the HTTP API, this makes sense but for the DB, we might be better of fetching the quantities of the related swaps?

PR #3104, opened 2020-08-31
Allow orders to be cancelled

thomaseizinger:


rishflab (approved):

da-kami (approved):
❓ I reckon we don't deal with partial orders yet? (there was no test - I assume one can only cancel an order that is open 100% atm)

thomaseizinger:
Well yes and no :D

Orders can be cancelled if they more than 0 percent of their amount is still open.
Upon cancelling, we take all that is currently open and move it to cancelled.

So, the cancel logic would support partial orders but they don't appear in the system yet because the matching logic is not implemented.

PR #3110, opened 2020-08-31
Notes on itegration of the SECT module

tcharding:
This PR is not a merge candidate.

Integration of the new `expiries` module into cnd and Nectar is non-trivial. This PR adds some comments about how we might go about it. Opening as draft for comments and/or historical value until we decide how/when to do the integration.


thomaseizinger:
I think we can close this. It is going to be here for reference anyway and we already have ideas about how to integrate this module.

PR #3112, opened 2020-09-01
Add a comment on the partial ordering of current_time

tcharding:
Add documentation for the constraint that return values of
`CurrentTime` from different implementors (i.e. alpha/beta
connectors) can not be compared.

Please see code change for full explanation.


comit/src/expiries.rs
```diff
@@ -70,6 +70,18 @@ pub fn to_timestamps(
 /// always refers to the time used by the op codes used in the COMIT contracts.
 #[async_trait]
 pub trait CurrentTime {
+    // Two types that implement this trait do not necessarily support total
```
D4nte, line 73: Should we use a different struct that `Timestamp` here and **not** impl `Ord` or `PartialOrd` on it?

D4nte (approved):

thomaseizinger (approved):

PR #3113, opened 2020-09-01
Remove custom thread stack size

D4nte:
I believe the stack was due to our heavy use of traits which is now
gone. Let's see what the CI says.


thomaseizinger (approved):
Sweet, thank you @D4nte !

While we are at it, we could actually convert our `main` to use `#[tokio::main]` and make it `async`. For when we need to retrieve a handle, we could use `Handle::current`.

That would allow us to even get rid of instantiating the runtime altogether :)

tcharding (approved):

PR #3119, opened 2020-09-02
Inline comit-js-sdk into api_tests

thomaseizinger:
As per discussion on the mailing list, we are deprecating the comit-js-sdk. For some the benefits, check the 2nd patch in this PR.


api_tests/src/cnd/swaps_payload.ts
```diff
@@ -0,0 +1,288 @@
...
+    amount: string;
+    identity: string;
+}
+
+interface Herc20Request {
+    contract_address: string;
+    absolute_expiry: number;
+}
+
+interface HalightRequest {
```
D4nte: Isn't this already defined in the e2e tests? I thought we dupe it already as we were doing iterations on the halight endpoints.

api_tests/src/cnd/swaps_payload.ts
```diff
@@ -1,288 +0,0 @@
-/**
```
D4nte: Ok this answers my other comment :)

D4nte (approved):

da-kami (approved):

PR #3127, opened 2020-09-03
Use noise instead of SECIO

tcharding:
SECIO is now deprecated by libp2p (see ref). Use noise instead to build our transports.

ref: https://github.com/libp2p/rust-libp2p/pull/1729
     https://blog.ipfs.io/2020-08-07-deprecating-secio/

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,37 +16,57 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ee2a4ec343196209d6594e19543ae87a39f96d5534d7174822a3ad825dd6ed7e"
 
 [[package]]
-name = "aes-ctr"
-version = "0.3.0"
+name = "aead"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2e5b0458ea3beae0d1d8c0f3946564f8e10f90646cf78c06b4351052058d1ee"
+checksum = "7fc95d1bdb8e6666b2b217308eeeb09f2d6728d104be3e31916cc74d15420331"
+dependencies = [
+ "generic-array 0.14.3",
+]
+
+[[package]]
+name = "aes"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f7001367fde4c768a19d1029f0a8be5abd9308e1119846d5bd9ad26297b8faf5"
 dependencies = [
  "aes-soft",
  "aesni",
- "ctr",
- "stream-cipher",
+ "block-cipher",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "86f5007801316299f922a6198d1d09a0bae95786815d066d5880d13f7c45ead1"
+dependencies = [
+ "aead",
+ "aes",
+ "block-cipher",
+ "ghash",
+ "subtle 2.2.3",
 ]
 
 [[package]]
 name = "aes-soft"
-version = "0.3.3"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cfd7e7ae3f9a1fb5c03b389fc6bb9a51400d0c13053f0dca698c832bfd893a0d"
+checksum = "4925647ee64e5056cf231608957ce7c81e12d6d6e316b9ce1404778cc1d35fa7"
 dependencies = [
- "block-cipher-trait",
+ "block-cipher",
  "byteorder",
  "opaque-debug 0.2.3",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


cnd/src/network/transport.rs
```diff
@@ -57,21 +60,24 @@ fn build_comit_transport(keypair: identity::Keypair) -> anyhow::Result<ComitTran
...
-    let transport = Socks5TokioTcpConfig::default().nodelay(true).onion_map(map);
-    let transport = DnsConfig::new(transport)?;
+    let dh_keys = noise::Keypair::<X25519Spec>::new().into_authentic(&id_keys)?;
+    let noise = NoiseConfig::xx(dh_keys).into_authenticated();
+
+    let socks = Socks5TokioTcpConfig::default().nodelay(true).onion_map(map);
+    let dns = DnsConfig::new(socks)?;
 
-    let transport = transport
+    let transport = dns
```
thomaseizinger, line 78: Seeing this, I think DNS doesn't make a lot of sense for the Tor transport :)

thomaseizinger (approved):

Cargo.lock
```diff
@@ -1534,6 +1652,7 @@ dependencies = [
  "libp2p-dns",
  "libp2p-gossipsub",
  "libp2p-mplex",
+ "libp2p-noise",
  "libp2p-request-response",
  "libp2p-secio",
```
thomaseizinger, line 1657: Why is this dependency not going away? :thinking:

cnd/src/network/transport.rs
tcharding: I noticed the same thing, its done so we can return the same type for both the tor transport builder function and the usual one. I elected to be lazy and not investigate further whether it was actually doing any damage since we don't use the tor one actively. Should probably be thought about more though. If you have an answer without too much brain power I'm happy to implement it.

Cargo.lock
tcharding: Sloppy jalopy - will fix. Thanks.

tcharding: oh not so sloppy, I did remove the feature from libp2p.

tcharding: Found the little blighter - it was in comit network tests.

tcharding: `git grep -i secio` now returns nothing :)

thomaseizinger:
@tcharding Do you mind also updating nectar? Otherwise the two applications will not be compatible :)
I guess this change should also make it into the changelog!

tcharding:
> @tcharding Do you mind also updating nectar? Otherwise the two applications will not be compatible :)
> I guess this change should also make it into the changelog!

I guess this PR is as good a place as any to make a decision on what goes where and who is responsible for patching what (cnd, nectar, etc). The Nectar upgrade is non-trivial, I am not that keen to do it without some precedent being formed so I know when I have to do this sort of thing and how it should be done i.e., PR order (cnd first then nectar?)

thomaseizinger:
> > @tcharding Do you mind also updating nectar? Otherwise the two applications will not be compatible :)
> > I guess this change should also make it into the changelog!
> 
> I guess this PR is as good a place as any to make a decision on what goes where and who is responsible for patching what (cnd, nectar, etc). The Nectar upgrade is non-trivial, I am not that keen to do it without some precedent being formed so I know when I have to do this sort of thing and how it should be done i.e., PR order (cnd first then nectar?)

The nectar upgrade should be pretty much the same as this no?
@rishflab recently updated nectar to the latest comit lib. Also, changing the transport should not need any change to the COMIT stuff. It is purely application-specific.

I am happy to merge this as is, I just wanted to avoid that we are being stuck on in-compatible software again at the start of the next project, hence I brought it up :)

tcharding:
> > > @tcharding Do you mind also updating nectar? Otherwise the two applications will not be compatible :)
> > > I guess this change should also make it into the changelog!
> > 
> > 
> > I guess this PR is as good a place as any to make a decision on what goes where and who is responsible for patching what (cnd, nectar, etc). The Nectar upgrade is non-trivial, I am not that keen to do it without some precedent being formed so I know when I have to do this sort of thing and how it should be done i.e., PR order (cnd first then nectar?)
> 
> The nectar upgrade should be pretty much the same as this no?

The secio stuff is but upgrading `Cargo.toml` to a more recent commit breaks a bunch of stuff. 

> @rishflab recently updated nectar to the latest comit lib. Also, changing the transport should not need any change to the COMIT stuff. It is purely application-specific.

Nectar does not build if the comit-rs dependencie is update to current tip of master.

> I am happy to merge this as is, I just wanted to avoid that we are being stuck on in-compatible software again at the start of the next project, hence I brought it up :)

I'm going to convert this to draft and leave it open for others to look at. I have other things on my priority list at the moment that I'd rather work on. Happy to pick this up once a resolution to cnd/nectar codebase alignment is sorted out.

D4nte (approved):

PR #3128, opened 2020-09-04
Filter out market orders with "unsafe" expiries

rishflab:
The GET markets/BTC-DAI endpoint in Cnd used to return all order including those that do not match the expiries determined by the expiry module. A filter has been added to remove orders that do not match these safe expiries.

### DIFF
--- a/cnd/src/http_api/markets/get_btc_dai.rs
+++ b/cnd/src/http_api/markets/get_btc_dai.rs
@@ -24,7 +24,7 @@ async fn handler(swarm: Swarm) -> Result<impl Reply> {
     let mut orders = siren::Entity::default();
     let local_peer_id = swarm.local_peer_id();
 
-    for (maker, order) in swarm.btc_dai_market().await {
+    for (maker, order) in swarm.btc_dai_market_safe_expiries().await {
         let market_item = siren::Entity::default()
             .with_properties(MarketItem {
                 id: order.id,
--- a/cnd/src/http_api/markets/get_btc_dai.rs
+++ b/cnd/src/http_api/markets/get_btc_dai.rs
@@ -24,7 +24,7 @@ async fn handler(swarm: Swarm) -> Result<impl Reply> {
     let mut orders = siren::Entity::default();
     let local_peer_id = swarm.local_peer_id();
 
-    for (maker, order) in swarm.btc_dai_market_safe_expiries().await {
+    for (maker, order) in swarm.btc_dai_market().await {
         let market_item = siren::Entity::default()
             .with_properties(MarketItem {
                 id: order.id,
--- a/cnd/src/http_api/markets/get_btc_dai.rs
+++ b/cnd/src/http_api/markets/get_btc_dai.rs
@@ -20,20 +20,19 @@ pub fn route(swarm: Swarm) -> impl Filter<Extract = impl Reply, Error = Rejectio
         })
 }
 
-/// Retrieves "executable" orders: orders that have expiries that match the safe
+/// Retrieves viable orders: orders that have expiries that match the safe
 /// expiries determined by the expiries module.
 async fn handler(swarm: Swarm) -> Result<impl Reply> {
     let mut orders = siren::Entity::default();
     let local_peer_id = swarm.local_peer_id();
 
-    let executable_orders = facade
-        .swarm
+    let viable_orders = swarm
         .btc_dai_market()
         .await
         .into_iter()
-        .filter(|(_, order)| has_executable_expiries(order));
+        .filter(|(_, order)| has_viable_expiries(order));
 
-    for (maker, order) in executable_orders {
+    for (maker, order) in viable_orders {
         let market_item = siren::Entity::default()
             .with_properties(MarketItem {
                 id: order.id,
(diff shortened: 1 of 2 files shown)


comit/src/orderpool.rs
```diff
@@ -509,6 +527,34 @@ mod tests {
...
+            order_with_safe_expiries.clone(),
+            order_with_unsafe_expiries,
+        ]);
+
+        let filtered_orders = pool
+            .safe_expiries()
+            .map(|(_, order)| order.id)
+            .collect::<Vec<OrderId>>();
+
+        assert_eq!(filtered_orders.len(), 1);
```
rishflab, line 554: I wanted to matching_does_not_contain here to be more explicity but could not figure out how to do it

comit/src/orderpool.rs
```diff
@@ -90,6 +91,23 @@ impl OrderPool {
             .flat_map(|(maker, orders)| iter::from_fn(move || Some(maker)).zip(orders.values()))
     }
 
+    pub fn safe_expiries(&self) -> impl Iterator<Item = (&PeerId, &BtcDaiOrder)> {
```
thomaseizinger, line 94: Hmm, can we do better with the naming here? The function reads like an accessor that returns you expiries but it actually returns orders where the expiries match!

I see two options:

1. Completely imply this check on our view on the orderbook. Going that way I would just add this check to the "all" iterator and hence we can just avoid the naming discussion.
2. Provide a dedicated accessor like here but change the name so that it is clear that we are returning orders here.

The downside of approach (1) is that we are putting this policy deep into the orderpool and it will get in the way once we want to give the user the option to also see orders that don't match their expiries.
We are not there yet but I guess we can take it as a hint that we might want to keep this out of the orderpool?

Re option (2): Maybe a name like "executable_orders" conveys better what we are returning here? A Rust-doc comment might also help :)

cnd/src/http_api/markets/get_btc_dai.rs
```diff
@@ -24,7 +24,7 @@ async fn handler(facade: Facade) -> Result<impl Reply> {
     let mut orders = siren::Entity::default();
     let local_peer_id = facade.swarm.local_peer_id();
 
-    for (maker, order) in facade.swarm.btc_dai_market().await {
+    for (maker, order) in facade.swarm.btc_dai_market_safe_expiries().await {
```
thomaseizinger, line 27: I think on this level I would probably just imply that all the orders are safe and just retain the name "btc_dai_market".

thomaseizinger (changes requested):

comit/src/orderpool.rs
rishflab: I'm in favour of option 2. I don't quite understand what you mean by approach 1 "putting this policy deep into the orderpool". Wouldn't approach 1 be moving the policy (return only executable orders) out of the orderpool?

thomaseizinger: If you provide an accessor on orderpool, the orderpool has knowledge about this policy of what an executable order is although that policy is actually something that only the caller is concerned about. Note that while the orderpool cares about the swap protocol for matching, it will happily publish and receive orders of with all kinds of swap protocols. We choose the offsets in the HTTP controller and hand the complete order over to the pool for publishing. One could argue that the choice of which orders are returned to the user should happen on the same level and hence not be part of the orderpool :)

rishflab: yep understood. Got confused by this line "The downside of approach (1) is that we are putting this policy deep into the orderpool". I think you meant approach(2). Ill move the filtering into the request handler

thomaseizinger: Yeah, filtering in the HTTP controller makes sense! It is slightly less efficient if we cannot compose the iterators in one go but currently, the orderpool is so small that it doesn't really matter :)

cnd/src/http_api/markets/get_btc_dai.rs
```diff
@@ -20,11 +20,20 @@ pub fn route(facade: Facade) -> impl Filter<Extract = impl Reply, Error = Reject
...
     let mut orders = siren::Entity::default();
     let local_peer_id = facade.swarm.local_peer_id();
 
-    for (maker, order) in facade.swarm.btc_dai_market().await {
+    let executable_orders = facade
+        .swarm
+        .btc_dai_market()
+        .await
+        .into_iter()
+        .filter(|(_, order)| has_executable_expiries(order));
```
D4nte, line 34:
```suggestion
        .filter(|(_, order)| has_viable_expiries(order));
```
I think you replaced the original `useful` terminology from the expiry module by `executable`. After a lot of back and forth, @tcharding and I settled on `viable` as the best descriptive for those expiries.

cnd/src/http_api/markets/get_btc_dai.rs
```diff
@@ -52,3 +74,64 @@ struct MarketItem {
...
+
+#[cfg(test)]
+mod tests {
+    use crate::http_api::markets::get_btc_dai::has_executable_expiries;
+    use comit::{asset, order::SwapProtocol, BtcDaiOrder, Position, Price, Quantity, Role};
+    use spectral::{assert_that, prelude::MappingIterAssertions};
+    use time::Duration;
+
+    #[test]
+    fn filter_out_orders_with_unexectubale_expiries() {
```
D4nte, line 86:
```suggestion
    fn filter_out_orders_with_inviable_expiries() {
```

cnd/src/http_api/markets/get_btc_dai.rs
rishflab: fixed

thomaseizinger (approved):

D4nte (approved):

PR #3129, opened 2020-09-04
Introduce `Price` and `Quantity` types

thomaseizinger:
These newtypes allow us to contain all logic around calculations
of different rates and quote prices in a way that enforces the
correct unit.

The only way to construct instances of `Price` is through a "loud"
API that should make clear to the developer, what unit the price is
in. Currently, the only supported one here is WEI/SAT.

To limit conversions even more, we move more of the types shared
among the different HTTP routes into the file `http_api.rs`. This
allows us to make the `Amount::btc` and `Amount::dai` functions private
and make sure the routes only use the provided `From` conversions.
These in turn help to maintain consistency among all the different
endpoints.


comit/src/order.rs
```diff
@@ -89,27 +84,93 @@ impl BtcDaiOrder {
...
-    /// The field `price` holds the value in WEI per SATOSHI, hence we need to
-    /// multiply it by 100_000_000 to get WEI per BTC.
-    pub fn price(&self, denom: Denomination) -> Erc20Quantity {
-        let price = self.price.clone();
-        match denom {
-            Denomination::WeiPerSat => price,
-            Denomination::WeiPerBtc => 100_000_000 * price,
+    /// This is how we store the data internally and hence we don't need to do
+    /// any conversions.
+    pub fn from_wei_per_sat(rate: Erc20Quantity) -> Self {
```
rishflab, line 135: I still think we should use a a u64 rather than u256. The issue is highly unlikely to arise but going with a u64 prevents us from creating values greater than what u256 supports when doing price * quantity calculations.

comit/src/order.rs
```diff
@@ -89,27 +84,93 @@ impl BtcDaiOrder {
             created_at: OffsetDateTime::now_utc(),
         }
     }
+}
+
+/// A newtype representing a quantity in a certain base currency B.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Quantity<B> {
+    value: B,
```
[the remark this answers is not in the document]
tcharding: BOOM! method below is called `to_inner`, surely this should be `inner` then?

comit/src/order.rs
```diff
@@ -89,27 +84,93 @@ impl BtcDaiOrder {
...
+}
+
+/// A newtype representing the price of one unit of the base currency B in the
+/// quote currency Q.
+///
+/// The core idea around of this type is to enforce the unit of the rate when
+/// doing calculations. We achieve that by adding "loud" constructors and
+/// accessors for combinations of base and quote currency.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Price<B, Q> {
```
tcharding, line 125: Perhaps re-arranging the order of the type parameters would make this type read easier. Since it is sats/wei it seems more logical to have `<Q, B>`.

comit/src/order.rs
tcharding: Max value of a u64:

((2**64) - 1) == 18446744073709551615

= 1.8 x 10e19

Price of 13000 DAI in wei/sat is 

= 13000 x 10e18 / 10e8
= 1.3 x 10e12

So it seems there is enough room in a u64.

I don't see any reason to use a u64 though, since this is DAI why use any type other than what we use for wei, if we did so it would need an explanation.

comit/src/order.rs
```diff
@@ -89,27 +84,93 @@ impl BtcDaiOrder {
...
-    pub fn price(&self, denom: Denomination) -> Erc20Quantity {
-        let price = self.price.clone();
-        match denom {
-            Denomination::WeiPerSat => price,
-            Denomination::WeiPerBtc => 100_000_000 * price,
+    /// This is how we store the data internally and hence we don't need to do
+    /// any conversions.
+    pub fn from_wei_per_sat(rate: Erc20Quantity) -> Self {
+        Price {
+            value: rate,
```
tcharding, line 137: Can this please be call `price`, or `inner`, or `p` for the same reasons as above - no need to have a second name for the same thing here :)

comit/src/order.rs
thomaseizinger: That would limit us to a price of 1.8 billion DAI/BTC. Surely Bitcoin is gonna crack that in the next couple of months. /s

I am not sure to be honest, what is everyone else thinking? @comit-network/comit-devs

comit/src/order.rs
```diff
@@ -89,27 +84,93 @@ impl BtcDaiOrder {
...
+
+    pub fn wei_per_sat(&self) -> Erc20Quantity {
+        self.value.clone()
+    }
+
+    pub fn wei_per_btc(&self) -> Erc20Quantity {
+        self.value
+            .clone()
+            .checked_mul(100_000_000)
+            .expect("the price of bitcoin to not go through the roof")
```
tcharding, line 150: lol, love this message.

tcharding (approved):

comit/src/asset/ethereum/erc20.rs
```diff
@@ -37,32 +39,32 @@ impl Erc20Quantity {
         self.0.to_bytes_le()
     }
 
-    #[cfg(test)]
-    pub fn meaningless_test_value() -> Self {
-        Erc20Quantity::from_wei(1_000u32)
+    pub fn checked_mul(self, factor: u64) -> Option<Self> {
```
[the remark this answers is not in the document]
thomaseizinger: Well, an Erc20Quantity has a max value: U256!

comit/src/order.rs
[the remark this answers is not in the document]
thomaseizinger: Well, I purposely wanted to keep it generic so that it doesn't stutter. Happy to rename it to inner if you like that better :)

comit/src/order.rs
thomaseizinger: So the price is actually wei/sat but I thought that base is more important and hence put that one first. When talking about forex, base + quote seems to be the more natural ordering, no?

comit/src/asset/ethereum/erc20.rs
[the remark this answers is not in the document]
tcharding: Oh that's were it comes from. Even though we use BigUint not U256 for our Erc20Quantity.

comit/src/order.rs
tcharding: Funny you should mention it because a bunch of places in code use the field order 
```
quantity: Foo
price: Bar
```
and it has been bothering me, did that order get chosen because it matches base/quote. I would call the 'natural' order of those two fields price/quantity because A) price is more important B) p comes before q alphabetically C) That's how I say them

Obviously point C does not carry that much weight.

Back to the point though, in code `Price<asset::Bitcoin, Erc20Quantity>` reads uncomfortably for me where as `Price<Erc20Quantity, asset::Bitcoin>` reads better because Erc20Quantity comes first and that's the unit of price we use. Its not a super important though.

thomaseizinger: I am not too fussed about the ordering to be honest so I will not object if someone chooses to reorder them but I am too lazy now to be honest :sweat_smile:

PR #3130, opened 2020-09-04
Utilize diesel's `deserialize_as` feature of Queryable

thomaseizinger:
This allows us directly use our desired type within all structs
that derive `Queryable` and are used to load data from the DB.
As a result, we get less usage of `.0` once the data is loaded
and for the `BtcDaiOrder` in particular, we can directly go into
a `Price` and `Quantity`!

### DIFF
--- a/cnd/src/storage/db/wrapper_types.rs
+++ b/cnd/src/storage/db/wrapper_types.rs
@@ -1,7 +1,11 @@
 use crate::asset;
 use std::{fmt, str::FromStr};
 
-pub mod custom_sql_types;
+mod text;
+mod u32;
+
+pub use self::{text::Text, u32::U32};
+use comit::{asset::Erc20Quantity, Price, Quantity};
 
 /// A wrapper type for representing satoshis
 ///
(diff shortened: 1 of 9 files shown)


cnd/src/storage/db/wrapper_types.rs
```diff
@@ -95,3 +105,26 @@ impl fmt::Display for Erc20Amount {
...
+}
+
+impl fmt::Display for WeiPerSat {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0.to_wei_dec())
+    }
+}
+
+impl From<Text<WeiPerSat>> for Price<asset::Bitcoin, Erc20Quantity> {
+    fn from(rate: Text<WeiPerSat>) -> Self {
```
tcharding, line 127: My current bug bear; can we not introduce multiple terms for the same thing please. We are creating a price from some input value, why use 'rate'? Perhaps 'price' or 'p' or even 'value' if you don't want to use price.

tcharding (approved):
Nice trick!

da-kami (approved):

PR #3131, opened 2020-09-04
Remove Facade in favor of accessing Storage, Connectors & Swarm directly

thomaseizinger:


tcharding (approved):
I looked at doing this work myself last week so I could tell what you were doing and can guess at more or less why but I would have liked to see some sort of explanation for this. Also there were at least 4 things going on in this PR

- Change const value of 'type length limit' - no idea why this was done
- Pass timestamps instead of connectors to function
- Convert storage/connectors conceptually into facades
- Remove facade and use new storage/connector facades

I'm keen to review comit-rs code that patches areas I'm not working on but to do so with the hope of adding any real value please add some context. Thanks :)

thomaseizinger:
> I looked at doing this work myself last week so I could tell what you were doing and can guess at more or less why but I would have liked to see some sort of explanation for this. Also there were at least 4 things going on in this PR
> 
>     * Change const value of 'type length limit' - no idea why this was done
> 
>     * Pass timestamps instead of connectors to function
> 
>     * Convert storage/connectors conceptually into facades
> 
>     * Remove facade and use new storage/connector facades
> 
> 
> I'm keen to review comit-rs code that patches areas I'm not working on but to do so with the hope of adding any real value please add some context. Thanks :)

Thanks for the feedback @tcharding !

We seem to have a different perception of what "enough context" is :)
I agree that the type length limit could have been a separate commit. However, [you actually approved the PR that introduced it](https://github.com/comit-network/comit-rs/pull/3106). Rust 1.46 had a regression here and we will need to bump this number to keep things compiling if the compiler complains as we make changes.

Regarding the remaining changes, the strategy I followed was: I am gonna make a patch that does what I want to achieve, i.e. remove `Facade`. The remaining diff in the patch was motivated by that but I only discovered that as I went along. To understand your feedback correctly: After all the changes that made it compiling + working, I should have gone back and commit it as individual patches?

tcharding:
> Regarding the remaining changes, the strategy I followed was: I am gonna make a patch that does what I want to achieve, i.e. remove Facade. The remaining diff in the patch was motivated by that but I only discovered that as I went along. To understand your feedback correctly: After all the changes that made it compiling + working, I should have gone back and commit it as individual patches?

I wouldn't expect one to go back and split this, if the work was exploratory like this I wouldn't split it myself. I would hope that I noticed that I was changing all sorts of stuff and then write a bit of a description as to what got changed and _why_, the why being the most important. Otherwise reviewers have two choices 1) think hard themselves 2) just hit ack/nack. Neither of these are good solutions if we want to optimise in any way for review, right?

We already discussed this on Riot though ...

cnd/src/hbit.rs
```diff
@@ -28,7 +28,7 @@ pub async fn new<C>(
     role: Role,
     side: Side,
     states: Arc<States>,
-    connector: Arc<C>,
+    connector: impl AsRef<C>,
```
D4nte, line 31: 👍

cnd/src/herc20.rs
```diff
@@ -30,7 +30,7 @@ pub async fn new<C>(
     role: Role,
     side: Side,
     states: Arc<States>,
-    connector: Arc<C>,
+    connector: impl AsRef<C>,
```
D4nte, line 33: Yes, less `Arc` and more `impl As Ref` in the codebase would be good actually. It would make the APIs more stable I think.

D4nte (approved):
Nice PR

D4nte:
bors retry

cnd/src/herc20.rs
thomaseizinger: I think so too yes, only depend on what you need :)

thomaseizinger:
bors retry

PR #3139, opened 2020-09-08
Polish the e2e test framework

thomaseizinger:
Given recent code churn and the inlining of the SDK into the API tests, I saw great potential for simplifying and clarifying a lot of our e2e test code.

I hope the diff is reviewable if you go through it patch-by-patch :)

Enjoy!

Note: I took care that at least tsc & lint pass at each commit, but I hope to also not have broken the e2e tests in between :)


api_tests/src/actors/balance_asserter.ts
```diff
@@ -0,0 +1,194 @@
...
+        private readonly tokenContract: string
+    ) {}
+
+    public async assertSwapped() {
+        const currentBalance = await this.wallet.getErc20Balance(
+            this.tokenContract,
+            0
+        );
+        const expectedBalance = this.startingBalance + this.expectedDiff;
+
```
tcharding, line 58: As I read this I wondered, because of the name, if this checked the balance of both wallets i.e., that funds have gone from one and arrived in the other (for only this actor). (Obviously this is impossible since we only have a handle to one wallet.) Since this is not the case and we only check one wallet, would `assertRedeemed` not be a better name?

tcharding (approved):
Nice PR man, it was a pleasure to review :)

I saw some lines of code added that were debug code commented out, when viewing the diff in the terminal but it doesn't show up in the web interface, just mentioning it in case you meant to remove it.

api_tests/src/actors/balance_asserter.ts
thomaseizinger: Good point, I will change the function name :)

api_tests/src/actors/balance_asserter.ts
```diff
@@ -0,0 +1,194 @@
...
+export class Erc20BalanceAsserter implements BalanceAsserter {
+    public static async forFunder(
+        wallet: EthereumWallet,
+        swapAmount: bigint,
+        tokenContract: string
+    ) {
+        await wallet.mintErc20(swapAmount, tokenContract); // As funder, we need money to spend
+
+        const startingBalance = await wallet.getErc20Balance(tokenContract, 0);
```
da-kami, line 19: What is the `0` that is passed as second parameter? Maybe we can be more explicit.

da-kami (approved):
Nice work!

api_tests/src/actors/cnd_actor.ts
```diff
@@ -565,48 +477,14 @@ export class CndActor {
...
-            this.logger.debug(
-                "Starting %s balance: ",
-                asset.name,
-                balance.toString()
-            );
-            this.startingBalances.set(
-                toKey(asset),
-                BigInt(balance.toString(10))
-            );
+            // this.logger.debug(
```
D4nte, line 480: Some comments to maybe delete?

D4nte (approved):
Looks good. Do we have a dep on `BigNumber`? If yes, are we able to remove it in favour of `bigint`?

api_tests/src/actors/balance_asserter.ts
thomaseizinger: That is the number of decimal points of the the contract. I think we can remove that now, good catch!

thomaseizinger:
I had to push a few more commits to stablise the e2e tests, hopefully it works this time :)

thomaseizinger:
Argh, the stupid Ethereum dev account is causing problems ...

rishflab (approved):

thomaseizinger:
Hey @D4nte, @da-kami, @tcharding !

I finally made the tests pass on this one! A couple of things:

All tests pass until [`552bab1` (#3139)](https://github.com/comit-network/comit-rs/pull/3139/commits/552bab18f953e8df14e610ec4dc0701b10bd7470). That one breaks the LN tests because we share a single LN channel among all tests. That is fixed in the last commit: [`268cab3` (#3139)](https://github.com/comit-network/comit-rs/pull/3139/commits/268cab33bf539fb138e7dd977f5b33455fd3378e)

Please see the commit message for further details :)

D4nte (approved):

PR #3140, opened 2020-09-08
Order state uses quantity instead of percentage

da-kami:
This has the advantage that we don't have to calculate percentage.
Makes our life easier on the UI and in cnd.
The fact that the fields are duplicated with every new trading pair is a bit annoying but acceptable. (And cannot be avoided if we go for quantities, because of the specific types.)

BtcDaiOrder Response from http API: 

```json
{
  "id": "00000000-0000-0000-0000-000000000000",
  "position": "sell",
  "price": {
    "currency": "DAI",
    "value": "9100000000000000000000",
    "decimals": 18
  },
  "quantity": {
    "currency": "BTC",
    "value": "10000000",
    "decimals": 8
  },
  "state": {
    "open": {
      "currency": "BTC",
      "value": "3000000",
      "decimals": 8
    },
    "closed": {
      "currency": "BTC",
      "value": "1000000",
      "decimals": 8
    },
    "settling": {
      "currency": "BTC",
      "value": "0",
      "decimals": 8
    },
    "failed": {
      "currency": "BTC",
      "value": "6000000",
      "decimals": 8
    },
    "cancelled": {
      "currency": "BTC",
      "value": "0",
      "decimals": 8
    }
  }
}
```

It's a bit redundant, because currency and decimals are implied by the quantity. But since we are dealing with amounts I think we should handle it like all other amounts.

### DIFF
--- a/cnd/src/http_api.rs
+++ b/cnd/src/http_api.rs
@@ -32,7 +32,7 @@ use crate::{
 };
 use anyhow::Result;
 use chrono::Utc;
-use comit::{order::btc_zero, OrderId, Position, Price, Quantity};
+use comit::{OrderId, Position, Price, Quantity};
 use libp2p::{Multiaddr, PeerId};
 use serde::{Deserialize, Serialize};
 use warp::http::Method;
@@ -180,7 +180,7 @@ fn make_order_entity(properties: OrderProperties) -> Result<siren::Entity> {
 }
 
 fn cancel_action(order: &OrderProperties) -> Option<siren::Action> {
-    if order.state.open != Amount::btc(btc_zero().to_inner()) {
+    if order.state.open != Amount::btc(asset::Bitcoin::ZERO) {
         Some(siren::Action {
             name: "cancel".to_string(),
             class: vec![],
--- a/cnd/src/http_api.rs
+++ b/cnd/src/http_api.rs
@@ -93,11 +93,11 @@ impl From<(tables::Order, tables::BtcDaiOrder)> for OrderProperties {
             price: Amount::from(btc_dai_order.price),
             quantity: Amount::from(btc_dai_order.quantity),
             state: State::new(
-                Amount::from(btc_dai_order.open),
-                Amount::from(btc_dai_order.closed),
-                Amount::from(btc_dai_order.settling),
-                Amount::from(btc_dai_order.failed),
-                Amount::from(btc_dai_order.cancelled),
+                btc_dai_order.open.to_inner(),
+                btc_dai_order.closed.to_inner(),
+                btc_dai_order.settling.to_inner(),
+                btc_dai_order.failed.to_inner(),
+                btc_dai_order.cancelled.to_inner(),
             ),
         }
     }
@@ -143,21 +143,26 @@ impl Amount {
 
 #[derive(Serialize)]
 struct State {
-    open: Amount,
-    closed: Amount,
-    settling: Amount,
-    failed: Amount,
-    cancelled: Amount,
+    #[serde(with = "asset::bitcoin::sats_as_string")]
+    open: asset::Bitcoin,
+    #[serde(with = "asset::bitcoin::sats_as_string")]
+    closed: asset::Bitcoin,
+    #[serde(with = "asset::bitcoin::sats_as_string")]
+    settling: asset::Bitcoin,
+    #[serde(with = "asset::bitcoin::sats_as_string")]
+    failed: asset::Bitcoin,
+    #[serde(with = "asset::bitcoin::sats_as_string")]
+    cancelled: asset::Bitcoin,
 }
 
 impl State {
     #[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)] // we only store positive values in the DB ranging from 0 - 100
     fn new(
-        open: Amount,
-        closed: Amount,
-        settling: Amount,
-        failed: Amount,
-        cancelled: Amount,
+        open: asset::Bitcoin,
+        closed: asset::Bitcoin,
+        settling: asset::Bitcoin,
+        failed: asset::Bitcoin,
+        cancelled: asset::Bitcoin,
     ) -> Self {
         Self {
             open,
@@ -180,7 +185,7 @@ fn make_order_entity(properties: OrderProperties) -> Result<siren::Entity> {
 }
 
 fn cancel_action(order: &OrderProperties) -> Option<siren::Action> {
-    if order.state.open != Amount::btc(asset::Bitcoin::ZERO) {
+    if order.state.open != asset::Bitcoin::ZERO {
         Some(siren::Action {
             name: "cancel".to_string(),
             class: vec![],
@@ -517,11 +522,11 @@ mod tests {
             price: Amount::dai(Erc20Quantity::from_wei_dec_str("9100000000000000000000").unwrap()),
             quantity: Amount::btc(Bitcoin::from_sat(10000000)),
             state: State {
-                open: Amount::btc(Bitcoin::from_sat(3000000)),
-                closed: Amount::btc(Bitcoin::from_sat(1000000)),
-                settling: Amount::btc(Bitcoin::from_sat(0)),
-                failed: Amount::btc(Bitcoin::from_sat(6000000)),
-                cancelled: Amount::btc(Bitcoin::from_sat(0)),
+                open: Bitcoin::from_sat(3000000),
+                closed: Bitcoin::from_sat(1000000),
+                settling: Bitcoin::from_sat(0),
+                failed: Bitcoin::from_sat(6000000),
+                cancelled: Bitcoin::from_sat(0),
             },
         };
 
@@ -543,31 +548,11 @@ mod tests {
     "decimals": 8
   },
   "state": {
-    "open": {
-      "currency": "BTC",
-      "value": "3000000",
-      "decimals": 8
-    },
-    "closed": {
-      "currency": "BTC",
-      "value": "1000000",
-      "decimals": 8
-    },
-    "settling": {
-      "currency": "BTC",
-      "value": "0",
-      "decimals": 8
-    },
-    "failed": {
-      "currency": "BTC",
-      "value": "6000000",
-      "decimals": 8
-    },
-    "cancelled": {
-      "currency": "BTC",
-      "value": "0",
-      "decimals": 8
-    }
+    "open": "3000000",
+    "closed": "1000000",
+    "settling": "0",
+    "failed": "6000000",
+    "cancelled": "0"
   }
 }"#
         );
--- a/cnd/src/http_api.rs
+++ b/cnd/src/http_api.rs
@@ -92,13 +92,13 @@ impl From<(tables::Order, tables::BtcDaiOrder)> for OrderProperties {
             position: order.position,
             price: Amount::from(btc_dai_order.price),
             quantity: Amount::from(btc_dai_order.quantity),
-            state: State::new(
-                btc_dai_order.open.to_inner(),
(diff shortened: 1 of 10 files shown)


cnd/src/storage/db/errors.rs
```diff
@@ -54,4 +54,4 @@ pub struct NoRefundIdentity;
 
 #[derive(thiserror::Error, Debug, Clone, Copy)]
 #[error("Order {0} is no longer open and can therefore not be cancelled")]
-pub struct NotOpen(pub OrderId);
+pub struct NotOpen(pub i32);
```
thomaseizinger, line 57: This should stay as the order id, the PK will not be very useful in error messages.

comit/src/order.rs
```diff
@@ -115,6 +115,12 @@ pub fn btc(btc: f64) -> Quantity<Bitcoin> {
     ))
 }
 
+pub fn btc_zero() -> Quantity<Bitcoin> {
+    Quantity::new(Bitcoin::from_sat(
+        bitcoin::Amount::from_btc(0.0).unwrap().as_sat(),
```
thomaseizinger, line 120: `bitcoin::Amount` has a ZERO constant.

thomaseizinger (changes requested):
Thanks for tackling this @da-kami !

Please change the response to this as discussed:

```json
{
  "id": "00000000-0000-0000-0000-000000000000",
  "position": "sell",
  "price": {
    "currency": "DAI",
    "value": "9100000000000000000000",
    "decimals": 18
  },
  "quantity": {
    "currency": "BTC",
    "decimals": 8,
    "value": "10000000",
  },
  "state": {
    "open": "3000000",
    "closed": "1000000",
    "settling": "0",
    "failed": "6000000",
    "cancelled": "0"
  }
}
```

cnd/src/http_api.rs
```diff
@@ -93,17 +93,17 @@ impl From<(tables::Order, tables::BtcDaiOrder)> for OrderProperties {
...
-                order.open,
-                order.closed,
-                order.settling,
-                order.failed,
-                order.cancelled,
+                Amount::from(btc_dai_order.open),
+                Amount::from(btc_dai_order.closed),
+                Amount::from(btc_dai_order.settling),
+                Amount::from(btc_dai_order.failed),
+                Amount::from(btc_dai_order.cancelled),
```
D4nte, line 100:
```suggestion
                btc_dai_order.open.into(),
                btc_dai_order.closed.into(),
                btc_dai_order.settling.into(),
                btc_dai_order.failed.into(),
                btc_dai_order.cancelled.into(),
```

cnd/src/http_api.rs
```diff
@@ -189,7 +180,7 @@ fn make_order_entity(properties: OrderProperties) -> Result<siren::Entity> {
 }
 
 fn cancel_action(order: &OrderProperties) -> Option<siren::Action> {
-    if order.state.open > 0 {
+    if order.state.open != Amount::btc(btc_zero().to_inner()) {
```
D4nte, line 183:
```suggestion
    if order.state.open != Amount::btc::ZERO {
```
If it does not exist yet, create this const :)

D4nte: Also, a helper method `State::is_open` may make more sense here instead of exposing the internals. If such helper was existing this line would not have changed ;)

cnd/src/http_api.rs
```diff
@@ -526,11 +517,11 @@ mod tests {
             price: Amount::dai(Erc20Quantity::from_wei_dec_str("9100000000000000000000").unwrap()),
             quantity: Amount::btc(Bitcoin::from_sat(10000000)),
             state: State {
-                open: 30,
-                closed: 10,
-                settling: 0,
-                failed: 60,
-                cancelled: 0,
+                open: Amount::btc(Bitcoin::from_sat(3000000)),
```
D4nte, line 520:
```suggestion
                open: Amount::btc::from_sat(3000000),
```
Surely, if such function does not exists, we need to create it.

cnd/src/http_api.rs
```diff
@@ -143,41 +143,32 @@ impl Amount {
...
     #[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)] // we only store positive values in the DB ranging from 0 - 100
-    fn new(open: i32, closed: i32, settling: i32, failed: i32, cancelled: i32) -> Self {
+    fn new(
+        open: Amount,
+        closed: Amount,
+        settling: Amount,
+        failed: Amount,
+        cancelled: Amount,
+    ) -> Self {
         Self {
```
D4nte, line 162: `State::new` is only called once in `http_api`. You don't need it, you can directly construct the State using the fields. Like it's done here: https://github.com/comit-network/comit-rs/blob/9d3157a0c088266c70057f0c2c7bc30ceb5d1772/cnd/src/http_api.rs#L519

cnd/src/http_api.rs
```diff
@@ -143,41 +143,32 @@ impl Amount {
...
-    cancelled: u8,
+    open: Amount,
+    closed: Amount,
+    settling: Amount,
+    failed: Amount,
+    cancelled: Amount,
 }
 
 impl State {
     #[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)] // we only store positive values in the DB ranging from 0 - 100
```
D4nte, line 154: 📣 
```suggestion
```

cnd/src/http_api.rs
```diff
@@ -93,17 +93,17 @@ impl From<(tables::Order, tables::BtcDaiOrder)> for OrderProperties {
             price: Amount::from(btc_dai_order.price),
             quantity: Amount::from(btc_dai_order.quantity),
             state: State::new(
-                order.open,
-                order.closed,
-                order.settling,
-                order.failed,
-                order.cancelled,
+                Amount::from(btc_dai_order.open),
```
D4nte, line 96: How is `OrderProperties` only ever instantiated for the DB? How is it saved to the DB in the first place?

cnd/src/http_api/orders/make_btc_dai.rs
```diff
@@ -95,14 +98,24 @@ fn save_order(
...
         let price = order.price.wei_per_sat();
 
-        move |order_fk| InsertableBtcDaiOrder::new(order_fk, quantity, price)
+        move |order_fk| {
+            InsertableBtcDaiOrder::new(
+                order_fk,
+                quantity,
+                price,
+                quantity,
+                btc_zero().to_inner(),
```
D4nte, line 113: `btc_zero()` should be a method `zero()` generically implemented on `Quantity`.

cnd/src/storage/db/tables.rs
```diff
@@ -500,6 +447,47 @@ impl BtcDaiOrder {
...
     }
+
+    /// Move the amount that is settling from open to settling.
+    ///
+    /// Whilst we don't have partial order matching, this simply means updating
+    /// `settling` to the amount of `open` and updating `open` to `0`.
+    ///
+    /// Once we implement partial order matching, this will need to get more
+    /// sophisticated.
+    pub fn mark_as_settling(&self, conn: &SqliteConnection) -> Result<()> {
```
D4nte, line 458:
```suggestion
    pub fn set_to_settling(&self, conn: &SqliteConnection) -> Result<()> {
```
`mark` is not a terminology usually used. `set` is more common to Rust.

cnd/src/storage/db/tables.rs
```diff
@@ -500,6 +447,47 @@ impl BtcDaiOrder {
...
+            .execute(conn)?;
+
+        if affected_rows == 0 {
+            anyhow::bail!("failed to mark order {} as settling", self.order_id)
+        }
+
+        Ok(())
+    }
+
+    pub fn cancel(&self, conn: &SqliteConnection) -> Result<()> {
```
D4nte, line 473:
```suggestion
    pub fn set_to_cancelled(&self, conn: &SqliteConnection) -> Result<()> {
```
It is best for the future dev to have some name consistency between method that act similarly.

cnd/src/storage/db/tables.rs
```diff
@@ -508,14 +496,34 @@ pub struct InsertableBtcDaiOrder {
...
+        order_fk: i32,
+        quantity: asset::Bitcoin,
+        price: Erc20Quantity,
+        open: asset::Bitcoin,
+        closed: asset::Bitcoin,
+        settling: asset::Bitcoin,
+        failed: asset::Bitcoin,
+        cancelled: asset::Bitcoin,
+    ) -> Self {
         Self {
```
D4nte, line 518: 📣 a check should be done to ensure that the quantities make sense (the sum of the state quantities equals the `quantity`).

cnd/src/storage/db/tables.rs
```diff
@@ -508,14 +496,34 @@ pub struct InsertableBtcDaiOrder {
     pub order_id: i32,
     pub quantity: Text<Satoshis>,
     pub price: Text<Erc20Amount>,
+    pub open: Text<Satoshis>,
```
D4nte, line 499: 📣 Because of the relation between `quantity` and the state quantities` I would suggest to use getters and checked setters instead.

D4nte (changes requested):
Blocking comment is the lack of checks on the consistency of the quantity.

cnd/src/storage/db/tables.rs
da-kami: Instead I moved the initialisation into the constructor; `open` is set to `quantity`, all other states are initialized with `ZERO`. That way the initial state is ensured.

cnd/src/storage/db/tables.rs
da-kami: I started doing the checked setters, but after a quick discussion on how to achieve this in the best way with @thomaseizinger we decided to go for a simpler solution for now. 

Since the functions that manipulate the state (`set_to_settling`, `set_to_cancelled`) are in full control of the values, it should be fine with the tests for these functions in place. The tests ensure, that the complete amount is moved correctly for the current logic that does not include partial filling yet. 
Even when adding partial filling, the functions can apply these checks and the tests can be adapted accordingly.

cnd/src/http_api.rs
da-kami: This check concerned the validity of the percentages and is not relevant any more.

D4nte: I guessed :)

cnd/src/http_api.rs
D4nte: @da-kami can you answer the question? I am curious :)

cnd/src/storage/db/tables.rs
D4nte: What I am trying to prevent is a future dev trying to change this values without using the setter.

da-kami: Note that testing the current logic is covered by `orderbook_dry.ts`. I think that is good enough for now. If the tests are not exhaustive enough please ping me.

D4nte (approved):

cnd/src/storage/db/tables.rs
```diff
@@ -508,14 +504,25 @@ pub struct InsertableBtcDaiOrder {
...
     pub price: Text<Erc20Amount>,
+    open: Text<Satoshis>,
+    closed: Text<Satoshis>,
+    settling: Text<Satoshis>,
+    failed: Text<Satoshis>,
+    cancelled: Text<Satoshis>,
 }
 
 impl InsertableBtcDaiOrder {
+    #[allow(clippy::too_many_arguments)]
```
thomaseizinger, line 515: Probably no longer needed :)

thomaseizinger (approved):
Thanks @da-kami !

PR #3152, opened 2020-09-11
Move Business logic out of network behaviour

D4nte:


nectar/src/network.rs
```diff
@@ -99,15 +93,15 @@ pub struct SetupSwapContext {
 /// A `NetworkBehaviour` that delegates to the `Orderbook` and `SetupSwap`
 /// behaviours.
 #[derive(NetworkBehaviour)]
-#[behaviour(out_event = "Event", poll_method = "poll")]
+#[behaviour(out_event = "orderbook::BehaviourOutEvent", poll_method = "poll")]
```
thomaseizinger, line 96: Use `event_process = false` :)

https://docs.rs/libp2p/0.28.1/libp2p/swarm/trait.NetworkBehaviour.html#deriving-networkbehaviour

D4nte: I don't think it behaves as you expect. `swarm.next()` ends up returning `()`.

thomaseizinger: Here is how it should behave: https://github.com/libp2p/rust-libp2p/blob/9647557f0d81a37f1c2062638aa10eba50d91631/misc/core-derive/tests/test.rs#L267-L308

Is this not what is happening?

D4nte: I'll look in more details but I am not yet sure how it is better than the status quo.

thomaseizinger: You should be get rid of your "events" field and the dedicated "poll" function :)

nectar/src/network.rs
```diff
@@ -59,7 +65,7 @@ pub struct SetupSwapContext {
 /// A `NetworkBehaviour` that delegates to the `Orderbook` and `SetupSwap`
 /// behaviours.
 #[derive(NetworkBehaviour)]
-#[behaviour(out_event = "BehaviourOutEvent", poll_method = "poll")]
+#[behaviour(out_event = "BehaviourOutEvent", event_process = false)]
```
thomaseizinger, line 68: :partying_face:

nectar/src/network.rs
```diff
@@ -90,19 +96,6 @@ impl Nectar {
...
-        if let Some(event) = self.events.pop_front() {
-            return Poll::Ready(NetworkBehaviourAction::GenerateEvent(event));
-        }
-
-        // We trust libp2p to poll us.
-        Poll::Pending
-    }
 }
 
 impl libp2p::swarm::NetworkBehaviourEventProcess<orderbook::BehaviourOutEvent> for Nectar {
```
thomaseizinger, line 101: These can go away now :)

nectar/src/command/trade/event_loop.rs
```diff
@@ -87,97 +87,95 @@ impl EventLoop {
                     }
                 },
                 event = self.swarm.next().fuse() => {
-                    self.handle_network_event(
-                            event,
-                        ).await?;
+                    let _ = self.handle_network_event(
+                        event,
+                    ).await.map_err(|err| tracing::error!("Network event handling failed: {:#}", err));
```
thomaseizinger, line 75: This use of the `map_err` combinator is a bit odd because you are causing a side-effect (logging). Combinators are usually (pure) data pipelines but in this case your are just throwing away the result.

An alternative version of this could be:

```suggestion
                    if let Err(e) = self.handle_network_event(event).await {
                    	tracing::error!("Network event handling failed: {:#}", err));
                    }
```

nectar/src/command/trade/event_loop.rs
```diff
@@ -87,97 +87,95 @@ impl EventLoop {
...
-                tracing::error!(
-                    "Unable to fetch latest rate! Fetching rate yielded error: {}",
-                    e
-                );
-            }
+    fn handle_rate_update(&mut self, new_rate: MidMarketRate) -> Result<()> {
+        let publish_order = self
+            .maker
+            .update_rate(new_rate)
+            .context("Unable to fetch latest rate! Fetching rate yielded error")?;
```
thomaseizinger, line 118: I think there is a copy-pasta error here! That error used to be emitted if an `Err` was passed to the function but that now happens outside.

thomaseizinger: I think the same applies for all other functions as well.

nectar/src/command/trade/event_loop.rs
```diff
@@ -426,7 +426,7 @@ impl EventLoop {
                 let bitcoin_transient_sk = self
                     .bitcoin_wallet
                     .derive_transient_sk(index)
-                    .map_err(|err| anyhow!("Could not derive Bitcoin transient key: {:?}", err))?;
+                    .map_err(|err| anyhow!("Could not derive Bitcoin transient key: {:#}", err))?;
```
thomaseizinger, line 409: This could/should be a `.context` because you are already acting on an error.

nectar/src/config/serde/bitcoin_amount.rs
```diff
@@ -31,7 +31,7 @@ where
                 Some(value) => {
                     let amount = Amount::from_btc(value).map_err(|error| {
                         serde::de::Error::custom(format!(
-                            "Could not deserialize bitcoin amount: {:?}",
+                            "Could not deserialize bitcoin amount: {:#}",
                             error
                         ))
                     })?;
```
thomaseizinger, line 37: This could just be `Amount::from_btc(value).map_err(Error::custom)?`, not sure if we need the extra string in there.

nectar/src/seed.rs
```diff
@@ -62,7 +62,7 @@ impl fmt::Debug for Seed {
 
 impl fmt::Display for Seed {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{:?}", self)
+        write!(f, "{:#}", self)
```
thomaseizinger, line 65: This looks like it is an unintended change :)

nectar/src/command/trade.rs
```diff
@@ -239,40 +238,60 @@ fn init_dai_balance_updates(
...
+}
+
+impl SwapExecutor {
+    async fn run(
+        mut self,
+        db: Arc<Database>,
+        bitcoin_wallet: Arc<bitcoin::Wallet>,
+        ethereum_wallet: Arc<ethereum::Wallet>,
+        bitcoin_connector: Arc<comit::btsieve::bitcoin::BitcoindConnector>,
+        ethereum_connector: Arc<comit::btsieve::ethereum::Web3Connector>,
```
thomaseizinger: Could the executor possibly own these?

nectar/src/command/trade/event_loop.rs
```diff
@@ -21,13 +21,10 @@ use comit::{
...
 };
-use futures::{
-    channel::mpsc::{Receiver, Sender},
-    FutureExt, StreamExt, TryFutureExt,
-};
+use futures::{channel::mpsc::Receiver, FutureExt, StreamExt, TryFutureExt};
 use std::sync::Arc;
 
-pub struct EventLoop {
+pub(super) struct EventLoop {
```
thomaseizinger, line 27: Why do you need `pub(super)`? You could just make the struct `pub`, the module private and not re-export it from the parent module. That should achieve the same thing without special visibility modifiers :)

thomaseizinger (approved):
Thanks for doing this cleanup!

Left a couple of comments here and there. Action at your liking :)

nectar/src/network.rs
thomaseizinger: + the `events` field as well but I couldn't comment on it :D

nectar/src/command/trade/event_loop.rs
D4nte: Interesting, I previously suggested the exact opposite to @luckysori because it was more compact :)

nectar/src/network.rs
D4nte: Ok but how can I be sure it still works afterwards as the network is not tested? I guess your e2e tests will tell me.

D4nte: I just want to be clear that I blindly following your suggestion here :)

thomaseizinger: > Ok but how can I be sure it still works afterwards as the network is not tested? I guess your e2e tests will tell me.

The libp2p project contains a test for this behaviour :)

nectar/src/command/trade/event_loop.rs
```diff
@@ -87,97 +87,95 @@ impl EventLoop {
...
+                new_rate = self.rate_update_receiver.next().fuse() => {
+                    if let Some(new_rate) = new_rate {
+                        match new_rate {
+                            Ok(new_rate) => {
+                                let _ = self.handle_rate_update(new_rate)
+                                    .map_err(|err| tracing::error!("Rate update handling failed: {:#}", err));
+                            }
+                            Err(err) => tracing::error!("Rate retrieval failed: {:#}", err),
+                        }
+                    }
```
thomaseizinger, line 86: You can combine this to:

```suggestion
if let Some(Ok(new_rate)) = new_rate {
	let _ = self.handle_rate_update(new_rate).map_err(|err| tracing::error!("Rate update handling failed: {:#}", err));
}
```

If you want to keep the error logging, use `match` and match on `Some(Err(e))`.

nectar/src/command/trade/event_loop.rs
thomaseizinger: Yeah I saw this pattern a number of times in nectar and always found it a bit irritating to be honest because we are not really mapping the error so it feels like an abuse of the combinator :/

I agree though that the alternative is also not really nice. Not sure how to solve this problem in a nice way :D

thomaseizinger (commented):
Spotted another possible improvement :)

nectar/src/seed.rs
D4nte: Thanks, sorry about that, I did try to check but I clearly missed one.

nectar/src/command/trade.rs
D4nte: yes I moved the connectors, I guess I did the commit after your review.

nectar/src/command/trade/event_loop.rs
D4nte: Hum, I'll double check what I was doing but rustc wasn't happy as `TaskExecutor` is part of the signature but it is private to `trade` module.

nectar/src/command/trade/event_loop.rs
D4nte: I would try to avoid getting irritated when reading code 😉

nectar/src/command/trade/event_loop.rs
D4nte: As discussed, we cannot go around this one.

nectar/src/command/trade/event_loop.rs
D4nte: and introducing a `map_err` that only has a side-effect with the data being dropped? 😱

nectar/src/command/trade/event_loop.rs
```diff
@@ -211,17 +198,33 @@ impl EventLoop {
...
             .remove_swap(&swap_id)
             .await
-            .map_err(|error| tracing::error!("Unable to delete swap from db: {}", error));
+            .context("Unable to delete swap from db");
+
+        if let Err(history_err) = history_res {
+            Err(history_err.context(format!(
+                "Issue inserting in history; peer db: {:?}; swap db: {:?}",
+                peer_db_res, swap_db_res
+            )))
```
thomaseizinger, line 200: Nit, if you use `format!` with `context`, you might want to get in the habit of using `with_context` to avoid the allocation if it is not actually executed (i.e. in the Ok) case :)

thomaseizinger (approved):

D4nte:
Alright, all done, feel free to have a last look @thomaseizinger but it should be good now.

bors r+

D4nte:
@comit-network/comit-devs Do note this. Instead of acting on a network event in the network trait implementation, we can do that when pulling on the stream. It has the following advantages:
- Better handling of error cases: errors can be returned, logging does not need to happen inside the network.
- Allows `async` call when processing the network event
- Keeps the swarm less `busy` (see all the fields removed from it).
- The point above allows for less indirection

Cheers.

PR #3153, opened 2020-09-11
Modularize `tables` module into one per database table

thomaseizinger:
This patch doesn't change any functionality but just moves code
around.

### DIFF
--- a/cnd/src/network/swarm.rs
+++ b/cnd/src/network/swarm.rs
@@ -284,11 +284,11 @@ async fn handle_new_match(
     let (order_hbit, order_herc20) = storage
         .db
         .do_in_transaction(|conn| {
-            use crate::storage::tables;
+            use crate::storage::*;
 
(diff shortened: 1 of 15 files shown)


cnd/src/network/swarm.rs
```diff
@@ -284,11 +284,11 @@ async fn handle_new_match(
     let (order_hbit, order_herc20) = storage
         .db
         .do_in_transaction(|conn| {
-            use crate::storage::tables;
+            pub use crate::storage::*;
```
D4nte, line 287: How is a `pub` keyword useful in a closure's block?

D4nte (approved):

cnd/src/network/swarm.rs
thomaseizinger: It is not :sweat_smile:

tcharding (approved):

PR #3167, opened 2020-09-14
Bump dependencies

thomaseizinger:
This should hopefully sort out the security alert on GitHub.
We also slightly limit our dependency footprint by only opting into
the features we need.


thomaseizinger:
bors merge

thomaseizinger:
bors merge

PR #3168, opened 2020-09-14
Allow user to specify desired swap protocol role through config

rishflab:
The role that nectar was previously hard coded to be bob. This PR adds an optional role field to the config file that allows the user to specify a role. If a role is not specified, Bob is chosen as a default to prevent a user unexpectedly losing fees from failed swaps.

### DIFF
--- a/nectar/sample-config.toml
+++ b/nectar/sample-config.toml
@@ -7,6 +7,7 @@
 [maker]
 # The spread to apply to the mid-market when publish an offer. It's a pyrimiad format, 12.34 = 12.34% spread.
 spread = 500
+role = "Bob"
 
 [maker.max_sell]
 # The maximum amount of bitcoin to sell in one order, optional field.
@@ -19,7 +20,7 @@ dai = 1000
 [maker.maximum_possible_fee]
 # An estimation of the maximum fee that we would expect to pay, used to ensure we always have enough
 # balance to execute an order we publish.
-bitcoin = 0.00009275 
+bitcoin = 0.00009275
 
 [network]
 # The libp2p socket on which nectar listens for COMIT messages.
--- a/nectar/src/command/trade.rs
+++ b/nectar/src/command/trade.rs
@@ -22,7 +22,7 @@ use crate::{
     maker::TakeRequestDecision,
     network::{new_swarm, ActivePeer, SetupSwapContext},
 };
-use comit::{Position, Role};
+use comit::Position;
 use std::{sync::Arc, time::Duration};
 
 const ENSURED_CONSUME_ZERO_BUFFER: usize = 0;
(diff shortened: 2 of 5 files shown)


nectar/sample-config.toml
```diff
@@ -19,7 +20,7 @@ dai = 1000
 [maker.maximum_possible_fee]
 # An estimation of the maximum fee that we would expect to pay, used to ensure we always have enough
 # balance to execute an order we publish.
-bitcoin = 0.00009275 
+bitcoin = 0.00009275
```
rishflab, line 23: seems to come back when I run make format

thomaseizinger (commented):
Is this really the only change we have to make?

If we want this feature (I am not fully convinced) then I think stuff needs to change here as well: https://github.com/comit-network/comit-rs/blob/b0c43253d9095a2a686a337a086bf7bcf072768d/nectar/src/network.rs#L254-L365

Currently, we are always defaulting to `Bob` here.

I know that @D4nte is also touching the network code at the moment so there might be bigger merge conflicts coming up if we change that now.

D4nte:
Can you please justify this change? We do not want/need this feature.

rishflab:
> Can you please justify this change? We do not want/need this feature.

> If we want this feature (I am not fully convinced)


I was thinking that the ability to choose risk vs order desirability would be something a maker might want to be able to configure. Given that nectar is a for bootstrapping the network, and we don't want to potentially lose money (by being Alice), maybe it is not that useful feature to have. Happy to close this after further consideration.

D4nte:
> and we don't want to potentially lose money (by being Alice)

Yes, absolutely, that is why nectar is always Bob and the taker/ambrosia has to move first.

If we were to implement the possibility to change roles, I'd see it as something more dynamic. e.g., posting offers as both Alice and Bob but putting a premium if we are Alice and also a lower cap.

For me the next step in this debate is to allow ambrosia to dynamically be either roles to allow transparent ambrosia to ambrosia trades, before considering allowing nectar to be alice.

thomaseizinger:
> > and we don't want to potentially lose money (by being Alice)
> 
> Yes, absolutely, that is why nectar is always Bob and the taker/ambrosia has to move first.
> 
> If we were to implement the possibility to change roles, I'd see it as something more dynamic. e.g., posting offers as both Alice and Bob but putting a premium if we are Alice and also a lower cap.
> 
> For me the next step in this debate is to allow ambrosia to dynamically be either roles to allow transparent ambrosia to ambrosia trades, before considering allowing nectar to be alice.

Given that, and that this PR already has conflicts, I believe closing it for now makes the most sense? We can always revisit at a later stage.

D4nte:
Thanks @rishflab for this contribution but it is currently not in line with our goals. Closing now.
Feel free to butt in the feast project if anything does not seem right to you.

D4nte:
Thanks @rishflab for this contribution but it is currently not in line with our goals. Closing now.
Feel free to butt in the feast project if anything does not seem right to you.

PR #3183, opened 2020-09-18
Cleanup implementation of Bitcoin amounts

thomaseizinger:


comit/src/network/swap_digest.rs
```diff
@@ -88,6 +88,10 @@ pub struct HbitHerc20 {
     pub token_contract: identity::Ethereum,
 }
 
+/// A new-type that allows us to implemented `ToDigestInput` on foreign types.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Digestable<T>(pub T);
```
D4nte, line 93: Does the inner need to be pub? Maybe better to implement `From`.

nectar/src/order.rs
```diff
@@ -66,7 +66,7 @@ impl BtcDaiOrderForm {
 
         Ok(BtcDaiOrderForm {
             position: Position::Sell,
-            quantity: base_amount.into(),
+            quantity: Quantity::new(base_amount),
```
D4nte, line 69: Is it a sign that `From` implementation should be done?

nectar/src/order.rs
```diff
@@ -225,7 +225,7 @@ mod tests {
         )
         .unwrap();
 
-        assert_eq!(bitcoin::Amount::from(order.quantity), btc(8.0));
+        assert_eq!(order.quantity.to_inner(), btc(8.0));
```
D4nte, line 228: Is it a sign that From implementation should be done?

D4nte (approved):
Noice! 👏

comit/src/network/swap_digest.rs
thomaseizinger: This struct doesn't enforce any invariants among its inner field so I don't see a reason why it should be private. Adding a `From` impl is just more code for no benefit IMO.

nectar/src/order.rs
thomaseizinger: I am not sure to be honest.

I am not abstracting over the construction of `Quantity` anywhere so I don't see anything wrong with a using an explicit constructor. For a type like `Quantity` that enforces invariants, I actually prefer it being explicit so it stands out among all the other `.into()` calls. Can't really present any better argument though :)

tcharding: I'm with @D4nte on this one, although I don't feel strongly, I prefer the from/into version than new/to_inner.

tcharding (approved):

PR #3184, opened 2020-09-18
Two fixes in nectar's network code

thomaseizinger:


nectar/src/command/trade/event_loop.rs
```diff
@@ -222,99 +222,33 @@ impl EventLoop {
...
-                            },
-                            secret_hash: exec_swap.hbit.secret_hash,
-                            start_of_swap,
-                            swap_id,
-                            taker: ActivePeer {
-                                peer_id: exec_swap.peer_id,
-                            },
-                        })
-                    }
+                let swap_kind = match exec_swap.swap_protocol {
```
D4nte, line 225: The `SwapParams` struct looks exactly the same so replace with:

```
let swap_params = ...;

let swap_kind = match exec_swap.swap_protocol {
	..::HbitHerc20 => SwapKind::HbitHerc20(swap_params),
	..
};
```

D4nte (approved):

nectar/src/command/trade/event_loop.rs
thomaseizinger: Oh, I didn't even realize! Good catch!

tcharding (approved):

PR #3186, opened 2020-09-20
[cnd] add peer bootstrap configuration feature

D4nte:
Add a new config option to pass an array of multiaddress to which the cnd node will try to connect when starting.

Went for addresses only because because at this stage the identification of the peers (via peer id) does not really matter. We just want to connect to another node to get order and execute trustless atomic swaps.

### DIFF
--- a/api_tests/src/actor_test.ts
+++ b/api_tests/src/actor_test.ts
@@ -186,7 +186,10 @@ async function newCndActor(role: Role, startCnd: boolean) {
         ethereum: await ethereumWallet,
         lightning: newLightningStubChannel(), // Lightning channels are initialized lazily
     });
-    await cndStarting;
+
+    if (cndStarting !== undefined) {
+        await cndStarting;
+    }
 
     const lndClient = (() => {
         switch (role) {
--- a/api_tests/src/actor_test.ts
+++ b/api_tests/src/actor_test.ts
@@ -26,17 +26,31 @@ declare var global: HarnessGlobal;
 export function startAlice(
     testFn: (alice: CndActor) => Promise<void>
 ): ProvidesCallback {
-    return cndActorTest(["Alice"], async ([alice]) => testFn(alice));
+    return startCndActorTest(["Alice"], async ([alice]) => testFn(alice));
 }
 
 /**
  * Instantiates two CndActors, the first one in the role of Alice and the second one in the role of Bob.
+ * Starts respective cnd instances.
  * @param testFn
  */
 export function startAliceAndBob(
     testFn: ([alice, bob]: CndActor[]) => Promise<void>
 ): ProvidesCallback {
-    return cndActorTest(["Alice", "Bob"], async ([alice, bob]) =>
+    return startCndActorTest(["Alice", "Bob"], async ([alice, bob]) =>
+        testFn([alice, bob])
+    );
+}
+
+/**
+ * Instantiates two CndActors, the first one in the role of Alice and the second one in the role of Bob.
+ * Does not start the cnd instances.
+ * @param testFn
+ */
+export function createAliceAndBob(
+    testFn: ([alice, bob]: CndActor[]) => Promise<void>
+): ProvidesCallback {
+    return createCndActorTest(["Alice", "Bob"], async ([alice, bob]) =>
         testFn([alice, bob])
     );
 }
@@ -50,23 +64,27 @@ export function startAliceAndBob(
 export function startConnectedAliceAndBob(
     testFn: ([alice, bob]: CndActor[]) => Promise<void>
 ): ProvidesCallback {
-    return cndActorTest(["Alice", "Bob"], async ([alice, bob]) => {
+    return startCndActorTest(["Alice", "Bob"], async ([alice, bob]) => {
         await alice.connect(bob);
         return testFn([alice, bob]);
     });
 }
 
-/*
+/**
  * Instantiates a set of CndActors with the given roles, executes the provided test function and tears the actors down again.
  *
  * This can be used to set up an arbitrary number of nodes by passing any combination of "Alice" or "Bob" within the `roles` array. For example: `cndActorTest(["Alice", "Alice", "Alice", "Bob"], ...)` will give you four nodes, with the first three being in the role of Alice and the fourth one in the role of Bob.
  */
-export function cndActorTest(
+export function startCndActorTest(
     roles: Role[],
     testFn: (actors: CndActor[]) => Promise<void>
 ): ProvidesCallback {
     return async (done) => {
-        const actors = await Promise.all(roles.map(newCndActor));
+        const actors = await Promise.all(
+            roles.map(async (role: Role) => {
+                return newCndActor(role, true);
+            })
+        );
         global
             .getLogger(["test_environment"])
             .info("All actors created, running test");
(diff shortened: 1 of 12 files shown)


cnd/src/config.rs
```diff
@@ -372,21 +381,46 @@ mod tests {
         let file_contents = vec![
             r#"
             listen = ["/ip4/0.0.0.0/tcp/9939"]
+            peers = [ { id = "QmfUfpC2frwFvcDzpspnfZitHt5wct6n4kpG5jzgRdsxkY", address_hints = [ "/ip4/1.1.1.1/tcp/9939" ] } ]
```
thomaseizinger, line 384: This structure is quite verbose.

Do we need the ability to provide multiple addresses for the same peer?
Do we even need to provide the PeerId? Asked differently, do we care about the validation that exactly the peer with this ID is behind the provided address?

Libp2p actually has a way of encoding the peerId of an address in a multi-address: https://docs.libp2p.io/concepts/peer-id/#peerids-in-multiaddrs

Maybe we don't even have to parse that manually and can just hand it off to libp2p?

cnd/src/http_api/markets/get_btc_dai.rs
```diff
@@ -1,12 +1,13 @@
 use crate::{
-    http_api::{problem, serde_peer_id, Amount},
+    http_api::{problem, Amount},
     network::Swarm,
+    serde,
 };
+use ::serde::Serialize;
```
thomaseizinger, line 6: Should we re-export everything from the `serde` crate from the `serde` module so we can avoid this?

cnd/src/network/peer_tracker.rs
```diff
@@ -9,24 +9,63 @@ use libp2p::{
...
 
 impl PeerTracker {
     pub fn connected_peers(&self) -> impl Iterator<Item = (PeerId, Vec<Multiaddr>)> {
         self.connected_peers.clone().into_iter()
     }
 
-    pub fn add_address_hint(&mut self, id: PeerId, addr: Multiaddr) -> Option<Multiaddr> {
-        self.address_hints.insert(id, addr)
+    /// Adds an address hint for the given peer id. The added address is
+    /// considered most recent and hence is added at the start of the list.
```
thomaseizinger, line 29: Can we encode this information in the function name? Like "add_recent_address"?

cnd/src/network/peer_tracker.rs
```diff
@@ -9,24 +9,63 @@ use libp2p::{
...
+        let old_addresses = self.address_hints.get_mut(&id);
+
+        match old_addresses {
+            None => {
+                let mut hints = VecDeque::new();
+                for _ in 0..addresses.len() {
+                    // Need to keep same order than the vec
+                    let addr = addresses.pop().expect("We loop on the length");
+                    hints.push_front(addr);
+                }
```
thomaseizinger, line 58: Does this do the trick as well?

```suggestion
                hints.extend(addresses);
```

cnd/src/network/peer_tracker.rs
```diff
@@ -9,24 +9,63 @@ use libp2p::{
...
+            Some(hints) => {
+                // Put the new address at the start of the list.
+                hints.push_front(addr);
+            }
+        }
+    }
+
+    /// Adds address hints for the given peer id. The added addresses are
+    /// considered most recent and hence are added at the start of the list.
+    pub fn add_address_hints(&mut self, id: PeerId, mut addresses: Vec<Multiaddr>) {
```
thomaseizinger, line 48: Instead of providing two functions we could make the caller always create a `Vec`. Is relatively concise with `vec![addr]`.

Could also think about accepting anything that is `IntoIterator` so we don't hardcode against a `Vec`. That allow callers to also pass a slice for example.

cnd/src/network/swarm.rs
```diff
@@ -83,6 +83,17 @@ impl Swarm {
                 .with_context(|| format!("Address is not supported: {:?}", addr))?;
         }
 
+        for peer in &settings.network.peers {
+            tracing::info!("Dialing peer {:#?} from config", peer.id);
```
thomaseizinger, line 87: Why alternate-debug? `Display` will be prettier I think :)

cnd/src/network/swarm.rs
```diff
@@ -152,6 +157,30 @@ impl Swarm {
...
 
+    pub async fn dial_peer(
+        &self,
+        peer_id: PeerId,
+        hint_addresses: Vec<Multiaddr>,
+    ) -> anyhow::Result<()> {
+        let mut swarm = self.inner.lock().await;
+
+        Self::dial_peer_on_swarm(&mut *swarm, peer_id, hint_addresses).await
+    }
```
thomaseizinger, line 168: Is this one actually used? Could remove it and rename the other function to `dial_peer` if not :)

thomaseizinger (commented):
Some initial feedback :)

Mostly nits apart from the config serialization!

cnd/src/config.rs
D4nte: We only need the ip for our purpose so I'll simplify the config file :]

api_tests/src/actor_test.ts
```diff
@@ -113,7 +170,12 @@ async function newCndActor(role: Role) {
         logger,
         finalConfig
     );
-    const cndStarting = cndInstance.start();
+
+    let cndStarting;
+
+    if (startCnd) {
+        cndStarting = cndInstance.start();
+    }
```
thomaseizinger, line 178: `CndActor` actually has a `start` function that starts the instance. Instead of making this configurable here, we could just remove the functionality from this fn and do it outside. Thoughts?

It seems like we are awaiting an undefined value on line 189 now in case `startCnd` is false. Not sure if that does any harm?

api_tests/tests/sanity.ts
```diff
@@ -152,4 +160,31 @@ describe("Sanity", () => {
...
+
+            const configOverride = {
+                network: {
+                    peer_addresses: aliceAddresses,
+                },
+            };
+
+            const currentConfig = bob.cndInstance.getConfigFile();
+
+            const updatedConfig = merge(configOverride, currentConfig);
```
thomaseizinger, line 181: This only works if the currentConfig doesn't already specify the `peer_addresses` field. I'd flip it so you an be sure that your overrides always apply.
```suggestion
            const updatedConfig = merge(currentConfig, configOverride);
```

api_tests/tests/sanity.ts
```diff
@@ -152,4 +160,31 @@ describe("Sanity", () => {
...
+
+            bob.cndInstance.setConfigFile(updatedConfig);
+
+            await bob.cndInstance.start();
+
+            await bob.pollUntilConnectedTo(aliceId);
```
thomaseizinger, line 187: Can remove some newlines to group related functionality :)

```suggestion
            await alice.cndInstance.start();

            const aliceAddresses = await alice.cnd.getPeerListenAddresses();
            const configOverride = {
                network: {
                    peer_addresses: aliceAddresses,
                },
            };

            const currentConfig = bob.cndInstance.getConfigFile();
            const updatedConfig = merge(configOverride, currentConfig);

            bob.cndInstance.setConfigFile(updatedConfig);
            await bob.cndInstance.start();

            const aliceId = await alice.cnd.getPeerId();
            await bob.pollUntilConnectedTo(aliceId);
```

thomaseizinger (approved):

cnd/src/network/peer_tracker.rs
```diff
@@ -9,24 +9,38 @@ use libp2p::{
...
+        let old_addresses = self.address_hints.get_mut(&id);
+
+        match old_addresses {
+            None => {
+                let mut hints = VecDeque::new();
+                hints.push_back(addr);
+                self.address_hints.insert(id, hints);
+            }
+            Some(hints) => {
+                // Put the new address at the start of the list.
```
tcharding, line 40: This comment states the obvious, perhaps better would be to describe how libp2p trys the addresses from the front of the list first.

tcharding: lol, you make this comment right below :)

tcharding (approved):

api_tests/src/actor_test.ts
D4nte: > CndActor actually has a start function that starts the instance. Instead of making this configurable here, we could just remove the functionality from this fn and do it outside. Thoughts?

I agree that the current code design is not great, however, with the current design we start the blockchain nodes and cnd in parallel.
If we were to move it outside, we would start the blockchain nodes together and then cnd.
For the sake of performance I'd be keen to keep this property. Not sure I can do that if I move it outside.

> It seems like we are awaiting an undefined value on line 189 now in case startCnd is false. Not sure if that does any harm?

I tried and was wondering what would happen. It seemed to work as linux CI tests pass but not mac os. I'll investigate but it does not seem related.

api_tests/tests/sanity.ts
D4nte: well spotted

cnd/src/network/peer_tracker.rs
D4nte: I sorted it.

PR #3192, opened 2020-09-21
Add an e2e test between nectar and cnd to the e2e test suite

thomaseizinger:
The actual test is in the last commit, everything prior to that is refactorings and/or fixes that I made on the way to make the test possible.


thomaseizinger:
Splitting this up into smaller PRs :D

PR #3195, opened 2020-09-21
Refactor nectar swap execution

thomaseizinger:
This patch serious refactors nectar's swap execution to more generally deal with scenarios in which we want to refund.

We also add logging so we can see what is going on.

Lastly, we refactor `SwapExecutor#execute` to no longer be an async function to avoid accidential blocking of the event-loop.

FYI: @luckysori

### DIFF
--- a/nectar/src/main.rs
+++ b/nectar/src/main.rs
@@ -13,7 +13,7 @@
 #![cfg_attr(not(test), warn(clippy::unwrap_used))]
 #![forbid(unsafe_code)]
 #![recursion_limit = "512"]
-#![type_length_limit = "1944624"]
+#![type_length_limit = "7743004"]
 
 mod bitcoin;
 mod command;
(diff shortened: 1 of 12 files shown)


nectar/src/main.rs
```diff
@@ -13,7 +13,7 @@
 #![cfg_attr(not(test), warn(clippy::unwrap_used))]
 #![forbid(unsafe_code)]
 #![recursion_limit = "512"]
-#![type_length_limit = "1944624"]
+#![type_length_limit = "7743004"]
```
D4nte, line 16: 💥

nectar/src/swap/comit.rs
```diff
@@ -8,3 +8,18 @@ mod herc20_hbit;
 pub use comit::{ethereum, *};
 pub use hbit_herc20::{hbit_herc20_alice, hbit_herc20_bob};
 pub use herc20_hbit::{herc20_hbit_alice, herc20_hbit_bob};
+
+use std::fmt::Debug;
```
D4nte, line 12: Nectar is in `comit-rs`. If that "could be upstreamed" why not upstreaming it directly? Feels very limbo-ish.

D4nte (approved):

nectar/src/swap/comit.rs
thomaseizinger: Yeah, I am planning to do that at some point, just didn't want to add even more noise to these PRs :D

nectar/src/main.rs
thomaseizinger: Yeah, don't know what is going on here :(

PR #3196, opened 2020-09-21
Improve logging

thomaseizinger:
This series of patches improves logging in nectar and the orderbook.

### DIFF
--- a/comit/src/order.rs
+++ b/comit/src/order.rs
@@ -42,7 +42,7 @@ impl From<Uuid> for OrderId {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct BtcDaiOrder {
     pub id: OrderId,
     pub position: Position,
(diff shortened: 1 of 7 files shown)


comit/src/order.rs
```diff
@@ -87,7 +87,7 @@ impl BtcDaiOrder {
 }
 
 /// A newtype representing a quantity in a certain base currency B.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
D4nte, line 90: Where is the `Hash` trait used?

comit/src/orderpool.rs
```diff
@@ -22,6 +23,22 @@ pub struct OrderPool {
     ///
     /// Allows us to filter out our own orders.
     me: PeerId,
+
+    /// A cache for storing which orders don't match.
```
D4nte, line 27: Can you explain more the functionality you are trying to achieve?

If an order does not match at a given point in time, it may match if the user adds a new order in their orderbook.

D4nte (commented):
Please explain the caching strategy

comit/src/order.rs
thomaseizinger: It is used for computing the key of the LruCache because the LruCache is basically a bounded hashmap.

comit/src/orderpool.rs
thomaseizinger: Which is why the key that is stored is a tuple of the two orders that don't match.

The idea was, if two orders don't match, I can store that for later and avoid checking them again in the next round. The primary motivation was reducing the log-spam of the same two orders constantly not matching over and over again.

D4nte: thanks!

D4nte (approved):

PR #3197, opened 2020-09-21
Make sure we always put the HTLC on out-index 0

thomaseizinger:
The implementation is not particularly pretty at the moment but at least it is contained :)

I opted for using PSBT's because it allows us to configure, where the change-output should be. I believe this is superior to taking the serialized transaction and "finding" the output again. Could be convinced otherwise though :)

### DIFF
--- a/nectar/src/bitcoin/bitcoind.rs
+++ b/nectar/src/bitcoin/bitcoind.rs
@@ -4,6 +4,7 @@ use crate::{
 };
 use ::bitcoin::{consensus::encode::serialize_hex, hashes::hex::FromHex, Transaction, Txid};
 use anyhow::Context;
+use bitcoin::OutPoint;
 use serde::Deserialize;
 
 pub const JSONRPC_VERSION: &str = "1.0";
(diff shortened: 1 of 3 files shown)


nectar/src/bitcoin/bitcoind.rs
```diff
@@ -188,6 +189,98 @@ impl Client {
...
+        amount: Amount,
+    ) -> anyhow::Result<OutPoint> {
+        let address = address.to_string();
+
+        let response: CreatePsbtResponse = self
+            .rpc_client
+            .send_with_path(
+                format!("/wallet/{}", wallet_name),
+                jsonrpc::Request::new(
+                    "walletcreatefundedpsbt",
```
D4nte, line 205: Remind me of thor code... time to have a dedicated bitcoin rpc lib 🙃

D4nte (approved):

nectar/src/bitcoin/bitcoind.rs
thomaseizinger: Once I have my lib ready, I was thinking of proposing to the `rust-bitcoin` lib to add the RPC stuff to it under a feature flag. Given that my lib is client-agnostic, the only dependency that would be added is `syn` but also that would be feature-flagged.

That would allow us/them to keep the types in the interface up to date and it would avoid situations in which we can no longer update rust-bitcoin because it would become incompatible with their rpc lib.

thomaseizinger: Unless we are just better at maintaining our bitcoin rpc lib, there isn't really an advantage to it as long as it is two different libraries.

thomaseizinger:
Kicking this in so we can make progress :)

bors r+

PR #3199, opened 2020-09-21
Add an e2e test between nectar and cnd

thomaseizinger:
Enjoy!

### DIFF
--- a/nectar/src/command/trade.rs
+++ b/nectar/src/command/trade.rs
@@ -64,7 +64,7 @@ pub async fn trade(
     let update_interval = Duration::from_secs(15u64);
 
     let (rate_future, rate_update_receiver) =
-        init_rate_updates(update_interval, settings.maker.kraken_api_host);
+        init_rate_updates(Duration::from_secs(5 * 60), settings.maker.kraken_api_host);
     let (btc_balance_future, btc_balance_update_receiver) =
         init_bitcoin_balance_updates(update_interval, Arc::clone(&bitcoin_wallet));
     let (dai_balance_future, dai_balance_update_receiver) =
--- a/tests/src/actors/nectar_actor.ts
+++ b/tests/src/actors/nectar_actor.ts
@@ -1,0 +1,74 @@
+import { DumpState, GetListenAddress, GetPeerId, Stoppable } from "./index";
+import { Logger } from "log4js";
+import { Balances, NectarInstance } from "../environment/nectar_instance";
+
+export default class NectarActor
+    implements Stoppable, DumpState, GetPeerId, GetListenAddress {
+    private startingBalances: Balances;
+
+    constructor(
+        public readonly logger: Logger,
+        public readonly nectarInstance: NectarInstance,
+        private readonly peerId: string
+    ) {
+        logger.info(
+            "Created new nectar actor with config",
+            nectarInstance.config
+        );
+    }
+
+    async dumpState(): Promise<void> {
+        return Promise.resolve(undefined);
+    }
+
+    async stop(): Promise<void> {
+        return this.nectarInstance.stop();
+    }
+
+    async getListenAddress(): Promise<string> {
+        // Very dirty hack that assumes we are listening on an IPv4 interface :crossed_fingers:
+        const address = this.nectarInstance.config.network.listen[0].replace(
+            "0.0.0.0",
+            "127.0.0.1"
+        );
+
+        return Promise.resolve(address);
+    }
+
+    async getPeerId(): Promise<string> {
+        return Promise.resolve(this.peerId);
+    }
+
+    public async saveBalancesBeforeSwap() {
+        this.startingBalances = await this.nectarInstance.balance();
+    }
+
+    public async assertBalancesChangedBy(expectedDiff: Partial<Balances>) {
+        const newBalances = await this.nectarInstance.balance();
+
+        if (expectedDiff.bitcoin) {
+            const expectedBalance =
+                this.startingBalances.bitcoin + expectedDiff.bitcoin;
+
+            expect(newBalances.bitcoin.toString()).toEqual(
+                expectedBalance.toString()
+            );
+        }
+        if (expectedDiff.ether) {
+            const expectedBalance =
+                this.startingBalances.ether + expectedDiff.ether;
+
+            expect(newBalances.ether.toString()).toEqual(
+                expectedBalance.toString()
+            );
+        }
+        if (expectedDiff.dai) {
+            const expectedBalance =
+                this.startingBalances.dai + expectedDiff.dai;
+
+            expect(newBalances.dai.toString()).toEqual(
+                expectedBalance.toString()
+            );
+        }
+    }
+}
--- a/nectar/src/network.rs
+++ b/nectar/src/network.rs
@@ -24,7 +24,7 @@ pub fn new_swarm(seed: Seed, settings: &crate::config::Settings) -> anyhow::Resu
 
     let transport = transport::build_transport(local_key_pair)?;
 
-    let mut swarm = libp2p::swarm::SwarmBuilder::new(transport, behaviour, local_peer_id)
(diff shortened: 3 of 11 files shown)


e2e_tests/package.json
```diff
@@ -5,7 +5,7 @@
     "main": "index.js",
     "scripts": {
         "check": "tsc && prettier --check '**/*.{ts,json,yml}' && tslint --project .",
-        "pretest": "cargo build --bin cnd && tsc",
+        "pretest": "cargo build --bin cnd && cargo build --bin nectar && tsc",
```
D4nte, line 8: Back to my other comment, this shows that we need to separate the tests.
One folder for cnd api
One folder for full e2e

If I am changing something on the http api of cnd, I don't want to compile nectar. The CI can do it but I don't want to be forced locally.

D4nte (commented):
We need to split the tests. Happy if you do it after we merge but it's going to be a PITA to compile nectar even when it's not needed.

e2e_tests/src/environment/nectar_instance.ts
```diff
@@ -0,0 +1,174 @@
...
+        return parseBalanceOutput(stdout);
+    }
+
+    /*
+     * Executes nectar's `trade` command.
+     *
+     * Returns the PeerId under which this nectar instance participates in the COMIT network.
+     */
+    public async trade(): Promise<string> {
+        const bin = process.env.NECTAR_BIN
```
D4nte, line 77: You are doing that few times, you may want to extract to a function

e2e_tests/tests/cnd_and_nectar.ts
```diff
@@ -0,0 +1,35 @@
...
+
+        await alice.assertAndExecuteNextAction("deploy");
+        await alice.assertAndExecuteNextAction("fund");
+        await alice.assertAndExecuteNextAction("redeem");
+
+        // Wait until the wallets sees the new balance.
+        await sleep(2000);
+
+        await alice.assertBalancesAfterSwap();
+        await bob.assertBalancesChangedBy({
```
D4nte, line 30: 🎉

nectar/src/command/trade.rs
```diff
@@ -69,7 +69,7 @@ pub async fn trade(
     let update_interval = Duration::from_secs(15u64);
 
     let (rate_future, rate_update_receiver) =
-        init_rate_updates(update_interval, settings.maker.kraken_api_host);
+        init_rate_updates(Duration::from_secs(5 * 60), settings.maker.kraken_api_host);
```
D4nte, line 72: Maybe it's time to put those durations in `const`s.

e2e_tests/src/environment/nectar_instance.ts
```diff
@@ -0,0 +1,174 @@
...
+import { parseFixed } from "@ethersproject/bignumber";
+
+export class NectarInstance {
+    private process: ChildProcess;
+
+    constructor(
+        private readonly cargoTargetDirectory: string,
+        private readonly logFile: string,
+        private readonly logger: Logger,
+        private readonly _config: NectarConfig
```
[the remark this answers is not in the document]
thomaseizinger: To avoid ambiguity with the getter that is named `config`.

e2e_tests/src/environment/nectar_instance.ts
thomaseizinger: Actually, I don't think we are using this functionality at all. Might as well remove it :D

thomaseizinger: Actually, I can't find it in the CI config. Are we no longer testing against the release build?

thomaseizinger:
> it's going to be a PITA to compile nectar even when it's not needed.

We seem to have very different definitions of what a PITA is :D

I completely agree that it is unnecessary to compile in some cases and can be optimized. However, I don't think it warrants splitting the tests into two directories. That is the wrong solution from my perspective. There is no intrinsic need why we should have two directories for e2e tests (which would imply two `node_modules`, two package.json, two dependabot configs, etc).

If you care strongly about optimizing this, I would suggest to remove the invocation of cargo from `package.json` and instead call `cargo build` from within the e2e test suite, for example from within `CndInstance#start` and `NectarInstance#start`. Any place where we know that we are going to need cnd/nectar to be built will suffice though.

That would also solve the problem of potentially forgetting to recompile cnd/nectar if the test is run from the IDE and not through `yarn`.

nectar/src/command/trade.rs
thomaseizinger: I don't mind honestly. I believe we are going to be tweaking those a lot so we may even want to put them into the config :)

e2e_tests/src/environment/nectar_instance.ts
D4nte: Testing against the release build is done as part of the pre-release CI as far as I remember.

D4nte:
> > it's going to be a PITA to compile nectar even when it's not needed.
> 
> We seem to have very different definitions of what a PITA is :D

It is a pain in the arse to have to wait for an unnecessary compilation when trying to run tests locally.

> If you care strongly about optimizing this, I would suggest to remove the invocation of cargo from `package.json` and instead call `cargo build` from within the e2e test suite, for example from within `CndInstance#start` and `NectarInstance#start`. Any place where we know that we are going to need cnd/nectar to be built will suffice though.

Lazily build each binary within the test suite. I like it.

D4nte:
The fact that you are renaming `api_tests` to `e2e_tests` because you are adding an integration test between nectar and cnd still makes me think that the best descriptor for this folder is simply `tests`.

It is not less an `api` test than before you make the change.

thomaseizinger:
> > > it's going to be a PITA to compile nectar even when it's not needed.
> > 
> > 
> > We seem to have very different definitions of what a PITA is :D
> 
> It is a pain in the arse to have to wait for an unnecessary compilation when trying to run tests locally.

Oh, I totally got why you think it is a pain in the arse :)
My comment was meant to point out that _how_ painful it actually is is a subjective matter. As a counter example, I find that having to maintain all the old /swaps endpoint to be a lot more painful than waiting for the compilation to finish.

> 
> > If you care strongly about optimizing this, I would suggest to remove the invocation of cargo from `package.json` and instead call `cargo build` from within the e2e test suite, for example from within `CndInstance#start` and `NectarInstance#start`. Any place where we know that we are going to need cnd/nectar to be built will suffice though.
> 
> Lazily build each binary within the test suite. I like it.

Happy you like the suggestion! Are you ok if we do that in a follow-up PR?

thomaseizinger:
> The fact that you are renaming `api_tests` to `e2e_tests` because you are adding an integration test between nectar and cnd still makes me think that the best descriptor for this folder is simply `tests`.
> 
> It is not less an `api` test than before you made the change.

I agree. I actually think it should have been named `e2e_tests` from the very beginning given that the makefile command is called `e2e` and even the `package.name` in the `package.json` already said `e2e_tests` before my change.

We can name it simply `tests` if you feel strongly. Might be a bit of surprise for newcomers to the project as such directory would usually contain Rust integration tests for the current crate. In our case, that happens to not exist because all our crates are workspace members and we don't have a root crate. Hence we don't have `src` and `tests` directories.

e2e_tests/src/environment/nectar_instance.ts
thomaseizinger: That is what I was thinking but I couldn't find that anywhere ...

tests/src/actors/nectar_actor.ts
```diff
@@ -0,0 +1,74 @@
...
+    async dumpState(): Promise<void> {
+        return Promise.resolve(undefined);
+    }
+
+    async stop(): Promise<void> {
+        return this.nectarInstance.stop();
+    }
+
+    async getListenAddress(): Promise<string> {
+        // Very dirty hack that assumes we are listening on an IPv4 interface :crossed_fingers:
```
D4nte, line 29: Note THAT dirty 😜

D4nte (approved):
🎉🎉🎉

PR #3204, opened 2020-09-22
Remove sha2 dependency

D4nte:
We don't really need it :)


thomaseizinger (approved):

tcharding (approved):

tcharding:
Can I ask a purely software engineering question @D4nte  and @thomaseizinger please. What is the benefit of doing this i.e., using a module that is re-exported by a dependency we have rather than depending on the library directly?

thomaseizinger:
> Can I ask a purely software engineering question @D4nte  and @thomaseizinger please. What is the benefit of doing this i.e., using a module that is re-exported by a dependency we have rather than depending on the library directly?
> 
> 

It is not the same dependency in this case, bitcoin_hashes has its own impl of sha2. Hence we can use that instead of adding another dependency.

D4nte:
> Can I ask a purely software engineering question @D4nte and @thomaseizinger please. What is the benefit of doing this i.e., using a module that is re-exported by a dependency we have rather than depending on the library directly?

If it were the same dependency, we would still save having to handle 2 dep and 2 versions. If we actually use both deps together then it is a PITA to upgrade them at the same time.

PR #3213, opened 2020-09-24
Remove Nectar wallet rescan because not needed in production

da-kami:
@D4nte has pointed out that `rescan` might be counterproductive. This was added before to help us with the dev-blockchain environment.

In prod the address to be funded is only known after wallet creation.
The rescan was convenient for the dev-environment, but Nectar should not be responsible for rescanning. For now, the user has to take care of running `rescan` for the dev-env.


D4nte (approved):
LGTM

thomaseizinger (approved):

PR #3216, opened 2020-09-24
Add test where cnd sells and nectar buys Bitcoin

thomaseizinger:


tests/src/actor_test.ts
```diff
@@ -90,8 +90,11 @@ export function startConnectedCndAndNectar(
     testFn: (actors: { alice: CndActor; bob: NectarActor }) => Promise<void>
 ): ProvidesCallback {
     return async (done) => {
-        const alice = await newCndActor("Alice", true);
-        const bob = await newNectarActor();
+        const startingAlice = newCndActor("Alice", true);
+        const startingBob = newNectarActor();
+
+        const alice = await startingAlice;
+        const bob = await startingBob;
```
D4nte, line 97:
```suggestion
        const [alice, bob] = await Promise.all([newCndActor("Alice", true), newNectarActor()]);
```
nitpick/individual taste.

tests/tests/cnd_and_nectar.ts
```diff
@@ -33,3 +33,29 @@ test(
         });
     })
 );
+
+test(
+    "given_cnd_and_nectar_when_cnd_publishes_a_matching_sell_order_then_successful_swap",
```
D4nte, line 38: BOOM! 💥

D4nte (approved):
This was hard 🙃

tests/src/actor_test.ts
thomaseizinger: I haven't tried it but I think that doesn't work because they are different types? Wouldn't TS get confused after this because both `alice` and `bob` are now of type `CndActor | NectarActor`?

D4nte: 🤷🏻‍♂️

PR #3217, opened 2020-09-24
Bitcoin network serialization

da-kami:
By using the comit library's ledger::Bitcoin type we fix a jsonrpc deserialization error for bitcoin mainnet `chain` in Nectar.

In order to re-use the serialization for `chain` properly the first commit pulls it into the `ledger` module.


nectar/src/command/wallet_info.rs
```diff
@@ -80,7 +78,7 @@ mod tests {
             Some(ethereum_wallet),
             Some(bitcoin_wallet),
             &seed,
-            bitcoin::Network::Regtest,
+            ledger::Bitcoin::Regtest,
```
tcharding, line 81: I like this usage, reads nicely.

nectar/src/bitcoin/bitcoind.rs
```diff
@@ -494,7 +496,7 @@ mod test {
 
         let network = client.network().await.unwrap();
 
-        assert_eq!(network, Network::Regtest)
+        assert_eq!(network, ledger::Bitcoin::Regtest)
```
tcharding, line 499: Did you consider bringing `ledger::Bitcoin` into scope with an alias
```
use ledger::Bitcoin as Network
```
Its a bit of a shame to have ta say `ledger::Bitcoin` when we are in a file that is solely bitcoin, (we have context from the filename already).

tcharding (approved):
Little comment on context, looks good to me though.

comit/src/ledger/bitcoin.rs
```diff
@@ -0,0 +1,114 @@
...
+/// ```rust
+/// use comit::ledger;
+///
+/// #[derive(serde::Serialize, serde::Deserialize, PartialEq, Debug)]
+/// #[serde(transparent)]
+/// struct Container(#[serde(with = "ledger::bitcoin::bitcoind_jsonrpc_network")] ledger::Bitcoin);
+///
+/// let container = Container(ledger::Bitcoin::Mainnet);
+///
+/// assert_eq!(json_sats, serde_json::to_string(&container).unwrap());
```
thomaseizinger, line 57: The variable you are asserting against `json_sats` is never defined.

Also, it should probably not be called "sats" because we are not deserializing an amount here :)

comit/src/ledger/bitcoin.rs
```diff
@@ -0,0 +1,114 @@
...
+    fn from(network: ::bitcoin::Network) -> Self {
+        match network {
+            bitcoin::Network::Bitcoin => Bitcoin::Mainnet,
+            bitcoin::Network::Testnet => Bitcoin::Testnet,
+            bitcoin::Network::Regtest => Bitcoin::Regtest,
+        }
+    }
+}
+
+/// Module specifically desgined for use with the `serde(with)` attribute.
```
thomaseizinger, line 44: Technically, this is not a module but just a single function. Hence, it is also not gonna work with the `serde(with)` attribute :)

Two options:

1. Fix the documentation
2. Make it a module and provide also provide the `serialize` variant.

thomaseizinger (approved):
Thanks @da-kami !

Some nits that are probably also the cause for your build failure :)

comit/src/ledger/bitcoin.rs
```diff
@@ -0,0 +1,118 @@
...
+        let got = serde_json::to_string(&ledger).expect("failed to serialize");
+
+        assert_that(&got).is_equal_to(&want);
+    }
+
+    #[test]
+    fn bitcoin_serialization_roundtrip() {
+        let ledger = Bitcoin::Mainnet;
+        let json = serde_json::to_string(&ledger).expect("failed to serialize");
+        let rinsed: Bitcoin = serde_json::from_str(&json).expect("failed to deserialize");
```
D4nte, line 114:
```suggestion
        let value: Bitcoin = serde_json::from_str(&json).expect("failed to deserialize");
```
rinsed?

D4nte (commented):
I am not convinced that the deserialisation of the bitcoind's output into a `comit::ledger` type is something that belong in the `comit` lib.

The usage of bitcoind by both cnd and nectar is incidental: we could have been using electrum server instead of example.

I don't think the `comit` lib should be concerned by how bitcoind encode the network on its rpc api.

Instead, I would expect a utility (or common bitcoind rpc lib) that can do the conversion from rpc api to `rust_bitcoin::Network` and then have a conversion from `rust_bitcoin::Network` to `comit::ledger::Bitcoin`.

thomaseizinger:
> The usage of bitcoind by both cnd and nectar is incidental: we could have been using electrum server instead of example.

I am with you but I don't see how the addition of a serde module violates that?
Also, the btsieve module currently depends on bitcoind. We could move the serde module under btsieve but not sure if that makes it any better?

> I don't think the `comit` lib should be concerned by how bitcoind encode the network on its rpc api.

But it is concerned about it through the btsieve module!

> Instead, I would expect a utility (or common bitcoind rpc lib) that can do the conversion from rpc api to `rust_bitcoin::Network` and then have a conversion from `rust_bitcoin::Network` to `comit::ledger::Bitcoin`.

We could use dedicated types yes. However, it only adds an additional layer of indirection.

> I am not convinced that the deserialisation of the bitcoind's output into a `comit::ledger` type is something that belong in the `comit` lib.

Fundamentally, our `ledger::Bitcoin` type and the value returned from the RPC API contain the same information. Why shouldn't we provide a module that plugs those two neatly together?
The way I see it is that we make our own life easier (using a serde module is more convenient over converting between types). At the same time, serde modules are not invasive at all and purely "additional" to a crate.

PR #3219, opened 2020-09-25
Make cnd republish open orders on startup

thomaseizinger:
The last commit adds the actual feature, all commits prior to that are refactoring that more or less works towards that :)


da-kami (approved):
Sweet to review in the way the commits are structure!

D4nte:
Just thinking a bit more about the feature. Let's imagine:
- I publish an order to sell BTC for 10,000 DAI
- No taker, I close ambrosia
- One week later, I tried to trade more, BTC is now USD20,000
- Orders are published, I have takers (of course) BUT I need to not proceed forward with the execution because, well, I would loose money.

Why do we want to republish the orders again?

thomaseizinger:
> Just thinking a bit more about the feature. Let's imagine:
> 
>     * I publish an order to sell BTC for 10,000 DAI
> 
>     * No taker, I close ambrosia
> 
>     * One week later, I tried to trade more, BTC is now USD20,000
> 
>     * Orders are published, I have takers (of course) BUT I need to not proceed forward with the execution because, well, I would loose money.

What you are describing sounds like orders should maybe have an expiry date?

> Why do we want to republish the orders again?

Because currently, publishing an order is just in-memory but we still save it to the DB. Not republishing creates an inconsistency where we have open orders but none else can see them because they are not published.

If a user no longer wants to trade an order, they should close it. That will probably update the state and remove it from the market.

D4nte:
> What you are describing sounds like orders should maybe have an expiry date?

Ok 👍

thomaseizinger:
> > What you are describing sounds like orders should maybe have an expiry date?
> 
> Ok +1

I will add it as a feature to our todo list.

thomaseizinger:
@D4nte this is still waiting for your review :)

D4nte (approved):
LGTM

thomaseizinger:
bors retry

thomaseizinger:
bors retry

PR #3220, opened 2020-09-25
Segregate trading strategy

D4nte:
The current trading strategy that takes for input:
- a spread
- mid-market rate
- available balances

Is now in a dedicated module. Allowing this concern to be removed from
the maker.
Some primitives such as the bitcoin fees and the reserved assets for
ongoing swaps are also moved in the strategy module. It could be argued
that they should stay with the maker or moved in yet another module.

I propose that such concern can be reviewed when another trading
strategy is implemented, validating the best scope for trading
strategies.

### DIFF
--- a/nectar/src/command/trade.rs
+++ b/nectar/src/command/trade.rs
@@ -6,6 +6,7 @@ use crate::{
     config::{KrakenApiHost, Settings},
     ethereum::{self, dai},
     history::History,
+    maker::strategy,
     mid_market_rate::get_btc_dai_mid_market_rate,
     network::{self, new_swarm},
     swap::{Database, SwapExecutor, SwapKind, SwapParams},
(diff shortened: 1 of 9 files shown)


nectar/src/command/trade.rs
```diff
@@ -136,14 +137,13 @@ async fn init_maker(
 
     let spread: Spread = settings.maker.spread;
 
+    let strategy = strategy::AllIn::new(btc_fee_reserve, btc_max_sell, dai_max_sell, spread);
```
thomaseizinger, line 140: `AllIn`, I like it :D

nectar/src/maker.rs
```diff
@@ -160,64 +139,31 @@ impl Maker {
         }
     }
 
-    /// Decide whether we should proceed with order,
-    /// Confirm with the order book
-    /// Re & take & reserve
     pub fn process_taken_order(
```
thomaseizinger, line 142: This should probably be renamed at some point :)

thomaseizinger (approved):
Nice, well done @D4nte !

nectar/src/maker.rs
D4nte: Yes, I'll leave it like that for now. There is a todo on the matching function of `strategy::AllIn`, we can improve this interface when tackling the todo.

PR #3221, opened 2020-09-25
Remove as much of the old swaps API as we currently can

thomaseizinger:


D4nte (approved):

D4nte:
Bold move.

thomaseizinger:
> Bold move.

Is it? Would you prefer we keep the API and re-implement it using `setup-swap` on the network layer?

D4nte:
> > Bold move.
> 
> Is it? Would you prefer we keep the API and re-implement it using `setup-swap` on the network layer?

I think it's bold but in a good way ;)

tcharding (approved):

thomaseizinger:
Obsolete with #3336.

PR #3227, opened 2020-09-28
Unify usage of time libraries

thomaseizinger:
cc @tcharding :)

### DIFF
--- a/cnd/src/hbit.rs
+++ b/cnd/src/hbit.rs
@@ -7,10 +7,10 @@ use crate::{
     LocalSwapId, Role, Side,
 };
 use bitcoin::{Address, Block, BlockHash};
-use chrono::{DateTime, Utc};
 use comit::{asset, htlc_location, transaction, LockProtocol, Secret};
 use futures::TryStreamExt;
 use std::collections::{hash_map::Entry, HashMap};
+use time::OffsetDateTime;
 use tokio::sync::Mutex;
 
 use crate::storage::{BtcDaiOrder, Order};
@@ -24,7 +24,7 @@ pub use comit::{hbit::*, identity};
 pub async fn new<C>(
     id: LocalSwapId,
     params: Params,
-    start_of_swap: DateTime<Utc>,
+    start_of_swap: OffsetDateTime,
     role: Role,
     side: Side,
     storage: Storage,
--- a/cnd/src/storage/db/schema.rs
+++ b/cnd/src/storage/db/schema.rs
@@ -7,7 +7,7 @@ table! {
        local_swap_id -> Text,
        role -> Text,
        counterparty_peer_id -> Text,
-       start_of_swap -> Timestamp,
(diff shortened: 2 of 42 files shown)


cnd/src/hbit.rs
```diff
@@ -6,14 +6,14 @@ use crate::{
...
 use bitcoin::{Address, Block, BlockHash};
-use chrono::{DateTime, Utc};
 use comit::{asset, htlc_location, transaction, LockProtocol, Secret};
 pub use comit::{hbit::*, identity};
 use futures::TryStreamExt;
 use std::{
     collections::{hash_map::Entry, HashMap},
     sync::Arc,
 };
+use time::OffsetDateTime;
```
D4nte, line 16: Do we have the same guarantees with `OffsetDateTime` than we use to have? e.g. ensuring there is no mix with local and utc?

Also, why not using `chrono` instead of `time`? You did not justify the choice.

cnd/src/storage/db/schema.rs
```diff
@@ -7,7 +7,7 @@ table! {
        local_swap_id -> Text,
        role -> Text,
        counterparty_peer_id -> Text,
-       start_of_swap -> Timestamp,
+       start_of_swap -> BigInt,
```
D4nte, line 10: Why a BigInt?

D4nte (approved):

cnd/src/storage/db/schema.rs
thomaseizinger: Because it is an `i64`.

cnd/src/hbit.rs
thomaseizinger: > Do we have the same guarantees with `OffsetDateTime` than we use to have? e.g. ensuring there is no mix with local and utc?

I am 100% sure but I think so. `OffsetDateTime` can be constructed with a specific offset but there is no guarantee that it was constructed with a specific one. It shouldn't matter much though because the instance does track the timezone and within `btsieve` we only compare unix timestamps which are always in UTC.

The problem we had previously was that we constructed a `DateTime` WITHOUT timezone information.

> Also, why not using `chrono` instead of `time`? You did not justify the choice.

Usage of `time` is more spread out in the codebase due to @tcharding `expiries` module, hence moving to `time` was easier than the other way round. Also, development of `time` seems to be more active and it is the more lightweight dependency.

I am not too fussed about it though, we can move back to `chrono` if we want at some point. I just wanted it to be uniform :)

thomaseizinger:
@D4nte I had to slightly adapt the history tests for nectar because `OffsetDateTime` doesn't print milliseconds for the Rfc3339 format. I think that is ok but wanted to let you know :)

thomaseizinger:
bors retry

tcharding:
LGTM, thanks for CC'ing me @thomaseizinger.

PR #3228, opened 2020-09-28
[nectar] mainnet Bitcoin fee

D4nte:
Improve fee management for nectar. Two configurations options are offered:
- static fee in sat/byte
- Use bitcoind estimate mode

Please note the bitcoind mode is not tested because I was not able to get it to work in regtest despite injecting a bunch of transactions in the network.

Follow-up PRs:
- Make the `maximum_possible_fee` option take a sat/byte fee and calculate internally what that would be in total tx fee instead of expecting the user to know how big the COMIT transactions are.

### DIFF
--- a/nectar/sample-config.toml
+++ b/nectar/sample-config.toml
@@ -23,7 +23,13 @@ dai = 1000
 [maker.maximum_possible_fee]
 # An estimation of the maximum fee that we would expect to pay, used to ensure we always have enough
 # balance to execute an order we publish.
-bitcoin = 0.00009275 
+bitcoin = 0.00009275
+
+# Strategies used for Bitcoin fee handling.
+[maker.fee_strategies.bitcoin]
+# A static sat/byte value can be specified or you can relay on bitcoind's estimatesmartfee feature.
+strategy = "static"
+sats_per_byte = 12
 
 [network]
 # The libp2p socket on which nectar listens for COMIT messages.
--- a/nectar/src/config/settings.rs
+++ b/nectar/src/config/settings.rs
@@ -1,6 +1,6 @@
 use crate::{
     bitcoin,
-    config::{file, Bitcoind, Data, File, MaxSell, Network},
+    config::{file, Bitcoind, Data, EstimateMode, File, MaxSell, Network},
     ethereum, Spread,
 };
 use anyhow::{Context, Result};
@@ -167,6 +167,8 @@ pub struct Maker {
     /// balance. Fees are in the nominal native currency and per
     /// transaction.
     pub maximum_possible_fee: Fees,
+    /// Fee strategies
+    pub fee_strategies: FeeStrategies,
     pub kraken_api_host: KrakenApiHost,
 }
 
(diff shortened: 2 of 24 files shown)


nectar/sample-config.toml
```diff
@@ -23,7 +23,13 @@ dai = 1000
...
 # An estimation of the maximum fee that we would expect to pay, used to ensure we always have enough
 # balance to execute an order we publish.
-bitcoin = 0.00009275 
+bitcoin = 0.00009275
+
+# Strategies used for Bitcoin fee handling.
+[maker.fee_strategies.bitcoin]
+# A static sat/byte value can be specified or you can relay on bitcoind's estimatesmartfee feature.
+strategy = "static"
+sats_per_bytes = 53
```
thomaseizinger: Naming nit :)

```suggestion
sats_per_byte = 53
```

thomaseizinger: Also, should we use a better value here? :D

53 seems a lot!

nectar/sample-config.toml
```diff
@@ -23,7 +23,13 @@ dai = 1000
 [maker.maximum_possible_fee]
 # An estimation of the maximum fee that we would expect to pay, used to ensure we always have enough
 # balance to execute an order we publish.
-bitcoin = 0.00009275 
+bitcoin = 0.00009275
+
+# Strategies used for Bitcoin fee handling.
+[maker.fee_strategies.bitcoin]
+# A static sat/byte value can be specified or you can relay on bitcoind's estimatesmartfee feature.
+strategy = "static"
```
thomaseizinger, line 31: I assume the keys (like `sats_per_byte`) vary based on this field? You could elevate this to a toml table and let that one control the keys:

```toml
[maker.fee_strategies.bitcoin.static]
sats_per_byte = 10

[maker.fee_strategies.bitcoin.bitcoind_estimate]
mode = "conservative"
```

thomaseizinger: This should represent nicely as an enum in Rust and hence guarantee you at compile time that you don't get a "mode" if the user selected "static" and vice versa.

nectar/src/config/settings.rs
```diff
@@ -193,6 +195,59 @@ impl Default for KrakenApiHost {
...
+
+/// Defaults to static fee mode
+/// Default value for static mode is 10 sat per byte
+impl From<file::BitcoinFee> for BitcoinFeeStrategy {
+    fn from(file: file::BitcoinFee) -> Self {
+        match file.strategy {
+            None | Some(file::BitcoinFeeStrategy::Static) => Self::SatsPerByte(
+                file.sats_per_byte
+                    .unwrap_or_else(|| bitcoin::Amount::from_sat(DEFAULT_BITCOIN_STATIC_FEE_SAT)),
+            ),
```
thomaseizinger, line 239: You should be able to use `map_or_else` here and re-use the `Default` impl from below.

thomaseizinger (approved):
Looks good to me!

I wasn't aware that you are actually going to integrate with bitcoind for nectar but I guess now it is done already :D

Some suggestions regarding the config file format :)

nectar/sample-config.toml
D4nte: Hum, it's a sample file, not default, but happy to change.

nectar/sample-config.toml
D4nte: > This should represent nicely as an enum in Rust and hence guarantee you at compile time that you don't get a "mode" if the user selected "static" and vice versa.

I tried and that is how I spent most of my Friday. The toml crate does not seem to allow that.
However I only tried an enum with an anonymous field (newtype) not with a struct.

I don't really want to fight again with the crate at this stage. Happy to either review at a later stage or for you to give it a go.

nectar/src/config/settings.rs
```diff
@@ -163,10 +163,14 @@ pub struct Maker {
     /// Spread to apply to the mid-market rate, format is permyriad. E.g. 5.20
     /// is 5.2% spread
     pub spread: Spread,
+    // TODO: Probably move that under something common to fee_strategies.
+    // I did not realize I would still need it.
```
da-kami, line 167:
```suggestion
```

Maybe remove the comment above as well?

da-kami (approved):

nectar/sample-config.toml
thomaseizinger: I think I remember doing something similar in cnd so I ought it would work. Happy to give it a go in a follow-up :)

nectar/src/config/settings.rs
D4nte: No, I want to clean that up as part the follow-up PR I am suggesting.

@thomaseizinger FYI, if you end up touching the settings before me we may want to re-think what should go where: fee strategy, max possible fee. Under maker? together?

PR #3229, opened 2020-09-28
Close order after successful swap

thomaseizinger:


da-kami (approved):

cnd/src/hbit.rs
```diff
@@ -2,20 +2,20 @@ use crate::{
...
 use futures::TryStreamExt;
-use std::{
-    collections::{hash_map::Entry, HashMap},
-    sync::Arc,
-};
+use std::collections::{hash_map::Entry, HashMap};
 use tokio::sync::Mutex;
 
+use crate::storage::{BtcDaiOrder, Order};
+pub use comit::{hbit::*, identity};
```
D4nte, line 17:
```suggestion
use crate::storage::{BtcDaiOrder, Order};

pub use comit::{hbit::*, identity};
```

cnd/src/storage/db/tables/btc_dai_orders.rs
```diff
@@ -3,7 +3,7 @@ use crate::storage::{
         schema::*,
         wrapper_types::{Erc20Amount, Satoshis, WeiPerSat},
     },
-    NotOpen, Order, Text,
+    NotOpen, NotSettling, Order, Text,
```
D4nte, line 6: Any reason why you defined this error outside the only file that uses it?

D4nte (approved):

cnd/src/hbit.rs
thomaseizinger: Upsi, will fix in a follow-up!

cnd/src/storage/db/tables/btc_dai_orders.rs
thomaseizinger: I was thinking about that actually, for now all errors are just inside the db `errors` module but I would actually want them to be co-located yes.

I'll do that in one of my follow-ups!

PR #3230, opened 2020-09-28
Remove fee subtraction from sell order creation

da-kami:
The max_sell represents the maximum amount to be traded for each
currency. This should not be concerned by the fees. Fees are on top
of the maximum amount.
Note that checks for sufficient funds are already in place, so we just
remove the fee subtraction, add unit tests to show the bahavior and
adap existing integration and e2e tests.

### DIFF
--- a/tests/tests/buy_and_sell_bitcoin_e2e_cnd_nectar.ts
+++ b/tests/tests/buy_and_sell_bitcoin_e2e_cnd_nectar.ts
@@ -16,7 +16,7 @@ test(
             (market) => market.entities.length > 0
         );
 
-        await alice.makeBtcDaiOrder(Position.Buy, "0.09990725", "9450"); // This matches what nectar publishes.
+        await alice.makeBtcDaiOrder(Position.Buy, "0.1", "9450"); // This matches what nectar publishes.
         await alice.waitForSwap();
 
         await alice.assertAndExecuteNextAction("deploy");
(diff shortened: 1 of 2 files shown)


tests/tests/buy_and_sell_bitcoin_e2e_cnd_nectar.ts
```diff
@@ -28,8 +28,8 @@ test(
 
         await alice.assertBalancesAfterSwap();
         await bob.assertBalancesChangedBy({
-            bitcoin: -(9990725n + 3060n), // nectar pays order quantity + the funding fee
-            dai: 944_123_512_500_000_000_000n, // = 0.09990725 * 9450 * 10^18
+            bitcoin: -(10_000_000n + 3060n), // nectar pays order quantity + the funding fee
+            dai: 945_000_000_000_000_000_000n, // = 0.1 * 9450 * 10^18
         });
```
thomaseizinger, line 33: Nice! :heart_eyes: 

Why doesn't the other test need changing?

thomaseizinger (approved):

tests/tests/buy_and_sell_bitcoin_e2e_cnd_nectar.ts
da-kami: The other test did not need changes because the buy `quantity` is not having positions after the point because of fees - we don't deal with fees for buy - but because `max_sell` is in DAI. The `quantity` of the order is calculated out of the fixed DAI and the price (rate in Nectar), hence it fluctuates with the price. 
I thought about creating a follow-up PR where I change the `max` to being always in the base, i.e. `max_sell` and `max_buy` both being a Bitcoin quantity. Given the current model that makes more sense I feel.

D4nte (approved):

PR #3231, opened 2020-09-28
Update order and swap state in DB based on result of swap execution

thomaseizinger:


cnd/src/spawn.rs
```diff
@@ -100,14 +108,17 @@ impl ProtocolContext<hbit::Params> {
...
 }
 
 impl ProtocolContext<halbit::Params> {
-    fn spawn(self, connectors: Connectors, storage: Storage, handle: Handle) -> Result<()> {
+    fn spawn(
+        self,
+        connectors: Connectors,
+        storage: Storage,
+        handle: Handle,
+    ) -> Result<JoinHandle<()>> {
```
da-kami: That's an interesting one, I have not used that so far :)

cnd/src/spawn.rs
```diff
@@ -58,13 +59,47 @@ pub async fn spawn(
...
+        let alpha_handle = alpha
             .spawn(connectors.clone(), storage.clone(), handle.clone())
             .context("failed to spawn protocol for alpha ledger")?;
-        beta.spawn(connectors.clone(), storage.clone(), handle.clone())
+        let beta_handle = beta
+            .spawn(connectors.clone(), storage.clone(), handle.clone())
             .context("failed to spawn protocol for alpha ledger")?;
+
+        handle.spawn(async move {
+            let result = future::join(alpha_handle, beta_handle).await;
```
da-kami: neat.

tests/src/actors/cnd_actor.ts
```diff
@@ -582,6 +583,22 @@ export class CndActor
         );
     }
 
+    public async waitUntilSwapped() {
```
da-kami, line 593: 👍 nice, this might make things less brittle!

cnd/src/spawn.rs
```diff
@@ -69,58 +70,60 @@ pub async fn spawn(
...
+                    Ok(())
+                })
+            }
+            // Join successful but one of the protocols failed
+            Ok((Err(e), _)) | Ok((_, Err(e))) => {
+                tracing::error!(swap = %swap_id, "failed to complete swap: {:#}", e);
+
+                Box::new(move |conn| {
+                    commands::update_order_of_swap_to_failed(conn, swap_id)?;
+                    // we don't mark a swap as completed in case of failure so that a
```
da-kami, line 109: If I am not mistaken this behavior might cause trouble when trying to close an order if retrying succeeds? 

`update_order_of_swap_to_closed` is not capable of dealing with `failed`:

https://github.com/comit-network/comit-rs/blob/only-active-swaps/cnd/src/storage/db/commands.rs#L66-L91

and `update_btc_dai_order_to_settling` does not seem to deal with `failed` neither:

https://github.com/comit-network/comit-rs/blob/only-active-swaps/cnd/src/storage/db/commands.rs#L26-L42

da-kami (commented):
Nice one! I am not sure I could think everything through 100%.

My concerns are mostly related to `failed` scenarios, but I am not sure how much time we want to invest in this right now.

To my understanding the table `completed_swaps` contains only swaps that are completed successfully, failed swaps are just identified by their state. Is the only way to determine that a swap has "failed" to check the corresponding order, since we don't "save" the failed swap?

I think that we might have to introduce a more elaborate strategy to deal with `failed` swaps. I think just restarting the node won't work, because the order state will not be updated properly.

thomaseizinger:
Yes the whole order state + swap state is super messy at the moment. I think we should replace this with a proper state machine in the long run but that requires more work :/

Should we remove that feature again or are we fine with a not-so-great working feature?

cnd/src/spawn.rs
thomaseizinger: Yes true. The proper way of doing this would be to not keep track of these individual states but store the whole swap execution in the DB and reason about it there.

As a patch, I could change the order state again once the swap is re-spawned?
Do we care much at this stage? :sweat_smile:

da-kami:
> Yes the whole order state + swap state is super messy at the moment. I think we should replace this with a proper state machine in the long run but that requires more work :/
> 
> Should we remove that feature again or are we fine with a not-so-great working feature?

I think we can roll with it for now, but we should record the follow up task. I would hope that we don't run into the failing scenario too often 😬

D4nte (approved):
LGTM

thomaseizinger:
> > Yes the whole order state + swap state is super messy at the moment. I think we should replace this with a proper state machine in the long run but that requires more work :/
> > Should we remove that feature again or are we fine with a not-so-great working feature?
> 
> I think we can roll with it for now, but we should record the follow up task. I would hope that we don't run into the failing scenario too often grimacing

I have a follow-up patch here: https://github.com/comit-network/comit-rs/pull/3240

PR #3232, opened 2020-09-28
Maximum buy amount in bitcoin to adhere to model

da-kami:
The maximum sell amount in dai was from a time where we used a
different model. Now that we use quantity and price it more natural
to have the max buy quantity in bitcoin as well.

This helps us with having "fixed" quantities on buy orders according to `max_buy` instead of `max_sell.dai` in Nectar.


nectar/sample-config.toml
```diff
@@ -16,9 +16,11 @@ kraken_api_host = "https://api.kraken.com"
 # The maximum amount of bitcoin to sell in one order, optional field.
 # If absent, orders are capped by the available balance.
 bitcoin = 0.1
-# The maximum amount of dai to sell in one order, optional field.
-# If absent, orders are capped by the available balance.
-dai = 1000
+
+[maker.max_buy]
+# The maximum amount of bitcoin to buy in one order, optional field.
+# If absent, orders are capped by the available quote balance.
```
da-kami, line 22:
```suggestion
# If absent, orders are capped by the available quote balance given the current price.
```

D4nte: I think it's actually clearer without the suggestion above :)

nectar/src/command/trade.rs
```diff
@@ -137,7 +137,7 @@ async fn init_maker(
 
     let spread: Spread = settings.maker.spread;
 
-    let strategy = strategy::AllIn::new(btc_fee_reserve, btc_max_sell, dai_max_sell, spread);
+    let strategy = strategy::AllIn::new(btc_fee_reserve, max_sell, max_buy, spread);
```
D4nte, line 140: Please, I have dealt with enough merge conflicts can you hold on with settings changes?

nectar/src/maker/strategy.rs
```diff
@@ -104,10 +104,10 @@ impl AllIn {
     /// Create a new buy order given the passed parameters.
     /// Bitcoin is always the base asset, buy order buys bitcoin.
     ///
-    /// The quote amount is the full available dai balance or the max dai sell
-    /// parameter, whichever is the lowest.
-    /// The spread parameter is then applied on the mid market rate to decide
-    /// the price.
+    /// The quantity is the full available dai balance in btc given the current
+    /// rater or the maximum buy quantity parameter, whichever is the
```
thomaseizinger, line 108:
```suggestion
    /// rate or the maximum buy quantity parameter, whichever is the
```

nectar/src/config/file.rs
```diff
@@ -283,7 +289,9 @@ kraken_api_host = "https://api.kraken.com/"
 
 [maker.max_sell]
 bitcoin = 1.23456
-dai = 9876.54
+
+[maker.max_buy]
+bitcoin = 1.23456
```
thomaseizinger, line 294: I feel like it would be more natural if the config would look something like:

[maker.btc_dai]
max_sell = "0.1"
max_buy = "0.1"

Grouping it under the trading pair makes it more obvious how this would be extended and how they relate to each other.

thomaseizinger (approved):
Nice, thank you @da-kami !

I have no opinion on how we handle merge conflicts etc but this is good to go in from my side :)
Arrange yourself as necessary!

nectar/src/config/file.rs
D4nte: Let's group when we add a second pair? :D

thomaseizinger: More of a philosophical question I guess :D
I don't feel strongly about it although personally, I would have probably gone the extra mile here and completely changed the model to fit better in with forex :)

nectar/src/command/trade.rs
da-kami: If this comment is just concerning this file, and this specific rename I think I can remove it and add a ToDo for later. 
I was hoping to change all the names consistently at once, but if it is causing too much friction I can remove that one. 
`dai_max_sell` is misleading, because the value is a BTC value now (that represents the maximum buy quantity).

nectar/src/config/file.rs
da-kami: I like the grouping, I already had the feeling that the `max_sell` and `max_buy` split was a bit weird, but did not invest brainpower in it yesterday. Will change it.

da-kami: Actually, on second thought, I think it depends on what we want.
The current model allows to set an overall base max limit, i.e. all BTC-XXX trading pairs would be affected. (we never buy/sell more than max bitcoin), if I change it to 

```
[maker.btc_dai]
max_sell = "0.1"
max_buy = "0.1"
```

I have to specify this limit for each trading pair. So, if we additionally have `BTC/USDT` I would have to specify:

```
[maker.btc_dai]
max_sell = "0.1"
max_buy = "0.1"

[maker.btc_usdt]
max_sell = "0.1"
max_buy = "0.1"
```

Is that what we want?

da-kami: How about:

```
[maker.btc]
max_sell = "0.1"
max_buy = "0.1"
```

For me the `max` limit was something to be defined globally within Nectar and not per trading pair. Thoughts?

nectar/src/command/trade.rs
da-kami: Added `no-mergify`, going in after your PR has merged.

nectar/src/config/file.rs
thomaseizinger: If find the per trading per one more natural but I am not a trader :grin: 
I can see the use case of specifying a complete upper bound but I am not sure that is needed because it feels like an optimization to the problem of having to duplicate configuration across several pairs. I would defer such optimizations until later or not do them at all if they would create too much of a complexity overhead. It is not like one would change this config all the time hence I think clarity is more important.

nectar/src/command/trade.rs
```diff
@@ -295,10 +295,8 @@ mod tests {
 
         let settings = Settings {
             maker: settings::Maker {
-                max_sell: MaxSell {
-                    bitcoin: None,
-                    dai: None,
-                },
+                max_sell: Max { bitcoin: None },
```
D4nte, line 298: Use default here

nectar/src/config.rs
```diff
@@ -30,13 +30,10 @@ pub struct Bitcoind {
 }
 
 #[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
-pub struct MaxSell {
+pub struct Max {
     #[serde(default)]
```
D4nte, line 34: Does this design still make sense? Looking at it I think I agree with @thomaseizinger, we should introduce the pairs in the config file.

nectar/src/config/file.rs
D4nte: I like your new proposals better. The only ambiguuty here is that it might not be clear what the amount refer to. I would try to introduce "BTC" or "base" in the config terminology.

D4nte (changes requested):
let's improve the config structure

nectar/src/config/file.rs
da-kami: hm... I am a bit divided here. I find the fact, that one cannot see `bitcoin` not perfect neither. It would be good to see that the value specified is in the base currency. I know that the currency pair would imply it but I still think that it would be nice to have the `bitcoin` there so it is absolutely clear. 
Furthermore, I would not invest time into changing this now. It is a refactoring on top of a nice to have feature... So I tend to roll with @D4nte 's initial suggestion to tackle this once we add another pair.

nectar/src/config.rs
da-kami: Do we really want to tackle this already? I added another comment above.

nectar/src/config/file.rs
thomaseizinger: > hm... I am a bit divided here. I find the fact, that one cannot see bitcoin not perfect neither. It would be good to see that the value specified is in the base currency. I know that the currency pair would imply it but I still think that it would be nice to have the bitcoin there so it is absolutely clear.

You could say `max_sell_quantity` and `max_buy_quantity` in addition to the currency pair to make it clear that we are talking about the quantity and not the price.

> Furthermore, I would not invest time into changing this now. It is a refactoring on top of a nice to have feature... So I tend to roll with @D4nte 's initial suggestion to tackle this once we add another pair.

The more I think about it the more I would want to change the config because the current one is super confusing :see_no_evil:

nectar/src/config.rs
thomaseizinger: Don't tackle it if you don't want to, it is up to you :)

D4nte (approved):

thomaseizinger (approved):
<3

PR #3240, opened 2020-09-30
Order is settling on respawn

thomaseizinger:


D4nte:
`Order is settling on respawn` what does it means? Please describe a bit more.

thomaseizinger:
> `Order is settling on respawn` what does it means? Please describe a bit more.

The commit messages should give a bit more context but I can describe it again here as well:

This PR grew out of @da-kami's feedback here: https://github.com/comit-network/comit-rs/pull/3231

Basically, if we respawn a swap, we need to set the state to settling again so that it can successfully be set to closed if it succeeds then.

D4nte:
Why do we set to settling when respawning? Shouldn't it be already set to settling before a shutdown (if it's settling)?

thomaseizinger:
> Why do we set to settling when respawning? Shouldn't it be already set to settling before a shutdown (if it's settling)?

Because it might be set to failed if it failed: https://github.com/comit-network/comit-rs/blob/e539de2f6cad0f68fe53d5815a539b958d806140/cnd/src/spawn.rs#L108

cnd/src/storage/db/commands.rs
```diff
@@ -117,6 +130,35 @@ pub fn update_order_of_swap_to_failed(conn: &SqliteConnection, swap_id: LocalSwa
...
+) -> Result<()> {
+    let (order, btc_dai_order) = orders::table
+        .inner_join(order_swaps::table.inner_join(swaps::table))
+        .inner_join(btc_dai_orders::table)
+        .filter(swaps::local_swap_id.eq(Text(swap_id)))
+        .select((orders::all_columns, btc_dai_orders::all_columns))
+        .first::<(Order, BtcDaiOrder)>(conn)
+        .with_context(|| NoOrderForSwap(swap_id))?;
+
+    let affected_rows = diesel::update(&btc_dai_order)
```
D4nte, line 145: I think we should add a `warn` if we move a previously failed swap to settling.

D4nte (approved):

cnd/src/storage/db/commands.rs
thomaseizinger: Why `warn`? I feel it is expected behaviour for swaps to be respawned after they are failed and here we are just making sure the accounting is correct :)

PR #3241, opened 2020-09-30
Use `comit::hbit::Params` throughout nectar

thomaseizinger:
In order to upstream nectar's swap execution to comit, it is
important that we only use data structures that belong there.

Currently, nectar uses a custom `hbit::Params` struct that contains
the transient key among the remaining params. This does not (yet)
belong in `comit` and hence needs to be removed.

### DIFF
--- a/nectar/src/command.rs
+++ b/nectar/src/command.rs
@@ -136,7 +136,7 @@ pub fn into_history_trade(
         base_symbol: Symbol::Btc,
         quote_symbol: Symbol::Dai,
         position,
-        base_precise_amount: swap.hbit_params.shared.asset.as_sat().into(),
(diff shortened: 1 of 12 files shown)


luckysori (dismissed):
LGTM!

D4nte (commented):
Looks neat. go for it (there is one remaining todo).

thomaseizinger:
Cool, thanks for approving the general idea @D4nte @luckysori !

I will go ahead and implement this some time then :)

thomaseizinger:
Actually, I believe it should be the other way round. `comit` should be aware of this secret key (and probably even generate it) similar to how we did in other protocols like a2l or xmr-btc.

PR #3244, opened 2020-10-01
[nectar] clean-up fee config

D4nte:
See commit messages. Did some restructure of the config file to, hopefully, get something that makes sense.
Feel free to be pendantic on the terminology.


nectar/src/config/file.rs
```diff
@@ -57,7 +57,7 @@ pub struct BitcoinFee {
     /// The static value to use if the selected strategy is "static"
     #[serde(default)]
     #[serde(with = "crate::config::serde::bitcoin_amount::sat_as_optional_unsigned_int")]
```
thomaseizinger, line 49: Now that we are just re-exporting `rust-bitcoin`'s `Amount`, I think we should be able to use this module here: https://github.com/rust-bitcoin/rust-bitcoin/blob/c1ae3b795547176d1403e12a6f708ffecfa52c1d/src/util/amount.rs#L938-L960

nectar/src/swap/bitcoin.rs
```diff
@@ -88,15 +88,15 @@ impl hbit::ExecuteRefund for Wallet {
 
         let refund_address = self.inner.new_address().await?;
 
-        let fee_rate_per_byte = self.fee.byte_rate().await?;
+        let fee_rate_per_vbyte = self.fee.vbyte_rate().await?;
```
thomaseizinger, line 91: Similar to the patch on the HTTP API module, I think we can shorten this variable name without loosing any information to something like `btc_per_vbyte` or `sats_per_vbyte`.

Alternatively, if the method is named `vbyte_rate`, why can't the variable be called the same? :)

nectar/sample-config.toml
```diff
@@ -20,10 +20,12 @@ max_buy_quantity = 0.1
...
-[maker.maximum_possible_fee]
-# An estimation of the maximum fee that we would expect to pay, used to ensure we always have enough
-# balance to execute an order we publish.
-bitcoin = 0.00009275
+[maker.btc_fee_to_reserve]
+# An estimation of the maximum network fee that we would expect to pay, used to ensure
+# we always have enough balance to execute an order we publish.
+sat_per_vbyte = 34
+# An estimation of the largest Bitcoin transaction we are likely to pay for.
+vbyte_transaction_weight = 850
```
thomaseizinger: Does this setting help with anything?

I am imagining the following scenarios:

1. I am concerned about wasting money on fees and hence I will investigate the config options of nectar to optimize this. In that case, I am ok with stopping to trade in case the network is congested. To configure this, the primary handle is the fee rate.

2. Let's say my configured fee rate is still within the limits of the current network but for some reason, bitcoind has to pull together several inputs to fund an HTLC an hence the transaction is too heavy and would be blocked by this setting.

I don't feel like that is what I want to happen. If my inputs are this scattered then I will have to pay higher fees eventually to use them but this is just how Bitcoin works. Keeping this low is primarily the job of wallets and their coinselection algorithms.

I don't think we should have a config entry for this.

nectar/sample-config.toml
```diff
@@ -20,19 +20,19 @@ max_buy_quantity = 0.1
...
 # Strategies used for Bitcoin fee handling.
 [maker.fee_strategies.bitcoin]
 # A static sat/byte value can be specified or you can relay on bitcoind's estimatesmartfee feature.
 strategy = "static"
 sat_per_vbyte = 12
 
+[maker.fee_strategies.bitcoin.fees_to_reserve]
+# An estimation of the maximum network fee that we would expect to pay, used to ensure
+# we always have enough balance to execute an order we publish.
+sat_per_vbyte = 34
```
thomaseizinger: Just from looking at the file, I am confused as to what this does.

If I configure a static fee above, why do I have to configure another fee here?

If this is supposed to be a maximum, I feel like it should be part of the `estimatesmartfee` strategy because it doesn't seem to be useful in the case of a static fee.

nectar/sample-config.toml
```diff
@@ -63,3 +45,21 @@ node_url = "http://localhost:18443/"
 chain_id = 1
 # The url to the web3 node, can include an infura key: `https://mainnet.infura.io/v3/YOUR-PROJECT-ID`
 node_url = "http://localhost:8545/"
+
+# Strategies used for Bitcoin fee handling.
+[fee_strategies.bitcoin]
```
thomaseizinger, line 50: This could also just be called `[fees.bitcoin]`.

thomaseizinger: We also already have a `bitcoin` section, could possibly also have `[bitcoin.fees]`? That would "repeat" the `fees` section with every ledger but at the same time would allow us to use ledger-specific terminology and for example do `[ethereum.gas]` instead.

Generally, I feel like we should strive to design things per-ledger and avoid having to many abstractions like "fees" in this case that span multiple ledgers.

thomaseizinger (approved):
Thank you @D4nte ! Definitely an improvement over what we had before :)

I've made some pedantic comments in-line as requested :grin: 
Also, is the "max tx weight" config actually used anywhere? Maybe I missed it but I could see any code where we are using this.

D4nte:
> Also, is the "max tx weight" config actually used anywhere? Maybe I missed it but I could see any code where we are using this.

Used here: https://github.com/comit-network/comit-rs/blob/ff2f12211b8f72ddbea81906db7f3f680a46c7f1/nectar/src/config/settings.rs#L390

nectar/src/config/file.rs
D4nte: > Now that we are just re-exporting `rust-bitcoin`'s `Amount`, I think we should be able to use this module here: [rust-bitcoin/rust-bitcoin@`c1ae3b7`/src/util/amount.rs#L938-L960](https://github.com/rust-bitcoin/rust-bitcoin/blob/c1ae3b795547176d1403e12a6f708ffecfa52c1d/src/util/amount.rs#L938-L960)

Oh, thanks, I did not know this one! I assumed they did not have an opt one.

nectar/sample-config.toml
D4nte: What is the alternative and what values would you set then?

The issue is that at this stage, we are not sure of what reasonable values are until we use it in mainnet.

da-kami (approved):

thomaseizinger (approved):
Well done @D4nte !

thomaseizinger:
bors retry

PR #3252, opened 2020-10-02
Don't spam Infura

da-kami:
Might require a few more changes...

### DIFF
--- a/comit/src/btsieve.rs
+++ b/comit/src/btsieve.rs
@@ -90,7 +90,7 @@ where
             seen_blocks.extend(missed_blocks);
 
             // The duration of this timeout could/should depend on the network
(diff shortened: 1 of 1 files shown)


comit/src/btsieve.rs
```diff
@@ -90,7 +90,7 @@ where
             seen_blocks.extend(missed_blocks);
 
             // The duration of this timeout could/should depend on the network
```
D4nte, line 92: Shall we tackle the comment above?

thomaseizinger: Definitely, I don't think there is a point in just changing the delay from 1 sec to 30 :D

thomaseizinger:
Superseded by https://github.com/comit-network/comit-rs/pull/3255.

PR #3255, opened 2020-10-06
Specify poll_interval for btsieve based on connected network

thomaseizinger:


comit/src/btsieve/ethereum.rs
```diff
@@ -56,3 +57,16 @@ pub struct Event {
...
 }
+
+async fn poll_interval<C>(connector: &C) -> Result<Duration>
+where
+    C: ConnectedNetwork<Network = ChainId>,
+{
+    let network = connector.connected_network().await?;
+    let seconds = match network {
+        ChainId::GETH_DEV => 1,
+        _ => 10,
```
D4nte, line 68: Didn't you introduced another matching pattern to determine the retry time to use for each Ethereum network in your other PR? Why not using the same times?

D4nte (approved):

comit/src/btsieve/ethereum.rs
thomaseizinger: It is not quite the same because one is in nectar's wallet and one is in btsieve. Not sure if unifying is worth it. Where would I put the code?

D4nte: Feel free not to unify but both were about knowing how often should one check the blockchain status.
btsieve holds blockchain knowledge. would it make sense to expose from there?

thomaseizinger: Maybe, I am not so sure. I didn't want `btsieve` to become this general "blockchain connector thingy" hence I opted to not export functionality like that.

Let's see how things evolve and consolidate eventually.

PR #3256, opened 2020-10-06
Fix fees

D4nte:
- Update time to mine values to some more realistic times from looking at mainnet Ethereum and Bitcoin fees.
- Did a fix in the `expiries` modules where "bob time to complete" would include "alice's time to act" when the clock started, making Bob abort when Alice would act after 50min (she is allowed 60min).
- Default to use `bitcoind estimatesmartfee` feature for nectar instead of static value.

### DIFF
--- a/comit/src/actions.rs
+++ b/comit/src/actions.rs
@@ -65,13 +65,9 @@ pub mod bitcoin {
     where
         C: secp256k1::Signing,
     {
-        let rate = byte_rate.as_sat();
-        // TODO: change this interface to accept an amount instead of a usize and remove
-        // this allow
-        #[allow(clippy::cast_possible_truncation)]
         primed_transaction
-            .sign_with_rate(secp, rate as usize)
-            .map_err(|_| anyhow::anyhow!("failed to sign with {} rate", rate))
+            .sign_with_rate(secp, byte_rate)
+            .map_err(|_| anyhow::anyhow!("failed to sign with {} rate", byte_rate))
     }
 
     #[derive(Debug, Clone, PartialEq)]
--- a/comit/src/expiries.rs
+++ b/comit/src/expiries.rs
@@ -6,7 +6,8 @@
 
 mod config;
 
-use self::config::{Config, Protocol};
+pub use self::config::*;
+
 use crate::{
     timestamp::{self, Timestamp},
     Network,
--- a/nectar/src/bitcoin/fee.rs
+++ b/nectar/src/bitcoin/fee.rs
@@ -49,7 +49,7 @@ impl Fee {
 impl crate::StaticStub for Fee {
     fn static_stub() -> Self {
         Self {
-            config: Default::default(),
+            config: crate::StaticStub::static_stub(),
             client: bitcoin::Client::new("http://example.com/".parse().unwrap()), // Not used
         }
     }
--- a/nectar/src/bitcoin/fee.rs
+++ b/nectar/src/bitcoin/fee.rs
@@ -1,7 +1,6 @@
 use crate::{bitcoin, bitcoin::Amount, config, Result};
 use anyhow::Context;
-
-const ESTIMATE_FEE_TARGET: u32 = 3;
+use comit::expiries::bitcoin_mine_within_blocks;
 
 #[derive(Clone, Debug)]
 pub struct Fee {
@@ -26,9 +25,12 @@ impl Fee {
         match self.config.fees {
             SatsPerByte(fee) => Ok(fee),
             BitcoindEstimateSmartfee { mode, .. } => {
+                let network = self.config.network.into();
+                let mine_within_blocks = bitcoin_mine_within_blocks(network) as u32;
+
                 let kvbyte_rate = self
                     .client
-                    .estimate_smart_fee(ESTIMATE_FEE_TARGET, Some(mode.into()))
+                    .estimate_smart_fee(mine_within_blocks, Some(mode.into()))
                     .await
                     .map(|res| res.kbyte_rate)?;
 
--- a/nectar/src/bitcoin/fee.rs
+++ b/nectar/src/bitcoin/fee.rs
@@ -5,14 +5,14 @@ const ESTIMATE_FEE_TARGET: u32 = 3;
 
 #[derive(Clone, Debug)]
 pub struct Fee {
-    config: config::BitcoinFees,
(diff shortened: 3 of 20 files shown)


tcharding (approved):

thomaseizinger (commented):
What you did in patch https://github.com/comit-network/comit-rs/pull/3256/commits/327375adaf8ccad955db66aa55f474bf7599114a is unfortunately not quite enough :)

I was working on this as well and included your patch in which you fix it here: https://github.com/comit-network/comit-rs/pull/3261

tcharding:
Added no-mergify because I'm half way through review and have a one-on-one call.

comit/src/expiries.rs
```diff
@@ -712,13 +716,17 @@ impl BobState {
         };
 
         match next_action {
+            Start => {
+                // This time is needed for Alice to act.
+                // When a swap in progress, this time is not taken in
+                // account to know whether Bob can still complete.
+                c.broadcast_alpha_fund_transaction()
```
tcharding, line 723: This should be `c.start()` along with
```
// This time is needed for Alice to start
```

comit/src/expiries.rs
```diff
@@ -712,13 +716,17 @@ impl BobState {
...
+            }
             // Once Alice starts we need at least this much time, note that c.start() is not
             // included otherwise the actual time Alice takes invalidates the swap for Bob.
             WaitForAlphaFundTransactionFinality => {
                 // Transition from Started to AlphaFunded
-                c.broadcast_alpha_fund_transaction()
-                    + c.mine_alpha_fund_transaction()
-                    + c.finality_alpha()
```
tcharding, line 721: I don't agree with this change, can you explain please? Transitioning from Started to AlphaFunded includes the broadcast_aplha_fun_transaction period.

tcharding (dismissed):
> - Update time to mine values to some more realistic times from looking at mainnet Ethereum and Bitcoin fees.
> - Did a fix in the expiries modules where "bob time to complete" would include "alice's time to act" when the clock started, making Bob abort when Alice would act after 50min (she is allowed 60min).
> - Default to use bitcoind estimatesmartfee feature for nectar instead of static value.

ACK the first and last. NACK the second, I do not think this is a good change mate. I understand from the comment that you saw unexpected results that Bob aborted when Alice takes a while to act. I do not think that we should be making tweaks to the expiries state transition periods just to fix edge cases where actors act close to deadlines. The whole module is hueristic at best and is no way safe to push to the edge. On top of this the transition period calculations changed in this PR are incorrect IMO, commented inline.

thomaseizinger:
FYI: https://github.com/comit-network/comit-rs/pull/3261 has been merged into `dev` already, you can now rebase and drop [`62dfe95` (#3256)](https://github.com/comit-network/comit-rs/pull/3256/commits/62dfe952d4c95c4d79a896d3bbacd5deb9d5465a).

comit/src/expiries.rs
```diff
@@ -899,8 +908,8 @@ mod tests {
     fn dev_net_herc20_hbit_expiries() {
         let (a, b) = expiry_offsets_herc20_hbit(Network::Dev);
 
-        assert_eq!(a, 44.seconds().into());
-        assert_eq!(b, 35.seconds().into());
+        assert_eq!(a, 218.seconds().into());
+        assert_eq!(b, 155.seconds().into());
```
thomaseizinger, line 912: This is not good. Why did these values go up that much? If we ever want to test refund in our e2e tests, this is amount of time we would have to wait. 218 seconds is almost 4 minutes of just waiting.

Can we somehow keep these numbers under 1 minute?

D4nte: > This is not good. Why did these values go up that much? If we ever want to test refund in our e2e tests, this is amount of time we would have to wait. 218 seconds is almost 4 minutes of just waiting.
> 
> Can we somehow keep these numbers under 1 minute?

Yes we can with separate mainnet and dev value. Will do.

thomaseizinger: Cool, thank you!

nectar/src/bitcoin/fee.rs
```diff
@@ -18,6 +18,7 @@ impl Fee {
 
     pub async fn kvbyte_rate(&self) -> Result<Amount> {
         let rate = self.vbyte_rate().await?;
+        dbg!(&rate);
```
thomaseizinger, line 20: Left-over `dbg`?

nectar/src/swap.rs
```diff
@@ -385,7 +385,7 @@ mod tests {
                 crate::ethereum::GasPrice::geth_url(ethereum_blockchain.node_url.clone());
 
             let bitcoin_fee = crate::bitcoin::Fee::new(
-                Default::default(),
+                crate::config::BitcoinFees::SatsPerByte(bitcoin::Amount::from_sat(50)),
```
thomaseizinger: Left-over `fixup` commit :)

comit/src/expiries.rs
```diff
@@ -4,7 +4,7 @@
 //! atomic. One expiry for the alpha ledger and one for the beta ledger. The
 //! beta ledger expiry time must not elapse before the alpha ledger expiry time.
 
-mod config;
+pub mod config;
```
thomaseizinger, line 7: We are re-exporting it now, this should go away :)

thomaseizinger (approved):
Looks good! Thank you :)

You've got a left-over `fixup` commit here: [`0e631ea` (#3256)](https://github.com/comit-network/comit-rs/pull/3256/commits/0e631ea9e756e85c1f8289cd1cff8ffbb1af69fd)

comit/src/expiries.rs
D4nte: ooops, slipped in. thanks.

comit/src/actions.rs
```diff
@@ -65,13 +65,10 @@ pub mod bitcoin {
     where
         C: secp256k1::Signing,
     {
-        let rate = byte_rate.as_sat();
-        // TODO: change this interface to accept an amount instead of a usize and remove
-        // this allow
         #[allow(clippy::cast_possible_truncation)]
```
thomaseizinger, line 68: This should not longer be needed I think.
```suggestion
```

thomaseizinger (approved):

PR #3258, opened 2020-10-06
Only talk to the blockchain nodes if really necessary

thomaseizinger:
Instead of pre-maturely fetching the current time from the blockchain
connectors, we pass the future along and only await it if we run
into a refund scenario and need to compare it to the expiry time.

This should significantly speed up calls to GET /swap in any other
case.

### DIFF
--- a/cnd/src/http_api/swaps.rs
+++ b/cnd/src/http_api/swaps.rs
@@ -24,6 +24,7 @@ use crate::{
 };
 use comit::Timestamp;
 use serde::Serialize;
+use std::future::Future;
 use warp::{http, Rejection, Reply};
 
 #[allow(clippy::needless_pass_by_value)]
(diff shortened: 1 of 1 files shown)


D4nte (approved):
I think this is good enough for now but I believe we'll have to optimise further to not have the hanging in a refund scenario.

- Could we cache the current time based on the block we received last?
- Could we check the system time first and only if we are "close enough" to the expiry then check blockchain time?

thomaseizinger:
> I think this is good enough for now but I believe we'll have to optimise further to not have the hanging in a refund scenario.
> 
>     * Could we cache the current time based on the block we received last?
> 
>     * Could we check the system time first and only if we are "close enough" to the expiry then check blockchain time?

Excellent ideas!
I was looking into caching it based on the last block but wasn't quite sure how and where I should be introducing an abstraction.

The 2nd idea is also good, we can keep that in mind in the future :)

D4nte:
> Excellent ideas!
> I was looking into caching it based on the last block but wasn't quite sure how and where I should be introducing an abstraction.
> 
> The 2nd idea is also good, we can keep that in mind in the future :)

D4nte:
> Excellent ideas!
Thanks 😬
> I was looking into caching it based on the last block but wasn't quite sure how and where I should be introducing an abstraction.

I don't think this idea holds if we move from block processing to filters anyway.

da-kami (approved):

tcharding (commented):
Nice idea!

PR #3260, opened 2020-10-06
Replace btsieve's `watch_for_event` implementation with `eth_getLogs`

thomaseizinger:
Instead of scanning the blockchain manually, we can leverage the `eth_getLogs` web3 API of the Ethereum node.

This allows us to directly ask for logs emitted by a contract.

Technically, this functionality should no longer be within the `btsieve` module because it doesn't have anything to do with "sieving transactions". I would defer the fix to this to a later point though.

### DIFF
--- a/comit/src/btsieve/ethereum/watch_for_event.rs
+++ b/comit/src/btsieve/ethereum/watch_for_event.rs
@@ -25,15 +25,15 @@ where
         connector,
         start_of_swap,
         expected_event.topics.clone(),
-        |receipt| find_log_for_event_in_receipt(&expected_event, receipt),
+        |receipt| find_log_for_event(&expected_event, receipt.logs),
     )
     .await
 }
 
-fn find_log_for_event_in_receipt(event: &Event, receipt: TransactionReceipt) -> Option<Log> {
+fn find_log_for_event(event: &Event, logs: Vec<Log>) -> Option<Log> {
     match event {
         Event { topics, .. } if topics.is_empty() => None,
-        Event { address, topics } => receipt.logs.into_iter().find(|log| {
+        Event { address, topics } => logs.into_iter().find(|log| {
             if address != &log.address {
                 return false;
             }
--- a/comit/src/btsieve/ethereum/watch_for_event.rs
+++ b/comit/src/btsieve/ethereum/watch_for_event.rs
@@ -1,33 +1,39 @@
 use crate::{
     btsieve::{
-        ethereum::{poll_interval, Event, ReceiptByHash, Topic},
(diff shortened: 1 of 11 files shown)


comit/src/herc20.rs
```diff
@@ -191,16 +197,18 @@ where
...
     use crate::btsieve::ethereum::Event;
 
     let event = Event {
         address: params.asset.token_contract,
         topics: vec![
-            Some(Topic(*TRANSFER_LOG_MSG)),
+            Some(*TRANSFER_LOG_MSG),
             None,
-            Some(Topic(deployed.location.into())),
+            Some(deployed.location.into()),
```
D4nte, line 211: Didn't you say it would not work for deployed?

D4nte (commented):
LGTM, take it with a pinch of salt as I am not that familiar with Ethereum events.

comit/src/herc20.rs
thomaseizinger: This is within `watch_for_funded` :)

In this case, `deployed.location` is a parameter of the event we are looking for. More concretely, we are looking for a `TRANSFER` event emitted by the token contract in which the 3rd argument is our HTLC.

In other words: We want a `TRANSFER` event in which our HTLC receives tokens :)

comit/src/btsieve/ethereum/watch_for_event.rs
```diff
@@ -1,39 +1,45 @@
...
-        expected_event.topics.clone(),
-        |receipt| find_log_for_event_in_receipt(&expected_event, receipt),
-    )
-    .await
+    let poll_interval = poll_interval(connector).await?;
+
+    loop {
+        let logs = connector.get_logs(expected_event.clone()).await?;
+
+        if let Some(log) = find_log_for_event(&expected_event, logs) {
```
D4nte, line 29: Regarding my other comment, you are using logs here for a generic event, aren't you?

thomaseizinger: I am not sure I understand your question. This code works for any event.

D4nte: So you are using for deployed? I thought it would not work for deployed.

Maybe I am just not familiar enough with btsieve code.

thomaseizinger: > So you are using for deployed? I thought it would not work for deployed.

I am not. We are not emitting an event for contracts that are deployed, hence we can't use `eth_getLogs` for that. We look for the bytecode that is being deployed instead.

da-kami (approved):
Sweet! Let's get it in and see how it performs :)

D4nte:
Updating the changelog would be good I think :)

PR #3262, opened 2020-10-06
Improve JSON-RPC error reporting

thomaseizinger:
Instead of using an untagged enum, we can model the JSON-RPC response
as a tagged enum that is flattened into another struct. This greatly
improves the error message we get from serde if we can't deserialize
a response.

This also allows us to directly go into JSON again within nectar's
RPC client because serde's error messages are actually pretty good
if we don't use untagged enums.


D4nte (approved):
Please do the same change in the nectar jsonrpc.

PR #3274, opened 2020-10-09
Use CypherBlock Bitcoin fee estimation service for redeem and refund transactions

D4nte:


cnd/src/bitcoin_fees.rs
```diff
@@ -11,3 +11,142 @@ impl BitcoinFees {
...
+        }
+
+        /// A rolling average of the fee (in satoshis) paid per kilobyte for
+        /// transactions to be confirmed within 1 to 2 blocks.
+        async fn high_fee_per_kb(&self) -> Result<Amount> {
+            let response = self.get().await?;
+
+            let fee = response.high_fee_per_kb;
+
+            tracing::info!(
```
thomaseizinger, line 68: I appreciate your effort in writing more and good log statements!

In my experience though, having log statements within "connectors" or "clients" adds up to a lot of noise because they may end up being called more often than their data is actually used to make decisions.

Hence, a log statement that goes next to where we use this data would probably be more informative and also less noisy.

If you want keep it here, please reduce this to at least `debug` but probably actually `trace`.

For `info` level, my go to rule of thumb is that it should be associated with a "business state change", like: "Closed order XYZ in swap" or "Swap ABC completed successfully". Ideally, running the software on `info` level tells you about the progress it makes in doing what you want it to do. That only works if we are not too noisy on info level.

Note that I am describing an ideal scenario here, there are probably several log statements that _don't_ adhere to this at the moment :grimacing:

cnd/src/bitcoin_fees.rs
```diff
@@ -11,3 +11,142 @@ impl BitcoinFees {
...
+    pub struct Client {
+        url: url::Url,
+        network: comit::Network,
+    }
+
+    /// https://www.blockcypher.com/dev/bitcoin/#blockchain
+    impl Client {
+        /// Full URL to access blockcypher API blockchain endpoint: http://api.blockcypher.com/v1/btc/main
+        pub fn new(url: url::Url, network: comit::Network) -> Self {
+            Self { url, network }
```
thomaseizinger, line 47: Should we choose the URL based on the network?

cnd/src/config/file.rs
```diff
@@ -45,6 +46,28 @@ pub struct Bitcoin {
...
 
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct BitcoinFees {
+    /// Static value to use
+    #[serde(default)]
+    #[serde(with = "::bitcoin::util::amount::serde::as_sat::opt")]
+    pub sat_per_vbyte: Option<bitcoin::Amount>,
+    /// XOR, use CypherBlock estimation service
+    #[serde(default)]
```
thomaseizinger, line 57: Do we need `serde(default)` for `Option`s?

cnd/src/config/settings.rs
```diff
@@ -159,6 +163,115 @@ impl From<file::Logging> for Logging {
...
+            fees: Some(settings.fees.into()),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum BitcoinFees {
+    StaticSatPerVbyte(bitcoin::Amount),
+    CypherBlock(
+        /// Full URL to access blockcypher API blockchain endpoint: http://api.blockcypher.com/v1/btc/main
```
thomaseizinger, line 181: This doc comment is getting a bit redundant. You could provide a newtype around `Url` and use that instead.

That would allow you to not repeat this documentation all over the place.

cnd/src/config/settings.rs
```diff
@@ -410,4 +521,48 @@ mod tests {
 
         assert_that(&settings).is_err();
     }
+
+    #[test]
+    fn given_network_on_cli_when_config_disagrees_then_error() {
```
thomaseizinger, line 465: Thank you for finally adding tests for this <3

tests/src/environment/cnd_config.ts
```diff
@@ -71,9 +71,19 @@ interface Bitcoind {
...
 }
 
+interface BitcoinFees {
+    sat_per_vbyte?: number;
+    cypherblock?: {
+        // No point giving the option of "false" as absent == false
+        enabled: true;
+        blockchain_endpoint_url?: string;
+    };
+}
```
thomaseizinger, line 81: What was the motivation for diverging from the design you used in nectar where the user chooses a 'strategy'?

thomaseizinger (approved):
Nice, thank you @D4nte !

cnd/src/bitcoin_fees.rs
D4nte: Oh, good idea.

cnd/src/config/file.rs
D4nte: Don't we? I'll check.

cnd/src/config/settings.rs
D4nte: I just moved the tests :)

tests/src/environment/cnd_config.ts
D4nte: The fact that the user could use a "strategy" and then fill fields that are not used by the strategy, making it just confusing.

But yeah, maybe the best of both world is:
- use strategy
- fail to convert to `Settings` if any field NOT used by the selected strategy is set.

cnd/src/bitcoin_fees.rs
D4nte: In this particular instance, I think it can be kept and downgraded to `debug`.

tests/src/environment/cnd_config.ts
D4nte: I think I came up with a best design.

cnd/src/config/settings.rs
D4nte: Not convinced about the newtype, I'll clean up the comment.

thomaseizinger: The idea of the newtype would be that the URL is statically linked to it and you for example can only instantiate it by passing a `comit::Network`. Up to you :)

PR #3294, opened 2020-10-14
Replace dedicated error structs with anyhow's .context

thomaseizinger:
We don't need to define a struct for the error if we all we want
is to add more information to the error stacktrace. Addititionally,
if we were to define a struct for the error, we shouldn't print the
inner error as part of this error struct's message otherwise the
message will get printed twice once anyhow prints the error stacktrace.
More importantly, we want to include context that would otherwise
be lost like the URL we made the request to.

### DIFF
--- a/cnd/src/bitcoin_fees.rs
+++ b/cnd/src/bitcoin_fees.rs
@@ -31,7 +31,7 @@ impl BitcoinFees {
 
 mod block_cypher {
     use ::bitcoin::Amount;
-    use anyhow::Result;
+    use anyhow::{Context, Result};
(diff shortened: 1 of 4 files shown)


bonomat (approved):
👍

D4nte (commented):
Good, I did not like this usage of structs.

thomaseizinger:
> Good, I did not like this usage of structs.

To be clear, I think using structs is good if we want to react on this error because it allows downcasting from `anyhow::Error`.

In any case, I would probably prefer .context though if we already return an `anyhow::Result`.

PR #3297, opened 2020-10-16
Nectar sentry support

da-kami:
Note: The text of the messages in the screenshot below is slightly outdated. Screenshot just to give an overview of how this looks in Sentry.

image


nectar/sample-config.toml
```diff
@@ -62,3 +62,8 @@ node_url = "http://localhost:8545/"
 [ethereum.gas_price]
 service = "geth"
 url = "http://some.geth.url:8545/"
+
+# Logging service for Sentry, if you specify a URL here certain events will be automatically logged to Sentry.
+[sentry]
+# The DSN URL as specified by your Sentry project. The DSN tells the  Sentry SDK where to send the events to.
```
thomaseizinger, line 68:
```suggestion
# The DSN URL as specified by your Sentry project. The DSN tells the Sentry SDK where to send the events to.
```

nectar/src/config/settings.rs
```diff
@@ -408,6 +414,10 @@ impl From<Settings> for File {
             }),
             bitcoin: Some(bitcoin.into()),
             ethereum: Some(ethereum.into()),
+            sentry: match sentry {
+                None => None,
+                Some(sentry) => Some(file::Sentry { url: sentry.url }),
+            },
```
thomaseizinger: Please follow how the existing code config code is handled here and extract the actual mapping into a function of some sorts.

You could also utilize `Option::map` here.

nectar/src/config/settings.rs
```diff
@@ -483,6 +494,10 @@ impl Settings {
                 || Ethereum::default_from_chain_id(comit_network.unwrap_or_default().into()),
                 |file| Ethereum::from_file(file, comit_network),
             )?,
+            sentry: match sentry {
+                None => None,
+                Some(sentry) => Some(Sentry { url: sentry.url }),
+            },
```
thomaseizinger: Same thing here, please extract the actual mapping to a dedicated function of some sorts.

nectar/src/swap.rs
```diff
@@ -716,5 +718,10 @@ async fn execute(
 
     tracing::info!("swap {} finished successfully", swap_id);
 
+    sentry::capture_message(
+        format!("Swap with swap-id {} finished successfully", swap_id).as_str(),
```
thomaseizinger: For consistency's sake :man_shrugging: 

```suggestion
        format!("Swap {} finished successfully", swap_id).as_str(),
```

nectar/src/swap.rs
```diff
@@ -19,7 +19,7 @@ pub use self::comit::{hbit, herc20};
 pub use crate::database::Database;
 use time::OffsetDateTime;
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, strum_macros::Display)]
```
thomaseizinger, line 22: Are you aware that the generated `Display` implementation will only print the variant name?

thomaseizinger (approved):
Thanks!

I am excited about landing this!

nectar/src/swap.rs
da-kami: Yes, the way I am using it is actually for printing the name for reporting to sentry. I was not planning to report the complete params. It should not be necessary. 
I think that should be fine for now, can optimize this later, what do you think?

nectar/src/config/settings.rs
da-kami: My brain does not work perfectly atm, does the last patch fix this?

nectar/src/swap.rs
thomaseizinger: Yes, I just wasn't sure if you were fully aware of what this does :)

nectar/src/config/settings.rs
```diff
@@ -414,10 +426,7 @@ impl From<Settings> for File {
             }),
             bitcoin: Some(bitcoin.into()),
             ethereum: Some(ethereum.into()),
-            sentry: match sentry {
-                None => None,
-                Some(sentry) => Some(file::Sentry { url: sentry.url }),
-            },
+            sentry: sentry.map(|sentry| sentry.into()),
```
thomaseizinger: Could also write this as: `sentry.map(file::Sentry::from)`

thomaseizinger (approved):

nectar/Cargo.toml
```diff
@@ -27,6 +27,7 @@ num = "0.3"
 pem = "0.8"
 reqwest = { version = "0.10", default-features = false, features = ["json", "native-tls-vendored"] }
 rust_decimal = "1.8"
+sentry = { version = "0.19.0", features = ["anyhow"] }
```
D4nte:
```suggestion
sentry = { version = "0.19", features = ["anyhow"] }
```

nectar/src/config/file.rs
```diff
@@ -171,6 +173,13 @@ pub enum AllowedOrigins {
     Some(Vec<String>),
 }
 
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct Sentry {
+    /// The url to the sentry project
```
D4nte: This comment seems redundant "`Sentry::url` is the url to sentry".

nectar/src/maker/strategy.rs
```diff
@@ -116,6 +122,10 @@ impl AllIn {
         mid_market_rate: Rate,
     ) -> Result<BtcDaiOrderForm> {
         if quote_balance <= self.dai_reserved_funds {
+            sentry::capture_message(
+                "DAI balance too low to create order",
```
D4nte, line 126: Reading that, it confirm my opinion that we should hook sentry to tracing warn and error entries.

Please do it, then review warn and error tracing entries and confirm they are relevant.

D4nte (commented):
Please hook sentry to tracing.

Doing logging right is hard, having an extra level of logging (sentry) just makes it harder to do right.

I think all error and warn tracing entries should be sent to sentry.

The opposite is also true: if there is a warn/error that would be sent to sentry but you don't like it, then it should not be a warn or error.

da-kami:
> Please hook sentry to tracing.
> 
> Doing logging right is hard, having an extra level of logging (sentry) just makes it harder to do right.
> 
> I think all error and warn tracing entries should be sent to sentry.
> 
> The opposite is also true: if there is a warn/error that would be sent to sentry but you don't like it, then it should not be a warn or error.

I explicitly don't want to log all logs to Sentry at the moment - no matter what level they are on. Eventually we could go more full-blown, but at the moment I don't want to burst through the free `Developer` account on Sentry. 

That said, I don't need all the log, but only:

- Trade started / finished
- Balance low
- Swap execution errors

Those are actions were I have to get active, so it should be good enough for now. I would test further Sentry logging at a later stage.

nectar/src/maker/strategy.rs
da-kami: If we get a feature request from someone else running the software we can do it, otherwise I would just do what I need at the moment, see my comment below :)

D4nte:
> > Please hook sentry to tracing.
> > Doing logging right is hard, having an extra level of logging (sentry) just makes it harder to do right.
> > I think all error and warn tracing entries should be sent to sentry.
> > The opposite is also true: if there is a warn/error that would be sent to sentry but you don't like it, then it should not be a warn or error.
> 
> I explicitly don't want to log all logs to Sentry at the moment - no matter what level they are on. Eventually we could go more full-blown, but at the moment I don't want to burst through the free `Developer` account on Sentry.
> 
> That said, I don't need all the log, but only:
> 
>     * Trade started / finished
> 
>     * Balance low
> 
>     * Swap execution errors
> 
> 
> Those are actions were I have to get active, so it should be good enough for now. I would test further Sentry logging at a later stage.

Ok.

Slightly modified proposal:

Still push all tracing warn/error to sentry via hook.

- Trade started / finished: dedicated sentry call for this one
- Balance low: tracing warn, pushed to sentry via hook
- Swap execution errors: tracing error, pushed to sentry via hook

Please note in nectar we only have ONE `tracing:warn!` call: https://github.com/comit-network/comit-rs/blob/0d637304d3bf535ddfd9f05b257888fa840e15bd/nectar/src/swap.rs#L630 and it should be an error actually.

I think my proposal matches your criteria.

da-kami:
> > > Please hook sentry to tracing.
> > > Doing logging right is hard, having an extra level of logging (sentry) just makes it harder to do right.
> > > I think all error and warn tracing entries should be sent to sentry.
> > > The opposite is also true: if there is a warn/error that would be sent to sentry but you don't like it, then it should not be a warn or error.
> > 
> > 
> > I explicitly don't want to log all logs to Sentry at the moment - no matter what level they are on. Eventually we could go more full-blown, but at the moment I don't want to burst through the free `Developer` account on Sentry.
> > That said, I don't need all the log, but only:
> > ```
> > * Trade started / finished
> > 
> > * Balance low
> > 
> > * Swap execution errors
> > ```
> > 
> > 
> > Those are actions were I have to get active, so it should be good enough for now. I would test further Sentry logging at a later stage.
> 
> Ok.
> 
> Slightly modified proposal:
> 
> Still push all tracing warn/error to sentry via hook.
> 
>     * Trade started / finished: dedicated sentry call for this one
> 
>     * Balance low: tracing warn, pushed to sentry via hook
> 
>     * Swap execution errors: tracing error, pushed to sentry via hook
> 
> 
> Please note in nectar we only have ONE `tracing:warn!` call:
> 
> https://github.com/comit-network/comit-rs/blob/0d637304d3bf535ddfd9f05b257888fa840e15bd/nectar/src/swap.rs#L630
> and it should be an error actually.
> 
> I think my proposal matches your criteria.

Maybe I did not fully understand what you requested earlier: Sentry does not support hooking into `tracing` out of the box. See: https://github.com/getsentry/sentry-rust/issues/180 

We could potentially still implement something to achieve that, but I really don't think it is worth investing our time into that at the moment. If we feel that this feature is "not clean enough" I can also fork the repo and integrate sentry in my fork to achieve the functionality I need :)

D4nte:
> Maybe I did not fully understand what you requested earlier: Sentry does not support hooking into `tracing` out of the box. See: [getsentry/sentry-rust#180](https://github.com/getsentry/sentry-rust/issues/180)

Yep, that is what I meant. Doing ONE function call that both cover tracing and sentry.

> We could potentially still implement something to achieve that, but I really don't think it is worth investing our time into that at the moment. If we feel that this feature is "not clean enough" I can also fork the repo and integrate sentry in my fork to achieve the functionality I need :)

Fair.

D4nte (approved):

PR #3313, opened 2020-10-21
Archive swap command

D4nte:
New command to archive swaps: `nectar archive-swap <swap id>`.
The command should be done while nectar is stopped as it updates the database.
Archiving a swap stops nectar to resume its automated execution, `create-transaction` command can then be used to recover funds.

### DIFF
--- a/nectar/CHANGELOG.md
+++ b/nectar/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+**This release includes changes to the DB schema, we recommend to backup your data directory before upgrading.**
+
+### Added
+
+-   New command to archive swaps: `nectar archive-swap <swap id>`.
+    The command should only be used while nectar is stopped as it updates the database.
+    Archiving a swap stops nectar to resume its automated execution, `create-transaction` command can then be used to recover funds.
+
 ## [nectar-0.1.0] - 2020-10-20
 
 ### Added
@@ -20,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 -   Update the expected network times to calculate the expiries: We expect Bitcoin's transactions to be included within 6 blocks and Ethereum's within 30 blocks.
--   By default, use bitcoind's `estimatsmartfee` feature to estimate Bitcoin fees.
+-   By default, use bitcoind's `estimatesmartfee` feature to estimate Bitcoin fees.
     For Ethereum, Eth Gas Station API is used.
 -   Change log level configuration format from capitalised (e.g. "Debug") to lowercase (e.g. "debug").
 
--- a/nectar/src/database.rs
+++ b/nectar/src/database.rs
@@ -38,8 +38,8 @@ impl Database {
     pub fn new(path: &std::path::Path) -> Result<Self> {
         let path = path
             .to_str()
-            .ok_or_else(|| anyhow!("The path is not utf-8 valid: {:?}", path))?;
-        let db = sled::open(path).with_context(|| format!("Could not open the DB at {}", path))?;
+            .ok_or_else(|| anyhow!("failed to convert path to utf-8 string: {:?}", path))?;
+        let db = sled::open(path).with_context(|| format!("failed to open DB at {}", path))?;
 
         if !db.contains_key(Self::ACTIVE_PEER_KEY)? {
             let peers = Vec::<ActivePeer>::new();
@@ -59,7 +59,7 @@ impl Database {
     pub fn new_test() -> Result<Self> {
         let tmp_dir = tempfile::TempDir::new().unwrap();
         let db = sled::open(tmp_dir.path())
-            .with_context(|| format!("Could not open the DB at {}", tmp_dir.path().display()))?;
+            .with_context(|| format!("failed to open DB at {}", tmp_dir.path().display()))?;
 
         let peers = Vec::<ActivePeer>::new();
         let peers = serialize(&peers)?;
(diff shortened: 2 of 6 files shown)


D4nte:
@da-kami @thomaseizinger What did we say about active peers? As the automation is off, it means nectar will not remove the peer from active peers once the swap is manually finished.

The easiest way would be to remove the peer as active when archiving a swap. Are we happy with that?

thomaseizinger:
> The easiest way would be to remove the peer as active when archiving a swap. Are we happy with that?

I think that makes the most sense yes. Once it is archived, it should be fine for nectar to start trading again with someone.

nectar/src/database.rs
```diff
@@ -171,8 +171,32 @@ impl Database {
...
+        let stored_swap = self.get_swap_or_bail(&swap_id)?;
+
+        if let Some(true) = stored_swap.archived {
+            anyhow::bail!("Swap is already archived");
+        }
+
+        let mut new_swap = stored_swap.clone();
+        new_swap.archived = Some(true);
+
+        let key = serialize(&swap_id).context("Could not serialize swap id for db storage")?;
```
thomaseizinger: Most of our error messages use the sentence pattern of "failed to XYZ". Not sure how fussed you are about consistency here but I mildly prefer the "failed to" spelling.

Additionally, we usually lowercase those messages.

nectar/src/database.rs
```diff
@@ -467,4 +494,20 @@ mod tests {
...
+        db.insert_swap(swap).await.unwrap();
+
+        db.archive_swap(&swap_id).await.unwrap();
+
+        // Archiving an already archived swap must fail
+        let res = db.archive_swap(&swap_id).await;
+
+        assert!(res.is_err());
+
+        true
```
thomaseizinger: This could just be:

```suggestion
        res.is_err()
```

D4nte:
Should I test that a DB file created with a previous version of nectar can be loaded with this change?
Should it be something we add to the CI (e.g. committing a test file)?

nectar/src/database.rs
```diff
@@ -145,6 +145,13 @@ impl Database {
                     let swap = deserialize::<Swap>(&value).context("Could not deserialize swap");
 
                     match (swap_id, swap) {
+                        (
+                            Ok(_),
+                            Ok(Swap {
+                                archived: Some(true),
```
thomaseizinger: Thinking out load: This means `archived: None` is equivalent to `archived: Some(false)`. I guess that makes sense.

thomaseizinger: I think this filtering would be clearer if you add another `.filter` where you test for the `archived` property explicitly because this whole match expression seems to be about deserializing and hence I wouldn't include this "business logic" here.

nectar/CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+-   New command to archive swaps: `nectar archive-swap <swap id>`.
+    The command should be done while nectar is stopped as it updates the database.
```
thomaseizinger:
```suggestion
    The command should only be used while nectar is stopped as it updates the database.
```

nectar/CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+    Archiving a swap stops nectar to resume its automated execution, `create-transaction` command can then be used to recover funds.
+
+### Changed
+
+-   **DB Change**: Add an "archived" field to stored swap data;
```
thomaseizinger: This is not a separate feature, hence I would mention it twice.

What do you think about including a general notice at the beginning of the changelog that says: "This release includes changes to the DB schema, we recommend to backup your data directory before upgrading"?

thomaseizinger (approved):
Thanks!

A note regarding the commits: I personally find it easier to work with if the last commit of a patch-series is the one that adds the new feature because it can then easily squash more commits into it.

For this particular PR, I would probably squash commits 2,4,5,6 together into one as they don't really add any value just by themselves.

thomaseizinger:
> Should I test that a DB file created with a previous version of nectar can be loaded with this change?
> Should it be something we add to the CI (e.g. committing a test file)?

I would probably just test that once locally. There isn't really a point in testing that continuously because it is not something that can easily break in the future right? Actually, I guess that is only true for migrations - they run only once and then never again ...

I wouldn't mind integration tests that test if we can load a database from a specific version, like:

```rust
#[test]
fn can_load_nectar_0_1_0_database_without_errors() {}

#[test]
fn can_load_nectar_0_2_0_database_without_errors() {}

// etc
```

nectar/src/database.rs
```diff
@@ -142,14 +142,15 @@ impl Database {
...
-                        (Ok(swap_id), Ok(swap)) => Some(Ok(SwapKind::from((swap, swap_id)))),
+                        (Ok(swap_id), Ok(swap)) => Some(Ok((swap, swap_id))),
                         (Ok(_), Err(err)) => Some(Err(err)), // If the swap id deserialize, then
                         // it should be a swap
                         (..) => None, // This is not a swap item
                     }
                 }
                 Err(err) => Some(Err(err).context("failed to retrieve swaps from DB")),
             })
+            .map(|res| res.map(|(swap, swap_id)| SwapKind::from((swap, swap_id))))
```
thomaseizinger, line 196: Does clippy not suggest you to use a function reference here?

```suggestion
            .map(|res| res.map(SwapKind::from))
```

thomaseizinger (approved):

nectar/src/database.rs
D4nte: You can see the CI passed :)

da-kami (approved):

D4nte:
> > Should I test that a DB file created with a previous version of nectar can be loaded with this change?
> > Should it be something we add to the CI (e.g. committing a test file)?
> 
> I would probably just test that once locally. There isn't really a point in testing that continuously because it is not something that can easily break in the future right? Actually, I guess that is only true for migrations - they run only once and then never again ...
> 
> I wouldn't mind integration tests that test if we can load a database from a specific version, like:
> 
> ```rust
> #[test]
> fn can_load_nectar_0_1_0_database_without_errors() {}
> 
> #[test]
> fn can_load_nectar_0_2_0_database_without_errors() {}
> 
> // etc
> ```

Will add this in follow up PR as I am encountering problems: #3314.

nectar/src/database.rs
D4nte: Will fix with follow up PR

PR #3316, opened 2020-10-22
Remove DB Bug + test backward compatibility

D4nte:
See changelog re bug.

Snapshot data taken on tag `nectar-0.1.0` and test that the current code can load said data as part of the CI.


nectar/src/database.rs
```diff
@@ -2,18 +2,23 @@ use self::{
...
-#[cfg(test)]
-use crate::StaticStub;
 use std::{collections::HashSet, iter::FromIterator};
 use time::OffsetDateTime;
 
 mod hbit;
 mod herc20;
 
+static ACTIVE_PEER_KEY: Lazy<Vec<u8>> =
+    Lazy::new(|| serialize(&"active_peer").expect("this slice can be serialized"));
```
thomaseizinger, line 20: Alternatively, we could just make these `const` byte slices:

```suggestion
const ACTIVE_PEER_KEY: &[u8] = b"active_peer";
```

nectar/src/database.rs
```diff
@@ -636,3 +643,77 @@ mod tests {
...
+
+        #[allow(clippy::cast_possible_truncation)]
+        let expected_index = SNAPSHOT_SIZE as u32;
+        assert_eq!(bitcoin_index, expected_index);
+        assert_eq!(stored_swaps.len(), SNAPSHOT_SIZE);
+        assert_eq!(stored_peers.len(), SNAPSHOT_SIZE);
+    }
+
+    #[tokio::test]
+    #[ignore]
```
thomaseizinger, line 679: Just an alternative idea:

We could move some code to a `lib.rs` and simply create multiple binaries that use the same code. Then we don't need to (ab)-use ignored tests for this :)

nectar/src/database/test_snapshot/v0_1_0.rs
```diff
@@ -0,0 +1,2037 @@
+use conquer_once::Lazy;
+
+pub static TAR: Lazy<Vec<u8>> = Lazy::new(|| {
```
thomaseizinger, line 3: Why didn't you just create a DB file with the 0.1.0 tag and check that into source-control? Feels a bit odd to have this embedded here like this.

thomaseizinger:
Sorry for the late review, I thought I had reviewed all your PRs. Must have missed this one :(

nectar/src/database.rs
D4nte: This would break backward compatibility which was not the intent at the time of this PR.

nectar/src/database/test_snapshot/v0_1_0.rs
D4nte: Because in the past, you have expressed that such files in source control were not that clean and hence did PRs to embed the data directly in Rust code.
I just followed your lead here.

PR #3317, opened 2020-10-22
Change software time to act to 24 hours

D4nte:


comit/src/expiries/config.rs
```diff
@@ -34,16 +34,25 @@ use time::Duration;
...
-    pub const ACT_IN_SOFTWARE_SECS: u32 = 15 * 60; // Value arbitrarily chosen.
-    pub const ACT_WITH_USER_INTERACTION_SECS: u32 = 60 * 60; // Value arbitrarily
-                                                             // chosen.
+    use super::*;
+    use conquer_once::Lazy;
+    use time::NumericalDuration;
+
+    // 10 minutes, average Bitcoin block time.
+    pub static BITCOIN_BLOCK_TIME: Lazy<Duration> = Lazy::new(|| 600.seconds());
+    // Conservative Ethereum block // time.
```
tcharding:
```suggestion
    // Conservative Ethereum block time.
```

comit/src/expiries/config.rs
```diff
@@ -34,16 +34,25 @@ use time::Duration;
...
+    pub static ETHEREUM_BLOCK_TIME: Lazy<Duration> = Lazy::new(|| 20.seconds());
+
+    // Standard in the Bitcoin ecosystem.
+    pub const BITCOIN_CONFIRMATIONS: u8 = 6;
+    // Value used by Kraken.
+    pub const ETHEREUM_CONFIRMATIONS: u8 = 30;
+
+    // Allows someone running this as a software to react to any failure within 24
+    // hours.
+    pub static ACT_IN_SOFTWARE: Lazy<Duration> = Lazy::new(|| 24.hours());
```
tcharding: This is a totally different meaning to what this const originally had, was that intended? It used to by 15 minutes (less than for user interaction) based on the assumption that software acts instantly and the only time we would wait is for the network (hence 15 minutes). Now its 24 hours so it seems either the assumptions have changed or the meaning has changed?

tcharding (commented):
There isn't much explanation for this change so its hard to tell what the reasoning is. I don't think this is what you want to do. Did you run the test that outputs calculated expiries to see what we get after this change? 
```
cargo test -p comit can_create_offsets  -- --nocapture
...
herc20-hbit expiry offsets: 
 alpha: (222000 total seconds) 2 days 13 hours 40 minutes 0 seconds 
 beta: (133200 total seconds) 1 days 13 hours 0 minutes 0 seconds
hbit-herc20 expiry offsets: 
 alpha: (303600 total seconds) 3 days 12 hours 20 minutes 0 seconds 
 beta: (116400 total seconds) 1 days 8 hours 20 minutes 0 seconds
```

comit/src/expiries/config.rs
D4nte: The software can fail and the operator may not be in front of a computer for 24 hours.
The user is expected to be in front of the computer.

The meaning hasn't changed, the only difference is now we are matching the concept to real life use.

D4nte:
> There isn't much explanation for this change so its hard to tell what the reasoning is.

From the commit:

> Change software time to act to 24 hours
> To allow operator to react to any (overnight) failures.

That's it, I am not sure how I can elaborate on that.

comit/src/expiries/config.rs
tcharding: Fair enough. This time period is added for every time there is an action in software - that's why the expiries are so long once this is applied. I'm not acking or nacking just making you aware of the ramifications in case you missed them :)

comit/src/expiries/config.rs
```diff
@@ -34,16 +34,25 @@ use time::Duration;
...
-
-    pub const ACT_IN_SOFTWARE_SECS: u32 = 15 * 60; // Value arbitrarily chosen.
-    pub const ACT_WITH_USER_INTERACTION_SECS: u32 = 60 * 60; // Value arbitrarily
-                                                             // chosen.
+    use super::*;
+    use conquer_once::Lazy;
+    use time::NumericalDuration;
+
+    // 10 minutes, average Bitcoin block time.
+    pub static BITCOIN_BLOCK_TIME: Lazy<Duration> = Lazy::new(|| 600.seconds());
```
thomaseizinger: `Duration` has `const` constructors that you can use: https://docs.rs/time/0.2.22/time/struct.Duration.html#method.seconds

thomaseizinger (approved):
Would be nice to have the `const`s back :)

comit/src/expiries/config.rs
D4nte: Oh, good tip. One `const` still has to go away tho.

PR #3322, opened 2020-10-23
Add unit tests for all OS as part of CI

D4nte:


Makefile
```diff
@@ -78,8 +78,13 @@ clippy: install_clippy
 	$(CARGO) clippy --all-targets -- -D warnings
 
 test:
-
 	$(CARGO) test --workspace
+
+test_no_docker:
```
thomaseizinger: I'd really wish we solved this in a less hacky way.

comit/tests/bitcoin_transaction_pattern_e2e.rs
```diff
@@ -1,97 +1,102 @@
-use bitcoincore_rpc::RpcApi;
-use comit::btsieve::bitcoin::{watch_for_created_outpoint, BitcoindConnector};
-use images::coblox_bitcoincore::BitcoinCore;
-use reqwest::Url;
-use std::time::Duration;
-use testcontainers::*;
-use time::OffsetDateTime;
+#[cfg(feature = "test-docker")]
```
thomaseizinger: You could also just cfg the actual test :)

comit/Cargo.toml
```diff
@@ -48,5 +48,10 @@ testcontainers = "0.10"
 tokio = { version = "0.2", features = ["macros"] }
 
 [features]
-default = []
+default = ["test-docker"]
 test = ["libp2p/yamux", "libp2p/noise"]
+
+# "test-docker" feature is related to test code
+# if it's enabled then tests needing docker will be ran
+test-docker = []
```
thomaseizinger: What do you think of just making `testcontainers` optional? Then you get a feature "for-free" by cargo and you don't have to declare one.

Unfortunately, dev-dependencies can't be optional so we'd have to move it to the normal dependencies. It would be good to make it non-default then, otherwise the dependency will be included in a normal build.

I know you had some concerns about wanting to run all tests locally with a single command. However, I would argue that idiomatic configuration is more important. If you care strongly about the "single command" feature, then feel free to add a Makefile entry or alternatively make yourself a local alias.

Another viewpoint onto this would be that running testcontainers-based tests is slow anyway, hence I _personally_ actually don't mind them being excluded by default in a `cargo test` run, similarly how running e2e tests is a dedicated command.

thomaseizinger (approved):
I would prefer more idiomatic cargo configuration to be honest.

comit/Cargo.toml
D4nte: I am sold.

It's not like I run much test locally anyway 🙃

comit/tests/bitcoin_transaction_pattern_e2e.rs
D4nte: I would get clippy warnings...

Makefile
```diff
@@ -78,8 +78,13 @@ clippy: install_clippy
 	$(CARGO) clippy --all-targets -- -D warnings
 
 test:
-
 	$(CARGO) test --workspace
+
+test_docker:
+	(cd ./comit; $(CARGO) test --features testcontainers)
+	(cd ./cnd; $(CARGO) test --features testcontainers)
```
thomaseizinger: You removed the dependency from `cnd` :)

```suggestion
	(cd ./cnd; $(CARGO) test)
```

thomaseizinger:
> bors r+

@D4nte I think you need to rebase on latest dev to get @bonomat's changes made to the CI configs in regards of the CLA bot.

PR #3335, opened 2020-10-26
Do not use compare_and_swap for db update

D4nte:
Backward compatibility is maintained to allow newer software to
deserialize older data.
We do not try to guarantee that older data serialised the same way than
newer data.
Which means that `compare_and_swap` fails if the swap was created with
an older version of the software.

As we do not cater for concurrent update of the db, and we just retrieved
the swap to update few lines ago, it is ok to not check again.

This resolves the issue encountered when trying to archive a swap
created before the archive swap feature was introduced.

### DIFF
--- a/nectar/src/database.rs
+++ b/nectar/src/database.rs
@@ -118,15 +118,13 @@ impl Database {
         new_swap.archived = Some(true);
 
         let key = serialize(&swap_id).context("failed to serialize swap id for db storage")?;
-        let old_value = serialize(&stored_swap)
-            .context("failed to serialize stored swap value for db update")?;
         let new_value =
             serialize(&new_swap).context("failed to serialize new swap value for db storage")?;
 
-        self.db
-            .compare_and_swap(key, Some(old_value), Some(new_value))
-            .context("failed to write in the DB")?
-            .context("failed to save in the DB, stored swap somehow changed")?;
+        let _ = self
+            .db
+            .insert(key, new_value)
+            .context("failed to write in the DB")?;
 
         let peer = stored_swap.active_peer;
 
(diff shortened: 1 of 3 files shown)


nectar/src/database.rs
```diff
@@ -537,4 +535,21 @@ mod tests {
 
         !db.contains_active_peer(&peer).unwrap()
     }
+
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    struct Swap0_1_0 {
```
thomaseizinger, line 540: What do we need this struct for?

thomaseizinger (changes requested):
Compare-and-swap is the basic primitive of any lock-free algorithm. Not using it without introducing any other kind of locking is opening the door to data races ...

If this is what we have to do to work with sled then I don't think it is a good fit for our usecase. We should strongly consider moving to something else. We don't have to address this immediately though.

Requested changes because I think there is stale code. I commented inline.

nectar/src/database.rs
D4nte: stale code indeed. Thanks!

D4nte:
> If this is what we have to do to work with sled then I don't think it is a good fit for our usecase. We should strongly consider moving to something else. We don't have to address this immediately though.

It looks like to me is that sled is lower level than I first anticipated.

The proper solution here (except for using something else than sled) would be to migrate existing data to the new format so that compare and swap can still be used.

thomaseizinger:
> > If this is what we have to do to work with sled then I don't think it is a good fit for our usecase. We should strongly consider moving to something else. We don't have to address this immediately though.
> 
> It looks like to me is that sled is lower level than I first anticipated.
> 
> The proper solution here (except for using something else than sled) would be to migrate existing data to the new format so that compare and swap can still be used.

I agree but I don't want us to invest a lot of time into that to be honest :/

D4nte:
> > > If this is what we have to do to work with sled then I don't think it is a good fit for our usecase. We should strongly consider moving to something else. We don't have to address this immediately though.
> > 
> > 
> > It looks like to me is that sled is lower level than I first anticipated.
> > The proper solution here (except for using something else than sled) would be to migrate existing data to the new format so that compare and swap can still be used.
> 
> I agree but I don't want us to invest a lot of time into that to be honest :/

I am pretty sure it's 10 line of code. I am keen to try. Let's see.

D4nte:
10 lines, more or less 200 https://github.com/comit-network/comit-rs/pull/3343

PR #3336, opened 2020-10-26
Resilient and more efficient swap execution for nectar and cnd

thomaseizinger:
Design aspects & features:

- swap execution functions only fail if the funding is incorrect
- swap execution functions are generators that yield actions, performing these actions can fail and hence abort the swap
- watching for events is retried until it succeeds with a constant backoff of 1 second
- watching for events in nectar is optimized to check if the TX ID is confirmed IF we have the transaction ID
- actions in nectar are idempotent even across restarts
- shared swap execution logic between cnd and nectar

Side effects that make this change possible:

- LN swaps are removed
- Directly creating a swap is removed
- Refund actions are removed


luckysori (commented):
Wow, this is really cool! 

It's very interesting how you're focusing solely on happy path execution in the swap generators. I guess if we ever want to support automatic refund again, that could be handled by a different component.

thomaseizinger:
> Wow, this is really cool!
> 
> It's very interesting how you're focusing solely on happy path execution in the swap generators. I guess if we ever want to support automatic refund again, that could be handled by a different component.

Thank you @luckysori !

I can read some (rightful) criticism between these lines of why I am removing refund :D 
The argument I would bring forward to justify is that we never actually supported refund properly in nectar. Given that nectar is always Bob, the only place where it has to perform refund is if Alice doesn't redeem within a certain time period. @tcharding's `expiries` module would give us this information but it is not yet integrated into the actual swap execution functions.

I would envision that we do add it at some point to the swap execution functions by `select!`ing on two futures: the one that watches for funding and one that monitors the blockchain time. I was thinking about adding the feature by I didn't want to let this explode in scope and given that we never actually supported it properly IMO, I just left it out in this implementation.

For cnd, the story is a bit different in that we did actually suggest `refund` to the client at the right point in time. However, given that the only client - Ambrosia - doesn't have any support for this,  we are not effectively losing anything by removing it. In addition, if someone else is using `cnd`, they are not being exposed to any effective risk because `cnd` acts in the role of `Alice` by default and hence never has to worry about losing money.

thomaseizinger:
Sorry for the force-push, I've now ran the all tests locally and everything should be passing now :)

luckysori:
>  I was thinking about adding the feature by I didn't want to let this explode in scope and given that we never actually supported it properly IMO, I just left it out in this implementation

I think wanting to keep this PR's scope as small as possible is totally valid.

da-kami (approved):
Great stuff! I went over it in some detail and understood the overall approach. Did not review all commits in full detail. Would bring it in and test in production asap.

PR #3350, opened 2020-10-30
Implement Arbitrary directly on types where possible

thomaseizinger:
With the introduction swap execution in the comit lib, we also
introduced a dependency on quickcheck. Previously, a lot of the
`Arbitrary` impls where defined within nectar. We can now move
them to the `comit` lib.

### DIFF
--- a/comit/src/lib.rs
+++ b/comit/src/lib.rs
@@ -14,6 +14,8 @@
 #![forbid(unsafe_code)]
 
 pub mod actions;
+#[cfg(feature = "quickcheck")]
+pub mod arbitrary;
 pub mod asset;
 pub mod bitcoin;
 pub mod btsieve;
--- a/comit/src/arbitrary.rs
+++ b/comit/src/arbitrary.rs
@@ -1,0 +1,27 @@
+use quickcheck::{Arbitrary, Gen};
+
+pub mod secp256k1 {
+    use super::*;
+    use ::bitcoin::secp256k1::SecretKey;
+
+    pub fn secret_key<G: Gen>(g: &mut G) -> SecretKey {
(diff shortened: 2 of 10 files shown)


luckysori (approved):

comit/src/lib.rs
```diff
@@ -15,6 +15,8 @@
 #![type_length_limit = "1049479"] // Regressed with Rust 1.46.0 :(
 
 pub mod actions;
+#[cfg(feature = "quickcheck")]
```
D4nte, line 17: As this is test code I would expect it to be behind `cfg(test)`

D4nte (approved):

comit/src/lib.rs
thomaseizinger: If it is behind `cfg(test)`, you cannot use it from a different crate. It is quite common for library crates to expose such implementations under the feature-flag of the library that enables them. It is similar to how implementations of `serde::Serialize` are exposed under a `serde` feature flag.

D4nte: > If it is behind cfg(test), you cannot use it from a different crate.

Can you clarify how it is behind `cfg(test)`?

thomaseizinger: It isn't. If it were, then all consumers of the `comit` crate would not be able to use the functions in this module. This is the **comit** crate in case you didn't notice :)

thomaseizinger: Are you saying that we don't need to expose this module publicly because neither `cnd` nor `nectar` are currently using the functions in this module?

D4nte: Ok, thinking about it. It's fine the way it is :)

PR #3351, opened 2020-10-30
Raspberry pi release

D4nte:
Attach armv7 assets at release time.

Working proof:
- nectar: https://github.com/D4nte/comit-rs/actions/runs/340861629 / https://github.com/D4nte/comit-rs/releases/tag/nectar-99.0.0
- cnd: https://github.com/D4nte/comit-rs/runs/1340021469 / https://github.com/D4nte/comit-rs/releases/tag/cnd-99.0.3

### DIFF
--- a/.github/actions/create-release-archive/action.yml
+++ b/.github/actions/create-release-archive/action.yml
@@ -7,6 +7,9 @@ inputs:
   binary:
     description: 'The name of the binary to pack into the archive'
     required: true
+  target:
+    description: 'The target triple, used to find the binary; pass it if the compilation was done with the `--target` argument'
+    required: false
 outputs:
   archive:
     description: 'The name of the archive'
@@ -20,10 +23,16 @@ runs:
         import platform
         os_info = platform.uname()
 
-        archive_name=f'${{ inputs.binary }}_${{ inputs.version }}_{os_info.system}_{os_info.machine}'
+        arch = os_info.machine
+
+        if "${{ inputs.target }}":
+          triple = "${{ inputs.target }}".split("-")
+          arch = triple[0]
+
+        archive_name=f'${{ inputs.binary }}_${{ inputs.version }}_{os_info.system}_{arch}'
 
         print(f'::set-output name=archive::{archive_name}')
 
     - name: Make archive
       shell: bash
-      run: tar -C ./target/release --create --file=${{ steps.create-archive-name.outputs.archive }} ${{ inputs.binary }}
+      run: tar -C ./target/${{ inputs.target }}/release --create --file=${{ steps.create-archive-name.outputs.archive }} ${{ inputs.binary }}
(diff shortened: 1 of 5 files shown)


thomaseizinger (changes requested):
Nice work @D4nte !

Two minor things:

1. The commit message of the first commit seems to be outdated.
2. In the cnd release, we gate the building and uploading of the docker image on the OS being `ubuntu-latest`. I think we should change to the x64 target triple.

PS: Can we have the same matrix in CI as well please? :)

.github/actions/create-release-archive/action.yml
```diff
@@ -7,6 +7,9 @@ inputs:
   binary:
     description: 'The name of the binary to pack into the archive'
     required: true
+  target:
+    description: 'The target triple, used to find the binary; pass it if it the compilation was done with the `--target` argument'
```
bonomat:
```suggestion
    description: 'The target triple, used to find the binary; pass it if the compilation was done with the `--target` argument'
```

bonomat (approved):

D4nte:
> Nice work @D4nte !
> 
> Two minor things:
> 
>     1. The commit message of the first commit seems to be outdated.
> 
>     2. In the cnd release, we gate the building and uploading of the docker image on the OS being `ubuntu-latest`. I think we should change to the x64 target triple.
> 
> 
> PS: Can we have the same matrix in CI as well please? :)

Yes to all

da-kami (approved):

.github/workflows/release-cnd.yml
```diff
@@ -58,21 +79,29 @@ jobs:
           asset_name: ${{ steps.create-archive.outputs.archive }}
           asset_content_type: application/gzip
 
+      - name: Check if publication to docker should be done
+        id: check-docker
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_REGISTRY_USERNAME }}
+        run: |
+          echo "${{ env.DOCKER_USERNAME != '' && matrix.docker }}"
+          echo "::set-output name=process::${{ env.DOCKER_USERNAME != '' && matrix.docker }}"
```
thomaseizinger: I don't fully understand why this is necessary? If you already have `matrix.docker`, why don't you just make everything conditional on that?

If the secrets are not present, the login will fail. Wouldn't that be preferable over the steps being silently skipped?

thomaseizinger (approved):

.github/workflows/release-cnd.yml
D4nte: This is because the job fails on forks (docker login not setup) and the fork may not want to publish a docker image.

D4nte: I also updated the commit message: 06083973

thomaseizinger: > This is because the job fails on forks (docker login not setup) and the fork may not want to publish a docker image.

Interesting. I would expect forks to then either delete the steps, add appropriate credentials or temporarily switch `docker` to false :D
Thanks for the clarification in the commit message!

PR #3359, opened 2020-11-02
Introduced configurable static rate

D4nte:
In preparation of same asset cross chain swaps such as wBTC<>BTC.

This patch is generic enough so it can be introduced in the codebase
as is (i.e., with a BTC<>DAI context).


D4nte:
Almost there, just need to fix a couple of todos. Input is welcome.

nectar/src/rate.rs
```diff
@@ -114,6 +125,29 @@ impl Spread {
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum RateRetrieval {
```
thomaseizinger, line 129: A more scalable approach could be to have a trait for this and simply pass around a trait object like: `Box<GetRate>` or `&dyn GetRate`.

This allows you to _not_ have a single place in which all these different variants are mentioned. That in turn makes it easier to parallelize changes because it is less likely that changes to different modules conflict with each other. Another way of putting it would be to "optimize for deletion". Optimize the design in a way so that individual features can easily be deleted without interfering with other changes.

nectar/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   New command to migrate the database to a new format;
     Use `nectar migrate-db status` to check if migration is needed;
     If so, backup your data and then execute `nectar migrate-db run` to proceed with the migration.
+-  **Config file breaking change**: Introduced the possibility to use a static rate instead of Kraken mid market rate.
```
thomaseizinger, line 24: If you fancy, you could simplify the e2e tests now because we no longer need to mock the kraken API and as such you can get rid of our beloved `FakeTreasuryService` and directly specify the fixed rate at the top of the test files using the override functionality :)

D4nte:
Replaced with https://github.com/bitcoin-unwrap/comit-rs/pull/1

Issue #3008, opened 2020-07-27
try_lock_exclusive() on data dir, causes stable Access denied (os error 5) on Windows 10


llilakoblock:
Windows 10, latest updates.
It's an winapi error, so compiler, rust toolchain, etc, does not really matter.

Error tag: 0.7.3, and all above.

Reproduce: build tag 0.7.3 or above, with either release or debug cfg. and start cnd.exe. 
Reproducible on clean virtual box machine with Windows 10, without any other software.

thomaseizinger:
Hi @llilakoblock,

sorry for the late reply!

Unfortunately, I don't have a Windows box around to debug this. However, I found this: https://github.com/danburkert/fs2-rs/issues/26

Is that the issue you are facing? If so, we may be able to fix that by opening the lock file with different flags.

PR #3364, opened 2020-11-03
Merge ARM binary build into CI workflow

thomaseizinger:
Thoughts @D4nte ?


.github/workflows/ci.yml
```diff
@@ -58,12 +58,33 @@ jobs:
...
+        target: [x86_64-unknown-linux-gnu, armv7-unknown-linux-gnueabihf, x86_64-apple-darwin, x86_64-pc-windows-gnu]
         include:
-          - os: ubuntu-latest
+
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
             e2e: true
-          - os: macos-latest
+            test_features: testcontainers
+            cnd_binary_path: target/debug/cnd
```
D4nte: if you path `--target` when building, even if it's not cross-compiling then the binary will endup in the `target/<target>/(debug|release)/` path.

So you only need a `binary_suffix` or `binary-ext` option to pass `.exe` for windows.

.github/workflows/ci.yml
```diff
@@ -96,43 +124,37 @@ jobs:
...
-      - name: Run unit tests (no docker)
-        if: matrix.os != 'ubuntu-latest'
-        run: make test
+        run: cargo doc
 
-      - name: Run unit tests (with docker)
-        if: matrix.os == 'ubuntu-latest'
-        run: make test_docker
+      - name: Run unit tests
+        if: matrix.target != 'armv7-unknown-linux-gnueabihf' # We don't want to run tests for ARM as we are only cross-compiling
```
D4nte: That could be an option in the matrix actually. The idea being "run only test only once per platform".

.github/workflows/ci.yml
```diff
@@ -96,43 +124,37 @@ jobs:
...
+        run: |
+          (cd ./comit; cargo test --features "${{ matrix.test_features }}")
```
D4nte, line 130: Is it fine to pass `""` to the `--features` option?

D4nte (approved):
Looks good, few comments.

.github/workflows/ci.yml
thomaseizinger: Yes, I tested it locally. Behaves as expected and doesn't activate any features!

.github/workflows/ci.yml
thomaseizinger: Oh cool, that is good to know! I was annoyed at having to do this :D

.github/workflows/ci.yml
thomaseizinger: I can make another entry in the matrix yeah. Wasn't sure if it is worth it :)

.github/workflows/ci.yml
```diff
@@ -58,12 +58,29 @@ jobs:
...
+
+          - target: armv7-unknown-linux-gnueabihf
+            os: ubuntu-latest
+            # This target is only for cross-compiling, skip all tests.
+            skip_unit_tests: true
+            skip_e2e_test: true
+
+          - target: x86_64-apple-darwin
+            os: macos-latest
+            e2e_tests: true
```
D4nte: I guess the idea is to remove this attribute and only have the `skip_e2e_test` one?

thomaseizinger: Yes, I thought it is more readable if the matrix highlights the exceptions we are making per platform, rather than what we want to do and from there you have to infer what happens when.

D4nte: My comment was more about the odd attribute than the approach :)

.github/workflows/ci.yml
```diff
@@ -58,12 +58,28 @@ jobs:
...
+            skip_unit_tests: true
+            skip_e2e_test: true
+
+          - target: x86_64-apple-darwin
+            os: macos-latest
+
+          - target: x86_64-pc-windows-gnu
+            os: windows-latest
+            skip_e2e_test: true # Our e2e test framework doesn't work on windows.
+            skip_unit_tests: true # Can't run linux docker images on windows.
```
D4nte: We run e2e tests without docker on windows now: https://github.com/comit-network/comit-rs/pull/3322

.github/workflows/ci.yml
```diff
@@ -58,12 +58,28 @@ jobs:
...
+            # This target is only for cross-compiling, skip all tests.
+            skip_unit_tests: true
+            skip_e2e_test: true
+
+          - target: x86_64-apple-darwin
+            os: macos-latest
+
+          - target: x86_64-pc-windows-gnu
+            os: windows-latest
+            skip_e2e_test: true # Our e2e test framework doesn't work on windows.
```
D4nte: The real issue is that our software does not work on windows (see #3008) :)

thomaseizinger: Ouch :D

.github/workflows/ci.yml
thomaseizinger: Yes thanks for pointing that out!

D4nte (approved):

D4nte:
just merge it :)

PR #3365, opened 2020-11-03
Generalize swap actions

thomaseizinger:
Instead of every client needing to repeat themselves in how specific
actions are constructed from the params, we construct the action
directly in the swap protocol functions and emit fairly simply actions
to the caller.

This simplifies the implementation of clients driving the swap.

Additionally, this is conceptually better from the perspective of the
protocol execution function. For this function to be coherent, it
should control all parts that are critical to protocol. Building the
actual action inside the protocol function provides this control.


D4nte (approved):

da-kami (approved):

thomaseizinger:
bors retry

PR #3375, opened 2020-11-05
Yield the blocks as they are mined whilst checking the history

thomaseizinger:
In order to be exhaustive, we need to go all the way back until the
start of the swap. However, most likely the transaction we are looking
for is somewhere within the next blocks.
To speed this process up, we periodically also check the last block.

### DIFF
--- a/comit/src/btsieve.rs
+++ b/comit/src/btsieve.rs
@@ -78,6 +78,7 @@ where
             block,
             |_| true, // initially, yield all blocks because we haven't seen any of them
             connector,
+            poll_interval,
             &co,
         )
         .await?;
@@ -91,6 +92,7 @@ where
                 block,
                 |b| !seen_blocks.contains(b), // only yield if we haven't seen the block before
                 connector,
+                poll_interval,
                 &co,
             )
             .await?;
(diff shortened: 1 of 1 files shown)


comit/src/btsieve.rs
```diff
@@ -172,3 +189,190 @@ where
...
+
+        let gen = fetch_blocks_since(&connector, start_of_swap, ZERO_POLL_INTERVAL);
+        let yielded_blocks = fallible_generator_to_try_stream(gen)
+            .map_ok(|b| b.number)
+            .try_take_while(|n| future::ready(Ok(*n != 52)))
+            .try_collect::<Vec<_>>();
+
+        let processed_blocks = tokio::time::timeout(Duration::from_secs(60), yielded_blocks)
+            .await
+            .expect("block 52 to be yielded within 60 seconds")
```
D4nte, line 227: Do you want to check it goes backward too?

D4nte (approved):

comit/src/btsieve.rs
thomaseizinger: We already check this with our e2e tests and with the integration tests of btsieve. We could in theory replace those tests with what we have here but I don't think it is worth it.

The point of this test was really to make sure the newly added functionality is properly covered!

PR #3380, opened 2020-11-10
Replace language-specific formatters with dprint

thomaseizinger:
This has several advantages:

1. dprint is a lot faster, checking if the whole codebase is
formatted is so fast it is barely noticable.
2. A single tool for all languages (including markdown!), this
simplifies configuration a lot.
3. dprint directly depends on rustfmt and not on the version
shipped with cargo. As such, we can opt into any feature, nightly
or not which means we can remove the dependency on the Rust nightly
toolchain :tada:


.dprintrc.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "includes": ["**/*.{ts,tsx,js,jsx,json,md,rs,toml}"],
+  "excludes": [
+    "**/node_modules",
+    "**/*-lock.json",
+    "**/target",
+    "LICENSE.md",
+    "tests/root.schema.json",
+    "tests/siren.schema.json",
+    "nectar/src/database.rs"
```
thomaseizinger, line 28: This is unfortunate for now, here is the explanation: https://github.com/dprint/dprint/issues/320

D4nte:
>     3. dprint directly depends on rustfmt and not on the version
>        shipped with cargo. As such, we can opt into any feature, nightly
>        or not which means we can remove the dependency on the Rust nightly
>        toolchain 🎉

Do you mean the rustfmt crate as a lib?

Makefile
```diff
@@ -74,31 +50,20 @@ all: format build clippy test doc e2e
...
-check_format: check_rust_format check_toml_format check_ts_format
-
-format: install_rustfmt install_tomlfmt
-	$(CARGO_NIGHTLY) fmt -- --files-with-diff | xargs -I{} git add {}
-	@$(foreach file,$(CARGO_TOML_FILES),$(CARGO) tomlfmt -p $(file) && git add $(file);)
-ifneq (,$(MODIFIED_TYPESCRIPT_FILES))
-	(cd ./tests; yarn install; yarn run fix)
-endif
+check_format:
+	dprint check
```
D4nte, line 64: How do I install dprint?

RELEASING.md
```diff
@@ -39,7 +38,7 @@ We split the release into different workflows for several reasons.
 
 1. Building and attaching the binary needs to happen for several platforms whereas the actual release (and tag) only needs to be created once.
 2. It allows for multiple entry-points into the release process:
-Building binaries is triggered by a "release" event, which could in theory be created manually and doesn't necessarily have to be created through the release automation.
+   Building binaries is triggered by a "release" event, which could in theory be created manually and doesn't necessarily have to be created through the release automation.
```
D4nte, line 41: wow, neat

cnd/CHANGELOG.md
```diff
@@ -9,194 +9,180 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
--   `create-transaction` sub-command: Create a signed transactions for redeeming or refunding `hbit` protocols.
--   `print-secret` sub-command: Prints the secret of a swap IF the node acts in the role of Alice for this swap.
+- `create-transaction` sub-command: Create a signed transactions for redeeming or refunding `hbit` protocols.
```
D4nte, line 12: My understanding is that the release git hub action **adds* the spaces when manipulating the file.

which mean dprint removes it, release adds them, bot fight!

You also setup the release action, I guess maybe tell dlint to put 4 spaces for bullet lists?

cnd/CHANGELOG.md
```diff
@@ -9,194 +9,180 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Ping btsieve on cnd startup, checking for presence and version compatibility.
 
 ### Changed
 
--   Move config files to standard location based on platform (OSX, Windows, Linux).
--   Align implementation with RFC-002 to use the decision header instead of status codes.
+- Move config files to standard location based on platform (OSX, Windows, Linux).
+- Align implementation with RFC-002 to use the decision header instead of status codes.
 
 [Unreleased]: https://github.com/comit-network/comit-rs/compare/cnd-0.9.0...HEAD
```
D4nte, line 174: Same issue than for the list, I believe the release github action added the new lines.

cnd/Cargo.toml
```diff
@@ -1,41 +1,41 @@
 [package]
-authors = ["CoBloX <hello@coblox.tech>"]
 name = "cnd"
```
D4nte, line 2: Let's remove cargo toml from the CI/makefile then.

comit/Cargo.toml
```diff
@@ -1,54 +1,52 @@
...
+libp2p = { version = "0.29", default-features = false, features = [ "yamux", "noise" ] }
 proptest = "0.10"
 spectral = { version = "0.6", default-features = false }
-tokio = { version = "0.2", features = ["macros"] }
+tokio = { version = "0.2", features = [ "macros" ] }
 
 [features]
-default = []
-test = ["libp2p/yamux", "libp2p/noise"]
+default = [ ]
```
D4nte, line 51: meh, this is odd. Can it be setup differently?

D4nte (changes requested):
There is a conflict for the markdown formatting between dprint and the github release action.

I am not convinced about the toml formatting.

Makefile
thomaseizinger: Which ever way you prefer: https://dprint.dev/install/

I've changed my mind on this auto-installing tools topic. What is important is that versions of tools are enforced, like we do for `node` for example but devs should be able to decide themselves, how they want to manage their environment.

cnd/CHANGELOG.md
thomaseizinger: Yes, if you check the release workflow, I am installing dprint there and format everything after the release action updated the file to make sure there isn't a bot fight :)

cnd/Cargo.toml
thomaseizinger: So actually, our manifest files are not formatted correctly given the guidelines that are recommended here: https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/cargo.md?rgh-link-date=2020-11-10T07%3A37%3A52Z

See also https://github.com/dprint/dprint/issues/321 for more discussion!

comit/Cargo.toml
thomaseizinger: At the moment no, I would need to fix this upstream in `toml-parse`. I don't mind much personally though.

cnd/Cargo.toml
D4nte: It's not what I meant. I added a comment on the offending line.

Makefile
```diff
@@ -5,29 +5,18 @@ RUSTUP = rustup
...
-CARGO_NIGHTLY = $(RUSTUP) run --install $(NIGHTLY_TOOLCHAIN) cargo --color always
-
 GIT_HOOKS_PATH = ".githooks"
 GIT_HOOKS = $(wildcard $(GIT_HOOKS_PATH)/*)
 
 INSTALLED_TOOLCHAINS = $(shell $(RUSTUP) toolchain list)
 INSTALLED_COMPONENTS = $(shell $(RUSTUP) component list --installed --toolchain $(TOOLCHAIN))
-INSTALLED_NIGHTLY_COMPONENTS = $(shell $(RUSTUP) component list --installed --toolchain $(NIGHTLY_TOOLCHAIN))
 
 CARGO_TOML_FILES = $(wildcard **/Cargo.toml)
```
D4nte: I believe this is not needed anymore

thomaseizinger:
> > ```
> > 3. dprint directly depends on rustfmt and not on the version
> >    shipped with cargo. As such, we can opt into any feature, nightly
> >    or not which means we can remove the dependency on the Rust nightly
> >    toolchain 🎉
> > ```
> 
> Do you mean the rustfmt crate as a lib?

Yes! See here: https://github.com/dprint/dprint-plugin-rustfmt/blob/7697eab89dc6ac2c548d6bbf53ccdecd0034c86a/Cargo.toml#L15

tests/src/environment/crash_listener.ts
```diff
@@ -1,12 +1,12 @@
 export function crashListener(
     pid: number,
     component: string,
-    logFile: string
+    logFile: string,
```
D4nte, line 4: Can we apply a configuration to get a result that does not change so many files?

thomaseizinger: Will try!

thomaseizinger: I tried with `"trailingCommas": "never"` but the diff shrinks only by a couple of lines because we are currently not consistent on this topic. I do think that `"trailingCommas": "onlyMultiLine"` - which is the default - is better though so I am gonna keep it I think!

Makefile
thomaseizinger: In the case of dprint, what is important are the versions + configuration of the plugins. Both are in the dprint configuration file so the actual version of `dprint` is not that important!

D4nte (approved):

